PR #6, opened 2019-10-21
Deprecated client deleteFolder()

corymartin:
The API endpoint `DELETE content/filemapper/v1/delete/folder/:path` is deprecated.
The endpoint `DELETE content/filemapper/v1/delete/:path` can now be used for both file and folder paths.

### DIFF
--- a/packages/cms-cli/commands/remove.js
+++ b/packages/cms-cli/commands/remove.js
@@ -1,6 +1,4 @@
-const path = require('path');
-
-const { deleteFile, deleteFolder } = require('@hubspot/cms-lib/api/fileMapper');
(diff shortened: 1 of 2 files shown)


packages/cms-cli/commands/remove.js
```diff
@@ -44,39 +42,18 @@ function configureRemoveCommand(program) {
...
-                request: hsPath,
-              })
-            );
-          });
-      } else {
-        deleteFile(portalId, hsPath)
-          .then(() => {
-            logger.log(`Deleted "${hsPath}" from portal ${portalId}`);
+      try {
+        await deleteFile(portalId, hsPath);
```
gcorne: :+1:

PR #8, opened 2019-10-24
Add create and fetch commands for managing HubDB tables

gcorne:
This PR will add an initial set of commands for working with HubDB tables and proposes a JSON structure for representing a table on the file system.

The structure avoids the awkwardness of column ids being used in values object for each row and combines the table structure with the rows into a single object. The current proposal is:

```json
{
  "name": "test",
  "columns": [
    {
      "label": "First name",
      "name": "firstName",
      "type": "TEXT"
    },
    {
      "label": "Last name",
      "name": "lastName",
      "type": "TEXT"
    },
    {
      "label": "Photo",
      "name": "photo",
      "type": "IMAGE"
    },
    {
      "label": "Bio",
      "name": "bio",
      "type": "RICHTEXT"
    }
  ],
  "rows": [
    {
      "path": null,
      "name": null,
      "values": {
        "firstName": "John",
        "lastName": "Smith",
        "bio": "A non-descript person of the world."
      }
    },
    {
      "path": null,
      "name": null,
      "values": {
        "firstName": "Cookie",
        "lastName": "Monster",
        "bio": "He likes cookies"
      }
    }
  ]
}
```

That said to minimize overloading, I may switch to something like:

```json
{
  "schema": {},
  "rows": {}
}
```

@anthmatic @axellabs @boulter

### DIFF
--- a/packages/cms-cli/bin/hs-hubdb-create.js
+++ b/packages/cms-cli/bin/hs-hubdb-create.js
@@ -1,0 +1,9 @@
+#!/usr/bin/en node
+
+const { Command } = require('commander');
+
+const { configureHubDbCreateCommand } = require('../commands/hubdb');
+
+const program = new Command('hs hubdb create');
+configureHubDbCreateCommand(program);
+program.parse(process.argv);
--- a/packages/cms-cli/bin/hs-hubdb-fetch.js
+++ b/packages/cms-cli/bin/hs-hubdb-fetch.js
@@ -1,0 +1,9 @@
+#!/usr/bin/en node
+
+const { Command } = require('commander');
+
+const { configureHubDbFetchCommand } = require('../commands/hubdb');
+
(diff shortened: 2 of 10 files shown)


boulter:
Do you need the "values" wrapper object? I assume the keys there must match the "name" attributes of each column?

gcorne:
> Do you need the "values" wrapper object? I assume the keys there must match the "name" attributes of each column?

Hmm…maybe not. Is it possible for there to be conflicts between the top-level properties for a row and the name of a column?

anthmatic:
You can name a column `name`. Depending on where you use it (hubl vs api?) the reserved `name` columns are prefixed with `hs_` 
image

gcorne:
I think that this PR as at a point where we can treat it as an MVP and play around with it a bit. One thing that I am not sure how to handle is associations between tables since those are id-based.

PR #15, opened 2019-11-09
Make removing files when watching an option

gcorne:
Not sure if it is better to make automatic removal of files opt-in or opt-out, but my hunch is that opt-in is better.

@williamspiro @corymartin


williamspiro:
Opt-in sounds less likely to cause confusion to me

PR #18, opened 2019-11-19
Adding hs init command to cli

miketalley:
Added the `hs init` command which prompts the user for portal name, portal id, and portal api key then uses this input to generate `hubspot.config.yaml` in the CWD.

- Execution will exit if `hubspot.config.yaml` already exists
- Only supports api key auth method(for now - oauth2 will be later)

Note:
- Extracted prompts from auth command file into a `prompts.js` file which is shared by `hs auth` and `hs init` commands


mattcoley:
At what point are we going to switch over to https://github.com/HubSpot/hubspot-cms-tools/issues/4 to improve our customers' security?

miketalley:
@mattcoley sounds like that would be a good improvement. I'm not sure about the timing of that -- any insight @gcorne?

gcorne:
I think that environment variable support is a separate concern. The purpose of adding environment variable support is make it more straightforward to use secrets are managed via a separate tool. When working locally, we still will need to store credentials somewhere.

package.json
```diff
@@ -33,6 +33,7 @@
     }
   },
   "dependencies": {
-    "jest": "^24.8.0"
+    "jest": "^24.8.0",
+    "js-yaml": "^3.13.1"
```
gcorne: I believe that this was added for use in `packages/cms-cli`. If so, it should be added to the package.json local to that package. For future reference, `lerna` has some commands for managing deps.

packages/cms-cli/bin/hs-init.js
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
+
+const { Command } = require('commander');
+
+const { initializeConfigCommand } = require('../commands/init');
+
+const program = new Command('hs init');
+initializeConfigCommand(program);
+program.parse(process.argv);
```
gcorne: We have been keeping `hscms` in sync, so you'll need to add `hscms-init.js` as well.

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,83 @@
...
+  return prompt(promptConfig);
+};
+
+const generateHubSpotConfigYAML = configData => {
+  logger.log('Generating hubspot.config.yaml');
+  try {
+    fs.writeFileSync(HUBSPOT_CONFIG_YAML_FILE_NAME, yaml.safeDump(configData));
+    logger.log('Successfully generated hubspot.config.yaml');
+  } catch (e) {
+    logger.error(e);
```
corymartin: We have some handlers to help with printing errors.  Here you could use [`logFileSystemErrorInstance`](https://github.com/HubSpot/hubspot-cms-tools/blob/7a3e83130b4427b36bf2aaffc512615eedf4fcaa/packages/cms-lib/errorHandlers.js#L266). For [example](https://github.com/HubSpot/hubspot-cms-tools/blob/d594ead3a24bbbef3ee9e3afea68dfb4882a7404/packages/cms-lib/lib/config.js#L33).

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,83 @@
...
+      {
+        name: portalName,
+        portalId,
+        authType: 'apikey',
+        apiKey,
+        env: 'PROD',
+      },
+    ],
+  };
+};
```
corymartin: I think we could move this to `cms-lib/lib/config.js`

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,83 @@
...
+function initializeConfigCommand(program) {
+  program
+    .version(version)
+    .description('Initialize hubspot.config.yaml for a HubSpot portal')
+    .action(async options => {
+      setLogLevel(options);
+      logDebugInfo(options);
+
+      if (fs.existsSync(HUBSPOT_CONFIG_YAML_FILE_NAME)) {
+        const configFileExistsError = `The config file '${process.cwd()}/${HUBSPOT_CONFIG_YAML_FILE_NAME}' already exists.`;
```
corymartin: Use `getCwd()` from `cms-lib/path`

packages/cms-cli/lib/prompts.js
```diff
@@ -0,0 +1,70 @@
+const API_KEY_REGEX = new RegExp(
+  '^([a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12})',
```
corymartin: Should this also have the ends-with `$` operator?

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,83 @@
...
+      generateHubSpotConfigYAML(
+        generateConfigData(
+          await promptUser([PORTAL_NAME, PORTAL_ID, PORTAL_API_KEY])
+        )
+      );
+      trackCommandUsage(COMMAND_NAME);
+    });
+
+  addLoggerOptions(program);
+  addConfigOptions(program);
```
gcorne: I don't think that this option makes sense for `init`.

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,56 @@
...
+
+const promptUser = async promptConfig => {
+  const prompt = inquirer.createPromptModule();
+  return prompt(promptConfig);
+};
+
+function initializeConfigCommand(program) {
+  program
+    .version(version)
+    .description('Initialize hubspot.config.yaml for a HubSpot portal')
```
corymartin: We should use `DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME` for the filename here.  The description here says we write a `*.yaml` file but the `DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME` is a `*.yml` file.

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,56 @@
...
+    .action(async options => {
+      setLogLevel(options);
+      logDebugInfo(options);
+
+      const configPath = getConfigPath();
+
+      if (configPath) {
+        const configFileExistsError = `The config file '${configPath}' already exists.`;
+
+        logFileSystemErrorInstance(configFileExistsError, {
```
corymartin: In this case you can just use `logger.error()`.  The handlers are for printing error instances.

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,56 @@
...
+          filepath: configPath,
+        });
+        process.exit(1);
+      }
+
+      writeNewPortalApiKeyConfig(
```
corymartin: This can throw an FS error (via an `fs.write` in `writeConfig()`). So here we need to catch and use `logFileSystemErrorInstance()` to log.

packages/cms-cli/commands/init.js
```diff
@@ -0,0 +1,69 @@
...
+  trackCommandUsage,
+  addHelpUsageTracking,
+} = require('../lib/usageTracking');
+const {
+  DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME,
+} = require('@hubspot/cms-lib/lib/constants');
+const { PORTAL_API_KEY, PORTAL_ID, PORTAL_NAME } = require('../lib/prompts');
+const { addLoggerOptions, setLogLevel } = require('../lib/commonOpts');
+const { logDebugInfo } = require('../lib/debugInfo');
+const { logger } = require('@hubspot/cms-lib/logger');
```
gcorne: nit: let's order the imports so that it is:

1. external npm packages
2. internal npm packages (e.g. `@hubspot/cms-lib/*`)
3. local

packages/cms-cli/package.json
```diff
@@ -12,6 +12,7 @@
     "@hubspot/cms-lib": "^1.0.2",
     "commander": "^2.16.0",
     "inquirer": "^6.3.1",
+    "js-yaml": "^3.13.1",
```
gcorne: looks like this can be removed now since all the yaml stuff is in cms-lib

packages/cms-cli/lib/prompts.js
```diff
@@ -0,0 +1,70 @@
+const API_KEY_REGEX = new RegExp(
+  '^([a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12})$',
+  'i'
+);
```
corymartin: I missed this before but this could be a regex literal.  Our prettier config is based on Airbnb's style guide which prefers literals over constructors when applicable.

gcorne:
I think setting `defaultPortal` using the provided name makes sense. I don't think that the others are necessary as they are mostly for documentation purposes and more advanced uses.

gcorne:
> I think setting defaultPortal using the provided name makes sense. I don't think that the others are necessary as they are mostly for documentation purposes and more advanced uses.

Just realized that ^^ is already implemented.

packages/cms-cli/bin/hs-init.js
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
```
gcorne: This file and `hscms-init` should be marked as executable (e.g. `chmod +x`).

packages/cms-cli/lib/prompts.js
```diff
@@ -54,7 +51,7 @@ const PORTAL_API_KEY = {
   name: 'apiKey',
   message: 'Enter the API key for your portal:',
   validate(val) {
-    if (!API_KEY_REGEX.test(val)) {
+    if (!val.match(API_KEY_REGEX)) {
```
corymartin: This will also work but what you had before (`rgx.test(str)`) is preferable since it always returns a bool.

packages/cms-cli/lib/prompts.js
```diff
@@ -1,7 +1,4 @@
-const API_KEY_REGEX = new RegExp(
-  '^([a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12})$',
-  'i'
-);
+const API_KEY_REGEX = /^([a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12})$/gi;
```
corymartin: The global match isn't needed since we're also capturing beginning/end of string.

miketalley:
@corymartin made the requested changes - look good to you?

Issue #3, opened 2019-10-18
Add `hs init` to provide a guided experience for creating a config 


gcorne:

gcorne:
Addressed via #18 and #19.

PR #19, opened 2019-12-05
Added oauth functionality to hs init command and made it the default

miketalley:
Relates to #3 

This will help users generate an oauth based hubspot.config.yml file by creating a blank file and then leveraging the existing `hs auth oauth` command to prompt for config inputs.

Cancelling out of the prompt will remove the blank file that is created.

### DIFF
--- a/packages/cms-cli/commands/auth.js
+++ b/packages/cms-cli/commands/auth.js
@@ -69,6 +69,15 @@ const addOauthToPortalConfig = (oauth, portalId) => {
   }
 };
 
+const addNewAuthorizedOauthToConfig = async () => {
+  const answers = await getAuthContext();
+  const portalId = parseInt(answers.portalId, 10);
+  const oauth = setupOauth(answers, portalId);
+
+  await authorizeOauth(oauth);
+  addOauthToPortalConfig(oauth, portalId);
+};
+
 async function authAction(type, options) {
   setLogLevel(options);
   logDebugInfo(options);
@@ -85,13 +94,9 @@ async function authAction(type, options) {
     );
     return;
   }
-  const answers = await getAuthContext();
-  const portalId = parseInt(answers.portalId, 10);
-  const oauth = setupOauth(answers, portalId);
 
   trackCommandUsage(COMMAND_NAME);
-  authorizeOauth(oauth);
-  addOauthToPortalConfig(oauth, portalId);
+  await addNewAuthorizedOauthToConfig();
 }
 
 function configureAuthCommand(program) {
--- a/packages/cms-cli/commands/auth.js
+++ b/packages/cms-cli/commands/auth.js
@@ -70,6 +70,7 @@ function configureAuthCommand(program) {
           portalId: answers.portalId,
         });
         logger.log('Configuration updated');
+        process.exit();
       } catch (err) {
         logErrorInstance(err);
       }
--- a/packages/cms-cli/commands/auth.js
+++ b/packages/cms-cli/commands/auth.js
@@ -1,5 +1,4 @@
 const { version } = require('../package.json');
-const inquirer = require('inquirer');
 const OAuth2Manager = require('@hubspot/api-auth-lib/OAuth2Manager');
 const {
   loadConfig,
@@ -21,21 +20,10 @@ const {
   trackCommandUsage,
   addHelpUsageTracking,
 } = require('../lib/usageTracking');
-const {
-  PORTAL_NAME,
-  PORTAL_ID,
-  CLIENT_ID,
-  CLIENT_SECRET,
-} = require('../lib/prompts');
+const { promptUser, OAUTH_FLOW } = require('../lib/prompts');
 
 const COMMAND_NAME = 'auth';
-
 const REQUIRED_SCOPES = ['content'];
-
-const getAuthContext = async () => {
-  const prompt = inquirer.createPromptModule();
-  return prompt([PORTAL_NAME, PORTAL_ID, CLIENT_ID, CLIENT_SECRET]);
-};
 
 const setupOauth = (answers, portalId) => {
   const config = getPortalConfig(portalId) || {};
@@ -70,7 +58,7 @@ const addOauthToPortalConfig = (oauth, portalId) => {
 };
 
 const addNewAuthorizedOauthToConfig = async () => {
-  const answers = await getAuthContext();
+  const answers = await promptUser(OAUTH_FLOW);
   const portalId = parseInt(answers.portalId, 10);
   const oauth = setupOauth(answers, portalId);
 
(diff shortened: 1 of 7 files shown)


packages/cms-cli/commands/init.js
```diff
@@ -20,18 +24,58 @@ const { addLoggerOptions, setLogLevel } = require('../lib/commonOpts');
...
 
 const promptUser = async promptConfig => {
   const prompt = inquirer.createPromptModule();
   return prompt(promptConfig);
 };
 
+const oauthConfigSetup = () => {
+  try {
+    createEmptyConfigFile();
+    const authProcess = spawn(HS_AUTH_OAUTH_COMMAND, {
```
gcorne: Instead of spawning a new process can we refactor the code for the oauth command so that we can do it as part of the same process?

packages/cms-cli/commands/init.js
```diff
@@ -20,18 +24,58 @@ const { addLoggerOptions, setLogLevel } = require('../lib/commonOpts');
...
 function initializeConfigCommand(program) {
   program
     .version(version)
     .description(
       `Initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} for a HubSpot portal`
     )
+    .option(
+      '--api',
+      `create ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} using API key for authentication`
+    )
```
gcorne: Can we make the choice a part of the prompt flow?

miketalley: I added a list selection prompt for choosing the authentication method as well as the option to pass flags(--api/--oauth) to override the prompt.

packages/cms-cli/commands/init.js
miketalley: Done.

packages/cms-cli/commands/init.js
```diff
@@ -26,38 +36,79 @@ const promptUser = async promptConfig => {
...
-      `Initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} for a HubSpot portal`
+      `initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} for a HubSpot portal`
+    )
+    .option(
+      '--api',
+      `initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} using ${AUTH_METHODS.api}`
+    )
+    .option(
+      '--oauth',
+      `initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} using ${AUTH_METHODS.oauth}`
```
gcorne: I think we should forgo adding options to set the auth method.

packages/cms-cli/lib/prompts.js
```diff
@@ -58,10 +59,28 @@ const PORTAL_API_KEY = {
...
+  choices: Object.keys(AUTH_METHODS).map(method => {
+    const authMethod = AUTH_METHODS[method];
+    return {
+      value: authMethod,
+      name: authMethod
+        .split(CAPITAL_LETTER_REGEX)
+        .join(' ')
+        .toLowerCase(),
+    };
+  }),
```
gcorne: Instead of generating the choices, I think this is a situation where it is better to just spell them out. Doing so makes it easy to change the label used, which I think should be adjusted to be "OAuth2" and "API Key."

packages/cms-cli/commands/auth.js
```diff
@@ -22,58 +22,60 @@ const {
...
+
+  if (!validateConfig()) {
+    process.exit(1);
+  }
+
+  if (type !== 'oauth2') {
+    logger.error("The only supported authentication protocol is 'oauth2'");
+    return;
+  }
+  const answers = await getAuthContext();
```
gcorne: If `getAuthContext()` is not `async`, can we remove `await` here?

packages/cms-cli/commands/auth.js
```diff
@@ -22,58 +22,60 @@ const {
...
 
-const getAuthContext = async () => {
+const getAuthContext = () => {
   const prompt = inquirer.createPromptModule();
   return prompt([PORTAL_ID, CLIENT_ID, CLIENT_SECRET]);
 };
 
 const COMMAND_NAME = 'auth';
 
+async function authAction(type, options) {
```
gcorne: I think that we should move this to a separate file outside of the `commands` and that we should adjust the code so that we're not doing stuff that only makes sense in the context of the auth command. For example, `--config` is an option just of `hs auth`. Maybe the solution is to make a function that is just being used to go through the prompted oauth flow.

packages/cms-cli/commands/auth.js
miketalley: we still need async here to await the async return from prompt, which is returned by getAuthContext

gcorne: Should `getAuthContext` be `async` then?

gcorne: I realized after my comment that the code had changed, so feel free to ignore.

packages/cms-lib/oauth.js
```diff
@@ -31,6 +32,34 @@ const getOauthManager = (portalId, portalConfig) => {
...
+const addOauthToPortalConfig = (portalId, oauth) => {
+  logger.log('Updating configuration');
+  try {
+    updatePortalConfig({
+      ...oauth.toObj(),
+      authType: AUTH_METHODS.oauth.value,
+      portalId,
+    });
+    logger.log('Configuration updated');
+    process.exit();
```
gcorne: This looks unexpected. Could we move this as well as the error handling to where this is called?

packages/cms-cli/commands/auth.js
```diff
@@ -20,60 +18,50 @@ const {
...
-const { PORTAL_ID, CLIENT_ID, CLIENT_SECRET } = require('../lib/prompts');
+const { promptUser, OAUTH_FLOW } = require('../lib/prompts');
+
+const COMMAND_NAME = 'auth';
+const REQUIRED_SCOPES = ['content'];
 
-const getAuthContext = async () => {
-  const prompt = inquirer.createPromptModule();
-  return prompt([PORTAL_ID, CLIENT_ID, CLIENT_SECRET]);
+const addNewAuthorizedOauthToConfig = async configData => {
```
gcorne: Maybe this should be renamed to "authenticateWithOauth" since it covers not only updating the config but the OAuth flow?

gcorne: Can we also move this out of the `auth` command and into a file that more clearly indicates its shared nature?

packages/cms-cli/commands/auth.js
miketalley: I've updated the code and extracted some of the oauth-specific logic into a separate oauth file, but I am still exporting a method from `auth` that I am using within `init` -- not sure if this is what you meant or not.

packages/cms-lib/oauth.js
```diff
@@ -31,6 +32,43 @@ const getOauthManager = (portalId, portalConfig) => {
   return oauthManagers.get(portalId);
 };
 
+const setupOauth = (portalId, portalConfig) => {
```
gcorne: I just noticed this, but I think that this could have used OAuth2Manager.fromConfig`.

Issue #4, opened 2019-10-18
Add support for using environment variables instead of a config file


gcorne:
In order to leverage the CLI, there are situations where environment variables are preferred to limit the exposure of secrets. For example, instead of having to generate a config file in a Github action, it would be nice if a developer could just use environment variables.

mattcoley:
Using environmental variables and having a configure step to set up these variables for users would be preferable. See [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html) for an example.

PR #21, opened 2019-12-12
Fixed tests, updated documentation, and added more tracking

miketalley:
Gregory noticed that a test was failing, which should have been removed with the previous PR. This PR removes that test and adds some additional tests as well as fixes a small bug that I discovered while adding tests.

I merged the documentation update into this PR as well, which will allow help text to show up for the `hs init` command and adds updated instructions to the README.md.

This PR also has added tracking for `hs init` for status: 'started' and status: 'complete' so we can monitor the usage of the OAuth2 and API Key flows independently to see their completion rates.

### DIFF
--- a/packages/cms-cli/commands/init.js
+++ b/packages/cms-cli/commands/init.js
@@ -4,7 +4,6 @@ const {
   updatePortalConfig,
   updateDefaultPortal,
   createEmptyConfigFile,
-  deleteEmptyConfigFile,
 } = require('@hubspot/cms-lib/lib/config');
 const {
   logFileSystemErrorInstance,
@@ -38,6 +37,9 @@ const oauthConfigSetup = async ({ configPath }) => {
   const configData = await promptUser(OAUTH_FLOW);
 
   try {
+    createEmptyConfigFile({
+      deleteOnExitIfBlank: true,
+    });
     await authenticateWithOauth(configData);
     process.exit();
   } catch (err) {
@@ -61,6 +63,9 @@ const apiKeyConfigSetup = async ({ configPath }) => {
   const configData = await promptUser(API_KEY_FLOW);
 
   try {
+    createEmptyConfigFile({
+      deleteOnExitIfBlank: true,
+    });
     updateDefaultPortal(configData.name);
     updatePortalConfig(configData);
   } catch (err) {
@@ -92,9 +97,6 @@ function initializeConfigCommand(program) {
         logger.error(`The config file '${configPath}' already exists.`);
         process.exit(1);
       }
-
-      createEmptyConfigFile();
-      process.on('exit', deleteEmptyConfigFile);
 
       const { authMethod } = await promptUser(AUTH_METHOD);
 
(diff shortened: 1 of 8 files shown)


packages/cms-cli/commands/init.js
```diff
@@ -31,11 +31,13 @@ const { logDebugInfo } = require('../lib/debugInfo');
 const COMMAND_NAME = 'init';
 
 const oauthConfigSetup = async ({ configPath }) => {
+  trackCommandUsage(COMMAND_NAME, {
```
gcorne: I think we will need to use a different event for tracking progress so that we don't inflate command usage and get an accurate breakdown of the commands.

packages/cms-lib/lib/config.js
```diff
@@ -243,9 +250,7 @@ const configFileIsBlank = () => {
 };
 
 const createEmptyConfigFile = () => {
-  if (!_configPath) {
-    setDefaultConfigPath();
-  }
+  setDefaultConfigPathIfUnset();
```
gcorne: What happens if there already is a config file? Will this overwrite it?

packages/cms-cli/commands/init.js
```diff
@@ -29,45 +28,54 @@ const { addLoggerOptions, setLogLevel } = require('../lib/commonOpts');
...
-      filepath: configPath,
-      configData,
-    });
-  }
-
-  trackCommandUsage(COMMAND_NAME, {
-    authType: AUTH_METHODS.oauth.value,
-  });
+const trackAuthMethodStatus = (authMethod, status) => {
+  return trackCommandUsage(`${authMethod} ${status}`);
```
gcorne: Since what we're tracking here is not usage of a command but progress within a flow initiated by a command, I think we should create a separate tracking helper with a different `action` property.

packages/cms-lib/lib/config.js
```diff
@@ -242,9 +261,15 @@ const configFileIsBlank = () => {
...
-    setDefaultConfigPath();
+const createEmptyConfigFile = ({ deleteOnExitIfBlank }) => {
+  setDefaultConfigPathIfUnset();
+
+  if (configFileExists()) {
+    return;
+  }
+
+  if (deleteOnExitIfBlank) {
+    process.on('exit', deleteEmptyConfigFile);
```
corymartin: This seems like side effect code for a lib function.  Couldn't the subscription be done from the setup code in the init.js command file?

miketalley: I had thought allowing this option would make the method more robust, but I can see how it adds unnecessary complexity to it. Moved the logic out back into init.

gcorne: Thanks for doing that. I think in general we want to minimize ties in `cms-lib` to execution context meaning that we shouldn't assume that it is used in a script. I suspect tackling #4 and the refactoring required could further decouple configuration handling from the CLI/script use case.

Issue #13, opened 2019-11-07
Add ability to upload and fetch files from the file manager


gcorne:
- [x] Upload a file/folder via `hs filemanager upload`
- [x] Fetch a file or folder via `hs filemanager fetch`
- [ ] Delete a file or folder via `hs filemanager remove`

gcorne:
@miketalley not a high priority, but I think you could work on implementing `hs filemanager fetch`.

gcorne:
Created #232 to cover the `hs filemanager remove`.

PR #24, opened 2019-12-16
Add new filemanager subcommand for uploading files to the File Manager

gcorne:
This PR adds a new `filemanager` subcommand for uploading files to the File Manager.

Part of #13 

- [x] Update documentation

### DIFF
--- a/packages/cms-cli/README.md
+++ b/packages/cms-cli/README.md
@@ -61,13 +61,13 @@ Show all commands
 hs help
 ```
 
-Upload a file or directory to HubSpot
+Upload a file or directory to the Design Manager
 
 ```bash
 hs upload --portal=DEV [src] [dest]
 ```
 
-Fetch a file or directory by path
+Fetch a file or directory by path from the Design Manager
 
 ```bash
 hs fetch --portal=DEV [path] [dest]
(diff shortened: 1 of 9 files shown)


packages/cms-cli/README.md
```diff
@@ -96,3 +96,11 @@ Authenticate against a portal using OAuth2
...
 hs auth oauth2
 ```
+
+Upload a file or directory to the File Manager
+
+```bash
+hs filemanager upload --portal=DEV [src] [dest]
+```
+
+**Note:** Uploading to the File Manager is only supported when using the `apiKey` at this point.
```
gcorne: @williamspiro how do you feel about documenting the command before we have done the work needed to support OAuth.

williamspiro: I think that is okay as it will allow us to get feedback faster  👍

williamspiro:
This is awesome 🚀  2 things I notice
1. It might be useful in logging to distinguish a file was uploaded to the File Manager vs. the Design Manager? 
2. What are the subcommands of `filemanager <subcommand>`? I tried `watch` and `fetch` but neither work, so I am guessing only upload is support at the moment.

gcorne:
> What are the subcommands of filemanager <subcommand>? I tried watch and fetch but neither work, so I am guessing only upload is supported at the moment.

Yeah, you're correct that `hs filemanager upload` is the only `filemanager` command implemented at this point.

PR #34, opened 2019-12-18
Added authType to API Key config and set order when writing config file

miketalley:


packages/cms-lib/lib/config.js
```diff
@@ -16,11 +16,33 @@ const {
...
+
+  return {
+    defaultPortal,
+    defaultMode,
+    httpTimeout,
+    allowsUsageTracking,
+    portals,
+    ...rest,
+  };
+};
```
corymartin: I was gonna bring up that key order isn't guaranteed, but it seems that may be more old knowledge I carry that is no longer valid.
https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/

PR #36, opened 2019-12-18
Removed brackets

kieranja:
Fix so yarn lint doesn't fail. Sorry! @miketalley.


gcorne:
@kieranja no worries. We had disabled GitHub Action status checks due to an issue with the HubSpot organization being able to run actions after their general release, so the lint violation wasn't caught before the merge. I have enabled that status check.

PR #41, opened 2019-12-18
Updates to walk and watch to fix issues with symlinks and hidden files

### LINKED ISSUE #39: Add files prefixed with `.` to the default ignore rules

### LINKED ISSUE #40: Walk method breaks when trying to process hidden files
The `processFiles` method in the walk module chokes when passed a list of files that includes hidden files/symlinks.

```
[Error: ENOENT: no such file or directory, stat '/Users/mtalley/HubSpot/Prod/hubspot-cms-tools/tmp/timeline-test-app/node_modules/node-releases/node_modules/.bin/semver'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/Users/mtalley/HubSpot/Prod/hubspot-cms-tools/tmp/timeline-test-app/node_modules/node-releases/node_modules/.bin/semver'
} /Users/mtalley/HubSpot/Prod/hubspot-cms-tools/tmp/timeline-test-app/node_modules/node-releases/node_modules/.bin/semver
/Users/mtalley/HubSpot/Prod/hubspot-cms-tools/packages/cms-lib/lib/walk.js:18
            if (stats.isDirectory()) {
                      ^

TypeError: Cannot read property 'isDirectory' of undefined
    at /Users/mtalley/HubSpot/Prod/hubspot-cms-tools/packages/cms-lib/lib/walk.js:18:23
    at FSReqCallback.oncomplete (fs.js:158:21)
error Command failed with exit code 1.
```

Talking to @gcorne it was suggested that we ignore the hidden files/symlinks here.

miketalley:
Fixes #39 
Fixes #40 

### What this PR does
- Skips symlinks when running walk on a dir
- Updates watch task so it ignores hidden files/folders
- Adds tests for ignoreRules
- Adds more rules to ignoreRules based on those in junk library(https://github.com/sindresorhus/junk/blob/master/index.js#L3)

### DIFF
--- a/packages/cms-lib/ignoreRules.js
+++ b/packages/cms-lib/ignoreRules.js
@@ -8,7 +8,7 @@ const findup = require('findup-sync');
 const ignoreList = [
   '**/node_modules', // dependencies
   '**/.*', // hidden files/folders
-  '**/npm-debug.log', // Error log for npm
+  '**/*.log', // Error log for npm
   '**/*.swp', // Swap file for vim state
 
   // # macOS
--- a/packages/cms-lib/ignoreRules.js
+++ b/packages/cms-lib/ignoreRules.js
@@ -3,7 +3,10 @@ const path = require('path');
 const ignore = require('ignore');
 const findup = require('findup-sync');
 
-const ignoreRules = ignore().add('node_modules');
+const hiddenFilesRegEx = /(^|\/)\.[^/.]/g;
+const ignoreRules = ignore()
+  .add('node_modules')
+  .add(hiddenFilesRegEx);
 
 let configPath = null;
 let loaded = false;
--- a/packages/cms-lib/ignoreRules.js
+++ b/packages/cms-lib/ignoreRules.js
@@ -3,10 +3,10 @@ const path = require('path');
 const ignore = require('ignore');
 const findup = require('findup-sync');
 
-const hiddenFilesRegEx = /(^|\/)\.[^/.]/g;
-const ignoreRules = ignore()
-  .add('node_modules')
-  .add(hiddenFilesRegEx);
+const ignoreRules = ignore().add([
+  'node_modules', // dependencies
+  '.*', // hidden files/folders
+]);
 
 let configPath = null;
 let loaded = false;
--- a/packages/cms-lib/ignoreRules.js
+++ b/packages/cms-lib/ignoreRules.js
@@ -3,10 +3,29 @@ const path = require('path');
 const ignore = require('ignore');
 const findup = require('findup-sync');
 
-const ignoreRules = ignore().add([
-  'node_modules', // dependencies
-  '.*', // hidden files/folders
-]);
+// Rules used by ignore do not allow RegEx -- we are basing most of this list
+// on the Junk library, which uses RegEx https://github.com/sindresorhus/junk/blob/master/index.js#L3
+const ignoreList = [
+  '**/node_modules', // dependencies
+  '**/.*', // hidden files/folders
+  '**/npm-debug.log', // Error log for npm
+  '**/*.swp', // Swap file for vim state
+
+  // # macOS
+  '**/Icon\\r', // Custom Finder icon: http://superuser.com/questions/298785/icon-file-on-os-x-desktop
+  '**/__MACOSX', // Resource fork
+
+  // # Linux
+  '**/~', // Backup file
+
+  // # Windows
+  '**/Thumbs.db', // Image file cache
+  '**/ehthumbs.db', // Folder config file
+  '**/Desktop.ini', // Stores custom folder attributes
+  '**/@eaDir', // Synology Diskstation "hidden" folder where the server stores thumbnails
+];
+
+const ignoreRules = ignore().add(ignoreList);
 
 let configPath = null;
 let loaded = false;
--- a/packages/cms-lib/lib/walk.js
+++ b/packages/cms-lib/lib/walk.js
@@ -1,7 +1,5 @@
 const fs = require('fs');
 const path = require('path');
-const { logger } = require('../logger');
-const { logFileSystemErrorInstance } = require('../errorHandlers');
 
 const STAT_TYPES = {
   FILE: 'file',
@@ -23,11 +21,11 @@ const generateFilePromise = (dir, file) => {
         });
       }
       if (stats.isDirectory()) {
-        walk(filepath).then(fileContents => {
+        walk(filepath).then(files => {
           return resolve({
             filepath,
             type: STAT_TYPES.DIRECTORY,
-            fileContents,
+            files,
           });
         });
       } else if (stats.isFile()) {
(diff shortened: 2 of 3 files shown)


gcorne:
Is there a reason we shouldn’t use `ignoreRules` so that ignoring hidden files happens everywhere?

packages/cms-lib/lib/walk.js
```diff
@@ -7,10 +7,13 @@ function walk(dir) {
       files.map(file => {
         return new Promise((resolve, reject) => {
           const filepath = path.join(dir, file);
-          fs.stat(filepath, (error, stats) => {
+          fs.lstat(filepath, (error, stats) => {
             if (error) {
               reject(error);
             }
+            if (stats.isSymbolicLink()) {
+              resolve(filepath);
```
gcorne: It looks like this only handles file links and will result in the linked file being uploaded. I am not sure that is what we want.

packages/cms-lib/ignoreRules.js
```diff
@@ -3,7 +3,10 @@ const path = require('path');
 const ignore = require('ignore');
 const findup = require('findup-sync');
 
-const ignoreRules = ignore().add('node_modules');
+const hiddenFilesRegEx = /(^|\/)\.[^/.]/g;
```
gcorne: I am pretty sure that this could be the string '.*' but might be worth adding a test or two to confirm.

packages/cms-lib/lib/walk.js
miketalley: added code to reject symlinks and also catch and log errors during walk process -- open to ideas for improving!

packages/cms-lib/lib/walk.js
```diff
@@ -1,22 +1,31 @@
...
+const { logger } = require('../logger');
 
 function walk(dir) {
   const processFiles = files => {
     return Promise.all(
       files.map(file => {
         return new Promise((resolve, reject) => {
           const filepath = path.join(dir, file);
-          fs.stat(filepath, (error, stats) => {
+          fs.lstat(filepath, (error, stats) => {
```
corymartin: FYI going forward, [`fs-extra`](https://www.npmjs.com/package/fs-extra) is already a project dep.  It promisifies most functions from `fs` and adds a some convenience functions.

packages/cms-lib/ignoreRules.js
```diff
@@ -3,7 +3,10 @@ const path = require('path');
 const ignore = require('ignore');
 const findup = require('findup-sync');
 
-const ignoreRules = ignore().add('node_modules');
+const ignoreRules = ignore().add([
+  'node_modules', // dependencies
+  '.*', // hidden files/folders
+]);
```
corymartin: Perhaps we should also use the [`junk`](https://www.npmjs.com/package/junk) package or something similar.
It blacklists some other things we don't care about https://github.com/sindresorhus/junk/blob/master/index.js#L3

miketalley: That is a useful package -- it looks like it only works with regex though. That may make it tricky to use with `ignore` because regex is not supported. :(

packages/cms-lib/lib/walk.js
miketalley: I switched over to using `fs-extra` and making everything sync, but then rolled back due to potential performance issues.

We are walking many files(potentially) and blocking the thread for each one could really slow things down. I think using `fs-extra` is a good idea going forward, but for now it's a simpler change without it.

packages/cms-lib/ignoreRules.js
gcorne: I think we can use it as a source of inspiration and add rules to our own system.

packages/cms-lib/lib/walk.js
```diff
@@ -1,38 +1,74 @@
...
-      })
-    );
+      } else if (stats.isFile()) {
+        resolve({
+          filepath,
+          type: STAT_TYPES.FILE,
+        });
+      }
+    });
+  }).catch(logger.error);
```
gcorne: I think we should remove the catch and logging here.

packages/cms-lib/lib/walk.js
```diff
@@ -1,38 +1,74 @@
...
+};
+
+const filesDataReducer = (all, fileData) => {
+  try {
```
gcorne: What kinds of exceptions do we expect to be thrown here?

packages/cms-lib/lib/walk.js
```diff
@@ -1,38 +1,74 @@
...
+          filepath,
+          type: STAT_TYPES.SYMBOLIC_LINK,
+        });
+      }
+      if (stats.isDirectory()) {
+        walk(filepath).then(fileContents => {
+          return resolve({
+            filepath,
+            type: STAT_TYPES.DIRECTORY,
+            fileContents,
```
gcorne: I don't think that `fileContents` is the best name here. I think either `files` or `folderContent` is more accurate.

packages/cms-lib/lib/walk.js
```diff
@@ -1,38 +1,74 @@
...
+        resolve({
+          filepath,
+          type: STAT_TYPES.FILE,
+        });
+      }
+    });
+  }).catch(logger.error);
+};
+
+const filesDataReducer = (all, fileData) => {
```
gcorne: I think `all` could be clearer maybe something like `allFiles`?

packages/cms-lib/lib/walk.js
miketalley: Good point, I think perhaps I left this in from a previous iteration. Will remove.

packages/cms-lib/ignoreRules.js
```diff
@@ -3,7 +3,29 @@ const path = require('path');
 const ignore = require('ignore');
 const findup = require('findup-sync');
 
-const ignoreRules = ignore().add('node_modules');
+// Rules used by ignore do not allow RegEx -- we are basing most of this list
+// on the Junk library, which uses RegEx https://github.com/sindresorhus/junk/blob/master/index.js#L3
+const ignoreList = [
+  '**/node_modules', // dependencies
+  '**/.*', // hidden files/folders
+  '**/npm-debug.log', // Error log for npm
```
gcorne: Let's change this to `**/*.log`

packages/cms-lib/lib/walk.js
```diff
@@ -1,38 +1,68 @@
...
+
+const filesDataReducer = (allFiles, fileData) => {
+  switch (fileData.type) {
+    case STAT_TYPES.FILE:
+      return allFiles.concat(fileData.filepath);
+    case STAT_TYPES.DIRECTORY:
+      return allFiles.concat(fileData.fileContents);
+    case STAT_TYPES.SYMBOLIC_LINK:
+      // Skip symlinks
+      return allFiles;
```
corymartin: Minor nit: this could fall-through to the default case.

PR #50, opened 2020-01-09
Added support for accessing function logs

miketalley:
### What this PR does

- Allows output of function logs to console or file(using --file)
  - Prompts user for path of function
  - Includes SUCCESS, HANDLED_ERROR, and UNHANDLED_ERROR statuses
  - Logs payload data for SUCCESS
  - Logs stackTrace for HANDLED_ERROR and UNHANDLED_ERROR 
- Allows output of all or most recent log only(using --latest)
- Handles invalid paths

Output to console
image

Latest log only(--latest)
image

Output to file(--file)
image
image

Logging of invalid path
image


.gitignore
```diff
@@ -9,4 +9,5 @@ npm-debug.log.*
 *.log
 .DS_Store
 coverage
+function.log
```
gcorne: Shouldn't this be covered by `*.log` above?

gcorne:
There a few things that I think we can adjust, but let's go ahead and merge this.

PR #52, opened 2020-01-15
Success message for `hs init` and various improvements to cleanup on exit and prompts

### LINKED ISSUE #43: Add an hs init success message
Tell the user they successfully set up their local authentication, and a hubspot.congif.yml file was created. 

@markhazlewood any suggestions on copy here?

### LINKED ISSUE #51: Improve clarity provided to the user with `hs init` prompts
Based on [this conversation about improving the prompts](https://hubspot.slack.com/archives/CQYN4AM3P/p1578940233005100).

The most confusing prompts seem to be `portalId` and `name`. We need to improve the wording and/or links provided at these prompts to improve the success rate of users when running the `hs init` command.

cc @williamspiro @TheWebTech @gcorne

### LINKED ISSUE #55: Make apikey top and default option in init
Per Lauren's research recommendation, in the `init` flow, we should have `apikey` be the top and default option.
Screen Shot 2020-01-21 at 8 46 34 AM

miketalley:
Fixes #43 
Fixes #51 
Fixes #55 

### What this PR does

- Adds a success message for both oauth2 and apiKey flows of `hs init` command
- Improves cleanup when exit/SIGINT is triggered
  - Closes express server(prevents orphaned process that needs to be killed)
  - Deletes empty config file
- Prevents portal name prompt from allowing spaces(`--portal=` flag does not accept spaces)
- Updates `authMethod` prompt to use 'rawlist' which allows both numerical and arrow key selection of choice
- Improves clarity of `portalId` and `name` prompts and makes validation wording consistent with them
- Makes apiKey the default auth method

### DIFF
--- a/packages/api-auth-lib/package.json
+++ b/packages/api-auth-lib/package.json
@@ -10,7 +10,6 @@
   "dependencies": {
     "express": "^4.16.3",
     "moment": "^2.24.0",
-    "node-cleanup": "^2.1.2",
     "open": "^6.4.0",
     "request": "^2.88.0",
     "request-promise-native": "^1.0.7"
--- a/packages/api-auth-lib/package.json
+++ b/packages/api-auth-lib/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "express": "^4.16.3",
     "moment": "^2.24.0",
+    "node-cleanup": "^2.1.2",
     "open": "^6.4.0",
     "request": "^2.88.0",
     "request-promise-native": "^1.0.7"
--- a/packages/cms-cli/commands/init.js
+++ b/packages/cms-cli/commands/init.js
@@ -1,3 +1,4 @@
+const nodeCleanup = require('node-cleanup');
 const { version } = require('../package.json');
 const {
   getConfigPath,
@@ -41,7 +42,7 @@ const AUTH_METHOD_FLOW = {
     },
     setup: async configData => {
       createEmptyConfigFile();
-      process.on('exit', deleteEmptyConfigFile);
+      handleExit();
       updateDefaultPortal(configData.name);
       updatePortalConfig({
         ...configData,
@@ -55,11 +56,18 @@ const AUTH_METHOD_FLOW = {
     },
     setup: async configData => {
       createEmptyConfigFile();
-      process.on('exit', deleteEmptyConfigFile);
+      handleExit();
       await authenticateWithOauth(configData);
       process.exit();
     },
   },
+};
+
+const handleExit = () => {
+  nodeCleanup(() => {
+    deleteEmptyConfigFile();
+    return true;
+  });
 };
 
 const completeConfigSetup = async ({ authMethod, configPath }) => {
--- a/packages/cms-cli/commands/init.js
+++ b/packages/cms-cli/commands/init.js
@@ -1,4 +1,3 @@
-const nodeCleanup = require('node-cleanup');
 const { version } = require('../package.json');
 const {
   getConfigPath,
@@ -70,10 +69,8 @@ const AUTH_METHOD_FLOW = {
 };
 
 const handleExit = () => {
-  nodeCleanup(() => {
-    deleteEmptyConfigFile();
-    return true;
-  });
+  process.on('exit', deleteEmptyConfigFile);
+  process.on('SIGINT', deleteEmptyConfigFile);
 };
 
 const completeConfigSetup = async ({ authMethod, configPath }) => {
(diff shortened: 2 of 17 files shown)


williamspiro:
Thanks. So to confirm, the message would look something like:

>Success: hubspot.config.yml created with apikey.

cc @markhazlewood @iamw11y

miketalley:
> Thanks. So to confirm, the message would look something like:
> 
> > Success: hubspot.config.yml created with apikey.
> 
> cc @markhazlewood @iamw11y

It would be either:
> Success: hubspot.config.yml created with API Key.

or

> Success: hubspot.config.yml created with Oauth2.

packages/api-auth-lib/package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "express": "^4.16.3",
     "moment": "^2.24.0",
+    "node-cleanup": "^2.1.2",
```
gcorne: I'd rather not introduce a dep that is basically unmaintained (last commit 2017). Do we need it? Could we do what we need ourselves? See https://github.com/jtlapp/node-cleanup/issues/16

miketalley: I switched to doing manual cleanup and removed the package.

packages/cms-cli/commands/init.js
```diff
@@ -55,13 +58,21 @@ const AUTH_METHOD_FLOW = {
...
+      );
       process.exit();
     },
   },
 };
 
+const handleExit = () => {
+  process.on('exit', deleteEmptyConfigFile);
+  process.on('SIGINT', deleteEmptyConfigFile);
+};
```
corymartin: Isn't there a hoisting issue with this down here?

miketalley: It works because it's called within a function that is the value of an object property that is declared above this code. I can move it up to avoid any confusion though.

packages/cms-cli/lib/prompts.js
```diff
@@ -63,17 +67,17 @@ const PORTAL_API_KEY = {
   message: 'Enter the API key for your portal:',
```
williamspiro: Should we align this with "found under the DOMAIN column at https://app.hubspot.com/myaccounts-beta" and add something like:

>(found at https://app.hubspot.com/l/api-key)

cc @iamw11y @markhazlewood

iamw11y: We should make sure these messages are consistent, for sure. Rather than tackling this piecemeal, should I start a reference doc that lists all of the command/response possibilities in the CLI, with both success and error responses? This would save us from having to comb through `.JS` files for all of the possibilities and could live in a similar space as the glossary that I'm putting together, where it would function as another internal reference to ensure that our docs and interfaces are informative and consistent.

packages/cms-cli/lib/prompts.js
```diff
@@ -41,18 +42,21 @@ const CLIENT_SECRET = {
...
     }
     return true;
   },
 };
 
 const PORTAL_NAME = {
   name: 'name',
-  message: 'Enter a name for your portal:',
+  message:
+    'Enter a unique nickname that will be used to reference this account:',
```
williamspiro: https://github.com/HubSpot/hubspot-cms-tools/issues/51#issuecomment-577310264

cc @erincouse

williamspiro:
I think it would be good to get this merged and into a beta release soon so we can test the entire flow out.

miketalley:
@williamspiro After speaking with @iamw11y we are going to add some final touches to the prompts and then work on getting this merged so we can test.

packages/cms-cli/lib/prompts.js
gcorne: Are we confident about how https://app.hubspot.com/myaccounts-beta works when a user has a single portal?

miketalley: I haven't tested this case, is there a way to create a test user to confirm this works?

williamspiro: This does work fine when a user is un one account, however, I do still think it is a little awkward given when you try and copy the HID you get taken into the account, and that it differs from what we say in https://designers.hubspot.com/guides/getting-started#quick_start

miketalley: I agree that this is not ideal, however, I'm not sure there is a simple solution at the moment. What about adding the ability to easily copy the HID in the account dropdown in the top right -- a small copy icon button perhaps?

iamw11y: > it differs from what we say in https://designers.hubspot.com/guides/getting-started#quick_start

The downside to this link is that it doesn't cover how to switch portals to get the ID for a different portal if the user has several, which isn't super difficult, but could still lead to confusion. The "myaccounts-beta" link at least shows all portals and their associated portal IDs.

PR #62, opened 2020-01-26
Add basic support for authenticating via a user token

gcorne:
TODO:

- [x] Add some test coverage
- [x] Add support in validation

@corymartin @miketalley


gcorne:
@gchomatas here is the CLI side.

PR #65, opened 2020-01-28
Added notify param to watch command

### LINKED ISSUE #2: Add support for live reload tools when watching a directory
Let's add `--notify=/path/to-file` and consider adding some queuing to the file watcher so that we can handle situations where a bunch of files change all at once like when a user switches `git` branches.

miketalley:
Adds notify param to watch command which accepts a path to a file that will have logs written to it each time changes are detected within the watched directory.

This allows a dev user to watch the file that they specify so they can automate tasks when the watch task within the hubspot-cms-tools takes an action(such as uploading or removing remote files). It is particularly useful for using automatic reload on the page that the changes will be reflected on using a tool like [browser-sync](https://browsersync.io/) or [livereload](http://livereload.com/).

  - The logs to the `--notify` target file get written in batches to allow watches to this file to be triggered once, after all of the tasks complete. This is good for actions that make changes to multiple files very quickly(such as switching branches).
    - This required adding [debounce](https://github.com/component/debounce)
    - The logic here is when watch is triggered(add/change/unlink) the promises from these tasks will get added to an array and a debounced function to write the output to disk is called, when the debounced method finally executes(after all remote actions have started but perhaps not finished) a Promise.all will be created on the outstanding promises, which when resolved will write to the `--notify` target file
  - Ignores notify files if they are within the specified watched directory(prevents infinite loop)

  ### Usage example
Run watch command:
```
hs watch tmp/testing/cms-theme-boilerplate boilerplate --notify=tmp/notify.log --portalId=7149391
```
Run browser-sync:
```
browser-sync start -f /tmp/notify.log --proxy http://mtalley-7149391.hs-sites.com/homepage
```

**Note:** I filed a bug that I found in browser-sync because the [--proxy flag does not recognize more than one URL parameter](https://github.com/BrowserSync/browser-sync/issues/1744) -- this prevents us from using a proxy URL of something like...
```
https://preview.hs-sites.com/_hcms/preview/template/multi?portalId=7149391&template_file_path=boilerplate/src/templates/homepage.html
```

image

Fixes #2

### DIFF
--- a/packages/cms-lib/lib/logs.js
+++ b/packages/cms-lib/lib/logs.js
@@ -3,6 +3,7 @@ const util = require('util');
 const path = require('path');
 const moment = require('moment');
 const { logger } = require('../logger');
+const { TIMESTAMP_FORMAT } = require('./constants');
 
 const FUNCTION_LOG_PATH = 'function.log';
 
@@ -34,7 +35,7 @@ const formatStackTrace = log => {
 };
 
 const formatTimestamp = log => {
-  return `[${moment(log.createdAt).format('MM/DD/YYYY hh:mm:ss a')}]`;
+  return `[${moment(log.createdAt).format(TIMESTAMP_FORMAT)}]`;
 };
 
 const formatPayload = log => {
--- a/packages/cms-lib/lib/logs.js
+++ b/packages/cms-lib/lib/logs.js
@@ -3,7 +3,6 @@ const util = require('util');
 const path = require('path');
 const moment = require('moment');
 const { logger } = require('../logger');
-const { TIMESTAMP_FORMAT } = require('./constants');
 
 const FUNCTION_LOG_PATH = 'function.log';
 
@@ -35,7 +34,7 @@ const formatStackTrace = log => {
 };
 
 const formatTimestamp = log => {
-  return `[${moment(log.createdAt).format(TIMESTAMP_FORMAT)}]`;
+  return `[${moment(log.createdAt).toISOString()}]`;
 };
 
 const formatPayload = log => {
--- a/packages/cms-lib/lib/notify.js
+++ b/packages/cms-lib/lib/notify.js
@@ -41,10 +41,9 @@ function waitForActionsToCompleteAndWriteQueueToFile(filePathToNotify) {
     notifyPromises.splice(0, notifyPromises.length)
   );
 
-  allNotifyPromisesResolution.then(() => {
-    const notifyOutput = `${moment().toISOString()} Notify Triggered\n`;
-    notifyFilePath(filePathToNotify, actionOutput.concat(notifyOutput));
-  });
+  allNotifyPromisesResolution.then(
+    notifyFilePath(filePathToNotify, actionOutput)
+  );
 }
 
 /**
--- a/packages/cms-lib/lib/notify.js
+++ b/packages/cms-lib/lib/notify.js
@@ -22,13 +22,10 @@ function waitForActionsToCompleteAndWriteQueueToFile(filePathToNotify) {
   const actionOutput = notifyQueue.join('');
   const allNotifyPromisesResolution = Promise.all(notifyPromises);
 
-  console.log(notifyPromises.length, notifyQueue.length);
-
   notifyPromises.length = 0;
   notifyQueue.length = 0;
 
   allNotifyPromisesResolution.then(() => {
-    console.log(notifyPromises.length, notifyQueue.length);
     const notifyOutput = `${moment().toISOString()} Notify Triggered\n`;
     notifyFilePath(filePathToNotify, actionOutput.concat(notifyOutput));
   });
--- a/packages/cms-lib/lib/notify.js
+++ b/packages/cms-lib/lib/notify.js
@@ -10,6 +10,16 @@ const debouncedWaitForActionsToCompleteAndWriteQueueToFile = debounce(
   500
 );
 
+/**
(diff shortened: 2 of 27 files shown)


packages/cms-lib/lib/logs.js
```diff
@@ -34,7 +35,7 @@ const formatStackTrace = log => {
 };
 
 const formatTimestamp = log => {
-  return `[${moment(log.createdAt).format('MM/DD/YYYY hh:mm:ss a')}]`;
+  return `[${moment(log.createdAt).format(TIMESTAMP_FORMAT)}]`;
```
corymartin: Since we don't know if users may be remotely reading/writing these files, shouldn't we be using  
`moment().toISOString()`
for portability and ISO8601?

miketalley: This is a good point. I will update both the logs and watch task to log datetime as ISOString.

packages/cms-lib/lib/watch.js
```diff
@@ -56,9 +98,43 @@ function uploadFile(portalId, file, dest, { mode, cwd }) {
   });
 }
 
-function watch(portalId, src, dest, { mode, cwd, remove, disableInitial }) {
+async function deleteRemotePath(portalId, filePath, remotePath, { cwd }) {
```
gcorne: I think that the name of this function should have "file" in its name since it only applies to files and not directories.

packages/cms-lib/lib/watch.js
```diff
@@ -9,16 +12,54 @@ const {
...
+    ACTION: 'Removed',
+  },
+  CHANGE: {
+    EVENT: 'change',
+    ACTION: 'Changed',
+  },
+};
+
+const notifyQueue = [];
+const debouncedWriteNotifyQueueToFile = debounce(writeNotifyQueueToFile, 1500);
```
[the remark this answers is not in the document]
gcorne: It might be worth looking at https://github.com/sindresorhus/promise-fun for some inspiration.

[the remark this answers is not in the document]
miketalley: Thanks for the link, I bookmarked that one, but ended up coming up with my own solution(although I'd imagine there is a similar pattern in that repo).

packages/cms-lib/lib/uploadFolder.js
```diff
@@ -13,19 +13,31 @@ const {
...
+
 /**
  *
  * @param {number} portalId
  * @param {string} src
  * @param {string} dest
  * @param {object} options
  */
-async function uploadFolder(portalId, src, dest, { mode, cwd }) {
+async function uploadFolder(portalId, src, dest, { mode, cwd, notify }) {
```
gcorne: What happens if `notify` is not passed?

miketalley: There is no file to notify, so the watch works normally without notifying anything

gcorne: That makes sense. I misread how it was working. I do think that instead of adding `notify` support to `uploadFolder` we should either leverage the `uploadFolder` promise that is returned and do something like `uploadFolder.then(notify)` or not notify after the initial upload. Given that the use case is to live reload browser window(s) after changes are saved and uploaded while developing, I don't think there is much value in notifying after the initial upload, so I think we should remove it.

miketalley: I thought about this as we would miss the case where someone is uploading new files for the first time. Should we ignore that case in favor of not having a refresh in cases where the files are not new?

gcorne: > Should we ignore that case in favor of not having a refresh in cases where the files are not new?
Yeah, I think that is fine.

packages/cms-lib/lib/notify.js
```diff
@@ -0,0 +1,68 @@
...
+ * @param {string} filePathToNotify
+ */
+function waitForActionsToCompleteAndWriteQueueToFile(filePathToNotify) {
+  const actionOutput = notifyQueue.splice(0, notifyQueue.length).join('');
+  const allNotifyPromisesResolution = Promise.all(
+    notifyPromises.splice(0, notifyPromises.length)
+  );
+
+  allNotifyPromisesResolution.then(() => {
+    const notifyOutput = `${moment().toISOString()} Notify Triggered\n`;
```
corymartin: I don't know that this is that useful. It's a bit redundant in that any write to the notify log means that it was triggered.

miketalley: I removed this, that is a good point. More useful for testing

packages/cms-lib/lib/uploadFolder.js
miketalley: After testing out the flow a bit with browser-sync, I tend to agree with your comment about not notifying after the initial upload, so I removed it.

packages/cms-lib/lib/watch.js
```diff
@@ -71,7 +107,7 @@ function watch(portalId, src, dest, { mode, cwd, remove, disableInitial }) {
 
   if (!disableInitial) {
     // Use uploadFolder so that failures of initial upload are retried
-    uploadFolder(portalId, src, dest, { mode, cwd }).then(() => {
+    uploadFolder(portalId, src, dest, { mode, cwd, notify }).then(() => {
```
gcorne: Looks like `notify` should be removed now.

PR #66, opened 2020-01-29
Improved server error console messaging

### LINKED ISSUE #45: Improve server error console messaging
```
[ERROR] Uploading file "/Users/wspiro/Desktop/local-cms-dev/...." failed
[ERROR] Unable to upload "/Users/wspiro/Desktop/local-cms-dev/...". The post to "dev-website..." on portal 6906341 failed due to a server error. internal error
```

We should make this error contain more useful information for users, noting to try again and reach out to support if the issue persists. 

I also think "... on portal" should be _in portal_? 

@iamw11y

miketalley:
- Removed [prettier](https://github.com/prettier/prettier) dep from cms-lib in favor of using the [root version](https://github.com/HubSpot/hubspot-cms-tools/blob/00ada4caaea9e98cb791b06630c8133cb536a49c/package.json#L12) to prevent conflicts in pre-commit hook
- Updated error text to be more helpful, providing a link to help.hubspot.com
- Fixed tests

Fixes #45


packages/cms-lib/errorHandlers.js
```diff
@@ -181,7 +183,7 @@ function logApiStatusCodeError(error, context) {
     const request = context.request
       ? `${action} ${preposition} "${context.request}"`
       : action;
-    messageDetail = `${request} on portal ${context.portalId}`;
+    messageDetail = `${request} in portal ${context.portalId}`;
```
corymartin: Didn't someone (@williamspiro @gcorne) propose we start subbing *"account"* for *"portal"*

miketalley: Yes, I believe that is the plan, but not _just yet_ AFAIK.

gcorne: Yeah, I think we should tackle that separately. This reminded me to file #68.

gcorne:
@miketalley can we remove `package-lock.json` from the PR?

packages/cms-lib/errorHandlers.js
```diff
@@ -14,6 +14,8 @@ const isApiUploadValidationError = err =>
 const isSystemError = err =>
   err.errno != null && err.code != null && err.syscall != null;
 const isFatalError = err => err instanceof HubSpotAuthError;
+const contactSupportString =
+  'Please try again or visit https://help.hubspot.com/ to submit a ticket or contact HubSpot Support if the issue persists.';
```
gcorne: I think we should recommend contacting support only if the issue persists. @iamw11y any other suggestions when it comes to copy?

miketalley: I copied this directly from @iamw11y 's comment https://github.com/HubSpot/hubspot-cms-tools/issues/45#issuecomment-579284315

gcorne: Ah, cool. Sorry for the noise.

PR #67, opened 2020-01-29
Decrease likelihood of dependencies being uploaded after dependents

gcorne:
When uploading a folder of files, it is possible for dependencies to be uploaded before files that depend on them. An example is a CSS file that references a raw asset or a template that uses a module. This decreases the likelihood that occurs through uploading file types that are often
dependencies first. This is not a perfect solution and something that should be revisited through a bulk upload endpoint that can validate after receiving all assets.

TODO:

- [x] Add some tests

@mattcoley @miketalley @corymartin

### DIFF
--- a/packages/cms-lib/lib/uploadFolder.js
+++ b/packages/cms-lib/lib/uploadFolder.js
@@ -7,7 +7,12 @@ const { upload } = require('../api/fileMapper');
 const { createIgnoreFilter } = require('../ignoreRules');
 const { walk } = require('./walk');
 const escapeRegExp = require('./escapeRegExp');
-const { convertToUnixPath, isAllowedExtension } = require('../path');
+const {
+  convertToUnixPath,
+  isAllowedExtension,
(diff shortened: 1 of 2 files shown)


packages/cms-lib/lib/uploadFolder.js
```diff
@@ -41,38 +70,45 @@ async function uploadFolder(portalId, src, dest, { mode, cwd }) {
...
+        }
+        failures.push({
+          file,
+          destPath,
+        });
+      }
+    };
+  };
+
+  for (let i = 0; i < filesByType.length; i++) {
```
gcorne: fyi, the old school for loop is necessary due to async/await.

miketalley: do you think it's worth adding this comment to the code?

gcorne: a comment wouldn't hurt so added it.

PR #69, opened 2020-01-30
Limit concurrent refresh requests and fix access token expiration

gcorne:
TODO:

- [x] Add test coverage

@miketalley @gchomatas


gcorne:
Going to merge and leave adding more test coverage to a separate PR so that we can do some testing.

PR #77, opened 2020-02-04
Personal access key auth

### LINKED ISSUE #71: Adjust init command to switch to the personal CMS access key flow

miketalley:
Fixes #71 

- Makes `personalaccesskey` available as authType for `hs init` using flag `--personal-access-key`
- Adds ability to use `hs auth personalaccesskey` command to add portals using `authType: personalaccesskey`

Prompt example:
image

**Notes:**
- **This is a breaking change!** Because of the rename of `usertoken` to `personalaccesskey`, all previous portals using `usertoken` will break when updating to the version containing these changes. This can be fixed manually by updating `usertoken` to `personalaccesskey` and `userToken` to `personalAccessKey` in the `hubspot.config.yml` file.
- Moves logic from `/http/userToken.js` into `/personalAccessKey.js` for consistency
- Adds handleExit helper method and updates uses in multiple locations
- Makes all values for `authType` in config lowercased(`oauth2`/`apikey`/`personalaccesskey`)


packages/cms-cli/commands/init.js
```diff
@@ -70,6 +74,30 @@ const AUTH_METHOD_FLOW = {
       process.exit();
     },
   },
+  [USER_TOKEN_AUTH_METHOD.value]: {
+    prompt: async () => {
+      const firstPrompt = await promptUser(USER_TOKEN_FLOW);
+      open(`https://local.hubspot.com/user-token/${firstPrompt.portalId}`);
```
gcorne: I think what we want to do here is leverage a shotlink and then only prompt for the user token. Once we have the user token, we can look up the portal when fetching the access token.

miketalley: I'm not sure that I'm following how we will get them to the user-token-ui with the correct portal without asking them for the portalId.

packages/cms-lib/userToken.js
```diff
@@ -0,0 +1,164 @@
...
+const { logger } = require('./logger');
+const { fetchAccessToken } = require('./api/localDevAuth');
+
+const refreshRequests = new Map();
+
+function getRefreshKeyForUserToken(userToken, expiration) {
+  return `${userToken}-${expiration || 'fresh'}`;
+}
+
+async function getAccessToken(userToken, env = 'QA') {
```
miketalley: some of this logic was merged in from /http/userToken

packages/cms-lib/userToken.js
```diff
@@ -0,0 +1,164 @@
...
+      userToken,
+      authTokenInfo && authTokenInfo.expiresAt,
+      env
+    );
+  }
+
+  return auth.tokenInfo.accessToken;
+}
+
+/**
```
miketalley: Here down is new.

packages/cms-cli/commands/init.js
```diff
@@ -93,20 +96,36 @@ function initializeConfigCommand(program) {
...
       `initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} for a HubSpot portal`
     )
+    .option(
+      `--${AUTH_METHODS.oauth.value}`,
+      `use ${AUTH_METHODS.oauth.name} for authType`
+    )
+    .option(
+      `--${AUTH_METHODS.api.value}`,
+      `use ${AUTH_METHODS.api.name} for authType`
+    )
```
gcorne: Let's start with making `usertoken` an optional flow and keep the existing initialization flow as the default.

packages/cms-cli/hubspot.sample.config.yml
```diff
@@ -32,3 +32,13 @@ portals:
         refreshToken: 84d22710-4cb7-5581-ba05-35f9945e5e8e
         accessToken: >-
           CJDVnLanLRICEQIYyLu8LyDh9E4opf1GMhkAxGuU5XN_O2O2QhX0khw7cwIkPkBRHye-OfIADgLBAAADAIADAAAAAAAAAAJCGQC8a5TlhtSU8T-2mVLxOBpxS18aM42oGKk
+  - name: QA
```
gcorne: I don't think we want to add another entry into the sample config. When we're ready to switch over to usertoken being the default, we can update the sample config and only user `usertoken` auth types.

packages/cms-cli/lib/__tests__/validation.js
```diff
@@ -136,7 +136,7 @@ describe('validation', () => {
       getPortalId.mockReturnValueOnce(123);
       getPortalConfig.mockReturnValueOnce({
         portalId: 123,
-        authType: 'apiKey',
+        authType: 'apikey',
```
gcorne: Thanks for fixing.

packages/cms-lib/userToken.js
```diff
@@ -0,0 +1,164 @@
...
+  };
+};
+
+/**
+ * Adds a portal to the config using authType: userToken
+ *
+ * @param {object} promptData inquirer prompt object containing userToken and name properties
+ * @param {boolean} makeDefault option to make the portal being added to the config the default portal
+ */
+const updateConfigWithUserTokenPromptData = async (promptData, makeDefault) => {
```
gcorne: I think that it would be better to generalize this a bit instead of couple it with `promptData`. What if this was just `updateConfigWithUserToken()` and expect a particular object shape?

packages/cms-lib/userToken.js
gcorne: It looks like there were changes above. Is that not correct?

miketalley: Yeah, this comment is outdated, you can ignore it. Will delete...

williamspiro:
Couple questions:
1. Jeff A. is working on making a video going through this flow, and is waiting until we get the user token flow complete to record. Will the `--user-token` option stick around even once it is default? Or should he wait until this is the default in `hs init`. Just something to consider as we work on this, he needs some lead time on the video. 
2. Is there a reason or assumption we are testing going down the route of opening up the user token UI for the user, as opposed to telling them what to do (what URL to go to)? Not saying this route is wrong, but it is a new complexity, compared to the existing experience/path, which has been working, where we instruct the user where to go. I ask, once again, for Jeff's sake, but also as a reminder we only have 1 round of user testing for this new flow and it adds a new variable. Either way, excited to try out this opening the user token UI for user's flow!

packages/cms-cli/commands/auth.js
```diff
@@ -29,16 +40,25 @@ async function authAction(type, options) {
...
   trackCommandUsage(COMMAND_NAME);
-  await authenticateWithOauth(configData);
+  let configData;
+  switch (authType) {
+    case OAUTH_AUTH_METHOD.value:
+      configData = await promptUser(OAUTH_FLOW);
+      await authenticateWithOauth(configData);
+      break;
+    case USER_TOKEN_AUTH_METHOD.value:
+      configData = await userTokenPrompt();
```
gcorne: It looks like this flow assumes that the user is adding a new portal and not modifying an existing entry.

miketalley:
> Couple questions:
> 
> 1. Jeff A. is working on making a video going through this flow, and is waiting until we get the user token flow complete to record. Will the `--user-token` option stick around even once it is default? Or should he wait until this is the default in `hs init`. Just something to consider as we work on this, he needs some lead time on the video.
> 2. Is there a reason or assumption we are testing going down the route of opening up the user token UI for the user, as opposed to telling them what to do (what URL to go to)? Not saying this route is wrong, but it is a new complexity, compared to the existing experience/path, which has been working, where we instruct the user where to go. I ask, once again, for Jeff's sake, but also as a reminder we only have 1 round of user testing for this new flow and it adds a new variable. Either way, excited to try out this opening the user token UI for user's flow!

@williamspiro AFAIK we are testing using the `--personal-access-key` flag(renamed yesterday from `--user-token` based on [this slack conversation](https://hubspot.slack.com/archives/CDW3REULW/p1581439008190200)) but this will become the default once we are comfortable with the testing. @gcorne LMK if I am incorrect on this

As for question #2 -- I had set up the personal access key flow based on the oauth2 flow, which opens the browser for a user to accept giving access to the app, and for simplicity for the user(helping them not need to copy the url, open their browser, and paste the url in manually). 

I'm open to changing it if we decide that is not the best option.

packages/cms-cli/commands/auth.js
miketalley: Looking at the code in master it seems like that is how it currently works. Do we want to allow this to also be used for updating? I thought that was going to be `hs config`'s job?

gcorne: I think that this must have changed at some point.

williamspiro:
👍 sounds good, thanks for the info @miketalley! Excited to get this out

packages/cms-cli/commands/auth.js
miketalley: I'm working on this in a separate PR https://github.com/HubSpot/hubspot-cms-tools/pull/87

PR #79, opened 2020-02-05
Update watch.js to watch and respond to deleted folders.

gkemp94:
Hi All, 
George from HubSpot web team, loving experimenting with these new tools, one thing I noticed was that deleting folders weren't reflected in the design manager. which left some empty folders lying around when reorganizing within the local environment. 

This seemed like an easy and quick fix, tested it locally for this use case but not sure if there are other use cases that this might conflict with.

Feel free to let me know what you think! :D 

Cheers,
George

### DIFF
--- a/packages/cms-lib/lib/watch.js
+++ b/packages/cms-lib/lib/watch.js
@@ -88,7 +88,7 @@ function watch(portalId, src, dest, { mode, cwd, remove, disableInitial }) {
   });
 
   if (remove) {
-    watcher.on('unlink', filePath => {
+    const deleteFileOrFolder = filePath => {
       const remotePath = getDesignManagerPath(filePath);
 
       if (shouldIgnoreFile(filePath, cwd)) {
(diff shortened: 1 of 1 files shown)


packages/cms-lib/lib/watch.js
```diff
@@ -113,7 +113,10 @@ function watch(portalId, src, dest, { mode, cwd, remove, disableInitial }) {
             );
           });
       });
-    });
+    };
+    
+    watcher.on('unlink', deleteFileOrFolder);
+    watcher.on('unlinkDir', deleteFileOrFolder);
```
corymartin: Thanks @gkemp94!

Code looks good and your branch is running as expected.

One improvement: we can still log a specific statement as to whether a file or a folder was deleted via partial application or a closure.

Something like this should work

```javascript
const deleteFileOrFolder = type => filePath => {
  /*****/
  logger.log('Deleted %s "%s"', type, remotePath);
  /*****/
}

watcher.on('unlink', deleteFileOrFolder('file'));
watcher.on('unlinkDir', deleteFileOrFolder('folder'));
```

gkemp94: Great Suggestion! That's been updated!

PR #87, opened 2020-02-18
Config refactor to allow auth to handle existing data

miketalley:
This solves an [issue that @gcorne raised with auth not handling overwriting existing portal config data](https://github.com/HubSpot/hubspot-cms-tools/pull/77#discussion_r377937837), however, many of these changes will also set the stage for work TBD on `hs config` as well as allowing the [use of environment variables in place of a config file](https://github.com/HubSpot/hubspot-cms-tools/issues/4).

- Breaks `cms-lib/lib/config.js` into separate files within `cms-lib/lib/config/`
- Breaks `cms-lib/lib/constants.js` into separate files within `cms-lib/lib/constants/`
- Simplifies logic to `updatePortalConfig` by separating logic for `apikey`, `oauth2`, and `personalaccesskey` into their own methods/files
- Moves name prompt for `personalAccessKey` to the end and only prompts if no existing name is found
- Reconciles the shape of the config data with the `toObj` method of OAuth2Manager
- Adds tests for config and auth command
- Adds `--qa` flag to `hs init` and `hs auth` command to support creating/adding qa portals
  - Adds helpers for obtaining correct domain based on env
- Adds `ENVIRONMENTS` constant for `QA` and `PROD`
- Adds helpers for getting different versions of the env


packages/api-auth-lib/OAuth2Manager.js
```diff
@@ -5,6 +5,12 @@ const open = require('open');
 
 const { HubSpotAuthError } = require('./Errors');
 const { handleExit } = require('@hubspot/cms-lib/lib/process');
+const {
+  getEnv,
+  getEnvForConfig,
+  getHubSpotWebsiteDomain,
+  getHubSpotApiDomain,
+} = require('@hubspot/cms-lib/lib/environment');
```
gcorne: Sorry for not noticing this when the `handleExit` change was made, but this creates a circular dependency relationship between `@hubspot/cms-lib` and `@hubspot/api-auth-lib`, which I am a bit surprised hasn't caused issues.

packages/cms-lib/lib/config/index.js
```diff
@@ -0,0 +1,36 @@
+const {
+  createEmptyConfigFile,
+  deleteEmptyConfigFile,
+  getAndLoadConfigIfNeeded,
+  getConfig,
+  getConfigPath,
+  isTrackingAllowed,
```
gcorne: See `isTrackingAllowed` in particular feels out of place when it comes to the `config/file`.

packages/cms-lib/personalAccessKey.js
```diff
@@ -50,15 +51,13 @@ async function getAccessToken(personalAccessKey, env = 'PROD') {
...
-async function refreshAccessToken(personalAccessKey, env = 'PROD') {
+async function refreshAccessToken(personalAccessKey, env = PROD) {
   const { accessToken, expiresAt, portalId } = await getAccessToken(
     personalAccessKey,
     env
   );
-  const config = getPortalConfig(portalId);
 
   updatePortalConfig({
-    ...config,
```
gcorne: It looks like after the changes to `updatePortalConfig` this call will not work as expected due to `tokenInfo` being a top-level key.

packages/cms-lib/lib/config/oauth2.js
```diff
@@ -0,0 +1,42 @@
...
+    portalId: configUpdates.portalId,
+    name: configUpdates.name || portalConfig.name,
+    env: configUpdates.env || portalConfig.env,
+    defaultMode: configUpdates.defaultMode || portalConfig.defaultMode,
+    authType: OAUTH_AUTH_METHOD.value,
+    auth,
+  };
+
+  delete config.apiKey;
+  delete config.personalAccessKey;
```
gcorne: Due to how the individual authType update functions are called from `updatePortalConfig`, I don't think we should have destructive operations like this. I think our philosophy when it comes to updating the config is that we should only do destructive operations when a user is going through a prompted flow like `hs auth` or `hs config` (when added).

packages/cms-lib/lib/config/portal.js
```diff
@@ -0,0 +1,160 @@
...
+ * @param {string} configOptions.personalAccessKey Personal Access Key used in authType: personalaccesskey
+ * @param {object} configOptions.auth Auth object used in oauth2 and personalaccesskey authTypes
+ * @param {string} configOptions.auth.clientId Client ID used for oauth2
+ * @param {string} configOptions.auth.clientSecret Client Secret used for oauth2
+ * @param {array} configOptions.auth.scopes Scopes that are allowed access with auth
+ * @param {object} configOptions.auth.tokenInfo Token Info used for oauth2 and personalaccesskey authTypes
+ * @param {string} configOptions.auth.tokenInfo.accessToken Access token used for auth
+ * @param {string} configOptions.auth.tokenInfo.expiresAt Date ISO of accessToken expiration
+ */
+const updatePortalConfig = configOptions => {
```
gcorne: It looks like the shape expected for `configOptions` has changed, but all of the calls to `updatePortalConfig` have not been updated.

miketalley: Here is the prior shape and changes made
```
const {	
    portalId,	
    authType,	
    environment,	// Renamed to env
    clientId,	        // Moved into auth prop
    clientSecret,	// Moved into auth prop
    scopes,	        // Moved into auth prop
    tokenInfo,	        // Moved into auth prop
    defaultMode,	
    name,	
    apiKey,	
    personalAccessKey,	
  } = configOptions;
```

For the items that were moved into the `auth` prop, they were coming from the `OAuth2Manager.toObj` and `personalAccessKey.updateConfigWithPersonalAccessKey` methods, which I updated with proper nesting of these props under `auth`.

I think this is a good change because it makes the shape consistent instead of passing all top level prop data to updatePortalConfig and then relying on on this method to properly nest 3 different shapes of data.

gcorne: When I was reading through the code there were calls that weren’t updated. I think that they were related to refreshing access tokens for both access keys and standard OAuth.

packages/cms-lib/lib/config/oauth2.js
miketalley: These methods are only accessed from `hs init` and `hs auth`.

The reason for the destruction(`delete`) is to remove any props that get passed in through the existing `portalConfig` and are not necessary for the new `authType` `portalConfig data shape. For example, when using `authType: apikey` and converting to `authType: personalaccesskey` the `apiKey` prop is no longer wanted, so we discard it.

Only the possible leftover keys from other portalConfig data shapes are explicitly deleted.

This allows the following `apikey` data shape...
image
...to be changed properly to `personalaccesskey` by running `hs auth personalaccesskey`...
image

My goal here was to create pure functions for `apikey`, `oauth2`, and `personalaccesskey` that take in the existing `portalConfig` as well as the desired `configUpdates` and return a new object with only valid properties for that `authType`, however, allowing as few or as many properties to be updated by passing them as `configUpdates`.

packages/cms-lib/lib/config/index.js
miketalley: I agree. I will move it somewhere more appropriate.

packages/cms-lib/lib/config/portal.js
miketalley: AH sorry, you are correct. I missed those two -- will fix.

packages/cms-lib/lib/config/oauth2.js
gcorne: > These methods are only accessed from hs init and hs auth.

Because the specialized updaters are called from `updatePortalConfig` based on the `authType`, this is not correct. This means that if for some reason I edit the config file to temporarily switch from using a personalaccesskey to using an api key that `personalAccessKey` will be blown away. I don't now how often that will happen in the real world, but it something that I do quite regularly when testing stuff.

miketalley: Okay I'll remove the `delete` calls to prevent the superfluous props from getting removed

packages/api-auth-lib/OAuth2Manager.js
miketalley: As per our conversation, I added these methods to api-auth-lib to avoid any potential cyclic dep

packages/cms-cli/lib/commonOpts.js
```diff
@@ -24,6 +24,10 @@ const addOverwriteOptions = program => {
   program.option('--overwrite', 'overwrite existing files', false);
 };
 
+const addTestingOptions = program => {
+  program.option('--qa', 'use QA environment for command', false);
+};
```
gcorne: Can we suppress this from the help since it is only relevant for internal testing?

miketalley:
> I am getting the follow error when testing `hs init --personal-access-key`
> 
> ```
> [ERROR] An error occurred while accessing a file or folder.
> ```

Does `yarn hs init --personal-access-key` work?

packages/cms-cli/lib/commonOpts.js
miketalley: I don't see a way in commander to supress any registered option. If I leave this off, the necessary data will not get passed along in the options data so the flag won't work.

PR #99, opened 2020-02-22
Add ability to create global partials via `hs create`

### LINKED ISSUE #97: Add `hs create global-partial`

gcorne:
Fixes #97

@williamspiro @corymartin

### DIFF
--- a/packages/cms-cli/commands/create.js
+++ b/packages/cms-cli/commands/create.js
@@ -22,11 +22,16 @@ const TYPES = {
   module: 'module',
   template: 'template',
   'website-theme': 'website-theme',
+  'global-partial': 'global-partial',
 };
 
 const ASSET_PATHS = {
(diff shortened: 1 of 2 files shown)


packages/cms-cli/commands/create.js
```diff
@@ -39,28 +44,28 @@ const createModule = (name, dest) => {
...
   fs.mkdirp(dest);
-  logger.log(`Copying boilerplate template to ${filePath}`);
+  logger.log(`Creating file at ${filePath}`);
   fs.copySync(assetPath, filePath);
 };
 
 function configureCreateCommand(program) {
   program
     .version(version)
-    .description('Create assets from boilerplate.')
```
gcorne: For the basic assets like modules, templates, and partials, I feel like `boilerplate` is not the right term because the expectation is that there will need to be tons of modification. They are more like "code templates."

PR #106, opened 2020-02-25
Add gitignore rule for HubSpot config

corymartin:
After loading a config from any command:
- The directory in which the config is located will be checked for git version control
- If it is, then a check will be done for inclusion in a `.gitignore` file
- If lacking that, the following warning will be printed

Screen Shot 2020-02-27 at 13 29 38


packages/cms-lib/lib/config.js
```diff
@@ -77,14 +79,77 @@ const getOrderedConfig = unorderedConfig => {
...
+    const ignoreFilePath = path.join(dir, '.gitignore');
+    await fs.ensureFile(ignoreFilePath);
+    let source = (await fs.readFile(ignoreFilePath)) || '';
+    if (source) {
+      const lines = source.split(/[\n\r]/);
+      const hasPattern =
+        lines.find(
+          CONFIG_GITIGNORE_PATTERN_REGEX.test,
+          CONFIG_GITIGNORE_PATTERN_REGEX
+        ) != null;
```
gcorne: I think it may be possible to use https://www.npmjs.com/package/ignore which we are already using for `.hsignore` for this.

gcorne:
I wonder if we should switch to `$HOME` as the default config path as part of this when running `hs init`.

packages/cms-lib/lib/config.js
```diff
@@ -308,7 +367,7 @@ const setDefaultConfigPathIfUnset = () => {
 };
 
 const setDefaultConfigPath = () => {
-  setConfigPath(`${getCwd()}/${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME}`);
+  setConfigPath(`${os.homedir()}/${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME}`);
```
corymartin: Default config dir is now user's home dir.

corymartin: I'm wondering if we should skip writing `.gitignore` if the config is in home dir.  Assuming that wouldn't be in source control.

gcorne: I don’t think we should write to gitignore and leave that to users.

PR #107, opened 2020-02-25
Adjust auth flow for access keys to account for existing config

gcorne:
@williamspiro @miketalley

### DIFF
--- a/packages/cms-cli/commands/init.js
+++ b/packages/cms-cli/commands/init.js
@@ -1,5 +1,10 @@
 const { version } = require('../package.json');
-const { getConfigPath } = require('@hubspot/cms-lib/lib/config');
+const {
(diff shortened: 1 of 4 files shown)


gcorne:
While working through this, I also realized that we also need to account for the `--config` and `--portal` options when going through `hs auth`.

packages/cms-cli/commands/init.js
```diff
@@ -49,11 +54,13 @@ function initializeConfigCommand(program) {
         TRACKING_STATUS.STARTED
       );
 
+      createEmptyConfigFile();
+      handleExit(deleteEmptyConfigFile);
```
gcorne: Moved this into `init` so that both are explicitly only part of the initialization flow.

PR #111, opened 2020-02-26
Updated error for deactivated personal access key

miketalley:


packages/cms-lib/personalAccessKey.js
```diff
@@ -33,9 +33,14 @@ async function getAccessToken(personalAccessKey, env = 'PROD') {
     response = await fetchAccessToken(personalAccessKey, env);
   } catch (e) {
     if (e.response) {
-      throw new HubSpotAuthError(
-        `Error while retrieving new access token: ${e.response.body.message}`
-      );
+      const errorOutput = `Error while retrieving new access token: ${e.response.body.message}.`;
```
corymartin: We have some conditionals elsewhere that indicate you could get a reference error here.
https://github.com/HubSpot/hubspot-cms-tools/search?q=response.body&unscoped_q=response.body

packages/cms-lib/personalAccessKey.js
```diff
@@ -33,9 +33,14 @@ async function getAccessToken(personalAccessKey, env = 'PROD') {
...
+      if (e.response.statusCode === 401) {
+        throw new HubSpotAuthError(
+          `${errorOutput} \nYour personal CMS access key is invalid. Please run "hs auth personalaccesskey" to reauthenticate. See https://designers.hubspot.com/docs/personal-access-keys for more information.`
```
corymartin: A couple enhancement ideas.
- Move the command to its own line (indented? chalk()?) so it is easier to select and copy.
- Clickable link https://www.npmjs.com/package/terminal-link

PR #112, opened 2020-02-26
Prompt for name only if not set

miketalley:
- Moves the name prompt after `personalAccessKey` for both `hs init` and `hs auth`
- Only displays prompt(`hs auth`) if name does not already exist for portal
- Adds validation for name when running validateConfig
- Adds tests for updatePortalConfig and validateConfig
- Checks for duplicate name before accepting input from name prompt

#### Requirements
- Needs to be released in sync with Docs update https://git.hubteam.com/HubSpot/hubspot-developers-website/issues/411

/cc @williamspiro


packages/cms-cli/commands/auth.js
```diff
@@ -41,14 +43,30 @@ async function authAction(type, options) {
...
-      await updateConfigWithPersonalAccessKey(configData);
+      updatedConfig = await updateConfigWithPersonalAccessKey(configData);
+
+      if (!updatedConfig.name) {
+        promptAnswer = await promptUser([PORTAL_NAME]);
+        updatePortalConfig({
+          ...updatedConfig,
+          environment: updatedConfig.env,
+          tokenInfo: updatedConfig.auth.tokenInfo,
+          name: promptAnswer.name,
```
gcorne: I think we should add some validation somewhere of the name to make sure it is not a duplicate and that it doesn't have spaces.

packages/cms-cli/commands/init.js
```diff
@@ -59,7 +60,18 @@ function initializeConfigCommand(program) {
...
+          {
+            ...configData,
+            name,
+          },
+          true
+        );
+
+        logger.log(
+          `Success: ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} created with ${PERSONAL_ACCESS_KEY_AUTH_METHOD.name}.`
+        );
```
gcorne: Not something for this PR, but I think we should consider adding `logger.success` and having a `[Success]` prefix that we color green.

packages/cms-lib/lib/config.js
```diff
@@ -271,11 +271,12 @@ const updatePortalConfig = configOptions => {
       tokenInfo,
     };
   }
+
   const env = getConfigEnv(environment || (portalConfig && portalConfig.env));
   const mode = defaultMode && defaultMode.toLowerCase();
   const nextPortalConfig = {
     ...portalConfig,
-    name,
+    name: name || (portalConfig && portalConfig.name),
```
gcorne: We should probably have some unit tests for `updatePortalConfig`.

packages/cms-cli/commands/auth.js
corymartin: @miketalley I recently added a validate function where you can add this.
https://github.com/HubSpot/hubspot-cms-tools/blob/d98f0449e0add7dc37205d1c8b614f6cb66256d1/packages/cms-lib/lib/config.js#L31
It works with the config data in memory whereas `updatePortalConfig` both updates and writes.  So something would need refactored for the user flow.  Like passing an optional `write: false` option to the update method.

packages/cms-cli/commands/init.js
iamw11y: > Not something for this PR, but I think we should consider adding logger.success and having a [Success] prefix that we color green.

@gcorne This is for the console, correct? Would the plan be to duplicate the message or provide additional info? I was thinking about this in the context of @HubSpotProtected/LocalDevAuth#31 where we have both an error and info. Should I be thinking about console logs in the same vein as I'm thinking about standardized terminal messages and prompts (e.g., ensuring consistency across scenarios)?

packages/cms-cli/commands/auth.js
miketalley: do we want to require a name for all portals? right now it is optional, and the `hs auth oauth2` flow does not include asking for a portal name.

personally, I like using it and do 99% of the time, but some developers may not like having it be required. it would cause errors for people with configs that have nameless portals. thoughts @corymartin @gcorne ?

packages/cms-cli/commands/init.js
gcorne: The idea is that in a similar fashion to errors that we have a system for displaying success messages in a somewhat similar fashion to how it works in Canvas.

The following is an example of an error:
image

packages/cms-cli/commands/auth.js
miketalley: I added name validation for portals into the validateConfig method to handle duplicate portal names and names with spaces

packages/cms-lib/lib/config.js
miketalley: Added tests for portal naming/renaming as well as various others for updatePortalConfig.

packages/cms-cli/commands/auth.js
gcorne: In a similar fashion to a form input, I think we should be validating during the prompt flow as it provides a better user experience.

miketalley: We have [validation with the prompt to disallow names that contain spaces](https://github.com/HubSpot/hubspot-cms-tools/blob/46c5945fe82259b354c3aee864637628b9ee5a23/packages/cms-cli/lib/prompts.js#L62). Do we want to add something else?

packages/cms-cli/commands/init.js
miketalley: Fixed in https://github.com/HubSpot/hubspot-cms-tools/pull/125

packages/cms-lib/lib/config.js
```diff
@@ -46,13 +47,28 @@ function validateConfig() {
...
-    portalsHash[cfg.portalId] = cfg;
+
+    if (cfg.name) {
+      if (portalNamesHash[cfg.name]) {
+        logger.error(
+          `config.name has multiple entries with portalId=${cfg.portalId}`
+        );
+        return false;
+      }
+      if (cfg.name.indexOf(' ') !== -1) {
```
corymartin: We should check for whitespace generally `/\s+/.test(cfg.name)`

gcorne:
Are we planning to move forward with this?

packages/cms-cli/commands/auth.js
gcorne: I was thinking that we could validate on duplicate names in the prompt flow.

miketalley: I added a check to ensure a duplicate name will not be accepted when running `hs auth`

PR #113, opened 2020-02-26
Added constants for environments and updated usage and helper methods

miketalley:
- Adds helpers for obtaining correct URL based on env
- Adds constants for `QA` and `PROD`
- Adds helpers for getting valid versions of the env and masking if necessary


packages/cms-lib/lib/environment.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Returns environment constant for QA and PROD or optional masked value for PROD
+ * @param {string} env Environment string, can be any case
+ * @param {(boolean|object)} shouldMaskProduction Returning alternate value for PROD
+ *    Can be used to hide env value in the config. Boolean can be used to simply mask 'prod' with ''. The default
+ *    is not to modify the returned value for production, and instead return 'prod'.
+ * @param {any} shouldMaskProduction.maskedProductionValue Alternate value to return in place of PROD
+ */
+const getValidEnv = (env, shouldMaskProduction) => {
```
gcorne, line 11: This function introduces a "boolean trap" that makes reading code using this function difficult. It also feels like all this environment stuff is more complex than it needs to be. Is there a reason for all the complexity that we could fix through adjustments elsewhere?

packages/cms-lib/personalAccessKey.js
```diff
@@ -127,16 +139,11 @@ const personalAccessKeyPrompt = async () => {
   const { name } = await promptUser(PERSONAL_ACCESS_KEY_FLOW);
   const portalId = getPortalId(name);
   const env = getEnv(name);
+  const websiteOrigin = getHubSpotWebsiteOrigin(env);
```
gcorne, line 142: Instead of `getHubSpotWebsiteOrigin` could we come up with a better name? Maybe `getHubSpotAppDomain`?

PR #118, opened 2020-03-02
Support environment variable configuration

### LINKED ISSUE #4: Add support for using environment variables instead of a config file

miketalley:
- Allows environment variable configuration with no `hubspot.config.yml` file(if a config file and necessary environment variables are both present, environment variables take precedence)
- Different auth methods require different environment variables to be present
  - personalaccesskey requires `HUBSPOT_PORTAL_ID` and `HUBSPOT_PERSONAL_ACCESS_KEY`
  - apikey requires `HUBSPOT_PORTAL_ID` and `HUBSPOT_API_KEY`
  - oauth2 requires `HUBSPOT_PORTAL_ID`, `HUBSPOT_CLIENT_ID`, `HUBSPOT_CLIENT_SECRET`, and `HUBSPOT_REFRESH_TOKEN`

Fixes #4


packages/cms-lib/lib/constants.js
```diff
@@ -73,6 +73,15 @@ const OAUTH_SCOPES = [
   },
 ];
 
+const HUBSPOT_API_KEY = 'HUBSPOT_API_KEY';
+const HUBSPOT_CLIENT_ID = 'HUBSPOT_CLIENT_ID';
+const HUBSPOT_CLIENT_SECRET = 'HUBSPOT_CLIENT_SECRET';
+const HUBSPOT_PERSONAL_ACCESS_KEY = 'HUBSPOT_PERSONAL_ACCESS_KEY';
+const HUBSPOT_PORTAL_ID = 'HUBSPOT_PORTAL_ID';
+const HUBSPOT_REFRESH_TOKEN = 'HUBSPOT_REFRESH_TOKEN';
```
gcorne: Instead of separate constants, I think an enum would be better for this and for the enum to have `EnvironmentVariables` in the name.

packages/cms-lib/lib/config.js
```diff
@@ -353,6 +377,118 @@ const deleteEmptyConfigFile = () => {
...
+          },
+        },
+      },
+    ],
+  };
+};
+
+const generateApiKeyPortalConfig = (portalId, apiKey) => {
+  return {
+    defaultPortal: ENVIRONMENT_VARIABLES_DEFAULT_PORTAL_NAME,
```
gcorne: What is the thinking behind incorporating a default portal into the environment variables system.

packages/cms-lib/lib/config.js
```diff
@@ -353,6 +377,118 @@ const deleteEmptyConfigFile = () => {
...
+    portals: [
+      {
+        inMemoryOnly: true,
```
gcorne: Is there a reason that this is being handled in the config object instead of elsewhere? I think we want to create a clear separation between the interface of the config data and other data?

packages/cms-lib/lib/config.js
```diff
@@ -353,6 +377,118 @@ const deleteEmptyConfigFile = () => {
...
+    apiKey: env[HUBSPOT_API_KEY],
+    clientId: env[HUBSPOT_CLIENT_ID],
+    clientSecret: env[HUBSPOT_CLIENT_SECRET],
+    personalAccessKey: env[HUBSPOT_PERSONAL_ACCESS_KEY],
+    portalId: parseInt(env[HUBSPOT_PORTAL_ID], 10),
+    refreshToken: env[HUBSPOT_REFRESH_TOKEN],
+  };
+};
+
+const generatePersonalAccessKeyPortalConfig = (portalId, personalAccessKey) => {
```
gcorne: The naming of these functions is a bit confusing because it says "PortalConfig` which elsewhere only refers to the configuration for an individual portal.

packages/cms-lib/lib/config.js
```diff
@@ -119,7 +130,7 @@ const parseConfig = configSource => {
   return { parsed, error };
 };
 
-const loadConfig = (path, options = {}) => {
+const loadFileConfig = (path, options = {}) => {
```
gcorne: Maybe `loadConfigFromFile` or `loadConfigFile`.

packages/cms-lib/lib/config.js
```diff
@@ -297,7 +317,11 @@ const updatePortalConfig = configOptions => {
       config.portals = [nextPortalConfig];
     }
   }
-  writeConfig();
+
+  // TODO - Implement a better solution for decoupling this
+  if (!nextPortalConfig.inMemoryOnly) {
```
corymartin: Perhaps we just follow the single responsibility principle. So we'd have update and write functions that we can call in succession in our command code.

packages/cms-lib/lib/config.js
miketalley: I was using this so the in-memory portal config that is generated using the environment variables would be used by default

packages/cms-lib/lib/config.js
miketalley: The handling of this is now done through setting `disableWritesToFile` to `true` when we are using an environment variable based config

packages/cms-lib/lib/config.js
miketalley: Renamed these

packages/cms-lib/lib/constants.js
miketalley: Fixed

packages/cms-lib/lib/config.js
gcorne: What if we modify https://github.com/HubSpot/hubspot-cms-tools/blob/0074dc1a177d4335fd2e838c75a20d51a76acb47/packages/cms-cli/lib/commonOpts.js#L46 instead? We probably also want to throw an error if `--portal` is passed as an option.

packages/cms-lib/lib/__tests__/config.js
```diff
@@ -73,4 +111,138 @@ describe('lib/config', () => {
...
+
+      beforeEach(() => {
+        process.env = {
+          HUBSPOT_PORTAL_ID: portalId,
+          HUBSPOT_CLIENT_ID: clientId,
+          HUBSPOT_CLIENT_SECRET: clientSecret,
+          HUBSPOT_REFRESH_TOKEN: refreshToken,
+        };
+        getAndLoadConfigIfNeeded();
+        portalConfig = getConfig().portals[0];
```
gcorne: I think it is worthwhile to test the API that we use, so `getPortalConfig()` instead of reaching into `portals`.

packages/cms-cli/lib/commonOpts.js
```diff
@@ -43,9 +51,16 @@ const setLogLevel = (options = {}) => {
...
+ */
 const getPortalId = (options = {}) => {
-  const { portal } = options;
-  return getPortalIdFromConfig(portal);
+  const envConfig = loadConfigFromEnvironment();
+  const { portal: portalName } = options;
+  return (
+    (envConfig && envConfig.portals[0].portalId) ||
+    getPortalIdFromConfig(portalName)
+  );
```
gcorne: Instead of a complex one liner here, let's have an early exit. I think it also is reasonable for the environment variable to always take precedence.
```suggestion
  if (process.env.HUBSPOT_PORTAL_ID) {
    return process.env.HUBSPOT_PORTAL_ID;
  }
  return getPortalIdFromConfig(portalName);
```

packages/cms-cli/commands/hubdb.js
```diff
@@ -44,7 +44,9 @@ function configureHubDbCreateCommand(program) {
       setLogLevel(command);
       logDebugInfo(command);
       const { config: configPath } = command;
-      loadConfig(configPath);
+      loadConfig(configPath, {
+        allowEnvironmentVariableConfig: true,
```
gcorne: Why do we need this flag?

packages/cms-cli/lib/commonOpts.js
```diff
@@ -3,13 +3,21 @@ const {
...
 
 const addPortalOptions = program => {
   program.option('--portal <portal>', 'HubSpot portal id or name from config');
+  program.on('option:portal', () => {
+    if (loadConfigFromEnvironment()) {
+      throw new Error(
+        'Cannot specify a portal when environment variables are supplied. Please unset the environment variables or do not use the "--portal" flag.'
+      );
+    }
+  });
```
gcorne: Instead of leveraging the event, what if we adjusted the portal option to have custom processing or updated `validatePortal`.

packages/cms-cli/lib/commonOpts.js
miketalley: This is the code that I previously had, but I changed it because if the `HUBSPOT_PORTAL_ID` is set, but not any of the other necessary environment variables needed to successfully load a config, we run into an error here because we are using the `HUBSPOT_PORTAL_ID` for `portalId` and the rest of the portal config is loaded from file.

packages/cms-lib/lib/config.js
```diff
@@ -13,10 +13,23 @@ const {
...
+  PERSONAL_ACCESS_KEY_AUTH_METHOD,
+  OAUTH_SCOPES,
+  ENVIRONMENT_VARIABLES: {
+    HUBSPOT_REFRESH_TOKEN,
+    HUBSPOT_API_KEY,
+    HUBSPOT_CLIENT_ID,
+    HUBSPOT_CLIENT_SECRET,
+    HUBSPOT_PERSONAL_ACCESS_KEY,
+    HUBSPOT_PORTAL_ID,
+  },
```
corymartin: Our constants file is getting heavy.  At some point we should consider managing enums separately.

Instead of nested de-structuring here, how about we export an `EnvironmentVariables` enum that has these values.

packages/cms-cli/commands/hubdb.js
miketalley: If we try running `hs auth` with environmental variable data present, it never writes to file.

packages/cms-lib/lib/config.js
```diff
@@ -449,6 +484,109 @@ const deleteEmptyConfigFile = () => {
...
 
+const getConfigVariablesFromEnv = () => {
+  const env = process.env;
+
+  return {
+    apiKey: env[HUBSPOT_API_KEY],
+    clientId: env[HUBSPOT_CLIENT_ID],
+    clientSecret: env[HUBSPOT_CLIENT_SECRET],
+    personalAccessKey: env[HUBSPOT_PERSONAL_ACCESS_KEY],
+    portalId: parseInt(env[HUBSPOT_PORTAL_ID], 10),
```
corymartin: Should the env variable just be `HUBSPOT_PORTAL`?

miketalley: I'd rather be verbose with the name so there is no question about it.

packages/cms-cli/commands/hubdb.js
gcorne: Ah, I understand the complexity now. I would think that we would treat running commands that don't support environment variables as an error and exit early since it would be a mistake to run `hs init` or `hs auth` with environment variables.

packages/cms-cli/lib/commonOpts.js
gcorne: Under what circumstances does the error occur?

miketalley: [Loading an environment variable config](https://github.com/HubSpot/hubspot-cms-tools/blob/886b154c23187c3c856f9741db163724a4f35739/packages/cms-lib/lib/config.js#L542) requires the `HUBSPOT_PORTAL_ID` plus at least one other environment variable. If only `HUBSPOT_PORTAL_ID` is present and I use the code above, `getPortalIdFromConfig` will use the portalId from the environment variable even though the config being used is being loaded from the `hubspot.config.yml` file, which can cause issues depending on where `getPortalId` is being called from.

By doing the check using `loadConfigFromEnvironment` we are verifying that the required environment variables to load a config are present, so we are in fact loading the config from the environment variables as opposed to the config file.

gcorne: I still would like to make this more straightforward. The `allowEnvironmentVariableConfig` is awkward to me because it feels like we should be able to handle the two scenarios in a more direct than indirect fashion. It isn't clear why something that is loading the config should need to be configured such that under some circumstances it should check environment variables and another it shouldn't.

From a user, it also isn't clear why I couldn't do `HUBSPOT_PORTAL_ID=123 hs fetch foo bar` if the portal has been configured. It is clear why `HUBSPOT_PORTAL_ID=123 hs fetch --portal 456 foo bar` should throw and error. And, it also is clear why certain sets of env variables shouldn't be allowed.

packages/cms-lib/lib/constants.js
```diff
@@ -73,6 +73,18 @@ const OAUTH_SCOPES = [
...
+const ENVIRONMENT_VARIABLES = {
+  HUBSPOT_API_KEY: 'HUBSPOT_API_KEY',
+  HUBSPOT_CLIENT_ID: 'HUBSPOT_CLIENT_ID',
+  HUBSPOT_CLIENT_SECRET: 'HUBSPOT_CLIENT_SECRET',
+  HUBSPOT_PERSONAL_ACCESS_KEY: 'HUBSPOT_PERSONAL_ACCESS_KEY',
+  HUBSPOT_PORTAL_ID: 'HUBSPOT_PORTAL_ID',
+  HUBSPOT_REFRESH_TOKEN: 'HUBSPOT_REFRESH_TOKEN',
+};
+
+const ENVIRONMENT_VARIABLES_DEFAULT_PORTAL_NAME =
```
gcorne: This is unused so should be removed.

packages/cms-cli/lib/commonOpts.js
gcorne: > can cause issues depending on where getPortalId is being called from.

What kinds of issues? I ask because the function is part of CLI and used to process the command options, so I am not sure what scenarios might be a concern.

miketalley: > I still would like to make this more straightforward. The `allowEnvironmentVariableConfig` is awkward to me because it feels like we should be able to handle the two scenarios in a more direct than indirect fashion. It isn't clear why something that is loading the config should need to be configured such that under some circumstances it should check environment variables and another it shouldn't.

I replaced `allowEnvironmentVariableConfig` with `ignoreEnvironmentVariableConfig` which only needs to be set by `hs auth` in order to allow writes to file when there are environment variables present when running the command.

> From a user, it also isn't clear why I couldn't do `HUBSPOT_PORTAL_ID=123 hs fetch foo bar` if the portal has been configured. It is clear why `HUBSPOT_PORTAL_ID=123 hs fetch --portal 456 foo bar` should throw and error. And, it also is clear why certain sets of env variables shouldn't be allowed.

I updated the way this works so that setting `HUBSPOT_PORTAL_ID` works similarly to `defaultPortal`.

packages/cms-cli/commands/hubdb.js
miketalley: I think with the ability to use `HUBSPOT_PORTAL_ID` similarly to `defaultPortal`, the possibility of wanting to run `hs auth` with `HUBSPOT_PORTAL_ID` set is much higher. The ability to handle this case gracefully is preferable to asking the user to `unset HUBSPOT_PORTAL_ID`, re-run `hs auth`, and `export HUBSPOT_PORTAL_ID` again.

With the new code, this case is handled so the user can run `hs auth` with `HUBSPOT_PORTAL_ID` set.

packages/cms-lib/lib/constants.js
gcorne: @miketalley did you happen to see this?

PR #121, opened 2020-03-04
Ability to tail logs

miketalley:
- Adds ability to tail(follow) logs using --follow or -f

### DIFF
--- a/packages/cms-cli/commands/logs.js
+++ b/packages/cms-cli/commands/logs.js
@@ -76,9 +76,8 @@ const getLogs = program => {
     .version(version)
     .description(`get logs for a function`)
     .arguments('<function_path>')
-    .option('-f, --file', 'output logs to file')
     .option('--latest', 'retrieve most recent log only')
-    .option('--tail', 'tail logs')
+    .option('-f', '--follow', 'tail logs')
     .action(async (functionPath, options) => {
       const { config: configPath } = options;
       const portalId = getPortalId(program);
--- a/packages/cms-cli/commands/logs.js
+++ b/packages/cms-cli/commands/logs.js
@@ -1,4 +1,5 @@
 const { version } = require('../package.json');
+const ora = require('ora');
 const {
   addLoggerOptions,
   addPortalOptions,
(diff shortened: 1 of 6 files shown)


packages/cms-cli/commands/logs.js
```diff
@@ -13,27 +14,74 @@ const { logDebugInfo } = require('../lib/debugInfo');
...
+        portalId,
+      });
+    }
+  }
+
+  return new Promise(() => {
+    setInterval(async () => {
+      // eslint-disable-next-line require-atomic-updates
+      after = await tailCall(after, spinner);
+    }, TAIL_DELAY);
```
corymartin: `setInterval` and async code can lead to inconsistent interval periods, and in some cases calls can occur out of desired order.  You can use a recursive `setTimeout` to ensure that at least `TAIL_DELAY` ms have elapsed since last invoked.

```javascript
const h = async () => {
  after = await tailCall(after, spinner);
  setTimeout(h, TAIL_DELAY);
};
h();
```

corymartin: Also, this promise isn't resolving?  Instead of promises this could be a use case for generators so you can just `yield` the results as they come in.

miketalley: I updated the async pattern with your suggestion as it is a safer bet. I think this could be a good case for a generator, but I think we should get this out the door and I can come back to it later to implement this as a generator.

packages/cms-lib/lib/encoding.js
```diff
@@ -0,0 +1,23 @@
+const base64EncodeString = valueToEncode => {
+  if (typeof valueToEncode !== 'string') {
+    return valueToEncode;
+  }
+
+  const stringBuffer = Buffer.from(valueToEncode);
+  return encodeURIComponent(stringBuffer.toString('base64'));
```
gcorne: I think it is a bit surprising that this is url-encoded in addition to base64 encoded. Isn't the base64 encoded string URL safe?

packages/cms-lib/lib/encoding.js
```diff
@@ -0,0 +1,23 @@
...
+
+const base64DecodeString = valueToDecode => {
+  if (typeof valueToDecode !== 'string') {
+    return valueToDecode;
+  }
+
+  const decodedURI = decodeURIComponent(valueToDecode);
+  const stringBuffer = Buffer.from(decodedURI, 'base64');
+  return stringBuffer.toString('ascii');
+};
```
gcorne: Since this is unused, let's remove it.

packages/cms-lib/lib/encoding.js
miketalley: No - it breaks because it sends `=` instead of `%3D` without URL encoding it

packages/cms-cli/commands/logs.js
```diff
@@ -13,27 +14,78 @@ const { logDebugInfo } = require('../lib/debugInfo');
...
+    // A 404 means no latest log exists(never executed)
+    if (e.statusCode !== 404) {
+      logApiErrorInstance(e, {
+        functionPath,
+        portalId,
+      });
+    }
+  }
+
+  return new Promise(() => {
```
gcorne: Something doesn't look quite right when it comes to the return value of a Promise here and what the calls are doing below with the return value. In general, we should also favor `async` over `new Promise()` unless there is a particular reason.

miketalley: I agree, this is unusual. Basically, the thread is being blocked, waiting for a promise that never resolves. This is done so we can tail the logs indefinitely, until the process is manually killed.

We could [spawn a child process](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) that would start the tailing, which would achieve the same result in a less hacky way.

Thoughts @gcorne @corymartin ?

corymartin: This is working POC code.

```javascript
async function fetchFunctionLogs() {
  return Promise.resolve((Math.random() * 1000) | 0);
}

const INTERVAL = 2000;

async function delayFetch() {
  return new Promise(resolve => {
    setTimeout(() => {
      debugger;
      fetchFunctionLogs().then(resolve);
    }, INTERVAL);
  });
}

async function* checkLogs() {
  for (;;) {
    const result = await delayFetch();
    yield result;
  }
}

async function tail() {
  for await (const log of checkLogs()) {
    console.log('>>>', log);
  }
}

tail();
```

corymartin: Try running that in Node env or in Chrome console.

gcorne: @miketalley One way to handle it is to not return from the async function in the `logs` subcommand action. I think one way of thinking that might lead to a solution is to consider how to handle error case. What happens if the user doesn't have permission to fetch logs or there is a network interruption while tailing?

packages/cms-lib/lib/encoding.js
gcorne: I guess I was thinking of the `base64url` encoding from https://tools.ietf.org/html/rfc4648. Not a big deal. That said, I do think that `encodeURIComponent` should be called closer to where the URL is constructed. I am also a bit unclear on why base64-encoding is necessary. Why do we need to encode the id of the log?

miketalley: This is a BE requirement in order to retrieve the subsequent log(s).

@bkrainer can you add some color?

packages/cms-cli/commands/logs.js
miketalley: This has been improved greatly.  I removed the unnecessary promise.

Looking at the generator a bit more and comparing it with the current code, I think it creates unnecessary complication without a major benefit. Take a look at the new code and LMK if you don't agree @corymartin.

PR #124, opened 2020-03-09
Add hs secrets list command

miketalley:
- Allows `hs secrets list` command to be used to display the names of the secrets that are set for the portal specified

image

### DIFF
--- a/packages/cms-cli/commands/secrets.js
+++ b/packages/cms-cli/commands/secrets.js
@@ -4,7 +4,11 @@ const {
   checkAndWarnGitInclusion,
 } = require('@hubspot/cms-lib');
 const { logger } = require('@hubspot/cms-lib/logger');
-const { addSecret, deleteSecret } = require('@hubspot/cms-lib/api/secrets');
+const {
+  addSecret,
+  deleteSecret,
+  getSecrets,
+} = require('@hubspot/cms-lib/api/secrets');
 
 const { validatePortal } = require('../lib/validation');
 const { addHelpUsageTracking } = require('../lib/usageTracking');
@@ -24,7 +28,8 @@ function configureSecretsCommand(program) {
     .version(version)
     .description('Manage secrets')
     .command('add <secret-name> <secret-value>', 'add a HubSpot secret')
-    .command('delete <secret-name>', 'delete a HubSpot secret');
+    .command('delete <secret-name>', 'delete a HubSpot secret')
+    .command('list', 'list all HubSpot secrets');
 
   addLoggerOptions(program);
   addHelpUsageTracking(program);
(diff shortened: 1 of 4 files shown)


packages/cms-cli/commands/secrets.js
```diff
@@ -24,7 +28,8 @@ function configureSecretsCommand(program) {
     .version(version)
     .description('Manage secrets')
     .command('add <secret-name> <secret-value>', 'add a HubSpot secret')
-    .command('delete <secret-name>', 'delete a HubSpot secret');
+    .command('delete <secret-name>', 'delete a HubSpot secret')
```
corymartin: Since the positional args are within the context of the secrets command, I think we can shorted their slug names to just `name` and `value`.

packages/cms-cli/commands/secrets.js
```diff
@@ -95,8 +100,39 @@ function configureSecretsDeleteCommand(program) {
   addConfigOptions(program);
 }
 
+function configureSecretsListCommand(program) {
+  program
+    .version(version)
+    .description('List all HubSpot secrets')
+    .action(async (options = {}) => {
```
corymartin: The last parameter passed to action is always the commander instance.
Thus `program === options`.  Given this I've been preferring to use `program`.

For example: https://github.com/HubSpot/hubspot-cms-tools/blob/master/packages/cms-cli/commands/fetch.js#L32-L39

packages/cms-cli/commands/secrets.js
```diff
@@ -95,8 +100,39 @@ function configureSecretsDeleteCommand(program) {
...
+
+      if (!(validateConfig() && (await validatePortal(options)))) {
+        process.exit(1);
+      }
+      const portalId = getPortalId(options);
+
+      try {
+        const { results } = await getSecrets(portalId);
+        logger.log(`Secrets for portal: ${portalId}:`);
+        results.forEach(secret => logger.log(secret));
```
corymartin: It might present better if we grouped.

```javascript
const groupLabel = `Secrets for portal: ${portalId}:`;
logger.group(groupLabel);
/* ...log each secret... */
logger.groupEnd(groupLabel);
```

packages/cms-cli/commands/secrets.js
```diff
@@ -95,8 +100,39 @@ function configureSecretsDeleteCommand(program) {
...
+      } catch (e) {
+        logger.error('Getting secrets failed');
+        logger.error(e.message);
```
corymartin: We can use `logErrorInstance()` here.

packages/cms-lib/api/secrets.js
```diff
@@ -18,7 +18,14 @@ async function deleteSecret(portalId, key) {
   });
 }
 
+async function getSecrets(portalId) {
```
corymartin: Let's rename this to `fetchSecrets`.  Most of our `GET` clients are named `fetch*`.

TheWebTech:
curious, wouldn't it make sense for the command to be `hs secrets` ?
akin to `git branch` or something similar

PR #125, opened 2020-03-09
Add logger.info and logger.success

miketalley:
- Adds `logger.info` command to log helpful information
- Adds `logger.success` command to log success messages

Fixes https://github.com/HubSpot/hubspot-cms-tools/issues/110

### DIFF
--- a/packages/cms-lib/logger.js
+++ b/packages/cms-lib/logger.js
@@ -14,6 +14,7 @@ const LOG_LEVEL = {
 const Styles = {
   debug: chalk.reset.blue,
   log: chalk.reset,
+  success: chalk.reset.green,
   info: chalk.reset.white,
   warn: chalk.reset.yellow,
   error: chalk.reset.red,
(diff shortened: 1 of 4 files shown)


packages/cms-lib/logger.js
```diff
@@ -37,6 +39,12 @@ class Logger {
   log(...args) {
     console.log(...args);
   }
+  success(...args) {
+    console.log(...stylize('[SUCCESS]', Styles.success, args));
+  }
+  info(...args) {
+    console.log(...stylize('[INFO]', Styles.info, args));
```
corymartin:
```suggestion
    console.info(...stylize('[INFO]', Styles.info, args));
```

https://developer.mozilla.org/en-US/docs/Web/API/Console/info

packages/cms-lib/logger.js
```diff
@@ -96,6 +104,16 @@ const logger = {
       currentLogger.log(...args);
     }
   },
+  success(...args) {
+    if (shouldLog(LOG_LEVEL.LOG)) {
+      currentLogger.success(...args);
```
gcorne: I think we probably need to check for `currentLogger.success` since that is a bit non-standard and fall back to `currentLogger.log()`. For example, I am curious about whether or not this impacts the logging in the webpack plugin.

PR #136, opened 2020-03-18
Add utility functions to manage HubDB tables in other scripts

anthmatic:
This PR will add an new utility functions for managing HubDb tables:
- Clear/delete all rows from a table
- Update a table from a file
- Add rows to a table

### DIFF
--- a/packages/cms-cli/bin/hscms-hubdb-clear.js
+++ b/packages/cms-cli/bin/hscms-hubdb-clear.js
@@ -1,9 +1,0 @@
-#!/usr/bin/env node
-
-const { Command } = require('commander');
-
-const { configureHubDbClearCommand } = require('../commands/hubdb');
-
-const program = new Command('hscms hubdb clear');
-configureHubDbClearCommand(program);
-program.parse(process.argv);
--- a/packages/cms-cli/bin/hscms-hubdb-clear.js
+++ b/packages/cms-cli/bin/hscms-hubdb-clear.js
@@ -1,0 +1,9 @@
+#!/usr/bin/env node
+
+const { Command } = require('commander');
+
+const { configureHubDbClearCommand } = require('../commands/hubdb');
+
+const program = new Command('hscms hubdb clear');
+configureHubDbClearCommand(program);
+program.parse(process.argv);
--- a/packages/cms-cli/commands/hubdb.js
+++ b/packages/cms-cli/commands/hubdb.js
@@ -9,11 +9,8 @@ const { logErrorInstance } = require('@hubspot/cms-lib/errorHandlers');
 const { getCwd } = require('@hubspot/cms-lib/path');
 const {
   createHubDbTable,
-  importHubDbTableRows,
   downloadHubDbTable,
-  clearHubDbTable,
 } = require('@hubspot/cms-lib/hubdb');
-const { deleteTable } = require('@hubspot/cms-lib/api/hubdb');
 
 const { validatePortal } = require('../lib/validation');
 const { addHelpUsageTracking } = require('../lib/usageTracking');
@@ -78,43 +75,6 @@ function configureHubDbCreateCommand(program) {
   addConfigOptions(program);
 }
 
-function configureHubDbImportCommand(program) {
-  program
-    .version(version)
-    .description('Import HubDB table rows')
-    .arguments('<tableId> <src>')
-    .action(async (tableId, src) => {
-      setLogLevel(program);
-      logDebugInfo(program);
-      const { config: configPath } = program;
-      loadConfig(configPath);
-      checkAndWarnGitInclusion();
-
-      if (!(validateConfig() && (await validatePortal(program)))) {
-        process.exit(1);
-      }
-      const portalId = getPortalId(program);
-
-      try {
-        const table = await importHubDbTableRows(
-          portalId,
-          tableId,
-          path.resolve(getCwd(), src)
-        );
-        logger.log(
-          `The table ${table.tableId} was updated in ${portalId} with ${table.rowCount} rows`
-        );
-      } catch (e) {
-        logger.error(`Updating the table at "${src}" failed`);
-        logErrorInstance(e);
-      }
-    });
-
-  addLoggerOptions(program);
-  addPortalOptions(program);
-  addConfigOptions(program);
-}
-
 function configureHubDbFetchCommand(program) {
   program
     .version(version)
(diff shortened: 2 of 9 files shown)


anthmatic:
@miketalley @corymartin would either of you mind taking a 👀 at this?

packages/cms-cli/bin/hscms-hubdb-clear.js
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env node
```
gcorne: It looks like `hscms-hubdb-delete` and `hscms-hubdb-import` need to be added.

packages/cms-cli/commands/hubdb.js
```diff
@@ -29,7 +32,10 @@ function configureHubDbCommand(program) {
     .version(version)
     .description('Manage HubDB tables')
     .command('create <src>', 'create a HubDB table')
-    .command('fetch <tableId> <dest>', 'fetch a HubDB table');
+    .command('import <tableId> <src>', 'import a HubDB table')
+    .command('fetch <tableId> <dest>', 'fetch a HubDB table')
```
corymartin: `import` and `fetch` are aliases? Do we want both? Using `import` makes sense to me in the context of tables/stores.

packages/cms-cli/commands/hubdb.js
```diff
@@ -71,6 +77,43 @@ function configureHubDbCreateCommand(program) {
...
+      if (!(validateConfig() && (await validatePortal(command)))) {
+        process.exit(1);
+      }
+      const portalId = getPortalId(command);
+
+      try {
+        const table = await importHubDbTableRows(
+          portalId,
+          tableId,
+          path.resolve(getCwd(), src)
```
corymartin: The pattern we use with the `fetch` command is to make the local destination optional with the cwd as default.
So in the commander pattern you could use `[src]` instead.

packages/cms-cli/commands/hubdb.js
```diff
@@ -71,6 +77,43 @@ function configureHubDbCreateCommand(program) {
   addConfigOptions(program);
 }
 
+function configureHubDbImportCommand(program) {
+  program
+    .version(version)
+    .description('Import HubDB table rows')
+    .arguments('<tableId> <src>')
+    .action(async (tableId, src, command = {}) => {
```
corymartin: The last argument passed to `action()` is always the Commander instance.  Thus `program === command`.
I've been preferring to treat the parameters explicitly and just use `program`.  Also if your last param is optional you can test it against `program`.

packages/cms-cli/commands/hubdb.js
```diff
@@ -71,6 +77,43 @@ function configureHubDbCreateCommand(program) {
...
+          portalId,
+          tableId,
+          path.resolve(getCwd(), src)
+        );
+        logger.log(
+          `The table ${table.tableId} was updated in ${portalId} with ${table.rowCount} rows`
+        );
+      } catch (e) {
+        logger.error(`Updating the table at "${src}" failed`);
+        logger.error(e.message);
```
corymartin: We have handlers to log `Error` instances.
Example here https://github.com/HubSpot/hubspot-cms-tools/blob/d98f0449e0add7dc37205d1c8b614f6cb66256d1/packages/cms-cli/commands/lint.js#L11

packages/cms-cli/commands/hubdb.js
```diff
@@ -104,8 +147,69 @@ function configureHubDbFetchCommand(program) {
...
+
+      if (!(validateConfig() && (await validatePortal(command)))) {
+        process.exit(1);
+      }
+      const portalId = getPortalId(command);
+      try {
+        await clearHubDbTable(portalId, tableId);
+        logger.log(`Delete rows in HubDB table ${tableId}`);
+      } catch (e) {
+        logger.error(e);
```
corymartin: Use error handler

packages/cms-lib/hubdb.js
```diff
@@ -64,6 +66,60 @@ async function createHubDbTable(portalId, src) {
...
+      childTableId: 0,
+      isSoftEditable: false,
+      ...row,
+      values,
+    };
+  });
+
+  let response;
+  if (rowsToUpdate.length > 0) {
+    response = await createRows(portalId, tableId, rowsToUpdate);
```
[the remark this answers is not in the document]
anthmatic: I assumed so. Did you have a different idea in mind?

packages/cms-cli/commands/hubdb.js
anthmatic: I think it's possible that the name for this command is misleading. This command is more like the `create` command where it "uploads" a table (or, actually just appends the rows).

packages/cms-lib/hubdb.js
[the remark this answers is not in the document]
anthmatic: I think the name "import" might be misleading (or it should maybe behave a little differently?)

packages/cms-cli/commands/hubdb.js
gcorne: Maybe it should be `hubdb replace`?

anthmatic: I think that with `replace` I would expect metadata and columns would also get affected. If you think that's more valuable, I can look at that.

Otherwise, `append` might be more accurate, but I'd defer to you guys.

gcorne: > Otherwise, append might be more accurate, but I'd defer to you guys.

Does it only add rows or does it also update them? What happens if the schema is different?

Is adding `hs hubdb import` a blocker to finishing up the sync script for the developers site? If not, perhaps we should remove it from this PR since there is some subtlety that needs to be worked out. I bring it up just so that we can keep things moving forward.

packages/cms-lib/hubdb.js
```diff
@@ -147,7 +165,16 @@ async function downloadHubDbTable(portalId, tableId, dest) {
   await fs.writeFileSync(dest, tableJson);
 }
 
+async function clearHubDbTable(portalId, tableId) {
+  const rows = await fetchRows(portalId, tableId);
```
gcorne: This code only clears the first page of rows not all rows since fetching won't automatically fetch all rows.

packages/cms-lib/hubdb.js
```diff
@@ -8,9 +8,10 @@ const {
   fetchTable,
   fetchRows,
   publishTable,
+  deleteRows,
 } = require('./api/hubdb');
 
-async function createHubDbTable(portalId, src) {
+function validateJson(src) {
```
gcorne: I think that this function should have file in its name since it is validating that it is a readable JSON file not just that it is JSON.

packages/cms-lib/hubdb.js
```diff
@@ -50,20 +48,40 @@ async function createHubDbTable(portalId, src) {
...
+}
+
+async function importHubDbTableRows(portalId, tableId, src) {
+  validateJson(src);
+
+  const table = fs.readJsonSync(src);
+  const { rows } = table;
+  const { columns } = await fetchTable(portalId, tableId);
+
+  return updateRows(portalId, tableId, rows, columns);
```
gcorne: This code is making an assumption that columns from the table that is fetched match up with the structure of the columns in the table file that rows are being imported from.

packages/cms-lib/hubdb.js
```diff
@@ -23,12 +24,9 @@ async function createHubDbTable(portalId, src) {
...
     throw new Error('The HubDB table file must be a ".json" file');
   }
+}
 
-  const table = fs.readJsonSync(src);
-  const { rows, ...schema } = table;
-
-  const { columns, id } = await createTable(portalId, schema);
-
+async function updateRows(portalId, tableId, rows, columns) {
```
gcorne: I think that this function name needs to be tweaked since it is adding rows.

packages/cms-lib/hubdb.js
```diff
@@ -50,20 +48,40 @@ async function createHubDbTable(portalId, src) {
...
-    response = await createRows(portalId, id, rowsToUpdate);
+    response = await createRows(portalId, tableId, rowsToUpdate);
   }
 
-  await publishTable(portalId, id);
+  await publishTable(portalId, tableId);
 
   return {
-    tableId: id,
+    tableId: tableId,
```
gcorne:
```suggestion
    tableId,
```

anthmatic:
Thanks for all the feedback! After chatting with @gcorne I've pulled out all of the commands but left the utilities in place to clear rows and update tables.

packages/cms-cli/commands/hubdb.js
anthmatic: For the sync script, we need a way to replace the table in its place (as opposed to deleting and creating). `hs import` isn't necessary if there is a way to delete all rows AND update a table.

I think `import` might get conflated with the actual `/import` hubdb endpoint.

packages/cms-lib/hubdb.js
anthmatic: Good point, I've updated this

packages/cms-lib/hubdb.js
anthmatic: Changed to `addRows`

packages/cms-lib/hubdb.js
```diff
@@ -50,20 +49,40 @@ async function createHubDbTable(portalId, src) {
...
+  validateJsonFile(src);
+
+  const table = fs.readJsonSync(src);
+  const { rows, ...schema } = table;
+  const { columns, id } = await createTable(portalId, schema);
+
+  await updateRows(portalId, id, rows, columns);
+}
+
+async function updateHubDbTable(portalId, tableId, src) {
```
anthmatic, line 76: @gcorne the idea behind this is that it would:
- Update the table meta/columns
- Add/append any rows that get passed to it

In the UI, you can either `append` or `replace` your rows. This solves for appending, but you'd need to couple this with `clearHubDbTableRows` in order to replace

gcorne: I think we should split the updating of the table from the adding of the rows, so remove line 83 and return the updated table and then in the code using these new functions add the rows separately, which will require exporting `addRows`.

gcorne:
FYI, it is worth renaming the PR to drop "[WIP]" once the PR is out of the draft phase.

boulter:
This is cool. Would it be possible to package these up in way that we could include them in the package available to serverless functions?

anthmatic:
@boulter are you thinking that it would be better to have a different package that's more focused on content APIs than `@hubspot/cms-lib`?

cc @gcorne

boulter:
Yeah, basically just all the stuff in https://github.com/HubSpot/hubspot-cms-tools/tree/master/packages/cms-lib/api

gcorne:
@boulter we may want to provide https://www.npmjs.com/package/@hubspot/api-client instead.

PR #140, opened 2020-03-22
Improve usage tracking of the authentication flow during init

gcorne:
@williamspiro


packages/cms-lib/lib/config.js
```diff
@@ -256,7 +256,7 @@ const getAndLoadConfigIfNeeded = () => {
       silenceErrors: true,
     });
   }
-  return _config;
+  return _config || {};
```
gcorne: @miketalley @corymartin I am curious about whether or not you think that this is reasonable. There are situations where `getEnv()` and `getRequestOptions()` are called prior to a config being created or initialized and so I thought falling back to an empty config was reasonable.

PR #141, opened 2020-03-25
Upgrade packages to fix minimist security advisory

gcorne:
See https://github.com/maxogden/extract-zip/releases/tag/v1.6.8

@miketalley

@axellabs you should be able to do the same for cms-theme-boilerplate now.


gcorne:
Turns out that this didn't fix it for us due to a dep on handlebars relatively deep in lerna's deps. https://github.com/wycats/handlebars.js/pull/1662 is intended to fix handlebars.

PR #143, opened 2020-03-26
Allow fetch to work with extended characters

### LINKED ISSUE #135: Fetching paths with extended characters is not working as expected
Reported via https://hubspotdev.slack.com/archives/CLJJXGTC0/p1584425976101100

image

miketalley:
Kanji characters in filenames were causing the `hs fetch` command to encounter an error due to "unescaped characters".

- Escape `uri` before sending request for file to backend

Fixes #135

### DIFF
--- a/packages/cms-lib/api/fileMapper.js
+++ b/packages/cms-lib/api/fileMapper.js
@@ -91,7 +91,7 @@ async function fetchFileStream(portalId, filePath, destination, options = {}) {
   const response = await http.getOctetStream(
     portalId,
     {
-      uri: `${FILE_MAPPER_API_PATH}/stream/${filePath}`,
+      uri: `${FILE_MAPPER_API_PATH}/stream${filePath}`,
       ...options,
     },
     destination
--- a/packages/cms-lib/api/fileMapper.js
+++ b/packages/cms-lib/api/fileMapper.js
@@ -91,7 +91,7 @@ async function fetchFileStream(portalId, filePath, destination, options = {}) {
   const response = await http.getOctetStream(
     portalId,
     {
-      uri: `${FILE_MAPPER_API_PATH}/stream${filePath}`,
+      uri: `${FILE_MAPPER_API_PATH}/stream/${encodeURIComponent(filePath)}`,
       ...options,
     },
     destination
--- a/packages/cms-lib/http.js
+++ b/packages/cms-lib/http.js
@@ -106,15 +106,8 @@ const createGetRequestStream = ({ contentType }) => async (
   options,
   destination
 ) => {
-  const { query, uri, ...rest } = options;
-  const requestOptions = addQueryParams(
-    {
-      ...rest,
-      // Handle extended characters(kanji, etc)
-      uri: encodeURI(uri),
-    },
-    query
-  );
+  const { query, ...rest } = options;
+  const requestOptions = addQueryParams(rest, query);
   // Using `request` instead of `request-promise` per the docs so
   // the response can be piped.
   // https://github.com/request/request-promise#api-in-detail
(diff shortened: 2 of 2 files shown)


packages/cms-lib/http.js
```diff
@@ -106,8 +106,15 @@ const createGetRequestStream = ({ contentType }) => async (
...
   destination
 ) => {
-  const { query, ...rest } = options;
-  const requestOptions = addQueryParams(rest, query);
+  const { query, uri, ...rest } = options;
+  const requestOptions = addQueryParams(
+    {
+      ...rest,
+      // Handle extended characters(kanji, etc)
+      uri: encodeURI(uri),
```
gcorne: I think that this is okay, although I wondered if `encodeURIComponent()` in `fetchFileStream` was the way to go like we do in the `upload` api helper.

packages/cms-lib/api/fileMapper.js
```diff
@@ -91,7 +91,7 @@ async function fetchFileStream(portalId, filePath, destination, options = {}) {
   const response = await http.getOctetStream(
     portalId,
     {
-      uri: `${FILE_MAPPER_API_PATH}/stream/${filePath}`,
+      uri: `${FILE_MAPPER_API_PATH}/stream${filePath}`,
```
gcorne: Is the assumption that `filePath` will always be prefixed with `/`? Are we sure that is true?

corymartin: @miketalley You could extract this logic into a reusable function to format and ensure slashes.
https://github.com/HubSpot/hubspot-cms-tools/blob/3c1cd1c4391189af2dbb61187632858789047ce8/packages/cms-lib/api/fileMapper.js#L16-L21

miketalley: I tried it with several files/directories/subdirectories and it works fine. This was just a superflous `/` that I removed, but I suppose it doesn't hurt to leave it either. Thoughts?

packages/cms-lib/http.js
miketalley: I did that initially but the requests all fail when using `encodeURIComponent` in place of `encodeURI`.

corymartin: if `uri` is a complete URI then we want to use `encodeURI`.  `encodeURIComponent` is for paths, queries, etc.

```javascript
encodeURIComponent('https://foo.com?x=a b') == "https%3A%2F%2Ffoo.com%3Fx%3Da%20b"

encodeURI('https://foo.com?x=a b') == "https://foo.com?x=a%20b"
```

packages/cms-lib/api/fileMapper.js
corymartin: The logic I linked above is already in this file. I'd suggest pulling extracting it to as helper to be called upon `filePath`.

packages/cms-lib/http.js
gcorne: @corymartin yeah, i was suggesting we do it only to the path like we do in the upload api helper via something like `${FILE_MAPPER_API_PATH}/download/${encodeURIComponent(filepath)}`. I actually think we have to do in the hlpers to avoid double encoding.

packages/cms-lib/api/fileMapper.js
```diff
@@ -91,7 +91,7 @@ async function fetchFileStream(portalId, filePath, destination, options = {}) {
   const response = await http.getOctetStream(
     portalId,
     {
-      uri: `${FILE_MAPPER_API_PATH}/stream/${filePath}`,
+      uri: `${FILE_MAPPER_API_PATH}/stream/${encodeURIComponent(filePath)}`,
```
gcorne: I think `download` also needs to be updated.

PR #149, opened 2020-03-30
Config source control warning only shows when necessary

miketalley:
From what I can tell, this was not working properly, but was masked by the fact that the config location was moved to `/Users/<username>`. If I have a `hubspot.config.yml` file in my project directory and have also added `hubspot.config.yml` to my `.gitignore` this message is still triggered(false positive).

I changed the way that we are checking if the config file is ignored properly and also made it more dynamic so it works with `hubspot.config.yml` and `hubspot.config.yaml`(based on what the config file extension is) as well as `hubspot.config.*`(which catches both).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,7 @@ node_modules
 tmp
 context
 coverage
-hubspot.config.yml
-hubspot.config.yaml
+hubspot.config.*
 lerna-debug.log
 npm-debug.log
 npm-debug.log.*
--- a/packages/cms-lib/lib/config.js
+++ b/packages/cms-lib/lib/config.js
@@ -125,19 +125,36 @@ const isConfigPathInGitRepo = () => {
   return configDir.startsWith(gitDir);
 };
 
-const CONFIG_GITIGNORE_PATTERN = 'hubspot.config.*';
+const configFilenameIsIgnoredByGitignore = () => {
+  const ignoreFiles = getGitignoreFiles();
+  const configPathSegments = _configPath.split('/');
+  const configFilename = configPathSegments[configPathSegments.length - 1];
+
+  return ignoreFiles.some(gitignore => {
+    const gitignoreContents = fs
+      .readFileSync(gitignore, 'utf8')
+      .split('\n')
+      .filter(i => i);
+    if (
+      ignore()
+        .add(gitignoreContents)
+        .ignores(configFilename)
+    ) {
+      // Has a gitignore rule
+      return true;
+    }
+    return false;
+  });
+};
 
 const shouldWarnOfGitInclusion = () => {
   if (!isConfigPathInGitRepo()) {
     // Not in git
     return false;
   }
-  const ignoreFiles = getGitignoreFiles();
-  for (const gitignore in ignoreFiles) {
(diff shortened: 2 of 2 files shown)


packages/cms-lib/lib/config.js
```diff
@@ -125,19 +125,38 @@ const isConfigPathInGitRepo = () => {
...
+  const ignoreFiles = getGitignoreFiles();
+  const configPathSegments = _configPath.split('/');
+  const configFilename = configPathSegments[configPathSegments.length - 1];
+
+  return ignoreFiles.some(gitignore => {
+    const gitignoreContents = fs
+      .readFileSync(gitignore, 'utf8')
+      .split('\n')
+      .filter(i => i);
+    const gitignoreConfig = ignore().add(gitignoreContents);
```
gcorne: FWIW, supposedly `ignore` can take a string directly via `ignore().add(fs.readFileSync(filename).toString())` based on https://github.com/kaelzhang/node-ignore#addignorefilepath

packages/cms-lib/lib/config.js
```diff
@@ -125,19 +125,38 @@ const isConfigPathInGitRepo = () => {
...
+    if (
+      gitignoreConfig.ignores(configFilename) ||
+      gitignoreConfig.ignores(CONFIG_FILE_WILDCARD_MATCHER)
```
gcorne: Why is this necessary?

packages/cms-lib/lib/config.js
```diff
@@ -125,19 +125,38 @@ const isConfigPathInGitRepo = () => {
   return configDir.startsWith(gitDir);
 };
 
-const CONFIG_GITIGNORE_PATTERN = 'hubspot.config.*';
+const CONFIG_FILE_WILDCARD_MATCHER = 'hubspot.config.*';
+
+const configFilenameIsIgnoredByGitignore = () => {
```
gcorne: It would be nice if there was a unit test or two covering this.

.gitignore
```diff
@@ -2,8 +2,7 @@ node_modules
 tmp
 context
 coverage
-hubspot.config.yml
-hubspot.config.yaml
+hubspot.config.*
```
gcorne: Not sure I agree with this change since the CLI only supports yaml/yml.

packages/cms-lib/lib/config.js
miketalley: The security message suggested adding `hubspot.config.*` to `.gitignore` if there was no line ignoring the config file -- since the config only supports `hubspot.config.yml` and `hubspot.config.yaml` and that is easier than explaining that they need two separate lines and the reason for it.

So, in addition to checking if the line in the `.gitignore` matches the actual filename of the config file(including extension). We also check for the existence of the wildcard match, which is also sufficient to make sure the config does not get checked into source control.

gcorne: I am not sure that makes a ton of sense. The thing we're trying to detect is situations where a user's config file is not in `.gitignore` and so could accidentally be committed to source control, which runs the risk of accidental exposure. It doesn't matter how they ignore it.

PR #151, opened 2020-04-01
Tweaks and tests for config source control warning

miketalley:
Related to PR https://github.com/HubSpot/hubspot-cms-tools/pull/149/files

### DIFF
--- a/packages/cms-lib/lib/config.js
+++ b/packages/cms-lib/lib/config.js
@@ -27,6 +27,12 @@ let environmentVariableConfigLoaded = false;
 
 const getConfig = () => _config;
 
+const getConfigFilename = () => {
+  const configPathSegments = _configPath.split('/');
+
+  return configPathSegments[configPathSegments.length - 1];
+};
+
 const setConfig = updatedConfig => {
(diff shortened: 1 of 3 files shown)


packages/cms-lib/lib/config.js
```diff
@@ -131,24 +137,12 @@ const isConfigPathInGitRepo = () => {
...
-      .filter(i => i);
+    const gitignoreContents = fs.readFileSync(gitignore).toString();
     const gitignoreConfig = ignore().add(gitignoreContents);
-    if (
-      gitignoreConfig.ignores(configFilename) ||
-      gitignoreConfig.ignores(CONFIG_FILE_WILDCARD_MATCHER)
-    ) {
-      // Has a gitignore rule
+
+    if (gitignoreConfig.ignores(getConfigFilename())) {
```
gcorne, line 145: I think this needs to be a path relative to the location of the gitignore file being processed.

corymartin, line 145: @miketalley We need to make sure `getConfigFilename()` is a relative path or `ignores()` will throw.

PR #152, opened 2020-04-02
Fixed issue with relative paths when checking if config file is ignored

miketalley:


packages/cms-lib/lib/config.js
```diff
@@ -27,10 +27,13 @@ let environmentVariableConfigLoaded = false;
 
 const getConfig = () => _config;
 
-const getConfigFilename = () => {
+const getRelativeConfigFilename = () => {
   const configPathSegments = _configPath.split('/');
 
-  return configPathSegments[configPathSegments.length - 1];
+  return path.relative(
+    process.cwd(),
```
gcorne, line 34: This needs to be relative to the ignore file and this code makes the assumption that the path is relative to process.cwd().

packages/cms-lib/lib/config.js
```diff
@@ -139,10 +142,10 @@ const isConfigPathInGitRepo = () => {
 
 const configFilenameIsIgnoredByGitignore = ignoreFiles => {
   return ignoreFiles.some(gitignore => {
-    const gitignoreContents = fs.readFileSync(gitignore).toString();
+    const gitignoreContents = fs.readFileSync(gitignore);
```
gcorne, line 145: Should this be changed? I thought that `gitignoreContents` needed to be a `string` not a `Buffer`.

gcorne:
I think we also should make sure that exceptions when checking `gitignore` do not impact execution of commands since if there is a violation it is just a warning.

packages/cms-lib/lib/config.js
```diff
@@ -27,10 +27,13 @@ let environmentVariableConfigLoaded = false;
 
 const getConfig = () => _config;
 
-const getConfigFilename = () => {
+const getRelativeConfigFilename = () => {
   const configPathSegments = _configPath.split('/');
```
gcorne, line 31: FWIW, this code assumes that `/` is a separator. I believe we have code elsewhere that determines the separator based on the operating system.

packages/cms-lib/lib/config.js
miketalley: I changed this back. Strange how this works on my mac, but not on windows.

packages/cms-lib/lib/config.js
```diff
@@ -142,7 +136,7 @@ const configFilenameIsIgnoredByGitignore = ignoreFiles => {
     const gitignoreContents = fs.readFileSync(gitignore).toString();
     const gitignoreConfig = ignore().add(gitignoreContents);
 
-    if (gitignoreConfig.ignores(getConfigFilename())) {
+    if (gitignoreConfig.ignores(path.basename(_configPath))) {
```
gcorne, line 139: While this works in most situations, this doesn't work in cases where the `.gitignore` rule only applies based on the path.

gcorne: This can be addressed with a followup.

PR #156, opened 2020-04-03
Calculate relative path using directory

gcorne:
TODO:

- [ ] Add test coverage

@miketalley this is something that I discovered when digging into #152


gcorne:
Actually, it looks like I jumped the gun on second-guessing the code for `.hsignore`. See https://github.com/HubSpot/hubspot-cms-tools/blob/master/packages/cms-lib/ignoreRules.js#L40

PR #157, opened 2020-04-03
Add ability to create a function

gcorne:
@williamspiro @miketalley @corymartin @mattcoley


packages/cms-cli/lib/createFunctionPrompt.js
```diff
@@ -0,0 +1,66 @@
...
+      return 'The name may not be blank. Please try again.';
+    } else if (!STRING_WITH_NO_SPACES_REGEX.test(val)) {
+      return 'The name may not contain spaces. Please try again.';
+    }
+    return true;
+  },
+};
+const ENDPOINT_PATH_PROMPT = {
+  name: 'endpointPath',
+  message: 'Path portion of the URL created for the function',
```
miketalley, line 21: Should we give more context to the user on how this works? Maybe a link?

packages/cms-cli/lib/createFunctionPrompt.js
```diff
@@ -0,0 +1,66 @@
...
+    } else if (!val.length) {
+      return 'The name may not be blank. Please try again.';
+    } else if (!STRING_WITH_NO_SPACES_REGEX.test(val)) {
+      return 'The name may not contain spaces. Please try again.';
+    }
+    return true;
+  },
+};
+const ENDPOINT_METHOD_PROMPT = {
+  type: 'list',
```
miketalley, line 34: Do we want to use `rawList` here? We had feedback from users of `list` that they would prefer to be able to select with both arrows as well as numbers.

packages/cms-lib/functions.js
```diff
@@ -0,0 +1,141 @@
...
+    secrets: [],
+    endpoints: {
+      [endpointPath]: createEndpoint(endpointMethod, filename),
+    },
+  };
+}
+
+function writeConfig(configFilePath, config) {
+  const configJson = JSON.stringify(config, null, '  ');
+  fs.writeFileSync(configFilePath, configJson);
```
corymartin, line 39: Both of these lines can throw but I don't see the consuming functions catching.

PR #159, opened 2020-04-03
Added list of types to help for hs create

### LINKED ISSUE #146: Add supported types to hs help create
We should add the supported `<type>`s to the CLI documentation when you run `hs help create`.

miketalley:
Fixes #146

### DIFF
--- a/packages/cms-cli/commands/create.js
+++ b/packages/cms-cli/commands/create.js
@@ -69,7 +69,7 @@ function configureCreateCommand(program) {
   program
     .version(version)
     .description(
-      `Create HubSpot CMS assets. Supported authentication protocols are ${Object.values(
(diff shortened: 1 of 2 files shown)


packages/cms-cli/commands/create.js
```diff
@@ -68,7 +68,11 @@ const createTemplate = (name, dest, type = 'template') => {
 function configureCreateCommand(program) {
   program
     .version(version)
-    .description('Create HubSpot CMS assets')
+    .description(
+      `Create HubSpot CMS assets. Supported assets are ${Object.values(
+        TYPES
+      ).join(', ')}.`
```
gcorne, line 74: Should there be an and after the last comma?

miketalley: Fixed in https://github.com/HubSpot/hubspot-cms-tools/pull/183

PR #165, opened 2020-04-06
Use same CWD logic when looking for config as we use elsewhere

gcorne:


gcorne:
I think we should move forward with this to make behavior more consistent.

PR #184, opened 2020-04-28
Added --qa flag to hs auth and hs init

miketalley:
- Adds `--qa` flag to `hs init` and `hs auth` commands to add portal entry in config as QA portal
  - Opens correct browser URL based on flag
- Portal config properties are ordered consistently when writing config
  1. [`name`]
  1. `portalId`
  1. [`env`]
  1. `authType`
  1. `...rest`

### DIFF
--- a/packages/cms-cli/lib/commonOpts.js
+++ b/packages/cms-cli/lib/commonOpts.js
@@ -34,6 +34,10 @@ const addModeOptions = (program, { read, write }) => {
   program.option('--mode <mode>', help);
 };
 
+const addTestingOptions = program => {
+  program.option('--qa', 'run command in qa mode', false);
+};
+
 const setLogLevel = (options = {}) => {
   const { debug } = options;
   if (debug) {
@@ -82,6 +86,7 @@ module.exports = {
   addConfigOptions,
   addOverwriteOptions,
   addModeOptions,
+  addTestingOptions,
   setLogLevel,
   getPortalId,
   getMode,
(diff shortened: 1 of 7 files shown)


packages/cms-cli/lib/commonOpts.js
```diff
@@ -34,6 +34,10 @@ const addModeOptions = (program, { read, write }) => {
   program.option('--mode <mode>', help);
 };
 
+const addTestingOptions = program => {
+  program.option('--qa', 'run command in qa mode', false);
```
gcorne, line 38: Can we somehow hide this from documentation?

packages/cms-lib/personalAccessKey.js
```diff
@@ -162,11 +163,12 @@ const personalAccessKeyPrompt = async () => {
  * @param {boolean} makeDefault option to make the portal being added to the config the default portal
  */
 const updateConfigWithPersonalAccessKey = async (configData, makeDefault) => {
-  const { personalAccessKey, name } = configData;
+  const { personalAccessKey, name, env } = configData;
+  const accessKeyEnv = env || getEnv(name);
```
gcorne, line 167: Why is this named `accessKeyEnv`?

packages/cms-lib/lib/config.js
```diff
@@ -109,7 +121,7 @@ const getOrderedConfig = unorderedConfig => {
     defaultMode,
     httpTimeout,
     allowsUsageTracking,
-    portals,
+    portals: portals.map(getOrderedPortal),
```
gcorne, line 124: In general, I am not super comfortable with auto-formatting data that is outside the context of the command.

PR #197, opened 2020-05-12
Added missing @hubspot/cms-lib dependency for api-auth-lib

miketalley:
/cc @levinkeo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Change Log
 ## 1.1.4
 
  * Option `--qa` is hidden from help text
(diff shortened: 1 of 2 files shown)


gcorne:
Doesn't this create a circular dependency?

gcorne:
Something doesn't make sense to me because `@hubspot/cms-webpack-plugins` only has a dep on `@hubspot/cms-lib`. I think that the actual cause of the reported issue may be due to having a dep on `@hubspot/cms-webpack-plugins` and `@hubpsot/cms-cli` and there being a mismatch with `@hubspot/cms-lib`. I think it would be good to test the webpack plugin independent of the CLI and see what happens.

PR #201, opened 2020-05-14
Add `create react-app`

### LINKED ISSUE #12: Add skeleton / boilerplate for creating a react app
Basically add `hs create react-app`.

anthmatic:
Adds `react-app` as a new type of project under the `create` command.
Adds a more generic `cms-lib/projects` to set us up to add more boilerplate projects to the CLI down the road.
Moves `website-theme` over to the `projects`

Resolves #12


packages/cms-cli/lib/createReactAppPrompt.js
```diff
@@ -0,0 +1,17 @@
...
+  choices: ['Template', 'Module'],
+};
+
+function createReactAppPrompt() {
+  const prompt = inquirer.createPromptModule();
+  return prompt([APP_TYPE_PROMPT]);
+}
+module.exports = {
+  createReactAppPrompt,
+};
```
anthmatic, line 17: Will remove this if we don't need it yet

packages/cms-cli/commands/create.js
```diff
@@ -125,6 +140,10 @@ function configureCreateCommand(program) {
         case TYPES['website-theme']:
           createTheme(dest, type, program);
```
anthmatic, line 141: TODO: Refactor `website-theme` to use `createProject` which needs a special case to only copy `/src`

packages/cms-cli/commands/create.js
```diff
@@ -39,6 +40,11 @@ const ASSET_PATHS = {
   ),
 };
 
+const PROJECT_REPOSITORIES = {
+  [TYPES['react-app']]: 'cms-react-boilerplate',
```
anthmatic, line 44: Discussed briefly with @levinkeo but we may want to use a different repo for the react module

miketalley: Would it make sense to have a mono repo? I'm wondering if down the road we have `hs create <insert framework of choice here>-app` it may be easier to manage them and have consistency when referencing things from other projects(like here for instance).

anthmatic: Good idea, that's definitely something we're going to have to consider because we'll definitely have at least a few flavors of this.

PR #202, opened 2020-05-14
Add clear and delete commands

### LINKED ISSUE #192: Add `hs hubdb clear` and `hs hubdb delete`

anthmatic:
Adds 2 new HubDB actions to for HubDB:

`clear`: looks up all rows, deletes them, publishes the table
`delete`: deletes a table
﻿﻿
image

Fixes #192

### DIFF
--- a/packages/cms-cli/commands/hubdb.js
+++ b/packages/cms-cli/commands/hubdb.js
@@ -12,7 +12,7 @@ const {
   downloadHubDbTable,
   clearHubDbTableRows,
 } = require('@hubspot/cms-lib/hubdb');
-const { publishTable } = require('@hubspot/cms-lib/api/hubdb');
+const { publishTable, deleteTable } = require('@hubspot/cms-lib/api/hubdb');
 
 const { validatePortal } = require('../lib/validation');
 const { addHelpUsageTracking } = require('../lib/usageTracking');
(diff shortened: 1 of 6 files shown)


packages/cms-cli/commands/hubdb.js
```diff
@@ -105,8 +109,79 @@ function configureHubDbFetchCommand(program) {
...
+          logger.log(
+            `Removed ${deletedRowCount} rows from HubDB table ${tableId}`
+          );
+          const { rowCount } = await publishTable(portalId, tableId);
+          logger.log(`HubDB table ${tableId} now contains ${rowCount} rows`);
+        } else {
+          logger.log(`HubDB table ${tableId} is already empty`);
+        }
+      } catch (e) {
+        logger.error(e);
```
gcorne, line 141: I suggest using `logErrorInstance` for this because it has some extra stuff that makes the output more friendly. We need to do the same for the other hubdb commands. See https://github.com/HubSpot/hubspot-cms-tools/blob/master/packages/cms-lib/errorHandlers.js#L241

anthmatic: Ah! Gotcha, I swapped those out. I there are a couple of `logger.log()`s that I kept since they provide more context. Is that ok?

PR #203, opened 2020-05-14
Support working with hubdb via table name

### LINKED ISSUE #185: Support HubDB fetch by name

anthmatic:
Changes `<tableId>` to `<table`> . Looks up table by name unless a number (tableId) is passed as an arg.

Fixes #185


gcorne:
What if we switch to the v3 HubDB endpoint instead since it supports names?

anthmatic:
That would certainly be a lot nicer. I'll give that a shot.

anthmatic:
Closing in favor or #219

PR #204, opened 2020-05-14
Converting --qa flag "inheritance" for commands into separate commands entirely

miketalley:
- Removes ability to dynamically add the `--qa` flag to commands
- Creates a `hs qa` command with subcommands that are used only for qa environment and not displayed in help output
  - `hs qa init`
  - `hs qa auth <type>`


gcorne:
What if we use an environment variable instead?

```
HS_ENV=QA hs init
```

miketalley:
Closing this since the yargs migration will allow us the ability to have qa functionality not clutter the help output -- https://github.com/HubSpot/hubspot-cms-tools/pull/282#discussion_r470768464

PR #206, opened 2020-05-15
Cross-package dependency simplification

miketalley:
## Brief Description
This PR looks to simplify dependencies,  and eliminate the `@hubspot/api-auth-lib` package by doing so. `OAuth2Manager` is responsible for both the initial creation of oauth configurations as well as providing the oauth information to `http` requests that are made to execute commands within the CLI. The initial creation is the only part that requires the use of `express` so we are looking to separate the two responsibilities in order for only `@hubspot/cms-cli` to have a dependency on `express`, keep `@hubspot/cms-lib` free of bloat, and remove the possibility of runtime errors within `@hubspot/cms-lib` due to no explicit dependency on `@hubspot/api-auth-lib`.

- Removed `OAuth2Manager` from `api-auth-lib`
  - `oauth2` creation logic that requires use of `express` moved into `@hubspot/cms-cli` package as it is only used for the `auth` command
  - `oauth2` auth logic remains in `OAuth2Manager` but is now located in `@hubspot/cms-lib/lib/models`
- Moved `Errors.js`(contains `HubSpotAuthError`) from `@hubspot/api-auth-lib` to `@hubspot/cms-lib`
- Moved `prompts.js` from `@hubspot/cms-cli` to `@hubspot/cms-lib`
- Moved `personalAccessKeyPrompt` from `personalAccessKey.js` into `prompts.js`
- Removed `@hubspot/api-auth-lib` package from repo


miketalley:
The tests are failing due to a dependency mismatch issue with lerna. Looking into it.

.github/workflows/build.yml
```diff
@@ -13,6 +13,9 @@ jobs:
     - name: Install deps
       run: |
           yarn install
+    - name: Force local package symlinks
+      run: |
+          yarn lerna link --force-local
```
gcorne, line 18: Should we also be running `yarn lerna bootstrap`?

packages/api-auth-lib/package.json
```diff
@@ -1,24 +0,0 @@
-{
-  "name": "@hubspot/api-auth-lib",
```
gcorne, line 2: TBH, I am not 100% sold on removing this package entirely.

packages/cms-cli/package.json
```diff
@@ -8,8 +8,7 @@
     "url": "https://github.com/HubSpot/hubspot-cms-tools"
   },
   "dependencies": {
-    "@hubspot/api-auth-lib": "^1.1.7",
-    "@hubspot/cms-lib": "^1.1.10-beta.0",
+    "@hubspot/cms-lib": "^1.1.4",
```
gcorne, line 11: Shouldn't this be `^1.1.10-beta.0`?

packages/cms-lib/package.json
```diff
@@ -13,12 +13,14 @@
     "chokidar": "^3.0.1",
     "content-disposition": "^0.5.3",
     "debounce": "^1.2.0",
+    "express": "4.16.3",
```
gcorne, line 16: One of the reasons that I am not fond of bringing `@hubspot/api-auth-lib` into `@hubspot/cms-lib` is that includes express. This makes using `@hubspot/cms-lib` in non-traditional ways inside of the context of a web app that is built on top of express. It would be preferrer if somehow the authentication flow that uses `open` and `express` were part of the `@hubspot/cms-cli` and the logic needed to make API requests using OAuth and refreshing token lived in `@hubspot/cms-lib`. Does that make sense?

packages/cms-cli/lib/oauth.js
```diff
@@ -0,0 +1,103 @@
+const express = require('express');
```
drewjenkins, line 1: Anything changed in this file or was it straight port?

packages/cms-lib/package.json
```diff
@@ -19,6 +19,7 @@
     "ignore": "^5.1.4",
     "js-yaml": "^3.12.2",
     "moment": "^2.24.0",
+    "open": "6.4.0",
```
drewjenkins, line 22: Is this needed? Looks like the `open` require is in cms-cli which has the dep already.

packages/cms-cli/package.json
```diff
@@ -8,9 +8,9 @@
     "url": "https://github.com/HubSpot/hubspot-cms-tools"
   },
   "dependencies": {
-    "@hubspot/api-auth-lib": "^1.1.7",
     "@hubspot/cms-lib": "^2.0.0-beta.2",
     "commander": "^2.16.0",
+    "express": "^4.17.1",
     "inquirer": "^6.3.1",
     "open": "^7.0.3",
```
drewjenkins, line 15: This used to be using open@6.4.0 in cms-lib.  Are there any breaking changes we are concerned with going to 7.0.3?

anthmatic:
Tested this out and it seems to behave the same as before!

packages/cms-cli/package.json
miketalley: Good catch, there was a change to use `{ url: true }` for the options. I will update the code .

packages/cms-lib/package.json
miketalley: `open` is used in both `cms-cli` and `cms-lib` now.

gcorne:
Are we sure that we want to move `prompts.js` to `@hubspot/cms-lib`? Doesn't that require adding dependencies to `@hubspot/cms-lib`?

miketalley:
> Are we sure that we want to move `prompts.js` to `@hubspot/cms-lib`? Doesn't that require adding dependencies to `@hubspot/cms-lib`?

It does require a dependency on `open`. Is this something we don't want?

gcorne:
> It does require a dependency on open. Is this something we don't want?

Yeah, I thought so. We intentionally moved prompts into CLI to keep them separated in 53e1d38d3c2a3a882595713a152ac4dc9590a7c.

gcorne:
@miketalley FWIW, it looks like the move also introduced a dependency on `inquirer` to `@hubspot/cms-lib` which isn't reflected in `package.json`.

PR #209, opened 2020-05-20
Require dest arg in create command

anthmatic:
Makes the `dest` argument mandatory when running `hs create`. Previously this was enforced which caused inconsistent results like errors or output to the CWD.


anthmatic:
Closing in favor of #210

PR #210, opened 2020-05-20
Add default folder name for website-theme, function, react-app projects

anthmatic:
If a user doesn't include a `dest` when running `hs create` the existing behavior is to create the project in the CWD. However, this led to unexpected results with some project types. This change adds a default `dest`, which is the `type` of project.

### DIFF
--- a/packages/cms-cli/commands/create.js
+++ b/packages/cms-cli/commands/create.js
@@ -106,12 +106,16 @@ function configureCreateCommand(program) {
       }
 
       // TODO: In yargs use `.implies()`
-      if (
-        [TYPES['website-theme'], TYPES.function, TYPES['react-app']].includes(
-          type
-        )
-      ) {
-        dest = name || type;
+      switch (type) {
+        case TYPES.function:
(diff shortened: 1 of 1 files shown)


packages/cms-cli/commands/create.js
```diff
@@ -111,7 +111,7 @@ function configureCreateCommand(program) {
           type
         )
       ) {
-        dest = name;
+        dest = name || type;
```
gcorne, line 114: I think this makes sense for `website-theme` and `react-app` but not for `function`. What if we did it just for `website-theme` and `react-app`?

anthmatic: I updated this. I am curious if the switch pattern works or if it would make sense to stick with the `[].includes()` pattern?

gcorne: The switch works fine. Another option would be to have some of config object that could be used to look up the config based on the type. If the different types continue to expand in complexity, it may make sense to go in that direction.

PR #215, opened 2020-05-27
Add filemanager fetch command

anthmatic:
TODO:
- Account for paged results
- Fix/Add logging

### DIFF
--- a/packages/cms-cli/commands/filemanager.js
+++ b/packages/cms-cli/commands/filemanager.js
@@ -21,6 +21,7 @@ const {
 } = require('@hubspot/cms-lib/errorHandlers');
 const { validateSrcAndDestPaths } = require('@hubspot/cms-lib/modules');
 const { shouldIgnoreFile } = require('@hubspot/cms-lib/ignoreRules');
+const { resolveLocalPath } = require('../lib/filesystem');
 
 const {
   addConfigOptions,
@@ -73,7 +74,7 @@ function configureFileManagerFetchCommand(program) {
         process.exit(1);
       }
 
-      // dest = resolveLocalPath(dest);
+      dest = resolveLocalPath(dest);
 
       const portalId = getPortalId(program);
 
--- a/packages/cms-cli/commands/filemanager.js
+++ b/packages/cms-cli/commands/filemanager.js
@@ -37,7 +37,8 @@ const {
   addHelpUsageTracking,
 } = require('../lib/usageTracking');
 
-const COMMAND_NAME = 'filemanager-upload';
+const UPLOAD_COMMAND_NAME = 'filemanager-upload';
+const FETCH_COMMAND_NAME = 'filemanager-fetch';
 
 function configureFileManagerCommand(program) {
   program
@@ -78,7 +79,7 @@ function configureFileManagerFetchCommand(program) {
 
       const portalId = getPortalId(program);
 
-      trackCommandUsage(COMMAND_NAME, null, portalId);
+      trackCommandUsage(FETCH_COMMAND_NAME, null, portalId);
 
       // Fetch and write file/folder.
       downloadFileOrFolder(portalId, src, dest, program);
@@ -87,7 +88,7 @@ function configureFileManagerFetchCommand(program) {
   addConfigOptions(program);
   addPortalOptions(program);
   addLoggerOptions(program);
-  addHelpUsageTracking(program, COMMAND_NAME);
+  addHelpUsageTracking(program, FETCH_COMMAND_NAME);
 }
 
 function configureFileManagerUploadCommand(program) {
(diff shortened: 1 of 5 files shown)


packages/cms-lib/fileManager.js
```diff
@@ -51,6 +58,77 @@ async function uploadFolder(portalId, src, dest, { cwd }) {
...
+ *
+ * @param {number} portalId
+ * @param {string} dest
+ * @param {string} folderPath
+ */
+async function getFolderContents(portalId, dest, folderPath) {
+  const files = await getFilesByPath(portalId, folderPath);
+  files.objects.forEach(async file => {
+    fetchFile(portalId, file, dest, folderPath);
+  });
```
gcorne, line 103: I think we're going to need to use an old school for loop here. See https://medium.com/dailyjs/the-pitfalls-of-async-await-in-array-loops-cf9cf713bfeb

packages/cms-lib/fileManager.js
```diff
@@ -51,6 +58,77 @@ async function uploadFolder(portalId, src, dest, { cwd }) {
...
+/**
+ *
+ * @param {number} portalId
+ * @param {object} file
+ * @param {string} dest
+ * @param {string} folderPath
+ */
+async function fetchFile(portalId, file, dest, folderPath) {
+  const relativePath = `${folderPath}/${file.name}.${file.extension}`;
+  const destPath = convertToUnixPath(path.join(dest, relativePath));
```
gcorne, line 70: You'll want to make sure to always use `path.join` when working with paths for the local file system as well as https://github.com/HubSpot/hubspot-cms-tools/blob/master/packages/cms-lib/path.js#L14-L23 so that everything will work as expected on windows.

packages/cms-cli/commands/filemanager.js
```diff
@@ -39,12 +43,53 @@ function configureFileManagerCommand(program) {
...
+      if (typeof src !== 'string') {
+        logger.error('A source to fetch is required');
+        process.exit(1);
+      }
+
+      dest = resolveLocalPath(dest);
+
+      const portalId = getPortalId(program);
+
+      trackCommandUsage(COMMAND_NAME, null, portalId);
```
gcorne, line 81: `COMMAND_NAME` will need to vary so that we can track fetch separately.

gcorne:
@anthmatic is this ready for review?

packages/cms-lib/fileManager.js
```diff
@@ -51,6 +60,119 @@ async function uploadFolder(portalId, src, dest, { cwd }) {
...
+  }
+
+  await http.getOctetStream(
+    portalId,
+    {
+      baseUrl: file.url,
+      uri: '',
+    },
+    writeStream
+  );
```
gcorne, line 100: I think that this is missing error handling.

packages/cms-lib/fileManager.js
```diff
@@ -51,6 +60,119 @@ async function uploadFolder(portalId, src, dest, { cwd }) {
...
 }
 
+/**
+ *
+ * @param {number} portalId
+ * @param {object} file
+ * @param {string} dest
+ * @param {string} folderPath
+ */
+async function fetchFile(portalId, file, dest, folderPath) {
```
gcorne, line 70: I think we should rename this so that it is clear that it is going to fetch and write the file. Often `fetch*` is indicative of a helper for making a GET request and returning the data.

packages/cms-lib/fileManager.js
```diff
@@ -51,6 +60,119 @@ async function uploadFolder(portalId, src, dest, { cwd }) {
...
+      logger.log(`File ${remotePath} was downloaded to ${localDest}`);
+    } catch (err) {
+      console.log(err);
+    }
+  } else if (folder) {
+    try {
+      await getFolderContents(portalId, localDest, folder.full_path);
+      logger.log(`Folder ${remotePath} was downloaded to ${localDest}`);
+    } catch (err) {
+      console.log(err);
```
gcorne, line 170: This should use an error handler.

packages/cms-lib/fileManager.js
```diff
@@ -51,6 +60,119 @@ async function uploadFolder(portalId, src, dest, { cwd }) {
...
+async function downloadFileOrFolder(portalId, remotePath, localDest) {
+  const { file, folder } = await getStat(portalId, remotePath);
+
+  if (file) {
+    const folderPath = path.dirname(remotePath);
+    try {
+      await fetchFile(portalId, file, localDest, folderPath);
+      logger.log(`File ${remotePath} was downloaded to ${localDest}`);
+    } catch (err) {
+      console.log(err);
```
gcorne, line 163: This should use an error handler.

gcorne:
Spent some testing this out and there are a couple of inconsistencies with `hs fetch` that i think we should address:

1. When downloading a folder, a folder is created in the destination instead of it being used directly.
2. The `hs filemanger fetch` command automatically overwrites files while `hs fetch` requires `--overwrite`.

@williamspiro @iamw11y I think we should revisit no. 2 since it has been a common source of confusion with `hs fetch`.

gcorne:
I also will add that I am okay with merging this and releasing a beta, so that it is easier for people to test.

williamspiro:
This is awesome @anthmatic! Thanks for getting this in. Going to file a couple of issues on things I notice

PR #236, opened 2020-06-24
Allow fetching of all FM files

### LINKED ISSUE #234: Cannot fetch filemanager root directory ("/")
`hs filemanager fetch /` results in a 404. We'll need to add some special handling to account for `/`

cc @ajlaporte

anthmatic:
Fixes #234 

If `/` is passed as a `src` arg in `filemanager fetch`, the entire file manager will be downloaded

This also explicitly filters out hidden files and, optionally, archived files.

cc @ajlaporte

### DIFF
--- a/packages/cms-lib/api/fileManager.js
+++ b/packages/cms-lib/api/fileManager.js
@@ -25,13 +25,13 @@ async function uploadFile(portalId, src, dest) {
   });
 }
 
-async function getStat(portalId, src) {
+async function fetchStat(portalId, src) {
   return http.get(portalId, {
     uri: `${FILE_MANAGER_API_PATH}/files/stat/${src}`,
   });
 }
 
-async function getFiles(portalId, folderId, { offset, archived }) {
+async function fetchFiles(portalId, folderId, { offset, archived }) {
   return http.get(portalId, {
     uri: `${FILE_MANAGER_API_PATH}/files/`,
     qs: {
@@ -43,7 +43,7 @@ async function getFiles(portalId, folderId, { offset, archived }) {
   });
 }
 
-async function getFolders(portalId, folderId) {
+async function fetchFolders(portalId, folderId) {
   return http.get(portalId, {
     uri: `${FILE_MANAGER_API_PATH}/folders/`,
     qs: {
@@ -55,7 +55,7 @@ async function getFolders(portalId, folderId) {
 
 module.exports = {
   uploadFile,
-  getStat,
-  getFiles,
-  getFolders,
+  fetchStat,
+  fetchFiles,
+  fetchFolders,
 };
--- a/packages/cms-lib/api/fileManager.js
+++ b/packages/cms-lib/api/fileManager.js
@@ -38,7 +38,7 @@ async function fetchFiles(portalId, folderId, { offset, archived }) {
       hidden: 0,
       offset: offset,
       folder_id: folderId || 'None',
-      ...(archived === 0 && { archived }),
+      ...(!archived && { archived: 0 }),
     },
   });
 }
--- a/packages/cms-lib/api/fileManager.js
+++ b/packages/cms-lib/api/fileManager.js
@@ -31,14 +31,14 @@ async function getStat(portalId, src) {
   });
 }
 
-async function getFiles(portalId, folderId, { offset }, showArchived) {
+async function getFiles(portalId, folderId, { offset, archived }) {
   return http.get(portalId, {
     uri: `${FILE_MANAGER_API_PATH}/files/`,
     qs: {
       hidden: 0,
       offset: offset,
       folder_id: folderId || 'None',
-      ...(!showArchived && { archived: 0 }),
+      ...(archived === 0 && { archived }),
     },
   });
 }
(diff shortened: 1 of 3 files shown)


packages/cms-lib/fileManager.js
```diff
@@ -186,52 +202,106 @@ async function fetchFolderContents(portalId, folder, dest, options) {
...
+ * @param {object} file
  * @param {object} options
  */
-async function downloadFileOrFolder(portalId, src, dest, options) {
-  const { file, folder } = await getStat(portalId, src);
-  if (file) {
+async function downloadFolder(portalId, src, dest, folder, options) {
+  if (isHidden(folder)) {
+    logger.error('Folder "%s" could not be found in portal %s', src, portalId);
+    return;
```
anthmatic, line 216: I thought about filtering out results at the request level (in `/api/filemanager`) but it seemed better to add messaging for the user similar to a regular 404.

packages/cms-lib/fileManager.js
```diff
@@ -170,6 +183,9 @@ async function getAllPagedFiles(portalId, folderPath) {
  * @param {object} options
  */
 async function fetchFolderContents(portalId, folder, dest, options) {
+  if (isHidden(folder)) {
+    return;
+  }
```
gcorne, line 188: Is it possible to handle these when fetching instead of filtering them out here?

anthmatic: I had looked at doing that briefly, will take another look

packages/cms-lib/fileManager.js
gcorne: I think I agree when it comes to the top-level folder, but I think we should be clearer with the messaging so that we're differentiating hidden folders from not found.

packages/cms-lib/fileManager.js
gcorne: I think doing it via the API request might match up well with #248 if we do something to provide the option of fetching hidden files.

anthmatic: I switched over to filtering hidden/archived out in the API request. The exception is that `files/stat` does not support filtering, so there I'm still checking manually for singe file fetches.

packages/cms-lib/api/fileManager.js
```diff
@@ -31,21 +31,31 @@ async function getStat(portalId, src) {
   });
 }
 
-async function getFilesByPath(portalId, src) {
+async function getFiles(portalId, folderId, { offset }, showArchived) {
```
gcorne, line 34: What if we added `archived` as a param like `offset`? The reason for suggesting that approach is that it avoids a [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap).

anthmatic: Thanks for pointing me to that article. I've updated this, I think its more readable now

packages/cms-lib/fileManager.js
```diff
@@ -170,68 +174,118 @@ async function getAllPagedFiles(portalId, folderPath) {
...
+ * Download a single file and write to local file system.
+ *
+ * @param {number} portalId
+ * @param {string} src
+ * @param {string} dest
+ * @param {object} file
+ * @param {object} options
+ */
+async function downloadSingleFile(portalId, src, dest, file, options) {
+  if (file.archived) {
```
gcorne, line 231: Should there be a check against `options` here to determine if the flag is passed?

packages/cms-lib/fileManager.js
```diff
@@ -170,68 +174,118 @@ async function getAllPagedFiles(portalId, folderPath) {
...
+    logger.error(
+      '"%s" in the File Manager is an archived file. Try fetching again with the "--inclulde-archived" flag.',
```
gcorne, line 233:
```suggestion
      '"%s" in the File Manager is an archived file. Try fetching again with the "--include-archived" flag.',
```

packages/cms-lib/fileManager.js
```diff
@@ -143,13 +144,16 @@ async function downloadFile(portalId, file, dest, options) {
...
+async function getAllPagedFiles(portalId, folderId, { includeArchived }) {
   let totalFiles = null;
   let files = [];
   let count = 0;
   let offset = 0;
   while (totalFiles === null || count < totalFiles) {
-    const response = await getFilesByPath(portalId, folderPath, { offset });
+    const response = await getFiles(portalId, folderId, {
+      offset,
+      archived: includeArchived ? 1 : 0,
```
gcorne, line 155: Might be worthwhile to use booleans until we get to the API helper since we already need code to handle the option.

packages/cms-lib/api/fileManager.js
```diff
@@ -31,21 +31,31 @@ async function getStat(portalId, src) {
...
   });
 }
 
 module.exports = {
   uploadFile,
   getStat,
-  getFilesByPath,
-  getFoldersByPath,
+  getFiles,
+  getFolders,
```
gcorne, line 60: FWIW, I tend to favor `fetch*` for GET request api helpers to help communicate their async nature when compared to typical "getters" that getting data in a synchronous fashion from somewhere.

PR #239, opened 2020-06-25
Ensure `fetch` and `filemanager fetch` subcommands match

### LINKED ISSUE #220: File Manager and Design Manager fetch logging parity
Some improvements for https://github.com/HubSpot/hubspot-cms-tools/pull/215 

Similar to how `upload` for File Manager and Design Manager looks/feels the same, we should make `fetch` behave and log similarly. 

Screen Shot 2020-06-04 at 5 37 20 PM
Right now, `filemanager fetch` is missing logging for the writing of files/folders, initial success, and completion message. Similar to `upload`, I think we should also specify for the of the `fetch`es, if the content was fetched from the Design Manager or File Manager:

Screen Shot 2020-06-04 at 5 45 40 PM

anthmatic:
image

Resolves #220

### DIFF
--- a/packages/cms-lib/fileMapper.js
+++ b/packages/cms-lib/fileMapper.js
@@ -393,12 +393,20 @@ async function downloadFile(input) {
     const localFsPath = convertToLocalFileSystemPath(filepath);
     await fetchAndWriteFileStream(input, input.src, localFsPath);
     await queue.onIdle();
-    logger.log('Completed fetch of file "%s" to "%s"', input.src, localFsPath);
+    logger.success(
+      'Completed fetch of file "%s" to "%s" from the Design Manager',
+      input.src,
+      localFsPath
+    );
   } catch (err) {
     if (isHubspot && isTimeout(err)) {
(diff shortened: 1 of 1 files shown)


packages/cms-lib/fileMapper.js
```diff
@@ -421,7 +429,11 @@ async function fetchFolderFromApi(input) {
     const node = isHubspot
       ? await downloadDefault(portalId, srcPath, { qs })
       : await download(portalId, srcPath, { qs });
-    logger.log('Fetched "%s" from portal %d successfully', src, portalId);
+    logger.log(
+      'Fetched "%s" from portal %d from the Desing Manager successfully',
```
anthmatic, line 433: @williamspiro this is still inconsistent because the actual functionality differs a bit here. I believe DM assets are downloaded in bulk and then written. FM currently performs download/write in series.

gcorne: I think we can follow up in a separate PR and switch to the bulk model for file manager with a concurrency of 10.

PR #240, opened 2020-06-25
Update upload/watch messaging to match filemanager

### LINKED ISSUE #221: File Manager and Design Manager upload logging parity
Similar to https://github.com/HubSpot/hubspot-cms-tools/issues/220

I notice `filemanager upload` and `upload` do not log in the same way:
Screen Shot 2020-06-04 at 5 45 40 PM

I think how filemanager logs `[SUCCESS]` at the end makes sense

anthmatic:
Updates logger message to match filemanager `upload` subcommand

Resolves #221

### DIFF
--- a/packages/cms-cli/commands/upload.js
+++ b/packages/cms-cli/commands/upload.js
@@ -144,7 +144,7 @@ function configureUploadCommand(program) {
             );
           });
       } else {
-        logger.success(
+        logger.log(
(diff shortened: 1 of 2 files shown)


packages/cms-cli/commands/upload.js
```diff
@@ -144,15 +144,15 @@ function configureUploadCommand(program) {
             );
           });
       } else {
-        logger.success(
```
gcorne, line 147: I have a sinking suspicion that this was my doing.

PR #242, opened 2020-06-25
Add hubdb api test coverage

anthmatic:
Before upgrading to a new version of the HubDB api, it made sense to write some tests. I'm pretty green when it comes to writing tests, so I'd appreciate any feedback!


packages/cms-lib/__tests__/hubdb.js
```diff
@@ -0,0 +1,55 @@
...
+    const projectCwd = '/home/tom/projects';
+
+    getCwd.mockReturnValue(projectCwd);
+
+    const { filePath } = await downloadHubDbTable(portalId, tableId, destPath);
+
+    describe('transforms column names to ids', () => {
+      expect(fs.outputFile).toHaveBeenCalledWith(
+        `${projectCwd}/${destPath}`,
+        expect.stringContaining('"name": "My Event",')
```
gcorne, line 24: Minor thing, but it might be work using `JSON.parse()` here and then inspecting the object instead of inspecting a string. The benefit is that then we can verify that the shape of the object matches the expected.

packages/cms-lib/api/__mocks__/hubdb.js
```diff
@@ -0,0 +1,538 @@
+const hubdb = jest.genMockFromModule('../hubdb');
```
gcorne, line 1: I think for api helpers that don't have side effects we shouldn't need to create our own mocks and can instead leverage jest mocks on their own together with `mockReturnValue` and other friends for working mocked functions. Make sense?

packages/cms-lib/__tests__/hubdb.js
anthmatic: Ah, I'm glad that you mentioned that. I updated this, I think that's a lot more clear now.

packages/cms-lib/api/__mocks__/hubdb.js
anthmatic: I think so. I updated this, curious if that's along the lines of what you were thinking?

gcorne: Looks good. It also is fine to keep test fixtures/data in json or js files separately. Doing so can be helpful, if the same test data is useful across multiple tests and test files.

anthmatic: Good point. I added these as separate files. Not sure if that's preferred vs keeping everything in one file (seen it done both ways).

PR #246, opened 2020-06-26
Fix `hs watch . my-website-theme`

abstraktor:
# Steps to Reproduce
Given I run `hs watch . my-website-theme`
Then I errorneously get this output:

```
➜  my-website-theme hs watch . my-website-theme
(node:28185) UnhandledPromiseRejectionWarning: TypeError: path must not be empty
    at throwError (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/ignore/index.js:366:9)
    at checkPath (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/ignore/index.js:379:12)
    at Ignore._test (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/ignore/index.js:502:5)
    at Ignore.ignores (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/ignore/index.js:541:17)
    at shouldIgnoreFile (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/@hubspot/cms-lib/ignoreRules.js:49:22)
    at ignored (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/@hubspot/cms-lib/lib/watch.js:101:22)
    at matchPatterns (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/anymatch/index.js:62:18)
    at FSWatcher._userIgnored (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/anymatch/index.js:94:14)
    at FSWatcher._isIgnored (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/chokidar/index.js:763:15)
    at FsEventsHandler._addToFsEvents (/usr/local/lib/node_modules/@hubspot/cms-cli/node_modules/chokidar/lib/fsevents-handler.js:459:18)
(node:28185) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:28185) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Yet I'd expect it to upload local changes

# Severity
I like to have my `hubspot.config.yml` and my console in the synced folder, so that I end up running exactly the given command. It would be sad having to deepen my directory structure just to mitigate this bug.

# Analysis

* The exception we see comes from the `ignore` package [here](https://github.com/kaelzhang/node-ignore/blob/55ecc53f23e4ba787fce051a2bcd9d863aacc1f6/index.js#L379).
* Ignore intentionally complains when checking the empty path on it.
* It asks for anything that comes out of `path.relative(…)`
* Yet `path.relative('.','.')` returns an empty string
* I guess the best way out is to not involve `ignore` when `chokidar` let's us check the local folder. That'd be weird to ignore it all anyway…

# Todo

* [x] fix it
* [x] run test
* [x] run linter
* [x] try manually locally if `hs watch . my-website-theme` works well
* I'm unsure about what's left in your process here…

### DIFF
--- a/packages/cms-lib/__tests__/ignoreRules.js
+++ b/packages/cms-lib/__tests__/ignoreRules.js
@@ -42,5 +42,9 @@ describe('ignoreRules', () => {
     it('does not ignore allowed files', () => {
       expect(shouldIgnoreFile(`${REPO_FOLDER}/hiddenFile.js`, CWD)).toBe(false);
     });
+
+    it('does not ignore the current folder', () => {
+      expect(shouldIgnoreFile('', '')).toBe(false);
+    });
   });
 });
--- a/packages/cms-lib/__tests__/ignoreRules.js
+++ b/packages/cms-lib/__tests__/ignoreRules.js
@@ -44,7 +44,7 @@ describe('ignoreRules', () => {
     });
 
     it('does not ignore the current folder', () => {
-      expect(shouldIgnoreFile('', '')).toBe(false);
+      expect(shouldIgnoreFile(CWD, CWD)).toBe(false);
     });
   });
 });
--- a/packages/cms-lib/__tests__/ignoreRules.js
+++ b/packages/cms-lib/__tests__/ignoreRules.js
@@ -44,7 +44,7 @@ describe('ignoreRules', () => {
     });
 
     it('does not ignore the current folder', () => {
-      expect(shouldIgnoreFile(CWD, CWD)).toBe(false);
+      expect(shouldIgnoreFile('.', process.cwd())).toBe(false);
     });
   });
 });
--- a/packages/cms-lib/ignoreRules.js
+++ b/packages/cms-lib/ignoreRules.js
@@ -46,7 +46,8 @@ function loadIgnoreConfig() {
 function shouldIgnoreFile(file, cwd) {
   loadIgnoreConfig();
   const relativeTo = configPath || cwd;
-  return ignoreRules.ignores(path.relative(relativeTo, file));
+  const relativePath = path.relative(relativeTo, file);
+  return !!relativePath && ignoreRules.ignores(relativePath);
 }
 
 function createIgnoreFilter(cwd) {


gcorne:
@abstraktor it looks like an extra commit unrelated to the fix snuck in via 
2ae138a. Mind removing it?

packages/cms-lib/ignoreRules.js
```diff
@@ -46,7 +46,8 @@ function loadIgnoreConfig() {
 function shouldIgnoreFile(file, cwd) {
   loadIgnoreConfig();
   const relativeTo = configPath || cwd;
-  return ignoreRules.ignores(path.relative(relativeTo, file));
+  const relativePath = path.relative(relativeTo, file);
+  return !!relativePath && ignoreRules.ignores(relativePath);
```
gcorne, line 50: I think it would be best to start with a test case here because this fix suggests that there is something off elsewhere.

abstraktor: I guess that was hidden in the chaos with the other commit. I actually added a test for this, as you might be able to see now that I removed the commit again

packages/cms-lib/__tests__/ignoreRules.js
```diff
@@ -42,5 +42,9 @@ describe('ignoreRules', () => {
     it('does not ignore allowed files', () => {
       expect(shouldIgnoreFile(`${REPO_FOLDER}/hiddenFile.js`, CWD)).toBe(false);
     });
+
+    it('does not ignore the current folder', () => {
+      expect(shouldIgnoreFile('', '')).toBe(false);
```
gcorne, line 47: I think that the second argument here should be CWD.

packages/cms-lib/ignoreRules.js
gcorne: While this change is a good one, I think that the issue is actually with how we're using `shouldIgnoreFile` in the `chokidar.watch()` call in `lib/watch.js`.

abstraktor: Yes the fix smells like there should be something cleaned up, I'm unsure what though. The emptystring seems to be returned when `file == cwd`. I'd expect shouldIgnoreFile in general to treat such input gracefully…

packages/cms-lib/__tests__/ignoreRules.js
```diff
@@ -42,5 +42,9 @@ describe('ignoreRules', () => {
     it('does not ignore allowed files', () => {
       expect(shouldIgnoreFile(`${REPO_FOLDER}/hiddenFile.js`, CWD)).toBe(false);
     });
+
+    it('does not ignore the current folder', () => {
+      expect(shouldIgnoreFile(CWD, CWD)).toBe(false);
```
gcorne, line 47: Sorry for not being clear, but I was thinking something like `shouldIgnoreFile('.', CWD)` or `shouldIgnoreFile('', CWD)`. The expectation is the path passed to the function is a path to a file so it feels a little odd to have a test where `CWD` is passed to both.

packages/cms-lib/ignoreRules.js
gcorne: Yeah, I agree. If you make the small adjustment to the test, I'll get this merged. We can improve `lib/watch.js` separately.

packages/cms-lib/__tests__/ignoreRules.js
abstraktor: I checked it and when the bug occurs, the first argument equals the CWD. Consider these semantics of path.relative:
```
> path.relative('.','/a/b')
'../../../a/b'
> path.relative('','/a/b')
'../../../a/b'
> path.relative('/a/b','/a/b')
''
```

So you're suggestion wouldn't test what I fixed…

packages/cms-lib/ignoreRules.js
abstraktor: Hi gcorne, As I said there, that adjustment won't do, as you suggest.

packages/cms-lib/__tests__/ignoreRules.js
gcorne: yeah, in this case, it would need to be `process.cwd()` instead of the hard-coded CWD.

gcorne:
```suggestion
      expect(shouldIgnoreFile('.', process.cwd())).toBe(false);
```

abstraktor: Now I understand your suggestion. Great idea.

PR #247, opened 2020-06-26
Add more actionable errors when running serverless commands without proper scopes

miketalley:
When running serverless related commands(`hs secrets` and `hs logs`) without permissions for serverless functions, the error message was not very helpful...

> The post to "add secret" in portal 103482808 was forbidden. This app hasn't been granted all required scopes to make this call. Read more about required scopes here: https://developers.hubspot.com/scopes

This has been updated, based on where the problem lies, to...

@achong07 do you have any input on the error copy?
/cc @williamspiro


packages/cms-lib/lib/errorHandlers/customErrors.js
```diff
@@ -0,0 +1,14 @@
+const getErrorMessage = error => {
+  const { category } = error;
+
+  switch (category) {
+    case 'MISSING_SCOPES':
+      return 'Your personal CMS access key is missing required permissions for this action. Your account may not have these permissions or you may need to generate a new access key to add them. You can do so by running "hs auth personalaccesskey". If you believe this is in error, please contact your administrator.';
```
gcorne, line 6: > Your personal CMS access key is missing required permissions for this action.
This sentence assumes that an access key is being used for authentication.

> Your account may not have these permissions or you may need to generate a new access key to add them. You can do so by running "hs auth personalaccesskey".

This is a situation where "account" is unclear. Are we referring to the hub or the user? If it is the hub, the action may be to either purchase CMS Hub enterprise, create a trial, or leverage a developer sandbox. I wonder if we should add some sort of backend API that we can use to determine if the issue is a portal issue or a user permission issue.

packages/cms-lib/lib/errorHandlers/customErrors.js
```diff
@@ -0,0 +1,14 @@
+const getErrorMessage = error => {
+  const { category } = error;
+
+  switch (category) {
+    case 'MISSING_SCOPES':
```
gcorne, line 5: I do wonder if we should somehow add this handling closer to the API being called so that we can provide clearer context based on the API being called.

miketalley: We could add a handler to [the catch block](https://github.com/HubSpot/hubspot-cms-tools/blob/62597483e8eda34d60578331200f72779a2e0262/packages/cms-cli/commands/secrets.js#L66) of `configureSecretsAddCommand` that would display a custom message instead of calling `logApiErrorInstance` and we would have access to the entire error context.

packages/cms-lib/lib/errorHandlers/customErrors.js
miketalley: @gchomatas do you think it would be possible to return more info along with the error specifying whether the `MISSING_SCOPES` are a user or hub problem?

packages/cms-lib/lib/errorHandlers/customErrors.js
gcorne: Is there a different approach that could work? Perhaps somehow the API helpers could be wrapped somehow? It feels to me like we should be able to handle this in the CLI in a similar fashion to how we do it in the UI where either there are upfront scope checks for certain situations or there are specific error handlers for specific contexts. For example, there are situations where a 404 has a specific meaning that could be conveyed more clearly to the user when we know the context.

miketalley:
I talked to @mattcoley about an endpoint for checking the access to certain scopes (in this case `serverless-functions`). We decided to create an issue to add an endpoint for this and we can then check for access prior to making a call to add/delete/list secrets and display a friendly message if there is no access.

https://git.hubteam.com/HubSpotProtected/LocalDevAuth/issues/50

miketalley:
This is still a WIP as I am working on updating the functionality of this to check if the proper function scopes exist for both the portal and user prior to running commands related to serverless.

A new endpoint to obtain this data has been created, but we are still smoothing out some issues with it.

williamspiro:
Thanks for working on this @miketalley - is this going to work for other serverless commands, like `logs` as well?

packages/cms-lib/lib/constants.js
```diff
@@ -87,6 +87,18 @@ const ENVIRONMENT_VARIABLES = {
...
 
+const SCOPE_GROUPS = {
+  functions: 'cms.functions.read_write',
+};
+
+const SCOPES = {
+  functions: {
+    access: 'SERVERLESS_FUNCTIONS',
+    read: 'SERVERLESS_FUNCTIONS_READ',
+    write: 'SERVERLESS_FUNCTIONS_WRITE',
```
gcorne, line 98: I am concerned that this will fall out of date when/if scopes or scope groups change.

miketalley:
> Thanks for working on this @miketalley - is this going to work for other serverless commands, like `logs` as well?

I'm still working on the functionality, but the plan is to add it to work with all serverless commands.

packages/cms-lib/lib/constants.js
miketalley: I removed the `SCOPES` constants. We still need the `SCOPE_GROUPS.functions` scope group in order to hit the endpoint properly.

achong07:
> #### Portal missing scopes
> > Your personal CMS access key is missing required permissions for this action. Your portal does not have these permissions. If you believe this is in error, please contact your administrator.

Would suggest simplifying to the following:
`Your account does not have access to this action. Talk to an account admin to request it.`

> #### User missing scopes
> > Your personal CMS access key is missing required permissions for this action. Your user account does not have the required permissions. If you believe this is in error, please contact your administrator.

Would suggest simplifying to the following:
`You don't have access to this action. Ask an account admin to change your permissions in Users & Teams settings.`

> #### Access token missing scopes
> > Your personal CMS access key is missing required permissions for this action. Your user access token does not have the required permissions. Please generate a new access key to add them by running "hs auth personalaccesskey". If you believe this is in error, please contact your administrator.

Would suggest simplifying to the following:
`Your access key has expired. Please generate a new access key by running "hs auth personalaccesskey".`

Welcome feedback on these if they don't feel right - the first two I pulled from some of our standard messaging but we can try to get more specific if it's not clear and actionable.

miketalley:
@achong07 thanks for the feedback on the errors! I think the first two are solid. The third one may be a bit misleading because the token has not actually expired, it just doesn't have the scopes necessary to successfully run the command.

Maybe something along the lines of `Your access key does not allow this action. Please generate a new access key by running "hs auth personalaccesskey".` would avoid potential confusion?

packages/cms-cli/commands/logs.js
```diff
@@ -68,6 +72,14 @@ const loadAndValidateOptions = async options => {
   }
 };
 
+const logError = async (error, portalId, functionPath) => {
+  return logServerlessFunctionApiErrorInstance(
+    error,
+    await getScopeDataForFunctions(portalId),
```
miketalley, line 78: Ideally, this could go within `errorHandlers.js` but it creates a circular dependency there.

achong07:
@miketalley let's go with your suggestion for the third! 

`Your access key does not allow this action. Please generate a new access key by running "hs auth personalaccesskey".`

packages/cms-lib/lib/scopes.js
```diff
@@ -0,0 +1,30 @@
...
+const {
+  logApiErrorInstance,
+  ApiErrorContext,
+} = require('@hubspot/cms-lib/errorHandlers');
+
+/**
+ *
+ * @param {number} portalId
+ */
+async function getScopeDataForFunctions(portalId) {
```
gcorne, line 13: Not something we need to do now, but seeing this makes me wonder what would need to change in order for the BE to return the missing scope group so that we could reuse the scope related enhanced error handling across all the places where we're using v3 endpoints.

packages/cms-lib/errorHandlers.js
```diff
@@ -300,6 +304,45 @@ function logFileSystemErrorInstance(error, context) {
...
 }
 
+/**
+ * Logs a message for an error instance resulting from API interaction
+ * related to serverless function.
+ *
+ * @param {Error|SystemError|Object} error
+ * @param {ApiErrorContext}          context
+ */
+function logServerlessFunctionApiErrorInstance(error, scopesData, context) {
```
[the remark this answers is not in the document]
miketalley: I will fix the scopesData issue.

Initially I wanted to make this async and fetch the data within, but set it up the way that I did to avoid a circular dependency issue that importing `http` creates.

PR #252, opened 2020-07-02
Add usage tracking for HubDB commands

### LINKED ISSUE #222: Add usage tracking to the `hs hubdb` commands

anthmatic:
Fixes #222

### DIFF
--- a/packages/cms-cli/commands/hubdb.js
+++ b/packages/cms-cli/commands/hubdb.js
@@ -15,7 +15,10 @@ const {
 const { publishTable, deleteTable } = require('@hubspot/cms-lib/api/hubdb');
 
 const { validatePortal } = require('../lib/validation');
-const { addHelpUsageTracking } = require('../lib/usageTracking');
+const {
+  trackCommandUsage,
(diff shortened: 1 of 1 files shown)


packages/cms-cli/commands/hubdb.js
```diff
@@ -27,6 +30,11 @@ const {
 } = require('../lib/commonOpts');
 const { logDebugInfo } = require('../lib/debugInfo');
 
+const CREATE_COMMAND_NAME = 'hubdb create';
+const FETCH_COMMAND_NAME = 'hubdb fetch';
+const CLEAR_COMMAND_NAME = 'hubdb clear';
+const DELETE_COMMAND_NAME = 'hubdb delete';
```
gcorne, line 36: Since we use hyphens (e.g. `hubdb-create`) for the filemanager subcommands, i think we should do the same here. I also think these constants are a little overkill since we're only using them in one place.

anthmatic: updated

PR #253, opened 2020-07-02
Add more options to "create template"

### LINKED ISSUE #251: Add `hs create email` for generating email templates
Our current commands let folks create a function, global-partial, module, template, website-theme, and react-app. We should also add an option for email that would create the email that is generated when you create one from the DM

source: https://hubspotdev.slack.com/archives/CDSFETF5X/p1593683444052600

### LINKED ISSUE #259: Include `label` in template annotations for hs create template
We should include the `label` annotation when creating a template using `hs create template`, so developers know it exists for coded files.

anthmatic:
Expands upon the `create template` command and allows the user to create:  page templates, partials, global partials, email templates, blog templates, and search results template. Other template types omitted on purpose for now, to keep the options list manageable

Resolves #251 
also fixes #259


packages/cms-cli/commands/create.js
```diff
@@ -27,6 +27,7 @@ const TYPES = {
   'global-partial': 'global-partial',
   module: 'module',
   template: 'template',
+  email: 'email',
```
gcorne, line 30: Not sure that "email" is the right term here since it is creating an email template not an email. I think we should use `email-template` or add a prompt flow.

packages/cms-cli/commands/create.js
```diff
@@ -134,16 +141,19 @@ function configureCreateCommand(program) {
...
         case TYPES.module:
           createModule(name, dest);
           break;
-        case TYPES.template:
-        case TYPES['global-partial']:
-          createTemplate(name, dest, type);
+        case TYPES.template: {
+          const { templateType } = await createTemplatePrompt();
+
+          commandTrackingContext.templateType = templateType;
```
anthmatic, line 153: Is this useful? Or would it be better to just append `type = template` with the template type? (ex. `template-page`)

packages/cms-cli/commands/create.js
anthmatic: I updated this and went down the prompt flow path.

packages/cms-cli/commands/create.js
anthmatic: @williamspiro do you have a preference here?

packages/cms-cli/lib/createTemplatePrompt.js
```diff
@@ -0,0 +1,21 @@
...
+
+const TEMPLATE_TYPE_PROMPT = {
+  type: 'list',
+  name: 'templateType',
+  message: 'Select the type of template to create',
+  default: 'PAGE',
+  choices: [
+    { name: 'PAGE', value: 'page-template' },
+    { name: 'EMAIL', value: 'email-template' },
+    { name: 'GLOBAL PARTIAL', value: 'global-partial' },
```
gcorne, line 11: Should we add "Standard partial"? I also think that we should not use all caps and instead use sentence case. In terms of design, I think we should think about these prompts like the new file wizard in the Design Manager and also probably consider adding additional types.

image

anthmatic: Yeah, will be simple to just account for all current template skeletons. I'll add those.

packages/cms-cli/commands/create.js
gcorne: As we add more wizard prompt flows, I do think tracking the options people use is useful.

packages/cms-cli/lib/createTemplatePrompt.js
anthmatic: I added _some_ of these. I opted not to include them all, mainly to keep the list short and manageable. If anyone feels strongly, I can add remaining system and membership pages but I'd rather figure out a better pattern for surfacing these first, personally.

packages/cms-cli/lib/createTemplatePrompt.js
```diff
@@ -0,0 +1,24 @@
+const inquirer = require('inquirer');
+
+const TEMPLATE_TYPE_PROMPT = {
+  type: 'list',
+  name: 'templateType',
+  message: 'Select the type of template to create',
+  default: 'PAGE',
```
gcorne, line 7: Should this be lowercase now?

packages/cms-cli/commands/create.js
```diff
@@ -18,14 +18,14 @@ const {
...
 const { createFunctionPrompt } = require('../lib/createFunctionPrompt');
+const { createTemplatePrompt } = require('../lib/createTemplatePrompt');
 const { createModulePrompt } = require('../lib/createModulePrompt');
 const { commaSeparatedValues } = require('../lib/text');
 
 const COMMAND_NAME = 'create';
 
 const TYPES = {
   function: 'function',
-  'global-partial': 'global-partial',
```
gcorne, line 28: Not sure how often `global-partial` has been used but we should probably add handling that communicates to the user that they should use `hs create template` now and choose "global partial."

packages/cms-cli/commands/create.js
```diff
@@ -141,9 +149,17 @@ function configureCreateCommand(program) {
...
+      if (type === 'global-partial') {
+        logger.error(
+          `The asset type ${type} has been deprecated. Please choose the "template" asset and select "global partial".`
+        );
+        return;
+      }
+
       if (!type || !TYPES[type]) {
         logger.error(
-          `The asset type ${type} is not supported. Supported authentication protocols are ${SUPPORTED_ASSET_TYPES}.`
```
gcorne, line 146: Good catch.

PR #254, opened 2020-07-06
Add prompt for `create module`

### LINKED ISSUE #243: Add label option with hs create
Idea: Being able to set the label of the module, template and so on, directly when creating the given type, e.g. `hs create module form --label="Form" <dest>`

The reasoning this idea is to keep the local code easy-to-navigate and keep in line with formatting like lowercase, no special chars, no spaces etc, for folders, but still be able to show the pretty version (label) to the end user.

An alternative solution, is to have the `"label": ""` displayed in the `meta.json` upon creation, so people unfamiliar with the `label` option, will know that they can use it.

anthmatic:
Adds a prompt for the creation of modules and writes a custom `meta.json` file. Currently adds options for: 
-label
-host_template_types

Resolves #243


packages/cms-cli/defaults/email.html
```diff
@@ -0,0 +1,47 @@
+<!--
```
anthmatic, line 1: hmm. not sure how this got here

packages/cms-cli/commands/create.js
```diff
@@ -49,7 +52,34 @@ const PROJECT_REPOSITORIES = {
 
 const SUPPORTED_ASSET_TYPES = commaSeparatedValues(Object.values(TYPES));
 
-const createModule = (name, dest) => {
+const createModule = (moduleDefinition, name, dest) => {
```
anthmatic, line 55: todo: break this out

packages/cms-cli/lib/createModulePrompt.js
```diff
@@ -0,0 +1,34 @@
...
+const CONTENT_TYPES_PROMPT = {
+  type: 'checkbox',
+  name: 'contentTypes',
+  message: 'What types of content will this module be used in?',
+  default: 'PAGE',
+  choices: [
+    { name: 'Page', value: 'PAGE' },
+    { name: 'Blog post', value: 'BLOG_POST' },
+    { name: 'Blog listing', value: 'BLOG_LISTING' },
+    { name: 'Email', value: 'EMAIL' },
```
gcorne, line 24: I wonder if we should automatically leave out `module.css` and `module.js` if a user checks `EMAIL` since `module.css` and `module.js` are not allowed in email.

anthmatic: 👍 that's a good call

PR #256, opened 2020-07-08
Allow user to functions in the same directory

### LINKED ISSUE #179: Improve handling of running hs create function when other .functions folders exist
If you run `hs create function` in a directory where an existing .functions folder might already exist, an error is thrown as we try and create the function inside the pre-existing .functions folder, even if it is not the CWD

Screen Shot 2020-04-24 at 11 35 51 AM

In the above case, I would expect a /folder/b.functions folder be created, not /folder/a.functions/b.functions

anthmatic:
Instead of looking for ancestral `*.function` directories (which wasn't working) this switches to looking for `serverless.json` config files

@gcorne I see that you assigned that issue to yourself, if you have another idea for this in mind let me know

### DIFF
--- a/packages/cms-lib/functions.js
+++ b/packages/cms-lib/functions.js
@@ -99,14 +99,16 @@ function createFunction(
   { functionsFolder, filename, endpointPath, endpointMethod },
   dest
 ) {
-  const ancestorFunctionsDir = findup('serverless.json', {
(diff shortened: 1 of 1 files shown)


anthmatic:
Makes sense. Revisiting #179, I swear that referenced the bug that this addresses but I can't find any issues that reference that now. I'll keep that issue open for now

PR #47, opened 2020-01-03
Yargs

corymartin:
Fixes #23 

Initially this Yargs implementation will exist alongside the Commander implementation.
The temp Yargs executables are `_hs` and `_hscms`.

# WIP
Added working `_hs fetch` and `_hs upload` commands for review and discussion before implementing other commands similarly.

/cc @gcorne @miketalley

PR #264, opened 2020-07-24
Start migrating to `yargs` via the remove subcommand

gcorne:
Let's kickstart the process to make the shift to [yargs](https://github.com/yargs/yargs) again. This PR proposes a migration strategy where we refactor [commands](https://github.com/HubSpot/hubspot-cms-tools/tree/master/packages/cms-cli/commands) to support both `Commander.js` and `yargs` through sharing the core action logic for each command between the two. It borrows from #47 with a few adjustments.

This will allow us to make the process gradually and not block any enhancements or bug fixes that we want to get out before we're ready to switch entirely over to `yargs`.

@miketalley @anthmatic

### DIFF
--- a/packages/cms-cli/bin/cli.js
+++ b/packages/cms-cli/bin/cli.js
@@ -1,0 +1,42 @@
+#!/usr/bin/env node
+
+const yargs = require('yargs');
+const updateNotifier = require('update-notifier');
+
+const { logger } = require('@hubspot/cms-lib/logger');
+const { logErrorInstance } = require('@hubspot/cms-lib/errorHandlers');
+
+const { setLogLevel, getCommandName } = require('../lib/commonOpts');
+const { trackHelpUsage } = require('../lib/usageTracking');
+const pkg = require('../package.json');
+
+const removeCommand = require('../commands/remove');
+
+const SCRIPT_NAME = 'banjo';
+const notifier = updateNotifier({ pkg });
+
+notifier.notify({
+  shouldNotifyInNpmScript: true,
+});
+
+const argv = yargs
+  .scriptName(SCRIPT_NAME)
+  .usage('Tools for working with the HubSpot CMS')
+  .middleware([setLogLevel])
+  .exitProcess(false)
+  .fail((msg, err /*, _yargs*/) => {
+    if (msg) logger.error(msg);
+    if (err) logErrorInstance(err);
+  })
+  .command(removeCommand)
+  .help()
+  .demandCommand(
+    1,
+    `Please specifiy a command or run \`${SCRIPT_NAME} --help\` for a list of available commands`
+  )
+  .recommendCommands()
+  .strict().argv;
+
+if (argv.help) {
+  trackHelpUsage(getCommandName(argv));
+}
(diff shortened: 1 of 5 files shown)


packages/cms-cli/commands/remove.js
```diff
@@ -26,11 +26,44 @@ const {
 } = require('../lib/usageTracking');
 
 const COMMAND_NAME = 'remove';
+const DESCRIPTION = 'Delete a file or folder from HubSpot';
+
+async function action(args, options) {
```
gcorne, line 31: One thing that I think we need to decide is what should the signature for the actions be. As a first pass, I split positional arguments from options, but I am not sure if that is the right move or not.

packages/cms-cli/package.json
```diff
@@ -16,12 +16,14 @@
...
     "shelljs": "0.8.3",
-    "update-notifier": "3.0.1"
+    "update-notifier": "3.0.1",
+    "yargs": "15.4.1"
   },
   "engines": {
     "node": ">=8.9.1"
   },
   "bin": {
+    "banjo": "./bin/cli.js",
```
gcorne, line 26: This is just a silly codename borrowed from the earlier days and our mascot Banjo.

image

packages/cms-cli/bin/cli.js
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env node
```
gcorne, line 1: When we're fully switched over all the other files in `cms-cli/bin` will be able to be deleted.

packages/cms-cli/bin/cli.js
```diff
@@ -0,0 +1,42 @@
...
+const argv = yargs
+  .scriptName(SCRIPT_NAME)
+  .usage('Tools for working with the HubSpot CMS')
+  .middleware([setLogLevel])
+  .exitProcess(false)
+  .fail((msg, err /*, _yargs*/) => {
+    if (msg) logger.error(msg);
+    if (err) logErrorInstance(err);
+  })
+  .command(removeCommand)
```
anthmatic, line 31: Is the thought that we'd explicitly define commands here instead of using `commandDir()`?

gcorne: Yeah, at least for now. Once we've moved off of commander, we could switch to `commandDir` if we wanted.

packages/cms-cli/commands/remove.js
```diff
@@ -42,32 +75,30 @@ function configureRemoveCommand(program) {
...
-      }
+      await action({ hsPath }, command);
     });
 
   addConfigOptions(program);
   addPortalOptions(program);
   addLoggerOptions(program);
   addHelpUsageTracking(program, COMMAND_NAME);
 }
+exports.command = `${COMMAND_NAME} <path>`;
```
drewjenkins, line 86: Why do we export command and describe?

gcorne: https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module

drewjenkins: I see... then what is the exports.configureRemoveCommand for?

gcorne: For the existing Commander.js version of the CLI.

drewjenkins: Ahh.  I need to look into commander.js then.  Thanks for the info, making more sense

PR #273, opened 2020-07-31
Change default `hs watch` behavior to no longer upload by default

### LINKED ISSUE #213: [Question] Should watch be updated to not upload all watched files/folder initially
There has been quite a few users who would prefer if the initial upload of watched files/folders only happened if they so specified that should happen. As in, rather than setting `--disable-initial` to disable the initial upload, the initial upload does not occur by default, and a user would need to run `watch` with `--enable-initial` (or something similar) for the initial upload to happen.

User sometimes forget to use the `--disable-initial` flag when starting a `watch`, which can cause a large number of files to quickly become incorrect without an easy way to roll back. 

This would be a pretty major departure from the existing behavior. 

Some conversation on the topic https://hubspotdev.slack.com/archives/CLJJXGTC0/p1590284153389900

cc @gcorne @miketalley @iamw11y 

Thoughts? 

**UPDATE: We are going to move forward with work to make the initial auto-upload opt-in**
Changes which need to be made at the same time:

- [ ]  `hs watch` does not auto-upload the watched dir by default (@gcorne) 
- [ ]  A new `--enable-initial` flag added to watch, to make the initial auto-upload happen (@gcorne) 
- [ ]  Some sort of log when running `hs watch` that mentions the change, for any existing users who are confused why an initial auto-upload is not happening. (@gcorne) 
- [ ]  CLI reference docs on `watch` updated with new behavior and flag. We should maybe also have a callout on this command for a while that explains this change, in case anyone looks at the docs, and they reflect different behavior than their version. (@TheWebTech)
- [ ]  The quick start guide updated to include the `--enable-initial` flag for uploading the project (we also need to do some digging to make sure no other docs need to be updated. For example, https://github.com/HubSpot/cms-theme-boilerplate#to-download-the-hubspot-cms-boilerplate-using-the-hubspot-cms-cli (@TheWebTech).
- [ ]  We could include this change in a changelog post, as well as a slack post (@TheWebTech)

gcorne:
Feedback from developers shows that they would prefer to upload
separately.

Fixes #213

image

image

@williamspiro @miketalley @anthmatic

### DIFF
--- a/packages/cms-cli/commands/watch.js
+++ b/packages/cms-cli/commands/watch.js
@@ -37,7 +37,11 @@ function configureWatchCommand(program) {
     )
     .arguments('<src> <dest>')
     .option('--remove', 'remove remote files when removed locally')
-    .option('--disable-initial', 'disable initial upload of watched directory')
+    .option('--initial-upload', 'upload directory before watching for updates')
+    .option(
+      '--disable-initial',
+      'disable initial upload of watched directory (default)'
+    )
     .option(
       '--notify <path/to/file>',
       'log to specified file when a watch task is triggered and after workers have gone idle'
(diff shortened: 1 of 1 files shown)


packages/cms-cli/commands/watch.js
```diff
@@ -37,18 +37,41 @@ function configureWatchCommand(program) {
...
       loadConfig(configPath);
       checkAndWarnGitInclusion();
 
+      if (disableInitial) {
+        logger.info(
+          'Passing the "--disable-initial" option is no longer necessary. Running "hs watch" no longer uploads the watched directory by default.'
+        );
+      } else {
+        logger.warn(
+          `The "watch" command no longer uploads the watched directory by default. The directory "${src}" was not uploaded.`
```
williamspiro, line 68: I wonder if this should be a bit more explicit, to reduce any confusion.  

>The "watch" command no longer initially uploads the entire watched directory by default.

PR #277, opened 2020-08-11
Yargs lint migration

drewjenkins:


packages/cms-cli/commands/lint.js
```diff
@@ -27,51 +27,83 @@ const {
...
+  } catch (err) {
+    logger.groupEnd(groupName);
+    logErrorInstance(err, { portalId });
+    process.exit(1);
+  }
+  logger.groupEnd(groupName);
+  logger.log('%d issues found', count);
+};
+
+const configureYargs = {
```
drewjenkins, line 65: I did this and the `configureCommander` below just to make the division a bit more clear until we are able to remove commander

gcorne: TBH, I am not a huge fan of the extra objects when it comes to readability but understand the intent. I also think it would be easier to clean up when we remove Commander.js if we didn't have to touch the yargs code. What if instead we used a comment for the commander helpers?

drewjenkins: Sure, I think it still makes sense to wrap do the `configureYargs` object, and one the migration is done we can just export that whole object.  I can unwrap commander and just use a comment for it.  You fine if I keep the yargs stuff grouped?

gcorne: It is actually the yargs grouping that I find awkward because it adds what feels like an unnecessary layer of indirection — over my career I have tended towards avoiding abstractions whenever possible. In this case, I think developers adding commands will need to understand the `yargs` command module interface, and once the migration to `yargs` is complete, I don't think we will want to export things other than what `yargs` expects so that if `yargs` adds additional properties we don't hit conflicts.

drewjenkins: I guess I'm not following.  I don't see how 

```js
const configureYargs = {,
   command: ...
   describe: ...
   ...
};

module.exports = configureYargs;
```

or even

```js
module.exports = {
   command: ...
   describe: ...
   ...
}
```

in the future after commander is removed would be much different than

```js
exports.command = ...;
exports.describe = ...;
...
```

To me, it is more clear, and not really an abstraction to do the first or second option, since it inherently shows what Yargs needs exported, while keeping the other functions in the file (.e.g `async function action`) are not something Yargs is expecting to have exported.

The other option would need to be something like the following until we remove commander

```js
// Yargs Exports
exports.command = ...;
exports.describe = ...;
...
...
...
...
// End Yargs Exports
```

I guess my overall point being is I don't see a big difference between `module.exports = {...}` vs `exports.command = ...; exports.describe = ...;`  We seem to do a combo of both from what I have seen

gcorne: I am fine with `module.exports` but I'd rather we compose the object directly than spread over another object.

gcorne: I also don't think this is that big of a deal if you prefer the way it is written. There are bigger things to tackle like finishing up the migration to `yargs`, and I don't want to get too bogged down bikeshedding.

drewjenkins: I agree :) I'll go with the composing module.exports directly

packages/cms-cli/commands/lint.js
```diff
@@ -27,51 +27,75 @@ const {
...
-    .version(version)
-    .description('lint a file or folder for HubL syntax')
-    .arguments('<path>')
-    .action(async filepath => {
-      setLogLevel(command);
-      logDebugInfo(command);
-      const { config: configPath } = command;
-      loadConfig(configPath);
-      checkAndWarnGitInclusion();
+const parseOptions = async options => {
```
gcorne, line 32: Since there are side effects, I am not sure `parseOptions` is the best name here.

drewjenkins: Changed

packages/cms-cli/commands/lint.js
```diff
@@ -27,51 +27,75 @@ const {
...
 
-      trackCommandUsage(COMMAND_NAME, {}, portalId);
+const action = async (args, options) => {
+  const portalId = getPortalId(options);
+  const localPath = resolveLocalPath(args.localPath);
+  const groupName = `Linting "${localPath}"`;
 
-      filepath = resolveLocalPath(filepath);
+  trackCommandUsage(COMMAND_NAME, {}, portalId);
+  await loadAndValidateOptions(options);
```
gcorne, line 50: I think moving this logic below `getPortalId()` is likely to cause problems because we haven't validated the portal.

drewjenkins: Ahhh that makes sense.  Thanks for that.  Putting it back at the top

packages/cms-cli/commands/lint.js
```diff
@@ -27,51 +27,76 @@ const {
...
+      .arguments('<path>')
+      .action(async (localPath, command = {}) =>
+        action({ localPath }, command)
+      );
+
+    addConfigOptions(commander);
+    addPortalOptions(commander);
+    addLoggerOptions(commander);
+    addHelpUsageTracking(commander, COMMAND_NAME);
+  },
```
gcorne, line 101: Sorry to nitpick so much, but I think having all this logic inside `module.exports` makes it harder to see at a glance what is being exported. I think that I didn't explain thoroughly how I thought using `module.exports` would work.
I was thinking something like:
```js
const command = `${COMMAND_NAME} <path>`;
const describe =  DESCRIPTION;
const builder = yargs => {
  addConfigOptions(yargs, true);
  addPortalOptions(yargs, true);
  addLoggerOptions(yargs, true);
  yargs.positional('path', {
    describe: 'Local folder to lint',
    type: 'string',
  });
  return yargs;
};
// etc…
module.exports = {
  // Yargs
  command,
  describe,
  builder,
  // Commander
  configureLintCommand,
};
```

drewjenkins: Okay, I'm not huge on that since it just mixes all of the configuration stuff in with functions that aren't getting exported, but I'll swap it

gcorne: Yeah, that is why `export.command` (or named ES6 exports) can be preferable.

PR #280, opened 2020-08-12
Added hs create vue-app that utilizes vue-cli init command

### LINKED ISSUE #217: Add `hs create vue-app`
When demoing `hs create react-app`, there was a lot of interest in [Vue.js](https://vuejs.org/), so it makes sense to create a similar demo app that helps users set up a CMS project using Vue.

miketalley:
This adds the `hs create vue-app <destinationFolder>` command which generates a boilerplate Vue module app using the [vue-cli](https://cli.vuejs.org/) and the [cms-vue-module-template repo](https://github.com/HubSpot/cms-vue-module-template). It is the equivalent of running `vue init HubSpot/cms-vue-module-template <destinationFolder>`.

image

image

Fixes #217

### DIFF
--- a/packages/cms-cli/commands/create.js
+++ b/packages/cms-cli/commands/create.js
@@ -7,7 +7,10 @@ const {
 } = require('@hubspot/cms-lib/errorHandlers');
 const { getPortalId } = require('@hubspot/cms-lib');
 const { logger } = require('@hubspot/cms-lib/logger');
-const { createProject } = require('@hubspot/cms-lib/projects');
+const {
+  createProject,
+  createVueProject,
+} = require('@hubspot/cms-lib/projects');
 const { createFunction } = require('@hubspot/cms-lib/functions');
 
 const { addLoggerOptions, setLogLevel } = require('../lib/commonOpts');
@@ -30,6 +33,7 @@ const TYPES = {
   template: 'template',
   'website-theme': 'website-theme',
   'react-app': 'react-app',
+  'vue-app': 'vue-app',
   'webpack-serverless': 'webpack-serverless',
 };
 
@@ -56,6 +60,7 @@ const ASSET_PATHS = {
 
 const PROJECT_REPOSITORIES = {
   [TYPES['react-app']]: 'cms-react-boilerplate',
+  [TYPES['vue-app']]: 'cms-vue-module-template',
   [TYPES['website-theme']]: 'cms-theme-boilerplate',
   [TYPES['webpack-serverless']]: 'cms-webpack-serverless-boilerplate',
 };
@@ -171,6 +176,7 @@ function configureCreateCommand(program) {
           break;
         case TYPES['website-theme']:
         case TYPES['react-app']:
+        case TYPES['vue-app']:
         case TYPES['webpack-serverless']:
           dest = name || type;
           break;
@@ -217,6 +223,10 @@ function configureCreateCommand(program) {
         case TYPES['react-app']:
         case TYPES['webpack-serverless']: {
           createProject(dest, type, PROJECT_REPOSITORIES[type], '', program);
+          break;
+        }
+        case TYPES['vue-app']: {
+          await createVueProject(dest, type, PROJECT_REPOSITORIES[type]);
           break;
         }
         case TYPES.function: {
--- a/packages/cms-cli/commands/create.js
+++ b/packages/cms-cli/commands/create.js
@@ -187,7 +187,7 @@ function configureCreateCommand(program) {
       dest = resolveLocalPath(dest);
 
       try {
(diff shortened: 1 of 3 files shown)


packages/cms-lib/projects.js
```diff
@@ -175,8 +176,37 @@ async function createProject(dest, type, repoName, sourceDir, options = {}) {
...
+ * @param {String} sourceDir - Directory in project that should get copied.
+ * @param {Object} options
+ * @returns {Boolean} `true` if successful, `false` otherwise.
+ */
+async function createVueProject(dest, type, repoName) {
+  const relativePath = path.relative(process.cwd(), dest);
+
+  return new Promise(resolve => {
+    const vueInit = spawn(
+      'vue',
```
gcorne, line 193: What if `vue` is not installed?

packages/cms-cli/commands/create.js
```diff
@@ -181,7 +187,7 @@ function configureCreateCommand(program) {
       dest = resolveLocalPath(dest);
 
       try {
-        await fs.ensureDir(dest);
+        await fs.exists(dest);
```
gcorne, line 190: What if `dest` is not a directory?

packages/cms-lib/functions.js
```diff
@@ -131,6 +131,8 @@ function createFunction(
   if (fs.existsSync(functionFilePath)) {
     logger.error(`The JavaScript file at "${functionFilePath}" already exists`);
     return;
+  } else {
+    fs.ensureFileSync(functionFilePath);
```
gcorne, line 135: What is the purpose of this change?

packages/cms-lib/projects.js
```diff
@@ -175,8 +176,37 @@ async function createProject(dest, type, repoName, sourceDir, options = {}) {
...
+        stdio: 'inherit',
+        shell: false,
+      }
+    );
+
+    vueInit.on('exit', () => {
+      resolve();
+    });
+  });
+}
```
gcorne, line 205: One thing that I think is unclear here is that `hs` is running `vue` under the hood. I am not 100% sure that this adds a ton of value since the user has to have the vue CLI installed. If it does add value, I think we need to make it as clear as we can that the `vue` CLI is be run by the `hs` because if there is a problem it needs to be clear that the problem is with `vue` not `hs`.

packages/cms-lib/projects.js
miketalley: Thanks! I had mean't to handle this -- I will look into it.

packages/cms-lib/functions.js
miketalley: So `ensureDir` in `commands/create.js` will check if the directory exists and, if not, create it. This was the one case that relied on the folder being created by that. So when I changed it from `fs.ensureDir` to `fs.exists` this needed to be added to create the folder.

miketalley:
I did some more research into vue including the `vue init` command and the upcoming changes to the [vue-cli](https://cli.vuejs.org/) and vue with the [upcoming release of vue v3 which is in RC](https://github.com/vuejs/rfcs/issues/189). It really changed my outlook on using a template to build a vue module project.

With the release of vue v3, the vue-cli will be [essentially deprecating](https://cli.vuejs.org/guide/creating-a-project.html#pulling-2-x-templates-legacy) the `vue init` command by moving it into a separate package [@vue/cli-init](https://www.npmjs.com/package/@vue/cli-init) which I think could cause some confusion with the functionality with this PR. We would need to handle checking for these deps(and versions of the vue-cli) to make sure the spawned process that executes the `vue` command under the hood works properly.

To @gcorne's [comment about clarity of using `vue` under the hood](https://github.com/HubSpot/hubspot-cms-tools/pull/280#discussion_r469576061), I think that if this process fails it could be unclear to the user where the problem occurred.

Creating a project through the [cms-vue-boilerplate](https://github.com/HubSpot/cms-vue-boilerplate) and the [cms-vue-module-template](https://github.com/HubSpot/cms-vue-module-template) both allow subsequent addition of plugins through the `vue add` command. So if the boilerplate does not contain a specific plugin, it _can_ be added later.

With these points in mind, I'm thinking that [the previous iteration](https://github.com/HubSpot/hubspot-cms-tools/pull/266) of the `hs create vue-app` command that generates a project using the [cms-vue-boilerplate](https://github.com/HubSpot/cms-vue-boilerplate) would be a better choice because it is more simple, less complex and more transparent, less magical.

I'm going to close this out in favor of the previous iteration that uses cms-vue-boilerplate which was [reverted](https://github.com/HubSpot/hubspot-cms-tools/pull/278) until we could look into templates more.
New PR: https://github.com/HubSpot/hubspot-cms-tools/pull/286

PR #281, opened 2020-08-12
Migrate logs command to yargs

drewjenkins:


packages/cms-cli/commands/logs.js
```diff
@@ -97,78 +116,96 @@ const tailLogs = async ({
...
+const describe = DESCRIPTION;
+const builder = yargs => {
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+  addLoggerOptions(yargs, true);
+  yargs.positional('path', {
+    describe: 'Path to serverless function',
+    type: 'string',
+  });
+  return yargs;
```
gcorne, line 180: It looks like the other options like `--latest`, `--compact`, etc… haven't been added to the yargs version of the command.

drewjenkins: Ah jeez.  Adding now

packages/cms-cli/commands/logs.js
```diff
@@ -97,78 +116,113 @@ const tailLogs = async ({
...
+  });
+  yargs.option('tail', {
+    alias: ['t', 'follow', 'f'],
+    describe: 'tail logs',
+    type: 'boolean',
+  });
+
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+  addLoggerOptions(yargs, true);
```
miketalley, line 195: Similar to [my other comment](https://github.com/HubSpot/hubspot-cms-tools/pull/282#discussion_r470200845), can we add `addHelpUsageTracking` here?

gcorne: I think `--help` and tracking is handled in the top-level `cli.js` setup. See https://github.com/HubSpot/hubspot-cms-tools/blob/master/packages/cms-cli/bin/cli.js#L42-L44

packages/cms-cli/commands/logs.js
```diff
@@ -4,6 +4,7 @@ const ora = require('ora');
 const {
   addLoggerOptions,
   addPortalOptions,
+  addConfigOptions,
```
gcorne, line 7: :+1:

PR #282, opened 2020-08-12
migrate init command to yargs

drewjenkins:


packages/cms-cli/commands/init.js
```diff
@@ -34,83 +34,101 @@ const {
...
+  return yargs;
+};
+const handler = async argv => action(argv);
+
+// Commander Configuration
+const configureCommanderInitCommand = program => {
+  program
+    .version(version)
+    .description(DESCRIPTION)
+    .action(async (command = {}) => action(command));
```
gcorne, line 119: What is the reason for the extra `() => {}`? Could the initialization of the param happen in the declaration of `action`?

packages/cms-cli/commands/init.js
```diff
@@ -34,83 +34,101 @@ const {
...
-          TRACKING_STATUS.ERROR
-        );
-      }
-    });
+// Yargs Configuration
+const command = `${COMMAND_NAME}`;
+const describe = DESCRIPTION;
+const builder = yargs => {
+  addLoggerOptions(yargs, true);
+  addTestingOptions(yargs, true);
```
miketalley, line 108: do we want `addHelpUsageTracking` here as well?

drewjenkins: Yeah I totally would, but I can't find a way to make it work with args.  Commander exposes an `on('--help', () =>` method, but I can't find an equivalent for Yargs.

packages/cms-cli/commands/init.js
drewjenkins: No reason other than I was just following convention of the other modules.  Sure, I can do the initialization over there

packages/cms-cli/commands/init.js
miketalley: Yeah I see what you mean about there not being a clear event for this in yargs to attach to.

Currently, we display help in two cases `--help` and if the arguments provided are not enough for the command/subcommand.

My thought is perhaps we can utilize [.middleware](https://yargs.js.org/docs/#api-reference-middlewarecallbacks-applybeforevalidation) to check the arguments provided are enough for the command, similarly to how we are currently doing it([create command example](https://github.com/HubSpot/hubspot-cms-tools/blob/49748935b7b965b28a8df2280749abb408386538/packages/cms-cli/bin/hs-create.js#L11)) and also check for the presence of `--help` option.

The tricky part will be that different commands require different arguments.

gcorne: I'd also prefer it if we moved away from using an option for `--qa`, which is a feature that is only for HubSpotters especially if we can't suppress it from the help. With `yargs`, can we hide an option from help?

drewjenkins: Yep, there's a `hidden` prop, I'll add it

PR #285, opened 2020-08-13
Migrate hubdb commands over to yargs

drewjenkins:
For testing this, I ran all of the commands across all 3 CLIs

Banjo
```
banjo hubdb create 
banjo hubdb fetch
banjo hubdb clear
banjo hubdb delete
```

HS
```
hs hubdb create 
hs hubdb fetch
hs hubdb clear
hs hubdb delete
```

HSCMS
```
hscms hubdb create 
hscms hubdb fetch
hscms hubdb clear
hscms hubdb delete
```

### DIFF
--- a/packages/cms-cli/commands/hubdb.js
+++ b/packages/cms-cli/commands/hubdb.js
@@ -1,198 +1,48 @@
-const path = require('path');
-const {
-  loadConfig,
-  validateConfig,
-  checkAndWarnGitInclusion,
-} = require('@hubspot/cms-lib');
-const { logger } = require('@hubspot/cms-lib/logger');
-const { logErrorInstance } = require('@hubspot/cms-lib/errorHandlers');
-const { getCwd } = require('@hubspot/cms-lib/path');
-const {
-  createHubDbTable,
-  downloadHubDbTable,
-  clearHubDbTableRows,
-} = require('@hubspot/cms-lib/hubdb');
-const { publishTable, deleteTable } = require('@hubspot/cms-lib/api/hubdb');
+const { addHelpUsageTracking } = require('../lib/usageTracking');
+const { addLoggerOptions } = require('../lib/commonOpts');
+const { configureCreate, CREATE_DESCRIPTION } = require('./hubdb/create');
+const { configureFetch, FETCH_DESCRIPTION } = require('./hubdb/fetch');
+const { configureClear, CLEAR_DESCRIPTION } = require('./hubdb/clear');
+const { configureDelete, DELETE_DESCRIPTION } = require('./hubdb/delete');
 
-const { validatePortal } = require('../lib/validation');
-const {
-  trackCommandUsage,
-  addHelpUsageTracking,
-} = require('../lib/usageTracking');
 const { version } = require('../package.json');
 
-const {
-  addConfigOptions,
-  addLoggerOptions,
-  addPortalOptions,
-  setLogLevel,
-  getPortalId,
-} = require('../lib/commonOpts');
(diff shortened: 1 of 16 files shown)


packages/cms-cli/commands/hubdb.js
```diff
@@ -1,198 +1,48 @@
...
-  downloadHubDbTable,
-  clearHubDbTableRows,
-} = require('@hubspot/cms-lib/hubdb');
-const { publishTable, deleteTable } = require('@hubspot/cms-lib/api/hubdb');
+const { addHelpUsageTracking } = require('../lib/usageTracking');
+const { addLoggerOptions } = require('../lib/commonOpts');
+const { configureCreate, CREATE_DESCRIPTION } = require('./hubdb/create');
+const { configureFetch, FETCH_DESCRIPTION } = require('./hubdb/fetch');
+const { configureClear, CLEAR_DESCRIPTION } = require('./hubdb/clear');
+const { configureDelete, DELETE_DESCRIPTION } = require('./hubdb/delete');
```
gcorne, line 6: Was there a reason for using the `configure*` approach instead of exporting a yargs command like we are doing elsewhere?

drewjenkins: The way I understood it is yargs needs you to export a single object containing those commands (command, describe, builder) so this approach seemed like the best way, since it builds up the whole of the commands within the one larger command.  

It also seemed a lot more clear and clean to me to introduce a folder-based structure for the subcommands with each of them separated.

packages/cms-cli/commands/hubdb/fetch.js
```diff
@@ -0,0 +1,96 @@
...
+
+    logger.log(`Downloaded HubDB table ${args.tableId} to ${filePath}`);
+  } catch (e) {
+    logErrorInstance(e);
+  }
+};
+
+const configureFetch = yargs => {
+  yargs.command({
+    command: 'fetch <tableId> <dest>',
```
gcorne, line 54: I believe `dest` is optional so this should be `[dest]`.

drewjenkins: Ah good call, fixed

packages/cms-cli/commands/hubdb.js
gcorne: I was thinking that https://github.com/HubSpot/hubspot-cms-tools/blob/e2385f61bffa2d05fea619a557b08471aae68aca/packages/cms-cli/commands/secrets.js#L32-L35 was more straightforward.

packages/cms-cli/commands/hubdb.js
```diff
@@ -1,198 +1,48 @@
...
-
-      trackCommandUsage('hubdb-fetch', null, portalId);
-
-      try {
-        const { filePath } = await downloadHubDbTable(portalId, tableId, dest);
+// Yargs Configuration
+const command = `${COMMAND_NAME}`;
+const describe = DESCRIPTION;
+const builder = yargs => {
+  addLoggerOptions(yargs, true);
```
gcorne, line 17: 🔪

packages/cms-cli/commands/hubdb/create.js
```diff
@@ -0,0 +1,90 @@
...
+    handler: async argv => action({ src: argv.src }, argv),
+    builder: () => {
+      yargs.positional('src', {
+        describe: 'local path to file used for import',
+        type: 'string',
+      });
+    },
+  });
+
+  addLoggerOptions(yargs, true);
```
gcorne, line 67: :hocho:

PR #290, opened 2020-08-17
Migrate watch to yargs

drewjenkins:


packages/cms-cli/commands/watch.js
```diff
@@ -28,13 +28,125 @@ const {
...
+    alias: 'i',
+    describe: 'upload directory before watching for updates',
+    type: 'boolean',
+  });
+  yargs.option('disable-initial', {
+    alias: 'd',
+    describe: 'retrieve most recent log only',
+    type: 'disable initial upload of watched directory (default)',
+  });
+  yargs.option('notify', {
```
drewjenkins, line 134: couldn't find a way to do this command like commander, where it shows `--notify <path/to/file>` in the name so ended up just adding it in the description

gcorne: I think we should remove `<` and `>` in the description since the documentation is outside the context of the structure of an option or argument. Instead I think an example might be best like `/tmp/hs-watch` or something.

drewjenkins: Understood, will change.

packages/cms-cli/commands/watch.js
```diff
@@ -28,97 +28,156 @@ const {
...
+    describe:
+      'Will cause watch to delete files in your HubSpot account that are not found locally.',
+    type: 'boolean',
+  });
+  yargs.option('initial-upload', {
+    alias: 'i',
+    describe: 'Upload directory before watching for updates',
+    type: 'boolean',
+  });
+  yargs.option('disable-initial', {
```
gcorne, line 131: Since #273 was merged, I think we should hide this option since it is the default.

packages/cms-cli/commands/watch.js
```diff
@@ -28,97 +28,156 @@ const {
...
+    alias: 'd',
+    describe: 'retrieve most recent log only',
+    type: 'Disable the initial upload when watching a directory (default)',
```
gcorne, line 134: Looks like `type` and `describe` are a little off.

packages/cms-cli/commands/watch.js
```diff
@@ -28,97 +28,156 @@ const {
...
+  addModeOptions(yargs, { write: true }, true);
+
+  yargs.positional('src', {
+    describe:
+      'Path to the local directory your files are in, relative to your current working directory',
+    type: 'string',
+    demand: true,
+  });
+  yargs.positional('dest', {
+    describe: 'Path in HubSpot Design Tools, can be a net new path',
```
gcorne, line 116: The comma here and in the `src` seem off.

packages/cms-cli/commands/watch.js
drewjenkins: oh wow hahahah that's what I get for copy/pasting from the docs.  Thanks

packages/cms-cli/commands/watch.js
drewjenkins: Agreed, had just copied from the online docs, will clean it up

packages/cms-cli/commands/watch.js
drewjenkins: Oh yeah, for sure I can do that

PR #291, opened 2020-08-18
Migrate create command to yargs

drewjenkins:


packages/cms-cli/commands/create.js
```diff
@@ -128,12 +129,126 @@ const createTemplate = (name, dest, type = 'page-template') => {
...
+  }
+
+  if (!type || !TYPES[type]) {
+    logger.error(
+      `The asset type ${type} is not supported. Supported asset types are ${SUPPORTED_ASSET_TYPES}.`
+    );
+    return;
+  }
+
+  // TODO: In yargs use `.implies()`
```
gcorne, line 178: Should we do this?

drewjenkins: Based on what we discussed in our 1:1 I think we should hold off on this for now and rework these commands to be subcommands based on asset type.  I'll remove the comment for now

PR #292, opened 2020-08-18
Migrate fetch to yargs

drewjenkins:


packages/cms-cli/commands/fetch.js
```diff
@@ -27,56 +27,91 @@ const {
...
+
+  yargs.positional('src', {
+    describe: 'Path in HubSpot Design Tools',
+    type: 'string',
+    demand: true,
+  });
+
+  yargs.positional('dest', {
+    describe:
+      'Path to the local directory you would like the files to be placed, relative to your current working directory. If omitted, this argument will default to your current working directory',
```
gcorne, line 85: I wonder if we should shorten this a bit. For me, the readability starts to suffer when it wraps across three lines.

image

drewjenkins: Sure I can do that.  I was just trying to keep consistency between the CLI and docs

gcorne: :+1: FWIW, I don't think addressing this is a blocker.

Longer term, we probably should think about some sort of way to generate the reference documentation. Perhaps there also is a way to adjust how `yargs` formats help so that multi-line descriptions work better.

drewjenkins: Shortened it up a bit

packages/cms-cli/commands/fetch.js
```diff
@@ -27,56 +27,91 @@ const {
...
+  downloadFileOrFolder({ portalId, src, dest, mode, options });
+};
+
+// Yargs Configuration
+const command = `${COMMAND_NAME} <src> [dest]`;
+const describe = DESCRIPTION;
+const builder = yargs => {
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+  addLoggerOptions(yargs, true);
```
gcorne, line 73: Since I merged #294, this should be removed.

PR #293, opened 2020-08-19
Yargs filemanager migrate

drewjenkins:


packages/cms-cli/commands/filemanager/fetch.js
```diff
@@ -0,0 +1,110 @@
...
+
+  // Fetch and write file/folder.
+  downloadFileOrFolder(portalId, src, dest, options);
+};
+
+const command = 'fetch <src> [dest]';
+const describe = FETCH_DESCRIPTION;
+const handler = async argv => action({ src: argv.src, dest: argv.dest }, argv);
+const builder = yargs => {
+  addOverwriteOptions(yargs, true);
```
gcorne, line 62: The file manager doesn't have a concept of `draft` and `public` modes so this should be removed.

packages/cms-cli/commands/filemanager/fetch.js
```diff
@@ -0,0 +1,110 @@
...
+    .version(version)
+    .description(FETCH_DESCRIPTION)
+    .arguments('<src> [dest]')
+    .option(
+      '--include-archived',
+      'Include files that have been marked as "archived"'
+    )
+    .action((src, dest) => action({ src, dest }, commander));
+
+  addOverwriteOptions(commander);
```
gcorne, line 94: :hocho:

packages/cms-cli/commands/filemanager/fetch.js
drewjenkins: Sure will do, it was already in there from the older commander version of the command so I just ported it over

PR #294, opened 2020-08-19
Add secrets to the `yargs` version of the CLI

gcorne:


packages/cms-cli/lib/commonOpts.js
```diff
@@ -13,20 +13,13 @@ const addPortalOptions = (program, useYargs = false) => {
...
 
-const addLoggerOptions = (program, useYargs = false) => {
-  if (useYargs) {
-    return program.option('debug', {
-      alias: 'd',
-      default: false,
-      describe: 'set log level to debug',
-      type: 'boolean',
-    });
-  }
```
gcorne, line 29: @drewjenkins made an adjustment here.

packages/cms-cli/bin/cli.js
```diff
@@ -34,13 +35,20 @@ const argv = yargs
     if (msg) logger.error(msg);
     if (err) logErrorInstance(err);
   })
+  .option('debug', {
```
drewjenkins, line 38: nice!

packages/cms-cli/commands/secrets/addSecret.js
```diff
@@ -0,0 +1,95 @@
...
+
+exports.builder = yargs => {
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+  yargs.positional('name', {
+    describe: 'Name of the secret',
+    type: 'string',
+  });
+  yargs.positional('value', {
+    describe: 'The secret to be stored',
```
drewjenkins, line 85: If we wanna be consistent with the docs, this should be `API key or other Auth detail`

packages/cms-cli/commands/secrets/updateSecret.js
```diff
@@ -0,0 +1,95 @@
...
+    .action(async (secretName, secretValue) => {
+      await action({ secretName, secretValue }, program);
+    });
+
+  addLoggerOptions(program);
+  addPortalOptions(program);
+  addConfigOptions(program);
+}
+
+exports.command = 'update <name> <value>';
```
drewjenkins, line 73: Hm we don't seem to have this command documented

packages/cms-cli/commands/secrets/updateSecret.js
```diff
@@ -0,0 +1,95 @@
...
+
+exports.describe = DESCRIPTION;
+
+exports.builder = yargs => {
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+  yargs.positional('name', {
+    describe: 'Name of the secret to be updated',
```
drewjenkins, line 81: On here and throughout, since these are required, you can add `demand: true` so it errors out if they aren't present

packages/cms-cli/lib/commonOpts.js
drewjenkins: Awesome change

packages/cms-cli/commands/secrets/updateSecret.js
gcorne: I think that is already handled via using `<>` in the command.

packages/cms-cli/commands/secrets/updateSecret.js
gcorne: cc @TheWebTech

TheWebTech: Thanks for tagging, updated!

PR #299, opened 2020-08-24
Bail on yargs errors

drewjenkins:
Yargs would continue to try to run the command even if required arguments were not present.  Now we bail and give the user information on why the command failed

### DIFF
--- a/packages/cms-cli/bin/cli.js
+++ b/packages/cms-cli/bin/cli.js
@@ -5,7 +5,6 @@ const updateNotifier = require('update-notifier');
 
 const { logger } = require('@hubspot/cms-lib/logger');
 const { logErrorInstance } = require('@hubspot/cms-lib/errorHandlers');
-
 const { setLogLevel, getCommandName } = require('../lib/commonOpts');
 const { trackHelpUsage } = require('../lib/usageTracking');
 const pkg = require('../package.json');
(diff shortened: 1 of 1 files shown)


packages/cms-cli/bin/cli.js
```diff
@@ -35,9 +34,16 @@ const argv = yargs
   .usage('Tools for working with the HubSpot CMS')
   .middleware([setLogLevel])
   .exitProcess(false)
-  .fail((msg, err /*, _yargs*/) => {
+  .fail((msg, err, _yargs) => {
+    // Preserve stack trace.
+    if (err) throw err;
+
     if (msg) logger.error(msg);
     if (err) logErrorInstance(err);
```
gcorne, line 42: How is this reachable if we are throwing the error on line 39?

drewjenkins: Hmmm, that's a good question.  I'll remove it since we are doing `logErrorInstance(err)` below.

packages/cms-cli/bin/cli.js
```diff
@@ -35,9 +34,16 @@ const argv = yargs
...
+
+    // Give command-specifc help
+    console.log(_yargs.help());
+    process.exit(0);
```
gcorne, line 46: I think we should be using `process.exit(1)` so that scripts are properly notified of the error.

drewjenkins: Ok will do.  I had done 0 since we already give error reasons at the top, but that makes sense for scripts

PR #305, opened 2020-08-31
Code review changes

drewjenkins:


packages/cms-cli/bin/cli.js
```diff
@@ -35,9 +35,6 @@ const argv = yargs
   .middleware([setLogLevel])
   .exitProcess(false)
   .fail((msg, err, _yargs) => {
-    // Preserve stack trace.
-    if (err) throw err;
```
gcorne, line 39: Left another suggestion related to this over in https://github.com/HubSpot/hubspot-cms-tools/pull/299/files#r480221294 after doing some testing and further reviewing.

drewjenkins: Changed to process.exit(1)

packages/cms-cli/bin/cli.js
```diff
@@ -35,15 +35,12 @@ const argv = yargs
...
-
     if (msg) logger.error(msg);
     if (err) logErrorInstance(err);
 
     // Give command-specifc help
     console.log(_yargs.help());
```
gcorne, line 42: I do wonder if we always want to be outputting help when there is an error.

drewjenkins: It seemed like it would be helpful since it is context sensitive to the command that failed.  I can take it out if you think so, but I liked it

gcorne: I think outputting help here suggests that the error is an error with the syntax. I think in most cases those errors are being reported sooner and if they are not, they should be.

PR #307, opened 2020-09-03
Added creation support

drewjenkins:
Allow for the creation of a schema given a JSON file defintion.


packages/cms-cli/lib/validation.js
```diff
@@ -130,7 +133,48 @@ function validateMode(command) {
   return false;
 }
 
+const getAbsoluteFilePath = _path => {
```
drewjenkins, line 136: I still need to move this to a different file, since it isn't validation related

gcorne:
I think we should create a top-level `objects` command and make this a subcommand.

packages/cms-cli/bin/cli.js
```diff
@@ -58,10 +59,11 @@ const argv = yargs
   .command(fetchCommand)
   .command(filemanagerCommand)
   .command(secretsCommand)
+  .command(schemaCommand)
```
gcorne, line 62: Can we hide this from help? Looks like we need to do the same for `logs` as well.

drewjenkins: It is hidden. Check out line 5 in schema.js

gcorne: Ah, cool. Didn't realize that is how `yargs` handles that.

packages/cms-cli/lib/validation.js
```diff
@@ -130,7 +133,48 @@ function validateMode(command) {
...
+    }
+  } catch (e) {
+    logger.error(`The path "${_path}" is not a path to a file`);
+    return null;
+  }
+
+  return absoluteSrcPath;
+};
+
+const isFileValidJSON = _path => {
```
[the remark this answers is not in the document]
gcorne: Is the plan to adjust the `hubdb` commands to use this?

packages/cms-cli/lib/validation.js
gcorne: I think `cms-lib/path` might be a good place for this.

gcorne: That said, it is a little bit surprising that this is mixing validation with resolving and that it returns `null` if the path is not a file.

drewjenkins: I can split it into 2 functions. `getAbsolutePath` which will just `return path.resolve(getCwd(), _path)` and `validatePath` as a separate function

packages/cms-cli/lib/validation.js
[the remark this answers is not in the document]
drewjenkins: Isn't HubDB CSV, not JSON?

gcorne:
> Currently there is a schema top level command. This is under that. I can switch it to objects, but my thought was for creating/updating/deleting schema we'd use this command, and then if we wanted to expose actual object instance creation/fetching/listing/etc that would be done under an objects command

Yeah, I realized that in retrospect. Don't feel super strongly at this stage, but `schema` is pretty generic. It might be nice if `schema` was under `objects`.

packages/cms-cli/bin/cli.js
```diff
@@ -58,10 +59,11 @@ const argv = yargs
...
   .help()
   .demandCommand(
     1,
-    `Please specifiy a command or run \`${SCRIPT_NAME} --help\` for a list of available commands`
+    `Please specify a command or run \`${SCRIPT_NAME} --help\` for a list of available commands`
```
gcorne, line 66: :+1:

packages/cms-cli/commands/schema/create.js
```diff
@@ -0,0 +1,82 @@
...
+  if (!filePath || !isFileValidJSON(filePath)) {
+    process.exit(1);
+  }
+
+  if (shouldIgnoreFile(filePath, getCwd())) {
+    logger.error(
+      `The file "${definition}" is being ignored via an .hsignore rule`
+    );
+    return;
+  }
```
gcorne, line 54: I don't think we should leverage `shouldIgnoreFile` here because it would be natural to want to avoid uploading the schemas via `.hsignore`.

drewjenkins: Ah yeah that makes sense.  Removing

packages/cms-cli/lib/validation.js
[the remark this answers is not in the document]
gcorne: The CLI is a little different than the UI in that it operates with a JSON format that combines both the schema and the rows. It is designed more for moving HubDB tables between portals and scaffolding out a table that is part of an app. The CSV import tool in the app is more for importing data from other sources, although the inconsistency is a good call out.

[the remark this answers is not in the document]
drewjenkins: Got it, then in that case yes we can totally do that

gcorne (commented):
Took a quick pass through the code.

packages/cms-cli/lib/validation.js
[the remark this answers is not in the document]
drewjenkins: Done

packages/cms-lib/api/schema.js
```diff
@@ -0,0 +1,67 @@
...
+
+const deleteSchema = async (portalId, schemaObjectType) =>
+  http.delete(portalId, {
+    uri: `${SCHEMA_API_PATH}/${schemaObjectType}`,
+  });
+
+const listSchemas = async portalId => {
+  const response = await fetchSchemas(portalId);
+  logSchemas(response.results);
+};
```
gcorne, line 57: I'd move this into `cms-lib/schemas` as well.

gcorne: And the `logSchemas` function.

packages/cms-cli/commands/customObject/schema/update.js
```diff
@@ -0,0 +1,69 @@
...
+  const { config: configPath } = options;
+  loadConfig(configPath);
+  checkAndWarnGitInclusion();
+
+  if (!(validateConfig() && (await validatePortal(options)))) {
+    process.exit(1);
+  }
+  const portalId = getPortalId(options);
+
+  trackCommandUsage('schema-update', null, portalId);
```
gcorne, line 37: I think this should be `custom-object-schema-update` and the same prefix should be used in the other commands so that the naming scheme is consistent with the other subcommands (e.g. `command-subcommand-subsubcommand`)

packages/cms-cli/commands/customObject/schema/update.js
```diff
@@ -0,0 +1,69 @@
...
+exports.builder = yargs => {
+  addTestingOptions(yargs, true);
+
+  yargs.positional('schemaObjectType', {
+    describe: 'Fully qualified name or object type ID of the target schema.',
+    type: 'string',
+  });
+
+  yargs.positional('definition', {
+    describe: 'local path to JSON file containing schema definition',
```
gcorne, line 66: Maybe "Local path to the JSON file"?

packages/cms-cli/commands/customObject/schema.js
```diff
@@ -0,0 +1,22 @@
+const createCommand = require('./schema/create');
+const fetchCommand = require('./schema/fetch');
+const fetchAllCommand = require('./schema/fetch-all');
+const deleteCommand = require('./schema/delete');
+const listCommand = require('./schema/list');
+const updateSchema = require('./schema/update');
+const associationsCommand = require('./schema/associations');
+
+exports.command = 'schema';
+exports.describe = 'Manage Custom Object Schema';
```
gcorne, line 10: Should probably be "Manage custom object schemas"? We aren't capitalizing "custom object" or "schema" in other  docs.

packages/cms-cli/commands/customObject/schema/fetch-all.js
```diff
@@ -0,0 +1,56 @@
...
+    ],
+    [
+      '$0 custom-object schema fetch-all my/folder',
+      'Fetch all schemas for a portal and put them in a directory named my/folder',
+    ],
+  ]);
+
+  yargs.positional('dest', {
+    describe:
+      'Local destination folder to write schemas to.  If omitted, current working directory will be used',
```
gcorne, line 53: Maybe "Local folder where schemas will be written." to avoid ending the sentence with a preposition?

PR #308, opened 2020-09-03
Alias HS to Banjo

drewjenkins:


packages/cms-cli/bin/cli.js
```diff
@@ -22,7 +22,7 @@ const fetchCommand = require('../commands/fetch');
 const filemanagerCommand = require('../commands/filemanager');
 const secretsCommand = require('../commands/secrets');
 
-const SCRIPT_NAME = 'banjo';
+const SCRIPT_NAME = 'hs';
```
gcorne, line 25: Can we assign this dynamically somehow based on the command?

packages/cms-cli/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "bin": {
     "banjo": "./bin/cli.js",
```
gcorne, line 26: I think we should go ahead and remove this now.

drewjenkins: Awe, I was going to keep it in as an easter egg

packages/cms-cli/bin/cli.js
drewjenkins: It's a tad hacky, but not too bad.  We can change the `bin` in the package.json to pass a flag

```
  "bin": {
    "banjo": "./bin/hs.js --command=banjo",
    "hs": "./bin/hs.js --command=hs",
    "hscms": "./bin/hs.js --command=hscms"
  },
````

and then parse it out and assign the script name to it

```
const commandName = process.argv.find(a => a.includes('--command='));
const SCRIPT_NAME = commandName
  ? commandName.split('--command=')[1] || 'hs'
  : 'hs';
```

Testing locally it worked

Comment (drewjenkins) [reply packages/cms-cli/bin/cli.js:25]: Actually.  `npm_lifecycle_event` is a thing.  Not really seeing much documentation on it, but I think it would work

Comment (gcorne) [reply packages/cms-cli/bin/cli.js]: What if we just use the default for `yargs` and don't call `yargs.scriptName()`? I think it works as expected.

File: packages/cms-cli/package.json (hunk 2 above)
Comment (gcorne) [reply packages/cms-cli/package.json]: The drawback with doing that is that it sort of pollutes the set of commands installed when a user installs globally, which is what we recommend.

File: packages/cms-cli/bin/cli.js (hunk 1 above)
Comment (drewjenkins) [reply packages/cms-cli/bin/cli.js]: image

That seems to make it use the file name instead.  The `npm_licecycle_event` thing seems to work well tho.  I'll try a couple older versions of node just to be safe

Comment (drewjenkins) [reply packages/cms-cli/bin/cli.js]: Tested with node 8.9.1 and works like a champ

Comment (gcorne) [reply packages/cms-cli/bin/cli.js]: Out of curiosity, how were you testing?

Comment (gcorne) [reply packages/cms-cli/bin/cli.js]: We may want to create an `hs` file inside `bin` and have it require cli.js. 

```bash
#!/usr/bin/env node

require('./cli');
```

drewjenkins: Swapped to node 8.9.1 and locally called the script

packages/cms-cli/bin/hs.js
```diff
@@ -1,34 +1,3 @@
 #!/usr/bin/env node
```
gcorne, line 1: Can we rename the file to `hs` or create a new `hs` file?

packages/cms-cli/package.json
```diff
@@ -23,8 +23,7 @@
     "node": ">=8.9.1"
   },
   "bin": {
-    "banjo": "./bin/cli.js",
-    "hs": "./bin/hs.js",
+    "hs": "./bin/cli.js",
```
gcorne, line 26: I'd switch this to `./bin/hs` once that file is in place.

PR #312, opened 2020-09-08
Improve messaging around the change in `watch` command behavior

gcorne:
image

@williamspiro @TheWebTech


williamspiro:
:shipit: thanks!

PR #313, opened 2020-09-09
Show help if user runs 'hs' with no command [YARGS]

anthmatic:
Previous: 
image
Current: 
image
This PR:
image

### DIFF
--- a/packages/cms-cli/bin/cli.js
+++ b/packages/cms-cli/bin/cli.js
@@ -64,4 +64,5 @@ if (argv.help) {
   trackHelpUsage(getCommandName(argv));
 } else if (argv['_'].length == 0) {
   yargs.showHelp();
+  logger.error('Please specify a command to run');
 }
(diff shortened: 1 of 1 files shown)


packages/cms-cli/bin/cli.js
```diff
@@ -57,13 +57,11 @@ const argv = yargs
...
-    1,
-    `Please specify a command or run \`${yargs.$0} --help\` for a list of available commands`
-  )
   .recommendCommands()
   .strict().argv;
 
 if (argv.help) {
   trackHelpUsage(getCommandName(argv));
+} else if (argv['_'].length == 0) {
+  yargs.showHelp();
```
gcorne, line 66: Makes sense. I wonder if we should also include some copy to indicate that a command is required sort of like the message that was used previously.

gcorne (approved):
One suggestion otherwise LGTM.

packages/cms-cli/bin/cli.js
anthmatic: As I was working on getting a message added, I might have found a less hacky way of showing help. Will have that up shortly

anthmatic: Never mind, I was going to add `showHelp()` to the `.fail()` cb, but it looks like we already went down that road in https://github.com/HubSpot/hubspot-cms-tools/pull/305#pullrequestreview-478853884

PR #317, opened 2020-09-16
Adding required --use-env option to utilize environment variable config settings

### LINKED ISSUE #257: Make using environment variable based config explicit when running commands
Currently, when you have the required environment variables set, it will override the `hubspot.config.yml` settings. This can be a problem if the variables are set unbeknownst to the user and commands are attempted to be run using the `hubspot.config.yml` settings.

We should look into adding an option to commands to explicitly use the environment variable config like `hs watch --use-env` or something similar. Research some different options for this.

miketalley:
### Context
Loading a config from environment variables is currently enabled as long as the required environment variables are set. The three authentication types and the required environment variables for each group are seen below.

Personal Access Key
 - HUBSPOT_PERSONAL_ACCESS_KEY=\<personalAccessKey>
 - HUBSPOT_PORTAL_ID=\<portalId>

OAuth2
 - HUBSPOT_PORTAL_ID=\<portalId>
 - HUBSPOT_CLIENT_ID=\<clientId>
 - HUBSPOT_CLIENT_SECRET=\<clientSecret>
 - HUBSPOT_REFRESH_TOKEN=\<refreshToken>

API Key
 - HUBSPOT_PORTAL_ID=\<portalId>
 - HUBSPOT_API_KEY=\<apiKey>

We want to make using the environment variable config explicit so as not to use it unbeknownst and create problems.

### Brief Description of PR
- Adds a `HUBSPOT_ENVIRONMENT` to allow specifying `qa` value for `env` in config
- Adds a `--use-env` option to applicable commands
- Passes options forward to `loadConfig` method
- Makes loading environment variable config disabled by default unless `--use-env` is specified when running command

Fixes #257 

/cc @TheWebTech


packages/cms-lib/lib/config.js
```diff
@@ -319,16 +319,15 @@ const loadConfigFromFile = (path, options = {}) => {
 const loadConfig = (
   path,
   options = {
-    ignoreEnvironmentVariableConfig: false,
+    useEnv: false,
```
miketalley, line 322: This should be `false` because we are changing it to be disabled by default and I am changing the truthiness of the variable diction.

packages/cms-cli/commands/fetch.js
```diff
@@ -72,6 +73,7 @@ const builder = yargs => {
   addPortalOptions(yargs, true);
   addOverwriteOptions(yargs, true);
   addModeOptions(yargs, { read: true }, true);
+  addUseEnvironmentOptions(yargs, true);
```
drewjenkins, line 76: Is this being used in every command, or just some?  If it is, we can probably add it more as a middleware like we do with the logger options for yargs

miketalley: Not all, but most

packages/cms-cli/lib/commonOpts.js
```diff
@@ -78,6 +78,21 @@ const addTestingOptions = (program, useYargs = false) => {
   program.option('--qa', 'run command in qa mode', false);
 };
 
+const addUseEnvironmentOptions = (program, useYargs = false) => {
+  const option = 'use-env';
+  const description = 'use environment variable config';
+  require('dotenv').config();
```
gcorne, line 84: Should this be removed?

PR #318, opened 2020-09-16
Cleanup prior to explicit environment variable option work

miketalley:
### Brief Description
I started on the [explicit environment variable config option](https://github.com/HubSpot/hubspot-cms-tools/issues/257) work, but ended up doing some cleanup in preparation and thought it warranted a separate PR to keep things clear.

I also noticed we have a lot of inconsistencies with the argument naming between commands and also found an unused file, so cleaned those up as well.

#### Included in this PR
- Removed unused [hscms-sync file](https://github.com/HubSpot/hubspot-cms-tools/blob/master/packages/cms-cli/bin/hscms-sync.js)
- Cleaned up naming of command parameters for command actions to have consistency and clearly reflect the actual values. The parameter patterns will now be one of...
  - `async function actionName options => {...}`
  - `async function actionName ({ some, arguments }, options) => {...}`
  - `async function actionName (someArgument, options) => {...}`
- Used destructuring for `{ some, options }` for first parameter and removed unnecessary variables where applicable


packages/cms-cli/commands/auth.js
```diff
@@ -48,13 +48,13 @@ const SUPPORTED_AUTHENTICATION_PROTOCOLS_TEXT = commaSeparatedValues(
   ALLOWED_AUTH_METHODS
 );
 
-async function authAction(type, options) {
```
drewjenkins, line 51: This is fine, but just fyi the reason I had changed it was because to me `command` may have a connotation of using `commander`, whereas with `yargs` I believe it is actually the options that are being passed

miketalley: Gotcha. I don't really care if we use `command` or `options` for this, to me having consistency is the most important thing. It was confusing when looking at different commands to see options be different things at times.

packages/cms-lib/lib/config.js
```diff
@@ -388,8 +389,12 @@ const getPortalConfig = portalId => {
   return config.portals.find(portal => portal.portalId === portalId);
 };
 
-const getPortalId = nameOrId => {
+/*
+ * Returns a portalId from the config if it exists, else returns null
+ */
+const getPortalId = (command = {}) => {
```
gcorne, line 395: Are we sure that we want to bring the concept of a command to `@hubspot/cms-lib`? I think that this also is a breaking change since the signature changes.

packages/cms-cli/commands/auth.js
miketalley: After putting more thought into this, I think `options` makes more sense because it could be either `Commander.command` or `object`(Yargs options) types. Updated them to `options` throughout.

packages/cms-lib/lib/config.js
miketalley: Reverted this after our discussion.

PR #320, opened 2020-09-17
Get rid of commander, alias hscms over to hs, clean up code

drewjenkins:


packages/cms-cli/bin/hscms
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+
+require('./cli');
```
drewjenkins, line 3: alas `hscms` to `hs`

packages/cms-cli/commands/auth.js
```diff
@@ -39,7 +34,7 @@ const {
 } = require('../lib/prompts');
 const { commaSeparatedValues } = require('../lib/text');
 
-const COMMAND_NAME = 'auth';
+const COMMAND_NAME = 'auth <type>';
```
drewjenkins, line 37: Here and throughout I cleaned up and normalized all of these.  Removed the `demand` option where relevant and instead put it here in the command name. SHould be consistent throughout now

drewjenkins: Also, I have this as <type> instead of [type] since it has a default... not sure if that is technically correct or not but it seems to work well

packages/cms-cli/commands/logs.js
```diff
@@ -125,15 +118,17 @@ const tailLogs = async ({
...
+exports.handler = async options => {
   loadAndValidateOptions(options);
 
-  const { functionPath } = args;
-  const { latest, file, follow, compact } = options;
+  const { latest, tail, compact, path: functionPath } = options;
   let logsResp;
   const portalId = getPortalId(options);
 
-  trackCommandUsage(COMMAND_NAME, { latest, file }, portalId);
```
drewjenkins, line 136: Removed file because it wasn't actually being captured or used

packages/cms-cli/commands/remove.js
```diff
@@ -10,28 +10,23 @@ const {
...
 const { validatePortal } = require('../lib/validation');
-const {
-  trackCommandUsage,
-  addHelpUsageTracking,
-} = require('../lib/usageTracking');
+const { trackCommandUsage } = require('../lib/usageTracking');
 
-const COMMAND_NAME = 'remove';
-const DESCRIPTION = 'Delete a file or folder from HubSpot';
+exports.command = 'remove <path>';
```
drewjenkins, line 23: This wasn't marked as required anywhere but it actually should be

packages/cms-cli/commands/server.js
```diff
@@ -1,69 +0,0 @@
-#!/usr/bin/env node
```
drewjenkins, line 1: Removed this file completely.  Are there any concerns over this command?  I think technically it would already be broken if being used since we swapped hs to yargs

gcorne:
Didn't run through the changes to see how much is left to do, but I think it would be great if we can get this in pretty soon since we released 2.1.0.

packages/cms-cli/commands/auth.js
```diff
@@ -29,19 +28,15 @@ const {
 } = require('../lib/prompts');
 const {
   addConfigOptions,
-  addLoggerOptions,
```
miketalley, line 32: Guessing this and `addHelpUsageTracking` are being done at a higher level now?

packages/cms-cli/commands/auth.js
```diff
@@ -84,6 +83,7 @@ async function authAction(type, options) {
   }
 
   trackCommandUsage(COMMAND_NAME);
```
miketalley, line 85: I think this needs to be changed so it passes `auth` instead of `auth <type>`

packages/cms-cli/commands/init.js
```diff
@@ -19,29 +19,18 @@ const { logger } = require('@hubspot/cms-lib/logger');
...
-  addHelpUsageTracking,
-  trackAuthAction,
-} = require('../lib/usageTracking');
-const {
-  addLoggerOptions,
-  setLogLevel,
-  addTestingOptions,
-} = require('../lib/commonOpts');
 const { logDebugInfo } = require('../lib/debugInfo');
-const { updatePortalConfig } = require('@hubspot/cms-lib/lib/config');
```
miketalley, line 40: 👍

packages/cms-cli/commands/logs.js
```diff
@@ -121,14 +114,17 @@ const tailLogs = async ({
...
 
-const action = async ({ functionPath }, options) => {
+exports.command = 'logs <path>';
+exports.describe = 'get logs for a function';
+
+exports.handler = async options => {
   loadAndValidateOptions(options);
 
-  const { latest, file, follow, compact } = options;
+  const { latest, tail, compact, path: functionPath } = options;
```
miketalley, line 123: I had set this to `tail` initially and after discussion we agreed on `follow`. Also this would be a breaking change, right?

miketalley (changes requested):

anthmatic (approved):
Aside from @miketalley's comments, this LGTM

packages/cms-cli/commands/auth.js
drewjenkins: Yep they are both added in cli.js

packages/cms-cli/commands/logs.js
drewjenkins: I have it aliased so it won't break, but sounds good. I'll swap it back and make tail the alias.

```
yargs.option('tail', {
    alias: ['t', 'follow', 'f'],
    describe: 'tail logs',
    type: 'boolean',
  });
```

packages/cms-cli/commands/auth.js
drewjenkins: Woops, good catch

PR #323, opened 2020-09-21
Improved hs logs command output and added additional options

miketalley:
### Description
- Improved output of `--compact` error logs to display as single line always
- Added `--limit`(alias `-n` and `--max-count`) that limits the number of logs displayed in output
- Increased depth of objects in output so nested JSON properties will be shown
- Added color to JSON output

#### Nested objects only outputs type and no color
image

#### Nested object display with color added
image

#### Using --limit=3 option displays 3 logs only
image

#### Compact errors now on one line
image

_Note:_ The order of the logs is currently oldest first, but there is an issue fixed to order by timestamp DESC [here](https://git.hubteam.com/HubSpot/ContentServerlessFunctions/issues/159).

/cc @williamspiro


packages/cms-cli/commands/logs.js
```diff
@@ -171,20 +169,26 @@ exports.builder = yargs => {
...
+      type: 'boolean',
+    },
+    follow: {
+      alias: ['t', 'tail', 'f'],
+      describe: 'follow logs',
+      type: 'boolean',
+    },
+    limit: {
+      alias: ['limit', 'n', 'max-count'],
+      describe: 'limit the number of logs to output',
```
drewjenkins, line 189: Does this interact with --follow?

miketalley: Good catch. I added `.conflict('follow', 'limit')`so they cannot be used together.

Issue #181, opened 2020-04-27
Add testing and content generation commands


axellabs:
For the past few weeks, I've been working on a small utility to generate pages and run automated tests on them. I think it could be useful to integrate that functionality into this app. 

Testing functions:
- generate sample pages out of templates
- test site updates with visual regressions
- generate blog posts and sample content automatically (referenced in https://github.com/HubSpot/hubspot-cms-tools/issues/38)
- delete test pages

[Link to repo of testing utility](https://git.hubteam.com/HubSpot/DesignAssetsAcceptanceTesting)

@ckconant @levinkeo

axellabs:
@gcorne

gcorne:
I think we should start with content generation. @axellabs is this something that you are planning to work on? If so, maybe we can discuss on Slack.

One thing that will need to be done before we release broadly is adjust personal access keys to add support for the `content` scope group so that the key can have permission to hit the blog and pages endpoints.

axellabs:
I'm finishing up content generation this week! It exists for blog posts right now, and I'm adding a feature for websites. I can add you to that PR. 

I know a little about personal access keys but I'd love to talk more about adding them to this application.

brandenrodgers:
Related: https://github.com/HubSpot/hubspot-cli/issues/38

TanyaScales:
Closing this out as we have moved to an internal testing package for content assets

Issue #38, opened 2019-12-18
Add `hs generate blog` to start building out commands for generating content for testing


gcorne:

gcorne:
I wonder if we should invert this a bit and have `hs blog fetch` and `hs blog generate`.

brandenrodgers:
Closing in favor of https://github.com/HubSpot/hubspot-cli/issues/181

PR #325, opened 2020-09-21
super basic object creation

drewjenkins:
Some basic CO object creation to get the demo unblocked


packages/cms-cli/commands/customObject/create.js
```diff
@@ -0,0 +1,57 @@
...
+  checkAndWarnGitInclusion();
+
+  if (!(validateConfig() && (await validatePortal(options)))) {
+    process.exit(1);
+  }
+  const portalId = getPortalId(options);
+
+  trackCommandUsage('custom-object-create', null, portalId);
+
+  const filePath = getAbsoluteFilePath(definition);
```
drewjenkins, line 33: I think ideally we'd probably want the option for an interaction key/value input, this was just to help us for now

packages/cms-lib/schema.js
```diff
@@ -25,11 +25,39 @@ const logSchemas = schemas => {
   logger.log(data.length ? table(data, tableConfig) : 'No Schemas were found');
 };
 
-const writeSchemaToDisk = (schema, dest) => {
+const cleanSchema = schema => {
```
drewjenkins, line 28: Not sure if there's a different way we wanna do this, I just took @anthmatic example from the Slack thread and grabbed those properties.  I tested and am able to create new schemas from the cleaned ones so it seems to work well

packages/cms-cli/commands/customObject/schema/associations/create.js
```diff
@@ -42,9 +43,9 @@ exports.handler = async options => {
   try {
     // const res = await createSchema(portalId, filePath);
     // logger.success(
-    //   `Schema can be viewed at ${
+    //   `Schema can be viewed at ${getHubSpotWebsiteOrigin(
     //     getEnv() === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD
```
anthmatic, line 47: Minor, but I noticed that when `qa` is added, this the path is incorrect (it would be `api.hubspotqa.com` or similar)

anthmatic: Scratch that, I see that this was fixed!

packages/cms-lib/schema.js
gcorne: I think we should add some unit tests for this.

packages/cms-cli/commands/customObject/batch.js
```diff
@@ -0,0 +1,9 @@
+const createCommand = require('./batch/create');
```
gcorne, line 1: Instead of having a batch command what if we have a single command that expects the JSON file to be an array of objects? The thinking is that we should strive for a minimal command surface area when we can.

drewjenkins: Sure we can do that.  Are you saying also combine the `batch/create` and `create`, or just bring `batch/create` up a level to be something like `create-multiple`?

gcorne: Combine into a single `hs custom-object create` command. The thinking is that the most common use case is creating multiple so a little friction for creating a single object is not the end of the world. It also mirrors how I image other situations might work like `hs blog-post create` or `hs page create` — both of which have issues (e.g. #38 and #181).

packages/cms-cli/commands/customObject/create.js
drewjenkins: removed in favor of always doing batch for now

PR #327, opened 2020-09-22
Adding apikey auth method to auth and init commands

### LINKED ISSUE #324: Add support for 'authType: apikey' to 'hs auth'
Currently you can't use `hs auth` to authorize using an API key, even though it is supported if you manually edit your `hubspot.config.yml`:
```
$ hs auth apikey
[ERROR] Unsupported auth type: apikey. The only supported authentication protocols are oauth2 and personalaccesskey.
```

However you can edit your config yml to include the following:
```
    authType: apikey
    apiKey: c8291c24-...
```
Since apikey is a accepted auth type, could we teach `hs auth` to support it?

cc @drewjenkins

miketalley:
### Description
- Adds `--auth` option to `hs init` command to create a new `hubspot.config.yml` config using the specified auth method
  - `hs init --auth=apikey`
  - `hs init --auth=oauth2`
  - `hs init --auth=personalaccesskey`
- Adds `hs auth apikey` auth method to add additional portal config using `apiKey` auth method

Fixes #324 

/cc @gmcnaughton


gcorne:
When going through `hs init --auth apikey`, can we set the `defaultPortal` to the name that user chooses?

packages/cms-cli/commands/auth.js
```diff
@@ -67,8 +88,25 @@ async function authAction(type, options) {
   trackCommandUsage(COMMAND_NAME);
   let configData;
   let updatedConfig;
-  let promptAnswer;
+  let validName;
   switch (authType) {
+    case API_KEY_AUTH_METHOD.value:
```
drewjenkins, line 93: The yargs command description below needs updated to reflect the new option

packages/cms-cli/commands/auth.js
```diff
@@ -8,6 +8,7 @@ const { logger } = require('@hubspot/cms-lib/logger');
 const {
```
drewjenkins, line 8: Not in this PR, but it looks like the endpoint it sends you to when entering your key is wrong.    

```
'Enter the API key for your portal (found at https://app.hubspot.com/l/api-key):'
```

https://app.hubspot.com/l/api-key should be https://app.hubspot.com/api-key

packages/cms-cli/commands/init.js
```diff
@@ -56,54 +118,36 @@ const action = async options => {
...
   }
 };
 
 // Yargs Configuration
 const command = `${COMMAND_NAME}`;
 const describe = DESCRIPTION;
 const builder = yargs => {
+  yargs.option('auth', {
+    describe:
+      'specify auth method to use ["personalaccesskey", "oauth2", "apikey"]',
```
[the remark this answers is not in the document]
drewjenkins: You can also set a default choice so it becomes optional

packages/cms-cli/commands/init.js
```diff
@@ -115,6 +159,10 @@ const configureCommanderInitCommand = program => {
   program
     .version(version)
     .description(DESCRIPTION)
+    .option(
```
drewjenkins, line 165: just an fyi, we are removing commander in the very near future https://github.com/HubSpot/hubspot-cms-tools/pull/320

packages/cms-cli/commands/auth.js
miketalley: Nice catch. Strange that it seems to work regardless of this value being in the `choices` array.

packages/cms-cli/commands/auth.js
miketalley: That's actually the correct link -- it will allow you to choose from different portals if you have multiple.

drewjenkins: Nevermind, Gregory showed me short links.  Mine was broken for some reason for awhile, it was redirecting to `https://app.hubspot.com/api-key)/8045710` ¯\_(ツ)_/¯

packages/cms-cli/commands/init.js
miketalley: Yeah I thought about this, but figured I might as well update them both if they are still there.

packages/cms-cli/commands/init.js
```diff
@@ -56,54 +108,43 @@ const action = async options => {
...
     );
 
-    trackAuthAction(
-      COMMAND_NAME,
-      PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
-      TRACKING_STATUS.COMPLETE,
-      portalId
-    );
+    trackAuthAction(COMMAND_NAME, authType, TRACKING_STATUS.COMPLETE, portalId);
+    process.exit();
```
gcorne, line 125: Why is this necessary?

packages/cms-lib/lib/models/OAuth2Manager.js
```diff
@@ -15,6 +15,7 @@ class OAuth2Manager {
       scopes,
       environment = ENVIRONMENTS.PROD,
       tokenInfo = { expiresAt: null, refreshToken: null, accessToken: null },
+      name,
```
gcorne, line 18: I am a bit worried that introducing `name` here may have unexpected side effects. For example, it not passed via `fromConfig()`. Are we sure that we have covered all the places where an `OAuth2Manager` is constructed?

gcorne: Sounds like we're covered. Adding unit test coverage would be a way to increase confidence.

packages/cms-cli/commands/init.js
gcorne: Alright. I would feel more comfortable if we knew "why" it was necessary and why we only need the exit in the `try` and not the `catch`.

PR #328, opened 2020-09-23
Moved personalaccesskey prompts into cms-cli

miketalley:
#### Description
Redoing what we [had previously done](https://github.com/HubSpot/hubspot-cms-tools/commit/f53e1d38d3c2a3a882595713a152ac4dc9590a7c) and was undone in a recent commit in order to reduce `@hubspot/cms-lib` dependencies.

- Moves `personalaccesskey` promps from `@hubspot/cms-lib` to `@hubspot/cms-cli`
- Removes `open` and `inquirer` deps from `@hubspot/cms-lib`


gcorne:
Is there a reason for only moving some of them?

PR #332, opened 2020-09-24
Various CO Hardening Tasks

drewjenkins:
- [x] ~Only Allow Alpha-Numeric Schema Names~ After discussing with CO backend team, I reverted this change.  They say the `-` `_` issue should be resolved soon, and only allowing a-z0-9 may cause issues with CJK (chinese japanese korean) customers.
- [x] Make it clear that CO is in beta
- [x] Make help commands more clear
- [x] Remove Associations while we rethink how they should work
- [x] Add Unit Tests

When running any command that requires a subcommand, it wasn't obvious what was wrong if you specified nothing.  Now we display help.  Also put custom-objects in the help since they are released

BEFORE
before

AFTER
after


packages/cms-cli/commands/customObject.js
```diff
@@ -1,15 +1,41 @@
...
+  console.warn(
+    chalk.reset.yellow(
+      'See https://developers.hubspot.com/docs/api/crm/crm-custom-objects to find out more.'
+    )
+  );
+  console.warn(
+    chalk.reset.yellow(
+      '***************************** WARNING WARNING WARNING ****************************'
+    )
+  );
```
gcorne, line 38: Can we use the `logger` for this?

drewjenkins: So I was going to at first, but it wasn't showing up cuz the default logger level didn't show warning.  I figured we always want this to show

miketalley (approved):

packages/cms-cli/commands/customObject.js
drewjenkins: I could add a new level to logger, like `logger.announce`?

gcorne: Yeah, I think something like that makes sense. Since this is on the CLI side instead of inside `cms-lib`, it isn't a big deal.

drewjenkins: 👍 I'ma keep it as is for now

package.json
```diff
@@ -20,7 +20,8 @@
     "prettier:write": "prettier --write 'packages/**/*.{js,json}'",
     "publish-release": "yarn lerna publish --conventional-graduate",
     "publish-prerelease": "yarn lerna publish prerelease --preid beta --dist-tag next",
-    "test": "jest",
+    "test": "jest --no-colors",
```
gcorne, line 23: Not a fan of colors? I kind of like them so am kind of bummed to see this.

packages/cms-lib/__tests__/__snapshots__/schema.js.snap
```diff
@@ -172,3 +172,32 @@ Array [
...
+      \\"name\\": \\"name\\",
+      \\"label\\": \\"Name\\",
+      \\"type\\": \\"string\\",
+      \\"fieldType\\": \\"text\\",
+      \\"description\\": \\"Name Field\\"
+    }
+  ]
+}
+"
+`;
```
gcorne, line 203: :+1:

package.json
drewjenkins: It's just for the tests.  The CI server doesn't support them, whereas locally our terminals do.  It was causing test failures.

PR #333, opened 2020-09-24
Improve how scope errors are handled to only fetch scopes when needed

gcorne:
While working on #331, I decided to improve how the error handling works to only fetch the scope info when the error is a scope error.

### Technical design

The basic idea was to break up `errorHandlers` so that the error handlers for the different layers were defined in separate modules (e.g. standard, file system, and API). Splitting this up made it possible for the `config` module to use only standard and file system error handlers, which makes sense since `config` is lower level. The benefit of making this split is that it avoids circular dependencies.

TODO:

 - [x] Add unit test coverage


gcorne:
Going to try and find some time to add some more unit test coverage, but I think that this is in pretty good shape. @drewjenkins @anthmatic might be worthwhile for one of you also taking a look.

PR #339, opened 2020-09-30
Only write files on successful fetch

### LINKED ISSUE #265: Investigate 502 Bad Gateway HTML overwriting image files when fetching
Reported https://hubspotdev.slack.com/archives/CLJJXGTC0/p1595969690247400 and https://hubspotdev.slack.com/archives/CLJJXGTC0/p1592250962497600

It sounds like when you fetch assets, if they return a 502 Bad Gateway, the file gets replaced by the HTML response
```
<head><title>502 Bad Gateway</title></head>
<body bgcolor="white">
<center><h1>502 Bad Gateway</h1></center>
</body>
</html>
```

anthmatic:
Previously, on fetch, files were getting written even when a bad request came through. Sometimes this would result in a file with the wrong content (like an HTML response body).

This moves the logic for writing files up a level and only creates a writeStream when a successful fetch has occurred.

Resolves #265

### DIFF
--- a/packages/cms-lib/__tests__/http.js
+++ b/packages/cms-lib/__tests__/http.js
@@ -1,4 +1,5 @@
 const requestPN = require('request-promise-native');
+const request = require('request');
 const fs = require('fs-extra');
 const moment = require('moment');
 const { getAndLoadConfigIfNeeded, getPortalConfig } = require('../lib/config');
@@ -12,13 +13,22 @@ jest.mock('request-promise-native', () => ({
 }));
 
 jest.mock('request', () => ({
-  get: jest.fn().mockReturnValue({
-    on: jest.fn((val, cb) => {
-      if (val === 'response') {
-        cb({ statusCode: 200 });
+  get: jest.fn(({ uri }) => {
+    const isValidPath = path => {
+      if (path === 'some/endpoint/path') {
+        return true;
+      } else {
+        return false;
       }
-    }),
-    pipe: jest.fn(),
+    };
+    return {
+      on: jest.fn((val, cb) => {
+        if (val === 'response') {
+          cb({ statusCode: isValidPath(uri) ? 200 : 404 });
+        }
+      }),
+      pipe: jest.fn(),
+    };
   }),
 }));
 jest.mock('../lib/config');
@@ -31,7 +41,7 @@ describe('http', () => {
     getPortalConfig.mockReset();
   });
   describe('getOctetStream()', () => {
-    it('not sure', async () => {
+    it('makes a get request', async () => {
       getAndLoadConfigIfNeeded.mockReturnValue({
         httpTimeout: 1000,
         portals: [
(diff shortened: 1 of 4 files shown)


packages/cms-lib/fileMapper.js
```diff
@@ -265,29 +265,12 @@ async function fetchAndWriteFileStream(input, srcPath, filepath) {
...
-      new FileSystemErrorContext({
-        filepath,
-        portalId,
-        write: true,
-      })
-    );
-  };
-  let writeStream;
-  try {
-    await fs.ensureFile(filepath);
```
anthmatic, line 280: I don't believe we need this, because `createWriteStream` will create the file by default

packages/cms-lib/http.js
```diff
@@ -129,18 +147,25 @@ const createGetRequestStream = ({ contentType }) => async (
...
+            reject(err);
+          });
+          writeStream.on('close', async () => {
+            logger.log('Wrote file "%s"', filepath);
+            resolve(res);
+          });
+        } else {
+          reject(res);
+        }
+      });
```
gcorne, line 168: I wonder if a better solution might be to "clean up" whenever there is an error with either streaming the file to disk or the HTTP request (probably via `fs.unlink` or the synchronous equivalent). This would still involve creating the write stream but doing so before making the request.

gcorne:
Since this is an important low-level portion of the codebase, I think we should make sure to add some test cases to cover the different scenarios.

packages/cms-lib/http.js
anthmatic: I did investigate that approach and it was a challenge with `--overwrite` because `createWriteStream` effectively overwrites the file immediately. Calling `unlink` does clean that up, but we've now removed an otherwise "good" file.

That would fix the issue cases where you're fetching a file for the first time, but doesn't help things if you're overwriting.

anthmatic: Another approach would be to download to a `tmp` dir and move successfully fetched files over to the irdestination

anthmatic:
> Since this is an important low-level portion of the codebase, I think we should make sure to add some test cases to cover the different scenarios.

Working on some test coverage now

miketalley (approved):

packages/cms-lib/http.js
gcorne: I think moving forward with the implementation in this PR is fine with me. I hadn't considered `--overwrite`. I am a bit curious about how others handle this but don't want to hold up a fix.

anthmatic: Sounds good. I'm working on adding some tests around this and will get this in on monday

anthmatic:
I added a few tests around these changes. @gcorne let me know if you have any feedback one those.

packages/cms-lib/__tests__/http.js
```diff
@@ -9,6 +11,26 @@ jest.mock('request-promise-native', () => ({
...
+        return true;
+      } else {
+        return false;
+      }
+    };
+    return {
+      on: jest.fn((val, cb) => {
+        if (val === 'response') {
+          cb({ statusCode: isValidPath(uri) ? 200 : 404 });
+        }
```
gcorne, line 28: Instead of the logic for faking a 404 response being at the level of the mock definition like this, I think it is a bit nicer if the mocked response is closer to the test. This makes it easier to see what is being tested and how.

packages/cms-lib/__tests__/http.js
```diff
@@ -18,6 +40,89 @@ describe('http', () => {
...
+          {
+            id: 123,
+            apiKey: 'abc',
+          },
+        ],
+      });
+      getPortalConfig.mockReturnValue({
+        id: 123,
+        apiKey: 'abc',
+      });
```
gcorne, line 57: I'd movie these into a `before()` section since they are shared across all tests in the `describe` block.

packages/cms-lib/__tests__/http.js
```diff
@@ -18,6 +22,86 @@ describe('http', () => {
...
+      try {
+        await http.getOctetStream(
+          123,
+          {
+            uri: 'some/nonexistent/path',
+          },
+          'path/to/local/file'
+        );
+      } catch (e) {
+        expect(e.statusCode).toBe(404);
```
gcorne, line 101: Might also add a check to verify that `fs.createWriteStream` is not called.

gcorne (approved):
One small suggestion to add coverage over not doing any writes when response is a 404.

PR #342, opened 2020-10-05
Add autocompletion script

drewjenkins:
Take advantage of yargs autcompletion script. I was looking to see if there was a way to modify the output so we could add a bit more detail about the installation instructions but I couldn't find anything in the docs about it.


anthmatic (approved):
It might be worth throwing a note in the readme about this. FWIW, the completion script guessed the wrong config file for my (`.bashrc` instead of `.zshrc`) which is probably an issue on my end, but worth noting.

Otherwise, LGTM!

PR #344, opened 2020-10-07
hs open command

drewjenkins:
Added new `hs open` command.  This command will open a shortcut in the user's web browser.  The list so far isn't the definitive list, and can be added to as we see fit.

There are a few different ways to use `hs open`.

- `hs open`/`hs open --list` - display a list of all available shortcuts
- `hs open shortcut-name` - open the shortcut with name `shortcut-name` in your browser
- `hs open sn` - open the shortcut with an alias of `sn` in your browser
- `hs open settings/navigation` - Some links are sublinks within larger sections.  Stylistically, I used `/` to show this. `settings/navigation` means that we are opening the navigation section within settings.
- `hs open 10` - open the shortcut at index 10 in your browser.  I added this because I see a use case of someone doing `hs open` to find the shortcut they want, then just wanting the fastest way to open it.  They can glance at the index and do this as shorthand.  The index is subject to change if we add more commands, so there is a tiny bit of danger if someone were to do something like `hs open 10` in a script, but I think that risk is pretty negligible and I would hope someone would use the full command in their script anyways.


anthmatic:
Nice! My only thought is that when I ran `hs open` I expected to be able to select a value by index (like and inquirer.js `rawlist` prompt). Even though, as far as I know, we don't use that pattern in this CLI, it might be confusing. That's definitely not a deal breaker, though.

packages/cms-cli/commands/open.js
```diff
@@ -0,0 +1,56 @@
...
+    type: 'boolean',
+  });
+
+  yargs.example([
+    ['$0 open --list'],
+    ['$0 open settings'],
+    ['$0 open settings/navigation'],
+    ['$0 open sn'],
+    ['$0 open 10'],
+  ]);
```
gcorne, line 49: Ah, this is cool. We should use it more.

packages/cms-lib/links.js
```diff
@@ -0,0 +1,138 @@
+const { getEnv } = require('./lib/config');
```
gcorne, line 1: I think we should keep this in `cms-cli` since it is unlikely to be useful outside the CLI.  Moving to CLI has the added benefit that it reduces the dependencies needed in `@hubspot/cms-lib`.

packages/cms-lib/links.js
```diff
@@ -0,0 +1,138 @@
...
+const { ENVIRONMENTS } = require('./lib/constants');
+const { getHubSpotWebsiteOrigin } = require('./lib/urls');
+const { logger } = require('./logger');
+const chalk = require('chalk');
+const { table, getBorderCharacters } = require('table');
+var opn = require('opn');
```
gcorne, line 7: This looks to be deprecated in favor of [open](https://www.npmjs.com/package/open), which we're using elsewhere.

gcorne:
> hs open 10 - open the shortcut at index 10 in your browser. I added this because I see a use case of someone doing hs open to find the shortcut they want, then just wanting the fastest way to open it. They can glance at the index and do this as shorthand. The index is subject to change if we add more commands, so there is a tiny bit of danger if someone were to do something like hs open 10 in a script, but I think that risk is pretty negligible and I would hope someone would use the full command in their script anyways.

Maybe I am alone in this thought, but I think having too many ways to accomplish a task is confusing. Is there a reason that we need the numeric form?

packages/cms-lib/links.js
drewjenkins: Ah cool, thanks. I'll change it.

packages/cms-cli/lib/links.js
```diff
@@ -0,0 +1,130 @@
...
+      alias: 's',
+      url: `${baseUrl}/settings/${portalId}`,
+    },
+    SETTINGS_NAVIGATION: {
+      shortcut: 'settings/navigation',
+      alias: 'sn',
+      url: `${baseUrl}/menus/${portalId}/edit/`,
+    },
+    SETTINGS_PAGE: {
+      shortcut: 'settings/page',
```
gcorne, line 86: Maybe `settings/website`?

gcorne (approved):

PR #345, opened 2020-10-07
Prompt for secret values

anthmatic:
Removes the `value` argument from `hs secrets add` and `hs secrets update` and replaces it with a prompt where the use can input their secret as a masked input like:
```
? Enter a value for your secret *****
```

### DIFF
--- a/packages/cms-cli/commands/secrets/addSecret.js
+++ b/packages/cms-cli/commands/secrets/addSecret.js
@@ -21,12 +21,13 @@ const {
   getPortalId,
 } = require('../../lib/commonOpts');
 const { logDebugInfo } = require('../../lib/debugInfo');
+const { secretValuePrompt } = require('../lib/secretPrompt');
 
 exports.command = 'add <name> <value>';
 exports.describe = 'Add a HubSpot secret';
 
 exports.handler = async options => {
-  const { config: configPath, name: secretName, value: secretValue } = options;
+  const { config: configPath, name: secretName } = options;
 
   setLogLevel(options);
   logDebugInfo(options);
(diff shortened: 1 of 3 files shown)


williamspiro:
Thanks for doing this. cc @TheWebTech this will require some doc updates, both for the CLI docs and sample repos like https://github.com/HubSpot/cms-event-registration#getting-started

packages/cms-cli/commands/secrets/updateSecret.js
```diff
@@ -40,6 +41,8 @@ exports.handler = async options => {
   trackCommandUsage('secrets-update', {}, portalId);
 
   try {
+    const { secretValue } = await secretValuePrompt();
```
drewjenkins, line 44: Are we doing this for a security reason?  Wouldn't the secret still be copied to your clipboard if you are passing it into inquirer?

anthmatic: Copied to clipboard, yes. But my prompting for the value like this we can keep the value out of bash `history`.

packages/cms-cli/commands/secrets/addSecret.js
```diff
@@ -21,12 +21,13 @@ const {
   getPortalId,
 } = require('../../lib/commonOpts');
 const { logDebugInfo } = require('../../lib/debugInfo');
+const { secretValuePrompt } = require('../../lib/secretPrompt');
 
-exports.command = 'add <name> <value>';
+exports.command = 'add <name>';
```
gcorne, line 26: I think we also need to add the positional argument from the builder.

anthmatic: I went ahead and removed those

gcorne (approved):

Issue #288, opened 2020-08-17
Expose a command to list custom endpoints in an account


williamspiro:
To help CLI developers when writing serverless functions, we could create a command that exposed configured custom endpoints in an account, with some helpful information about them, like what their URL is, secrets/env variables scoped to them and request method.

miketalley:
### Data
```
{
  route:"randomStarWarsCharacter",
  rawAssetPath:"/serverless/starwars.functions/characters.js",
  method:"GET",
  secretNames:[],
  created:1581347465625 (February 10, 2020, Mon, 10:11:05am UTC-05:00),
  updated:1581347465625 (February 10, 2020, Mon, 10:11:05am UTC-05:00),
},
{
  route:"events",
  rawAssetPath:"/serverless/empathy.functions/action1.js",
  method:"GET",
  secretNames:["secretName" ],
  created:1581347466054 (February 10, 2020, Mon, 10:11:06am UTC-05:00),
  updated:1602172350278 (October 8, 2020, Thu, 11:52:30am UTC-04:00),
},
{
  route:"events",
  rawAssetPath:"/serverless/empathy.functions/action1.js",
  method:"POST",
  secretNames:["secretName" ],
  created:1602172350253 (October 8, 2020, Thu, 11:52:30am UTC-04:00),
  updated:1602172350253 (October 8, 2020, Thu, 11:52:30am UTC-04:00),
},
{
  route:"events/update",
  rawAssetPath:"/serverless/empathy.functions/action1.js",
  method:"POST",
  secretNames:[
    "secretName",
    "googlekeyName",
    "otherkeyname"
  ],
  created:1581347469481 (February 10, 2020, Mon, 10:11:09am UTC-05:00),
  updated:1602172348372 (October 8, 2020, Thu, 11:52:28am UTC-04:00),
}
```

### Output
`hs serverless api`
```
GET /randomStarWarsCharacter (https://mtalley-101867970.hs-sitesqa.com/_hcms/api/randomStarWarsCharacter)
GET /events (https://mtalley-101867970.hs-sitesqa.com/_hcms/api/events)
POST /events
POST /events/update
```

`hs serverless api --detail`
```
/randomStarWarsCharacter
URL: https://mtalley-101867970.hs-sitesqa.com/_hcms/api/randomStarWarsCharacter
Method: "GET"
Secrets: ["otherkeyname", "secretName", "googlekeyName"]
[...]

PR #352, opened 2020-10-08
hs functions list command

miketalley:
## Description and Context
Adds `hs functions list` command that will output information about currently deployed functions.
Related to #288 

## Screenshots
`hs functions list`
image

`hs functions list --compact`
image

`hs functions list --json`
image

## TODO
- ~Determine if we want to include a URL for the function or not~ Removed URL to keep this MVP
  - ~If we do want to include this, we will need the `content` scope on the personalaccesskey to obtain the `domain`~
- Have `/routes` endpoint updated to allow external auth on production before release

### DIFF
--- a/packages/cms-cli/commands/functions.js
+++ b/packages/cms-cli/commands/functions.js
@@ -13,7 +13,12 @@ exports.builder = yargs => {
   addConfigOptions(yargs, true);
   addPortalOptions(yargs, true);
 
-  yargs.command(list).demandCommand(1, '');
+  yargs
+    .command({
+      ...list,
+      aliases: 'ls',
+    })
+    .demandCommand(1, '');
 
   return yargs;
 };
--- a/packages/cms-cli/commands/functions.js
+++ b/packages/cms-cli/commands/functions.js
@@ -1,0 +1,19 @@
+const {
+  addConfigOptions,
+  addPortalOptions,
+  addOverwriteOptions,
+} = require('../lib/commonOpts');
+const list = require('./functions/list');
+
+exports.command = 'functions';
+exports.describe = 'Commands for working with functions';
+
+exports.builder = yargs => {
+  addOverwriteOptions(yargs, true);
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+
+  yargs.command(list).demandCommand(1, '');
+
+  return yargs;
+};
--- a/packages/cms-cli/commands/functions/list.js
+++ b/packages/cms-cli/commands/functions/list.js
@@ -69,6 +69,7 @@ exports.builder = yargs => {
   yargs
     .options({
       compact: {
+        alias: 'c',
         describe: 'output compact data',
         type: 'boolean',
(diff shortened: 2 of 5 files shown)


packages/cms-cli/commands/functions/list.js
```diff
@@ -0,0 +1,81 @@
...
+  loadAndValidateOptions(options);
+
+  const { json, compact } = options;
+  const portalId = getPortalId(options);
+
+  trackCommandUsage('functions-list', { json, compact }, portalId);
+
+  logger.debug('Getting currently deployed functions');
+
+  const routesResp = await getRoutes(portalId).catch(async e => {
```
drewjenkins, line 50: hm I had no idea you could mix async/await and `.catch` from promises like that.

packages/cms-cli/commands/functions/list.js
```diff
@@ -0,0 +1,81 @@
...
+
+exports.builder = yargs => {
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+  addUseEnvironmentOptions(yargs, true);
+
+  yargs
+    .options({
+      compact: {
+        describe: 'output compact data',
```
drewjenkins, line 72: `-c` alias?

packages/cms-cli/commands/functions.js
```diff
@@ -0,0 +1,19 @@
...
+
+exports.command = 'functions';
+exports.describe = 'Commands for working with functions';
+
+exports.builder = yargs => {
+  addOverwriteOptions(yargs, true);
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+
+  yargs.command(list).demandCommand(1, '');
```
drewjenkins, line 16: `ls` alias may be cool

packages/cms-cli/commands/functions/list.js
```diff
@@ -0,0 +1,81 @@
...
+    process.exit();
+  });
+
+  logger.debug(`Retrieving logs for functionId: ${routesResp.id}`);
+
+  return outputFunctions(routesResp, options);
+};
+
+exports.builder = yargs => {
+  addConfigOptions(yargs, true);
```
drewjenkins, line 65: Not sure if it would work well or not, but it'd be sweet to be able to filter results like you can in bash.

E.g. if you type `ls he` and press `tab`, if a folder matches `/.*(he)+.*/gi` it would show that.  So a folder named `help` would display, but a folder named `foo` would not.

Could even be cool to have the `tab` command trigger the API call (and maybe cache results to avoid excessive calls), on which you can locally filter based on the user input from the positional

drewjenkins: The idea being to make this functionality work similarly to native terminal `ls`

packages/cms-cli/commands/functions/list.js
miketalley: This is already reserved for `--config`

PR #354, opened 2020-10-09
hs ls command to check contents of directories

miketalley:
## Description and Context
New command `hs ls [path]` allows listing directory contents by path. Default path is root.

## Screenshots
image

image

image

### Questions
Do we want to group folders/files and output folders first(perhaps with `/` prefix)?
Any other changes to the output(I made this as MVP as possible for now)?
Do we want to look at the BE endpoint to increase speed(it's a bit slow ATM)?

## Who to Notify
/cc @TheWebTech

### DIFF
--- a/packages/cms-cli/commands/ls.js
+++ b/packages/cms-cli/commands/ls.js
@@ -1,0 +1,81 @@
+const {
+  addPortalOptions,
+  addConfigOptions,
+  setLogLevel,
+  getPortalId,
+  addUseEnvironmentOptions,
+} = require('../lib/commonOpts');
+const { trackCommandUsage } = require('../lib/usageTracking');
+const { logDebugInfo } = require('../lib/debugInfo');
+const {
+  loadConfig,
+  validateConfig,
+  checkAndWarnGitInclusion,
+} = require('@hubspot/cms-lib');
+const { logger } = require('@hubspot/cms-lib/logger');
+const {
+  logServerlessFunctionApiErrorInstance,
+  ApiErrorContext,
+} = require('@hubspot/cms-lib/errorHandlers');
+const {
+  getDirectoryContentsByPath,
+} = require('@hubspot/cms-lib/api/fileMapper');
+const { validatePortal } = require('../lib/validation');
+
+const loadAndValidateOptions = async options => {
+  setLogLevel(options);
+  logDebugInfo(options);
+  const { config: configPath } = options;
+  loadConfig(configPath, options);
+  checkAndWarnGitInclusion();
+
+  if (!(validateConfig() && (await validatePortal(options)))) {
+    process.exit(1);
+  }
+};
+
+exports.command = 'ls [path]';
+exports.describe = 'get remote contents of a directory';
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { path } = options;
+  const directoryPath = path || '/';
+  const portalId = getPortalId(options);
+
+  trackCommandUsage('ls', {}, portalId);
+
+  logger.debug(`Getting contents of ${directoryPath}`);
+
+  const contentsResp = await getDirectoryContentsByPath(
+    portalId,
+    directoryPath
+  ).catch(async e => {
+    await logServerlessFunctionApiErrorInstance(
+      portalId,
+      e,
+      new ApiErrorContext({ portalId, directoryPath })
+    );
+    process.exit();
+  });
+
+  if (contentsResp.children.length) {
+    logger.log(contentsResp.children.sort().join('\n'));
+  } else {
+    logger.info(`No files found in ${directoryPath}`);
+  }
+};
+
+exports.builder = yargs => {
+  yargs.positional('path', {
+    describe: 'Remote directory to list contents',
+    type: 'string',
+  });
+
+  addConfigOptions(yargs, true);
+  addPortalOptions(yargs, true);
+  addUseEnvironmentOptions(yargs, true);
+
+  return yargs;
+};
--- a/packages/cms-cli/commands/ls.js
+++ b/packages/cms-cli/commands/ls.js
@@ -22,6 +22,8 @@ const {
 } = require('@hubspot/cms-lib/api/fileMapper');
 const { validatePortal } = require('../lib/validation');
 
+const FOLDER_DOT_EXTENSIONS = ['functions'];
+
 const loadAndValidateOptions = async options => {
   setLogLevel(options);
   logDebugInfo(options);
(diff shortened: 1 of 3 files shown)


packages/cms-cli/commands/ls.js
```diff
@@ -0,0 +1,81 @@
...
+    logger.info(`No files found in ${directoryPath}`);
+  }
+};
+
+exports.builder = yargs => {
+  yargs.positional('path', {
+    describe: 'Remote directory to list contents',
+    type: 'string',
+  });
+
```
drewjenkins, line 75: This is great! Could you add some example usages? Something like 

```
yargs.example([
    ['$0 ls'],
    ['$0 ls /'],
    ['$0 ls serverless']
  ]);
```

anthmatic:
Nice! Re: grouping, I do think something there would be beneficial. Maybe some way to differentiate between folders and files?

miketalley:
> Nice! Re: grouping, I do think something there would be beneficial. Maybe some way to differentiate between folders and files?

I was thinking of changing the output to look for a file extension and then assume that if one exists(minus special case of `.functions`) we can assume those are files. Any others would tack on a slash to denote a folder -- \`/${name}\`

packages/cms-cli/commands/ls.js
miketalley: Good idea. Done.

anthmatic:
2 things I noticed, neither of which are deal breakers:
- the `@hubspot` folder doesn't show up. I'm not sure if we NEED to support that, but it might be nice
- `/Coded files` and `/Custom` aren't in alphabetical order (which is a departure from DM)

miketalley:
> * the `@hubspot` folder doesn't show up. I'm not sure if we NEED to support that, but it might be nice
Good point -- I can check with the BE team to see if this can be added to the endpoint

> * `/Coded files` and `/Custom` aren't in alphabetical order (which is a departure from DM)
I'm not sure that I follow, they appear in the same order for me in the CLI and DM.
image
image

anthmatic:
Re: the order, this is what I see (with other folders that start with "c")
image
 
vs

 image

miketalley:
Fixed the incorrect sorting and pinned @hubspot and @marketplace folders to top of the list
image

miketalley:
@anthmatic I filed an issue for the filemapper endpoint to include the `@hubspot` folder and it should be pinned to the top when included in the results from the endpoint response.

packages/cms-cli/commands/ls.js
```diff
@@ -0,0 +1,131 @@
...
+      const splitName = fileOrFolder.split('.');
+
+      if (
+        splitName.length > 1 &&
+        FOLDER_DOT_EXTENSIONS.indexOf(splitName[1]) === -1
+      ) {
+        return fileOrFolder;
+      }
+
+      return `/${fileOrFolder}`;
```
gcorne, line 80: Is there a particular reason for prepending `/` to some items but not others? If we decide to keep this logic, one thing to keep in mind is that modules also have a postfix (e.g. `.module`.

packages/cms-cli/commands/ls.js
```diff
@@ -0,0 +1,131 @@
...
+  const portalId = getPortalId(options);
+
+  trackCommandUsage('ls', {}, portalId);
+
+  logger.debug(`Getting contents of ${directoryPath}`);
+
+  const contentsResp = await getDirectoryContentsByPath(
+    portalId,
+    directoryPath
+  ).catch(async e => {
```
gcorne, line 60: Best to use `try/catch` instead of mixing promises with async/await.

packages/cms-cli/commands/ls.js
```diff
@@ -0,0 +1,131 @@
+const {
```
gcorne, line 1: I wonder if we should call the command `list` and make `ls` an alias sort of like what we have done with `remove` and `rm`.

packages/cms-cli/commands/ls.js
```diff
@@ -0,0 +1,131 @@
...
+      }
+
+      return `/${fileOrFolder}`;
+    });
+    const hubspotFolder = `/${HUBSPOT_FOLDER}`;
+    const marketplaceFolder = `/${MARKETPLACE_FOLDER}`;
+
+    logger.log(
+      mappedContents
+        .sort(function(a, b) {
```
gcorne, line 87: I sort of wonder if we should use `String.localeCompare` to account for characters common to other languages.

packages/cms-cli/commands/ls.js
miketalley: This was done to make it clearer which items are folders(prepended with `/`) and which are files.

miketalley: Looks like it works properly with modules
image

gcorne: Isn't `mymodule.module` a folder?

miketalley: I was trying to mimic the display of the finder in DM where modules are listed more like files, but that is a good point -- it should be listed like a folder.

image

gcorne:
I wonder if we should borrow from `ls` and colorize folders instead of using `/`. I say that because I think prepending with `/` is confusing especially when listing folders deeper in the directory structure because the path to the file in HubSpot is different from what is presented in the output. The path separator also applies to both files and folders, so I don't think the purpose is obvious.

packages/cms-cli/commands/list.js
```diff
@@ -0,0 +1,124 @@
...
+      return chalk.reset.blue(fileOrFolder);
+    });
+    const hubspotFolder = `/${HUBSPOT_FOLDER}`;
+    const marketplaceFolder = `/${MARKETPLACE_FOLDER}`;
+
+    logger.log(
+      mappedContents
+        .sort(function(a, b) {
+          // Pin @hubspot folder to top
+          if (a === hubspotFolder) {
```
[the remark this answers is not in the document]
miketalley: Yes, it looks to be pinning these folders to the top.

Note: The BE is not currently returning the `@hubspot` folder(I've filed an issue for this in the filemapper repo).

I tested with the following code...
```
contentsResp = await getDirectoryContentsByPath(portalId, directoryPath);
  const randomIndex = Math.floor(
    Math.random() * contentsResp.children.length
  );
contentsResp.children.splice(randomIndex, 0, '@hubspot');
```

miketalley:
Updated output to use `chalk` to colorize folders instead of prepending `/`

image

packages/cms-cli/commands/list.js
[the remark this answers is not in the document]
gcorne: Were you testing with a root directory that had folders starting with numbers such as `1folder`?

PR #358, opened 2020-10-13
Fix circular dep check

anthmatic:


packages/cms-lib/http.js
```diff
@@ -9,7 +9,7 @@ const { logger } = require('./logger');
 const {
   FileSystemErrorContext,
   logFileSystemErrorInstance,
-} = require('./errorHandlers');
+} = require('./errorHandlers/fileSystemErrors');
```
anthmatic, line 12: @drewjenkins this seemed like the simplest fix for now. Does this seem reasonable?

gcorne (approved):

PR #362, opened 2020-10-16
Support loading the config relative to a specified directory

gcorne:
## Description and Context

VS Code is built using electron and electron always returns `/` as the CWD when `process.cwd()` is run. This will allow us to control where to start looking for the the config file.

See https://github.com/HubSpot/hubspot-cms-vscode/pull/83 and https://github.com/electron/electron/issues/2108


gcorne:
Going to go ahead with this.

PR #363, opened 2020-10-16
hs mv command

miketalley:
## Description and Context
Adds `hs mv <sourcePath> <destinationPath>` command which will move a file or folder from the `sourcePath` to the `destinationPath`.

#### Renaming
image

#### Moving files/renaming
image

#### Moving folder w/ contents
image

## Who to Notify
/cc @williamspiro

### DIFF
--- a/packages/cms-cli/commands/mv.js
+++ b/packages/cms-cli/commands/mv.js
@@ -20,6 +20,7 @@ const {
 const { logDebugInfo } = require('../lib/debugInfo');
 const { validatePortal } = require('../lib/validation');
 const { trackCommandUsage } = require('../lib/usageTracking');
+const { isPathFolder } = require('../lib/filesystem');
 
 const loadAndValidateOptions = async options => {
   setLogLevel(options);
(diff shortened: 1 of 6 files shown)


packages/cms-cli/commands/mv.js
```diff
@@ -0,0 +1,88 @@
...
+const getCorrectedDestPath = (srcPath, destPath) => {
+  if (!isPathFolder(srcPath)) {
+    return destPath;
+  }
+
+  // Makes sure that nested folders are moved independently
+  return `${destPath}/${srcPath.split('/').pop()}`;
+};
+
+exports.command = 'mv <srcPath> <destPath';
```
drewjenkins, line 46: Missing a `>`

packages/cms-cli/commands/mv.js
```diff
@@ -0,0 +1,88 @@
...
+  loadAndValidateOptions(options);
+
+  const { srcPath, destPath } = options;
+  const portalId = getPortalId(options);
+
+  trackCommandUsage('mv', {}, portalId);
+
+  try {
+    await moveFile(portalId, srcPath, getCorrectedDestPath(srcPath, destPath));
+    logger.log(`Moved "${srcPath}" to "${destPath}" in portal ${portalId}`);
```
drewjenkins, line 59: Not sure if we have an agreed on convention for this, but should we use `logger.success` for these types of messages?

miketalley: That is a good point, I will make the change.

PR #366, opened 2020-10-20
Improved output of hs functions list command by using table

miketalley:
## Description and Context
Improves output of `hs functions list` command(https://github.com/HubSpot/hubspot-cms-tools/pull/352) by utilizing `table` dependency.

- Creates `@hubspot/cms-cli/lib/table.js` with helper methods for using table output
- Refactors `hs open` command to use helpers
- Refactors `hs functions list` to use helpers
- Removes `--compact` option from `hs functions list` command as it is no longer much use

### Before
`hs functions list`
image

`hs functions list --compact`
image

### After
image

## Who to Notify
@abelb -- any feedback on this?
@williamspiro

### DIFF
--- a/packages/cms-cli/commands/functions/list.js
+++ b/packages/cms-cli/commands/functions/list.js
@@ -4,7 +4,7 @@ const {
   logApiErrorInstance,
   ApiErrorContext,
 } = require('@hubspot/cms-lib/errorHandlers');
-const { outputFunctions } = require('@hubspot/cms-lib/lib/functions');
+const { getFunctionArrays } = require('@hubspot/cms-lib/lib/functions');
 const {
   loadConfig,
   validateConfig,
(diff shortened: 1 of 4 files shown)


miketalley:
> Also, do we need the unformatted timestamp? To save on horizontal space maybe we can format it just as something like `YYYY-MM-DD@H:M:S`

Good call. I was thinking about this but wanted to get input from @abelb

miketalley:
> Nice man. Do you know if it'd be possible to construct the endpoint for the function? like the `https://{subdomain}.hs-sites.com/_hscms/api/{route}` ?

It is, but it would require the `content` scope group, so I am leaving this out for now -- see https://github.com/HubSpot/hubspot-cms-tools/issues/288#issuecomment-707850719

anthmatic:
> Nice man. Do you know if it'd be possible to construct the endpoint for the function? like the https://{subdomain}.hs-sites.com/_hscms/api/{route} ?

This does make me think that `hs functions exec` should be a thing

miketalley:
> > Nice man. Do you know if it'd be possible to construct the endpoint for the function? like the https://{subdomain}.hs-sites.com/_hscms/api/{route} ?
> 
> This does make me think that `hs functions exec` should be a thing

Yeah this will totally be doable for `GET`s once the `content` scope group is available on the personal access keys.

miketalley:
I'm going to merge this as the output is far superior to the previous output

PR #370, opened 2020-10-23
Updated method and endpoint for updateTable function

darkmavis1980:
## Description and Context

Updated function `updateTable` under the `api/hubdb.js` file to use the appropriate method and endpoint for the [hubdb API specifications for version 3](https://developers.hubspot.com/docs/api/cms/hubdb), more details can be found on the related [issue](https://github.com/HubSpot/hubspot-cms-tools/issues/367)

## TODO

Nothing

## Who to Notify
@gcorne @drewjenkins

### DIFF
--- a/packages/cms-lib/api/hubdb.js
+++ b/packages/cms-lib/api/hubdb.js
@@ -21,8 +21,8 @@ async function createTable(portalId, schema) {
 }
 
 async function updateTable(portalId, tableId, schema) {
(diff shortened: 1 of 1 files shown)


anthmatic:
Thanks for the fix @darkmavis1980. I just tested this vs `hubdb/api/v2/tables/((tableId))/` and it works as expected

PR #377, opened 2020-10-29
Add hs functions package command

miketalley:
## Description and Context
Adding `hs functions package <path>` command to run new build from package.json on specified path(expects .functions folder).

## Screenshots

output

## Who to Notify
@bkrainer @williamspiro


packages/cms-cli/commands/functions.js
```diff
@@ -18,6 +19,7 @@ exports.builder = yargs => {
       ...list,
       aliases: 'ls',
     })
+    .command(build)
```
anthmatic, line 22: I do wonder if there's a better command name for this. Perhaps `package`, `bundle`, or `install-deps`?

packages/cms-cli/commands/functions/build.js
```diff
@@ -0,0 +1,73 @@
...
+  const { path: functionPath } = options;
+  const portalId = getPortalId(options);
+  const spinner = makeSpinner(functionPath, portalId);
+
+  trackCommandUsage('functions-build', { functionPath }, portalId);
+
+  logger.debug(`Starting build for function(s) with path: ${functionPath}`);
+
+  spinner.start();
+  await buildPackage(portalId, `${functionPath}/package.json`);
```
drewjenkins, line 54: Do we need to handle the scenario where a package.json doesn't exist?  Since legacy functions won't have one

packages/cms-cli/commands/functions.js
drewjenkins: I like `package`.  Then we could have one that is `hs functions deploy` for the second step

packages/cms-cli/commands/functions/build.js
```diff
@@ -0,0 +1,73 @@
...
+  spinner.stop();
+  logger.success(
```
drewjenkins, line 56: I know I've asked this before, but is there any chance we can output a direct link to the invocation of the function?

packages/cms-cli/commands/functions/build.js
```diff
@@ -0,0 +1,73 @@
...
+    `Successfully built bundle from package.json for ${functionPath} on portal ${portalId}.`
+  );
+};
+
+exports.builder = yargs => {
+  yargs.positional('path', {
+    describe: 'Path to serverless function folder',
+    type: 'string',
+  });
+  yargs.example([['$0 functions build myFunctionFolder.functions']]);
```
drewjenkins, line 66: I don't think you need the inner array unless you want to add a description to the example, though that could be a good addition.

`Package/Build a deployable bundle for myFunctionFolder.functions`

packages/cms-cli/commands/functions/build.js
```diff
@@ -0,0 +1,73 @@
...
+  loadConfig(configPath, options);
+  checkAndWarnGitInclusion();
+
+  if (!(validateConfig() && (await validatePortal(options)))) {
+    process.exit(1);
+  }
+};
+
+exports.command = 'build <path>';
+exports.describe =
```
drewjenkins, line 39: I wonder if this needs more/different description.  I don't know how to succinctly say it for the CLI, but what I am wondering is, is someone going to understand what is happening here vs actually deploying the function?  Will they know this is a required step?  Once this action happens, the user doesn't have any insight into what just happened, what it really means, or why it is required.  

This may be a larger discussion, it just seems a little odd to me.  Not saying we should do this now, but maybe it would be helpful to add capabilities to know the status of the current deployable unit.

Like be able to see what version or identifier of the bundle is ready to be deployed.  Maybe also allow them to download a deployable bundle to make rollbacks/backups easy and smooth

packages/cms-cli/commands/functions/build.js
```diff
@@ -0,0 +1,73 @@
...
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { path: functionPath } = options;
+  const portalId = getPortalId(options);
+  const spinner = makeSpinner(functionPath, portalId);
+
+  trackCommandUsage('functions-build', { functionPath }, portalId);
+
+  logger.debug(`Starting build for function(s) with path: ${functionPath}`);
```
drewjenkins, line 51: Probably not part of this PR specifically, but I feel like any extra insight we can give to the user that is happening here would he helpful.  

e.g.
```
[spinner] Building myFunctionFolder.functions/function1
```

```
[✅] Building myFunctionFolder.functions/function1
[✅] Building myFunctionFolder.functions/function2
[spinner] Building myFunctionFolder.functions/function3
```

packages/cms-cli/commands/functions/build.js
drewjenkins: Also just an idea, but since a CLI user may not know off of the top of their head the remote path to the function they want to build, making an interactive selector would be sweet.  It could leverage the `hs functions list` to get the remote paths

williamspiro:
Excited to see this being worked on!

I am having a hard time wrapping my head around the greater workflow/concepts we are introducing here, and how they impact the existing workflows. Some questions that come to mind:

1. How does this interface with upload/watch? 
2. What does this mean for https://developers.hubspot.com/docs/cms/guides/creating-an-efficient-development-workflow and https://developers.hubspot.com/docs/cms/guides/github-integration? 
3. What does this mean for the Design Manager? Is there a build scenario in the Design Manager? 
4. How does this interface with customers who do not have package.json files in a functions folder

@miketalley are you still planning on writing up a doc with the broader thinking for how we approach this? I think that would make sure the workflow and experience is high quality throughout the various commands/workflows this will impact, and make sure @abelb could weigh in from a UX perspective. 

I also think an interesting angle could be to start with documentation - how would we explain this concept and how it interfaces with existing concepts in writing? Does it make sense to developers? Is it intuitive? Does it sound reasonable to ask developers to do? 

I think fleshing out some of these questions will ensure a well thought out experience and workflow that makes developers love serverless functions ❤️

packages/cms-cli/commands/functions/build.js
miketalley: > Do we need to handle the scenario where a package.json doesn't exist? Since legacy functions won't have one

Added.

> It could leverage the hs functions list to get the remote paths

Love this idea! The endpoint does not currently support passing an endpoint ATM, but we should add that capability so we can then use the list to create a interactive selector @bkrainer

packages/cms-cli/commands/functions/build.js
miketalley: Not until we have the `content` scope group available to the personal-access-key -- I think Allison is working on this(can't find her public github).

packages/cms-cli/commands/functions/build.js
miketalley: It looks like it has to be an array of arrays. It breaks if I remove the nested array:
image

packages/cms-cli/commands/functions/build.js
miketalley: I totally agree, and there will likely be more steps in the future to complete the deploy process(which we can discuss streamlining). This is only part of an alpha release so we can get more feedback on how the preferred solution should function.

packages/cms-cli/commands/functions/build.js
miketalley: Yeah that would be helpful, this is not currently polling, the response is just delayed until the build completes. Something we could def add to vNext @bkrainer

packages/cms-cli/commands/functions/build.js
drewjenkins: Hm ok my bad.  :) Still do we should add a description?

packages/cms-cli/commands/functions/build.js
bkrainer: Agreed. I see this moving towards a fully asynchronous build process, where we poll based on a build ID. Could maybe even stream the `npm install` output so we could show exactly how far along the build process is

miketalley:
@williamspiro great insight and questions!

I have created a google doc with a lot of my ideas on these and shared it with the team in slack. This PR was intended to be a MVP for the CLI to enable testing to begin and feedback to be gathered. There are a lot of ideas in the google doc that we can continue to refine to create a great developer experience when it comes to building and deploying functions.

packages/cms-cli/commands/functions.js
miketalley: Renamed it to `hs functions package`

packages/cms-cli/commands/functions/build.js
kemmerle: I can get back on that PR, if it's blocking you. Completely slipped my mind!

packages/cms-cli/commands/functions/build.js
miketalley: Ah gotcha, added.

packages/cms-cli/commands/functions/build.js
miketalley: I think it would be a good improvement. Once the PAK has `content` we can retrieve the website `origin` to generate a link that executes the individual functions(`GET`s of course). We have the `/_hcms/api/endpointName` part accessible already through the response that `hs list` gets, and with the `origin` we will have a full URL.

PR #379, opened 2020-11-03
Fix log tailing

### LINKED ISSUE #378: hs logs --follow not working as docuemnted
## Description and Context
With version 2.2.0 of the CLI, `hs logs --follow` is no longer working as documented. When you run this command, old logs are getting printed to the console (not sure if this change was intended or not) and new executions are not showing up.

Possibly related to work in https://github.com/HubSpot/hubspot-cms-tools/pull/323

### Steps to reproduce
Run `hs logs --follow` for an endpoint on version 2.2.0. Notice old logs get printed to the console and new executions do not show up.

## Expected behavior
Previously, this command did not print any old logs to the console, and showed new executions of the endpoint after the command has been run. 

## Screenshots
logs

## Who to Notify
@HubSpot/cms-developer-experience

anthmatic:
## Description and Context
Fixes #378 
There were two issues preventing logs from displaying properly:
1. The first call to get the latest log result was improperly returning the first result (API has been fixed)
2. The paging `after` token was not getting passed while polling for results, which caused the first 5 results to return each time.

### DIFF
--- a/packages/cms-lib/api/results.js
+++ b/packages/cms-lib/api/results.js
@@ -3,12 +3,13 @@ const http = require('../http');
 const RESULTS_API_PATH = 'cms/v3/functions/results';
 
 async function getFunctionLogs(portalId, functionId, query = {}) {
-  const { limit = 5 } = query;
+  const { limit = 5, after } = query;
 
   return http.get(portalId, {
     uri: `${RESULTS_API_PATH}/${functionId}`,
     query: {
       limit,
+      after,
     },
   });
 }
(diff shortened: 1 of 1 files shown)


packages/cms-lib/api/results.js
```diff
@@ -3,12 +3,13 @@ const http = require('../http');
 const RESULTS_API_PATH = 'cms/v3/functions/results';
 
 async function getFunctionLogs(portalId, functionId, query = {}) {
-  const { limit = 5 } = query;
+  const { limit = 5, after } = query;
 
   return http.get(portalId, {
     uri: `${RESULTS_API_PATH}/${functionId}`,
     query: {
```
gcorne, line 10: I'd expect `query` to support all the options, which makes me wonder if it should be passed through.

drewjenkins (approved):

PR #383, opened 2020-11-09
Show destination account when uploading via webpack plugin

### LINKED ISSUE #350: Make it clear which account (portal) is the destination of uploads
When using the webpack `HubSpotAutoUploadPlugin`, it is not clear what the destination is for the uploads.

anthmatic:
## Description and Context
Fixes #350 

Prints the destination `accountId` each time a file is uploaded.

## Screenshots

image

## TODO
An alternative approach is that we could log this when the command is first run, but I found that this info kind of gets lost pretty quickly if we do that. 
image

### DIFF
--- a/packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
+++ b/packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
@@ -47,7 +47,9 @@ class HubSpotAutoUploadPlugin {
           const dest = `${this.dest}/${filename}`;
           upload(this.accountId, asset.existsAt, dest)
             .then(() => {
-              webpackLogger.info(`Uploaded ${dest} to ${this.accountId}`);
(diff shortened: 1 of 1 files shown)


packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
```diff
@@ -47,7 +47,7 @@ class HubSpotAutoUploadPlugin {
           const dest = `${this.dest}/${filename}`;
           upload(this.accountId, asset.existsAt, dest)
             .then(() => {
-              webpackLogger.info(`Uploaded ${dest}`);
+              webpackLogger.info(`Uploaded ${dest} to ${this.accountId}`);
```
gcorne, line 50: Can we prefix with "account" so that it is clearer that the number is a portal id?

gcorne (approved):

PR #384, opened 2020-11-09
Fix for #381 - downloadHubDbTable would download incorrect results for large tables

### LINKED ISSUE #381: Hubdb function `downloadHubDbTable` incorrectly download table rows
## Description and Context

I just found out an issue with the `downloadHubDbTable` in the `packages/cms-lib/hubdb.js` file in the cms-lib package.
So essentially while in the API `v2` implementation we used to fetch table rows using limit and offset parameters to fetch large tables,  now in the new version `v3` you must pass the limit and after params, where the after is a pointer to the next set of records and not just a number as the offset.
This cause an issue with the `downloadHubDbTable`  where if you have 1001 records, the downloaded table will have 2000 records as it downloads the first 1000 twice.

I'm using the latest version of the cms-cli `2.1.1-beta.10`

### Steps to reproduce

Use the downloadHubDbTable method to download a table with more than 1000 rows, once downloaded open the file and you should have the first 1000 rows repeated as per many times the total rows of the table divided by 1000 (so if you have 2020 rows, you will get 3000 rows).

## Expected behavior

It should download all the rows on a file.

## Who to Notify

@drewjenkins @anthmatic

drewjenkins:
## Description and Context
HubDB rows api changed from v2 to v3, updated the way we call it to reflect this

Resolves #381

@darkmavis1980

### DIFF
--- a/packages/cms-lib/hubdb.js
+++ b/packages/cms-lib/hubdb.js
@@ -135,20 +135,22 @@ async function downloadHubDbTable(accountId, tableId, dest) {
     validateJsonPath(dest);
   }
 
-  let totalRows = null;
   let rows = [];
-  let count = 0;
-  let offset = 0;
-  while (totalRows === null || count < totalRows) {
-    const response = await fetchRows(accountId, tableId, { offset });
-    if (totalRows === null) {
-      totalRows = response.total;
-    }
+  let after = null;
+  do {
+    const { paging, results } = await fetchRows(
+      accountId,
+      tableId,
+      after ? { after } : null
+    );
 
-    count += response.results.length;
-    offset += response.results.length;
-    rows = rows.concat(response.results);
-  }
+    rows = rows.concat(results);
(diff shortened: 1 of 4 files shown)


anthmatic:
I think this also affects https://github.com/HubSpot/hubspot-cms-tools/blob/26088f3b254c014618ca28f821eb8230fc325de7/packages/cms-lib/hubdb.js#L171

darkmavis1980 (approved):

packages/cms-lib/hubdb.js
```diff
@@ -124,6 +124,23 @@ function convertToJSON(table, rows) {
   };
 }
 
+async function _fetchRows(accountId, tableId) {
```
anthmatic, line 127: Could this be something like `fetchAllRows`?

anthmatic (approved):
One note about function name, otherwise looks slick 🚀

PR #387, opened 2020-11-12
Fix filemanger fetch error

anthmatic:
## Description and Context
While tracking down another bug, I found that the filemanger fetch command was failing when fetching folders due to a change in #339 that moved the logic for creating writeStreams and it looks like I missed updating this command as well.

### DIFF
--- a/packages/cms-lib/fileManager.js
+++ b/packages/cms-lib/fileManager.js
@@ -104,26 +104,6 @@ async function downloadFile(accountId, file, dest, options) {
     return;
   }
 
-  const logFsError = err => {
-    logFileSystemErrorInstance(
-      err,
-      new FileSystemErrorContext({
-        destPath,
-        accountId,
-        write: true,
-      })
-    );
-  };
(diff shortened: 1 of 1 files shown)


packages/cms-lib/fileManager.js
```diff
@@ -174,6 +154,19 @@ async function fetchAllPagedFiles(accountId, folderId, { includeArchived }) {
  * @param {object} options
  */
 async function fetchFolderContents(accountId, folder, dest, options) {
+  try {
+    await fs.ensureDir(dest);
+    logger.log('Wrote folder "%s"', dest);
```
gcorne, line 159: Minor, but I think that this assumes that the folder needs to be created.

drewjenkins (approved):

packages/cms-lib/fileManager.js
anthmatic: Good point, I just removed for now since I directory creation is implied

PR #389, opened 2020-11-17
Local testing for functions using hs functions test command

miketalley:
## Description and Context
Allows local testing of serverless functions using `hs functions test <localDotFunctionsFolderPath>` command. This runs an express server on port `5432` (or can be specified with `--port=<port>`) that will host the endpoints configured within `serverless.json` of the specified .functions folder. By default the specified .functions folder is watched for changes, which will restart the server.

The endpoints can be hit(`localhost:5432/_hcms/api/myEndpoint`) to execute the function and the context passed into the function will mimic the data shape/properties that are used in production. The output from the function will be logged to the console. CORS has been disabled on the local server, which allows pointing to the local endpoint from a DesignManger module to test the functionality.

This library can also be used for an external project. For example...
```
const { start } = require('@hubspot/serverless-dev-runtime');

start({
  accountId: 123456,
  port: 5678,
  path: './tmp/localTesting.functions'
});
```

The `limits` and `contact` property values are using mocked data, which can be modified using `.env` file variables.

#### Notes
- A `.env` file can be included in the local .functions folder to mimic secrets
- `environment` config object can be added at both the root level and at the function level in `serverless.json`(function level overwrites root)
- `headers`, `params`, `body`(from `POST`s), `secrets`, `accountId`, `limits`(mocked), and `contact`(mocked) get passed into the serverless execution context
- `console.log`s within the function will be caught and added to output
- Mocked data can be customized by setting the values of the following environment variables within a `.env` file
  - HUBSPOT_LIMITS_TIME_REMAINING
  - HUBSPOT_LIMITS_EXECUTIONS_REMAINING
  - HUBSPOT_CONTACT_VID
  - HUBSPOT_CONTACT_IS_LOGGED_IN
  - HUBSPOT_CONTACT_LIST_MEMBERSHIPS
- Mocking no contact data can be done using the `--contact=false` option when running the command
- Watch can be turned off(it's on by default) by using the `--watch=false` option
- Warnings will be displayed when hitting limits
  - secrets > 50
  - runTime > 3s

image

image

## TODO
- Think about creating a BE endpoint to serve default `package.json` content so we don't keep a copy in this repo

## Who to Notify
@williamspiro @gcorne @bkrainer


packages/cms-cli/package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "@hubspot/cms-lib": "^2.2.1-beta.3",
     "chalk": "^4.1.0",
+    "dotenv": "^8.2.0",
```
bkrainer, line 13: do we need to do anything to prevent `.env` files from being uploaded?

packages/cms-cli/commands/functions/test.js
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  await installDeps(functionPath);
+  handleExit(() => {
+    cleanupArtifacts(functionPath);
+  });
+
+  const app = express();
+  routes.forEach(route => {
+    const { method, file } = endpoints[route];
```
bkrainer, line 96: we can also have environment variables configured for a specific route (which could also override the top-level environment variables)

packages/cms-cli/package.json
miketalley: hmm good point 🤔

miketalley: Added `.env` to the ignoreRules.

packages/cms-cli/commands/functions/test.js
miketalley: Local environment config will now be loaded and will overwrite global environment config.

miketalley:
This is in a pretty decent state to start testing things if anyone wants. Still need to look at passing `Cookie` header and `contact` values(fake them maybe?). Also some things to work on...
- proxying
- moving this to a new lerna package

miketalley:
- Added `cors` to allow modules to successfully access a locally running test function
- Added environment variables that can be used to modify the mocked data that is passed to the test function
  - HUBSPOT_LIMITS_TIME_REMAINING
  - HUBSPOT_LIMITS_EXECUTIONS_REMAINING
  - HUBSPOT_CONTACT_VID
  - HUBSPOT_CONTACT_IS_LOGGED_IN
  - HUBSPOT_CONTACT_LIST_MEMBERSHIPS

packages/cms-cli/jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  testPathIgnorePatterns: ['commands/functions/test.js'],
```
drewjenkins, line 2: Not a big deal but what is this for?  Was it breaking jest?

packages/serverless-dev-runtime/README.md
```diff
@@ -0,0 +1,67 @@
+# @hubspot/serverless-dev-runtime
```
drewjenkins, line 1: Nice! I love how you split out the server code into a separate package

packages/serverless-dev-runtime/lib/files.js
```diff
@@ -0,0 +1,68 @@
+const os = require('os');
+const { spawn } = require('child_process');
+const fs = require('fs-extra');
+const tmp = require('tmp');
+const { logger } = require('@hubspot/cms-lib/logger');
+const defaultFunctionPackageJson = require('./templates/default-function-package.json');
+
+const installDeps = folderPath => {
+  const npmCmd = os.platform().startsWith('win') ? 'npm.cmd' : 'npm';
```
drewjenkins, line 9: Nice, good call

packages/cms-cli/jest.config.js
miketalley: Jest automatically looks at `test.js` files as tests and tries to run them, so had to add this to prevent that.

packages/serverless-dev-runtime/lib/files.js
miketalley: TY stackoverflow

miketalley:
> One thing that would be nice would be to allow for specifying the "name" of the function, rather than the full `name.functions`. Not a huge deal, but it did trip me up cuz my flow was
> 
> `hs create function` - named it `test-function`
> `hs functions test test-function`

Good call, I will add a check for both before displaying an error

miketalley:
> I put some invalid code in my `.functions/index.js` file, and I get this error when running it
> 
> ```
> [ERROR] Unable to process log {"executionTime":null,"log":"","duration":"NaN ms","status":"UNHANDLED_ERROR","createdAt":1608660747691,"memory":"74/128 MB","id":-1,"payload":1608660747676}
> ```
> 
> Is there any way we can capture the stack/give more info about the error?

Taking a look. Thanks for sharing.

miketalley:
> There is a difference in what is returned in QA vs what is returned locally.
> 
> QA
> 
> ```json
> {"response":{"total":0,"offset":0,"limit":10,"results":[],"cacheKey":"103705330::BLOG_POST:SITE_PAGE:LISTING_PAGE:10:::1:searchterm:0:LONG:0:1.0:5.0:0.0:0::::0:0:/54.166.131.109:[]::0:UNKNOWN::::0:0","searchTerm":"searchterm","page":0}}
> ```
> 
> Locally
> 
> ```json
> {"body":{"response":{"total":0,"offset":0,"limit":10,"results":[],"cacheKey":"103705330::BLOG_POST:SITE_PAGE:LISTING_PAGE:10:::1:searchterm:0:LONG:0:1.0:5.0:0.0:0::::0:0:/73.61.234.183:[]::0:UNKNOWN::::0:0","searchTerm":"searchterm","page":0}},"statusCode":200}
> ```
> 
> Local has the extra `body`, and also has the `statusCode`. I believe those are the only two differences

This is fixed, needed to send the `body` prop instead of the root in the data.

miketalley:
> One thing I notice is even if I throw an error, the status-code on the call is always a 200. Can we persist the status code?
> 
> Local
> image
> 
> QA
> image

Fixed

PR #392, opened 2020-11-20
Fix bug when fetching filemanger to `./`

anthmatic:
## Description and Context
A developer in the HS Slack community discovered an issue where you could not download all filemanager files to the current working directory (`./`). When constructing a local path, if the folder name argument wasn't passed, it failed.

### DIFF
--- a/packages/cms-lib/api/fileManager.js
+++ b/packages/cms-lib/api/fileManager.js
@@ -37,7 +37,7 @@ async function fetchFiles(accountId, folderId, { offset, archived }) {
     qs: {
       hidden: 0,
       offset: offset,
-      folder_id: folderId || 'None',
+      folder_id: folderId,
       ...(!archived && { archived: 0 }),
     },
   });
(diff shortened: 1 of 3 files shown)


gcorne:
This isn't working for me. It looks to be trying to fetch the files to `/` instead of resolving to the CWD. It also looks like we're missing some error handling somewhere.

image

anthmatic:
@gcorne I fixed the error handling, but I'm a little unsure what the desired behavior should be when fetching TO `/`.  I would think if I wanted to fetch to the CWD, I would run `hs filemanager fetch <remote> .` I'm not aware of any places in the CLI where we treat `/` as anything other than the root directory.

gcorne:
@anthmatic the above example has a destination of `./` which is an alias for `.`. I think something is up with how we're resolving paths if it is resolving to `/` but didn't dig into it.

anthmatic:
@gcorne ah, that's my mistake, I missed the `.`. I also misunderstood how `path.resolve()` works, but that should be fixed up now.

gcorne (approved):

PR #394, opened 2020-11-23
Delete portalId/defaultPortal if account is present

drewjenkins:
## Description and Context
Remove portalId/defaultPortal when we update the configs

### DIFF
--- a/packages/cms-lib/lib/__tests__/config.js
+++ b/packages/cms-lib/lib/__tests__/config.js
@@ -808,7 +808,6 @@ describe('lib/config', () => {
               },
               "authType": "oauth2",
               "name": "OAUTH2",
-              "portalId": 2223,
             }
           `);
         });
@@ -942,12 +941,9 @@ describe('lib/config', () => {
           setConfig(CONFIG);
         });
 
-        it('supports both defaultPortal and defaultAccount', () => {
+        it('supports defaultAccount', () => {
           updateDefaultAccount(LEGACY_ACCOUNTS[0].portalId);
           expect(getConfig().defaultAccount).toEqual(
-            LEGACY_ACCOUNTS[0].portalId
-          );
-          expect(getConfig().defaultPortal).toEqual(
             LEGACY_ACCOUNTS[0].portalId
           );
         });
--- a/packages/cms-lib/lib/__tests__/config.js
+++ b/packages/cms-lib/lib/__tests__/config.js
@@ -780,38 +780,6 @@ describe('lib/config', () => {
           expect(getAccountConfig(2222)).toMatchInlineSnapshot(`undefined`);
         });
       });
-
-      describe('combined', () => {
-        beforeEach(() => {
-          const CONFIG = {
-            portals: COMBINED_ACCOUNTS,
-          };
-
-          setConfig(CONFIG);
-        });
-
-        it('favors account over portal', () => {
-          expect(getAccountConfig(2222)).toMatchInlineSnapshot(`
-            Object {
-              "accountId": 2222,
-              "auth": Object {
-                "clientId": "fakeClientId",
-                "clientSecret": "fakeClientSecret",
-                "scopes": Array [
-                  "content",
-                ],
-                "tokenInfo": Object {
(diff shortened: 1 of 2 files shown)


packages/cms-lib/lib/config.js
```diff
@@ -546,12 +556,10 @@ const updateDefaultAccount = defaultAccount => {
...
   const config = getAndLoadConfigIfNeeded();
-  if (config.defaultAccount) {
-    config.defaultAccount = defaultAccount;
-  }
-  // Keep for backcompat
+  config.defaultAccount = defaultAccount;
+
   if (config.defaultPortal) {
-    config.defaultPortal = defaultAccount;
+    delete config.defaultPortal;
```
anthmatic, line 562: Since we're deleting, should we tell the user things are changing?

drewjenkins: Oh yeah good call, will do

packages/cms-lib/lib/config.js
```diff
@@ -417,11 +418,19 @@ const getEnv = nameOrId => {
...
-  getConfigAccounts(getAndLoadConfigIfNeeded()).find(
+const getAccountConfig = accountId => {
+  const account = getConfigAccounts(getAndLoadConfigIfNeeded()).find(
     account => account.accountId === accountId || account.portalId === accountId
   );
 
+  if (account && account.portalId) {
+    account.accountId = account.accountId || account.portalId;
+    delete account.portalId;
+  }
```
gcorne, line 429: Hmm...I was thinking that we would only do this when a user is going through a flow that is explicitly a write operation. If I understand how this is working, it will rewrite the config when going through the write process when exchanging an access key for an access token which happens transparently in the background when the access token is about to expire.

packages/cms-lib/lib/config.js
gcorne: Where are we copying `defaultPortal` to `defaultAccount`?

gcorne (changes requested):

packages/cms-lib/lib/config.js
drewjenkins: Yeah I was thinking about this too.  I moved the delete to the updateAccountConfig function

packages/cms-lib/lib/config.js
drewjenkins: We aren't copying it over, but this only runs when a new value is being passed in, so a couple lines above we are doing `config.defaultAccount = defaultAccount;` which takes in the arg passed to the function

packages/cms-lib/lib/config.js
gcorne: Can we move all side effects to the update? I think it makes more sense to keep this function pure.

packages/cms-lib/lib/config.js
gcorne: Got it. I didn't read the change with enough context visible.

packages/cms-lib/lib/__tests__/config.js
```diff
@@ -942,14 +909,11 @@ describe('lib/config', () => {
           setConfig(CONFIG);
         });
 
-        it('supports both defaultPortal and defaultAccount', () => {
+        it('supports defaultAccount', () => {
```
gcorne, line 912: Can we add a test to cover backwards compatibility?

drewjenkins: I have one [here](https://github.com/HubSpot/hubspot-cms-tools/blob/8dd0aa1a45bddcab06b709655f7dd38cb6a9a56f/packages/cms-lib/lib/__tests__/config.js#L891), this was just removing the test case around supporting both at the same time

packages/cms-lib/lib/config.js
drewjenkins: My concern of moving the `account.accountId = account.accountId || account.portalId` line would be that a function calls into `getAccountConfig`, making the assumption that `accountId` will be present, when in reality the config file hadn't been updated yet

PR #396, opened 2020-12-04
Fix ignore

drewjenkins:
## Description and Context
When looking at what to ignore, we were only checking CWD and descendants of this.  Now we start at root, to catch the case where a user can add items outside of the working dir


packages/cms-lib/http.js
```diff
@@ -96,7 +96,8 @@ const getRequest = async (accountId, options) => {
 };
 
 const postRequest = async (accountId, options) => {
-  return requestPN.post(await withAuth(accountId, options));
+  const data = await withAuth(accountId, options);
+  return requestPN.post(data);
```
anthmatic, line 100: Curious why this is broken out?

packages/cms-lib/ignoreRules.js
```diff
@@ -43,16 +43,17 @@ function loadIgnoreConfig() {
   loaded = true;
 }
 
-function shouldIgnoreFile(file, cwd) {
+function shouldIgnoreFile(file) {
   loadIgnoreConfig();
-  const relativeTo = configPath || cwd;
+  const relativeTo = configPath || '/';
```
anthmatic, line 48: minor, but the name `configPath` seemed to cause a lot of confusion

prettier.config.js
```diff
@@ -7,4 +7,5 @@ module.exports = {
   tabWidth: 2,
   trailingComma: 'es5',
   useTabs: false,
+  arrowParens: 'avoid',
```
miketalley, line 10: 👍

miketalley (commented):
LGTM other than @anthmatic 's comments

packages/cms-lib/http.js
drewjenkins: Cause I was logging and forgot to change it back, thanks for catching it :)

PR #397, opened 2020-12-04
Revert config migration

drewjenkins:
## Description and Context
We decided to revert migrating portal->account in the hubspot.config.yml.  This is because we cannot account for the following scenario

1. User has a hubspot.config.yml at a top level directory, `~/`
2. Project 1 has the latest version of the CLI with the migrated changes.  It works fine
3. Project 2 has an old version of the CLI. It is still expecting portal, not account.  It will fail because Project 1 already migrated portal->account


packages/cms-cli/lib/commonOpts.js
```diff
@@ -85,10 +85,6 @@ const getAccountId = (options = {}) => {
     return parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
   }
 
-  if (options.useEnv && process.env.HUBSPOT_ACCOUNT_ID) {
```
drewjenkins, line 88: Wasn't sure if we wanted to keep this.  I took it out just for consistency

anthmatic:
General question: should we still support reading `accountId` in config files?

packages/cms-cli/lib/commonOpts.js
miketalley: Looks like we should be good to remove this because `process.env.HUBSPOT_PORTAL_ID` is still supported

miketalley (changes requested):
The CLI readme needs to be updated https://github.com/HubSpot/hubspot-cms-tools/blob/master/packages/cms-cli/README.md

Also, we need to update the `hs upload` command to use `portalId` instead of `accountId` in https://github.com/HubSpot/hubspot-cms-tools/blob/master/CONTRIBUTING.md

Also, should we change back `multiple accounts` to `multiple portals` in https://github.com/HubSpot/hubspot-cms-tools/blob/master/docs/TechnicalDesign.md

May want to check out this PR for other potential changes that we need to change back https://github.com/HubSpot/hubspot-cms-tools/commit/fd031f0bbb24e43b1472fb8db478418ba0d0bf90#

TheWebTech:
Just making sure everyone related to this update is aware.

since this change was reverted, the website changelog announcement does not feel necessary. All of the other changes going out are either bug fixes, or for alpha functionality. There's no blocker from DevRel on the remainder of the updates  going for the CLI.

gcorne:
> since this change was reverted, the website changelog announcement does not feel necessary. All of the other changes going out are either bug fixes, or for alpha functionality. There's no blocker from DevRel on the remainder of the updates going for the CLI.

Correct me if this is not true, but I think the other info is still needed. The revert only impacts the config file.

TheWebTech:
@gcorne I think there was miscommunication. It was my original understanding we were just reverting the config file change but it appeared based on the comments and some 1:1 conversations that we reverted basically everything to do with `--account` `accountId` and `account`.



>Also, we need to update the hs upload command to use portalId instead of accountId in https://github.com/HubSpot/hubspot-cms-tools/blob/master/CONTRIBUTING.md
> 
> Also, should we change back multiple accounts to multiple portals in https://github.com/HubSpot/hubspot-cms-tools/blob/master/docs/TechnicalDesign.md
> 
> May want to check out this PR for other potential changes that we need to change back 

> Talking with mike over slack about his comments. Decided to change the --account flag to default to --portal instead, since it is directly referencing the hubspot.config.yml file. All other wordings are staying "account"

I somehow missed that "all other wordings are staying 'account'" bit.
I also directly asked Mike about this but I realize looking back the way I framed the question accidentally caused him to give an answer that further confused it. (that's completely on me)

So I think this is on me. If this is true that  `--account` and most of the account references are still working but `portal` is still used in the config file - do we think that the experience is more confusing as a result?

gcorne:
> So I think this is on me. If this is true that --account and most of the account references are still working but portal is still used in the config file - do we think that the experience is more confusing as a result?

It is more confusing than it is today, but my assumption is that most if not all developers are using `hs init` and `hs auth` to modify the config instead of manually editing the config file.

PR #403, opened 2020-12-14
Added delete option to hs functions deploy command to support BE in alpha release

miketalley:
## Description and Context
Adds a `--delete` option to the `hs functions deploy <remoteDotFunctionsFolderPath>` command to manually delete the current build associated with a functions folder.

## Screenshots
image
image

## Who to Notify
@bkrainer

### DIFF
--- a/packages/cms-cli/commands/functions/deploy.js
+++ b/packages/cms-cli/commands/functions/deploy.js
@@ -18,12 +18,15 @@ const {
   ApiErrorContext,
 } = require('@hubspot/cms-lib/errorHandlers');
 const { logger } = require('@hubspot/cms-lib/logger');
-const { buildPackage } = require('@hubspot/cms-lib/api/functions');
+const {
+  buildPackage,
+  deletePackage,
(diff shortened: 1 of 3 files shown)


gcorne:
Is the idea that this would un-deploy the endpoints? Does it revert to the previous build? It seems to me that `delete` would be an action taken on a build not a deploy and that `revert` would be a separate action if we want to provide that functionality.

miketalley:
> Is the idea that this would un-deploy the endpoints? Does it revert to the previous build? It seems to me that `delete` would be an action taken on a build not a deploy and that `revert` would be a separate action if we want to provide that functionality.

The endpoints remain, but the associated build is removed. Maybe @bkrainer has more specifics on this?

The reason I am adding this is to support the alpha version of `hs functions deploy`. In the current state, there is a limit of 3 `package.json` built functions per portal. If I try to build a 4th, I will get an error. However, there is not a way to specify that one of the builds should be removed to make space for a different build to occur.

It's unlikely that we will run into this, but I wanted to make sure that we have a way to manually remove the build.

PR #407, opened 2020-12-21
add alwaysStat flag to ensure we have fs.Stats available.  Without it…

drewjenkins:
adding this flag provides fs.Stats is passed in since it is needed in some scenarios.  Without this, the watch command would hang on uploading for some people (myself included)

### DIFF
--- a/packages/cms-lib/api/fileMapper.js
+++ b/packages/cms-lib/api/fileMapper.js
@@ -1,4 +1,5 @@
-const fs = require('fs-extra');
+// We are purposely using fs instaed of fs-extra here. See https://github.com/HubSpot/hubspot-cms-tools/pull/407
+const fs = require('fs');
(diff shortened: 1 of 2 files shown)


gcorne:
Is there some info on why it is happening and how this change fixes the issue?

anthmatic:
Something changed between node v12 and node v15 which causes the watch command to fail. I've been digging it to see what changed, but haven't come up with anything yet

gcorne:
@drewjenkins is the issue with the `fs.createReadStream()` call in `api/FileMapper#upload` which is coming from `fs-extra`? Is it fixed in a more recent version of `fs-extra`? Could we switch to the builtin `fs`? It does look like there have been some changes to read/write streams in v15+ such as https://github.com/nodejs/node/commit/54b36e401d.

PR #411, opened 2021-01-05
Updated endpoint used for build/deploy which supports polling and added polling

miketalley:
## Description and Context
Changed custom serverless build endpoint from `cms/v3/functions/package` to `cms/v3/functions/build` which returns a `buildId`. This is used to poll `cms/v3/functions/build/((id))/poll` for build status.

This update does not change anything visibly or from a flow standpoint, it is done to allow polling to be done instead of having the BE endpoint only respond when the build has completed.

The `--delete` option has also been removed from the `hs functions deploy` command as the backend handles properly deleting associated built packages when serverless function folders are deleted.

## Who to Notify
@bkrainer @gcorne @williamspiro

### DIFF
--- a/packages/cli/commands/functions/deploy.js
+++ b/packages/cli/commands/functions/deploy.js
@@ -21,7 +21,7 @@ const { logger } = require('@hubspot/cli-lib/logger');
 const {
   buildPackage,
   deletePackage,
-  pollBuild,
+  getBuildStatus,
 } = require('@hubspot/cli-lib/api/functions');
 const { validateAccount } = require('../../lib/validation');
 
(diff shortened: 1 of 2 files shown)


packages/cli/commands/functions/deploy.js
```diff
@@ -30,6 +30,23 @@ const makeSpinner = (actionText, functionPath, accountIdentifier) => {
   );
 };
 
+const pollBuildStatus = (accountId, buildId) => {
```
drewjenkins, line 33: Nice. Is 1000 too fast?

miketalley: I set it to a faster poll because the build seems to happen pretty quickly in some cases. Although, this is a good point because the BE only has `PROCESSING`, `ERROR`, and `SUCCESS` states ATM, so we can dial this back. I'll go with `5000`

PR #417, opened 2021-01-14
Build logs are displayed on build completion

miketalley:
## Description and Context
On `ERROR` and `SUCCESS` of build/deploy, there is a `cdnUrl` property that is attached that links to the log output of the build. This PR fetches this content and displays it for the user when running the `hs functions deploy` command.

## Screenshots
Success output
image

Error output
image

## Who to Notify
@bkrainer @williamspiro

### DIFF
--- a/packages/cli/commands/functions/deploy.js
+++ b/packages/cli/commands/functions/deploy.js
@@ -66,6 +66,11 @@ const logBuildOutput = async resp => {
     try {
       https
         .get(resp.cdnUrl, response => {
+          // If the cdnUrl is not found, just display success message
+          if (response.statusCode === 404) {
+            resolve('');
+          }
+
           let data = '';
           response.on('data', chunk => {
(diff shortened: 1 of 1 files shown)


packages/cli/commands/functions/deploy.js
```diff
@@ -60,6 +61,26 @@ const loadAndValidateOptions = async options => {
...
+const logBuildOutput = async resp => {
+  return new Promise((resolve, reject) => {
+    try {
+      https
+        .get(resp.cdnUrl, response => {
+          let data = '';
+          response.on('data', chunk => {
+            data += chunk;
+          });
+          response.on('end', () => {
```
anthmatic, line 73: Would it be worth it to handle 404s in case the log is not available?

miketalley: Good call, I added handling for this.

PR #419, opened 2021-01-14
Remove log output payload

miketalley:
## Description and Context
The `payload` property has been removed from the BE code(see PR 176 in the BE repo). The output is being updated to prevent `undefined` from being logged every time because there is no longer `payload` data.

- Also fixed a small visual bug with error output due to `/n` being used instead of `\n`.

## Who to Notify
@bkrainer -- Thanks for your help in tracking this down!


gcorne:
Ah, great to see this. I hit this recently when doing some Serverless testing.

PR #420, opened 2021-01-15
Add messaging to use new package

anthmatic:
## Description and Context
Steers `update-notifier` towards the new package `@hubspot/cli` and adds a custom message with upgrade instructions. Open to feedback on this.

## Screenshots

image

### DIFF
--- a/packages/cms-cli/bin/cli.js
+++ b/packages/cms-cli/bin/cli.js
@@ -8,7 +8,7 @@ const { logger } = require('@hubspot/cms-lib/logger');
 const { logErrorInstance } = require('@hubspot/cms-lib/errorHandlers');
 const { setLogLevel, getCommandName } = require('../lib/commonOpts');
 const { trackHelpUsage } = require('../lib/usageTracking');
-const { version } = require('../package.json');
+const pkg = require('../package.json');
 
 const removeCommand = require('../commands/remove');
 const initCommand = require('../commands/init');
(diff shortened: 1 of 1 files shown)


packages/cms-cli/bin/cli.js
```diff
@@ -27,10 +28,16 @@ const listCommand = require('../commands/list');
 const openCommand = require('../commands/open');
 const mvCommand = require('../commands/mv');
 
+// Point update-notifier towards @hubspot/cli
+const pkg = { name: '@hubspot/cli', version };
```
anthmatic, line 32: One thing I'm not clear on is if `cms-cli` version will continue to increment with `cli`? If so, we'll probably want to hard code this to a lower version number

gcorne: Plan is to continue using the same version scheme so that the tags / versions in the repo continue to be meaningful. We'll start with 3.0 for the `@hubspot/cli` release so that there is adequate space in the versioning scheme for `@hubspot/cms-cli` should a patch be needed.

packages/cms-cli/bin/cli.js
```diff
@@ -27,10 +28,16 @@ const listCommand = require('../commands/list');
...
+
 const notifier = updateNotifier({ pkg });
 
 notifier.notify({
   shouldNotifyInNpmScript: true,
+  message:
+    chalk.bold('The CMS CLI is now the HubSpot CLI') +
+    '\n\nTo upgrade, run:\n\nnpm uninstall -g @hubspot/cms-cli\nand npm install -g @hubspot/cli',
```
gcorne, line 40: What if we adjust the logic such that we're detecting that this code is part of `@hubspot/cms-cli` so that we can merge the changes back into `master` so that the process of merging changes from `cms-cli` branch back into `master` is straightforward?

anthmatic: That's a good point, updated

gcorne (approved):

PR #421, opened 2021-01-20
Create api samples

asalauyeu-hubspot:
## Description and Context
This is the first PR about integration with SDK sample-apps.
Currently initial flow is the following:
1) The following [repo](https://github.com/HubSpot/sample-apps-list) includes samples.json config, which will be downloaded each time user runs `hs create api-sample` to check, which samples are currently available on which languages.
2) After that we have a prompt with several questions to the user, which sample type on which language they want to download (using samples config for selection choices)
3) Then download repo with sample app and put it to destination folder the same way, as it's done for react-app / vue-app / webpack-serverless / website-theme.

We've added one [repo](https://github.com/HubSpot/sample-apps-oauth) with first oauth sample app on node language to test .

## Screenshots
image
image
image
image


packages/cli/lib/configs.js
```diff
@@ -0,0 +1,53 @@
+const request = require('request-promise-native');
+const { logger } = require('@hubspot/cli-lib/logger');
+const { logErrorInstance } = require('@hubspot/cli-lib/errorHandlers');
+
+// https://developer.github.com/v3/#user-agent-required
+const USER_AGENT_HEADERS = { 'User-Agent': 'HubSpot/hubspot-cms-tools' };
```
gcorne, line 6: I think we should create a constant in `packages/cli-lib/http/requestOptions` for the constant and share it between here and `getRequestOptions`.

packages/cli/lib/configs.js
```diff
@@ -0,0 +1,53 @@
...
+ * @param {String} filePath - path where config file is stored
+ * @param {String} repoName - name of the github repository
+ * @returns {Buffer|Null} Zip data buffer
+ */
+async function downloadConfig(repoName, filePath) {
+  try {
+    const configData = await fetchRawConfig(repoName, filePath);
+    if (!configData) return;
+    const { download_url: downloadUrl, name } = configData;
+    logger.log(`Fetching ${name}...`);
```
gcorne, line 37: Since this is not something that the user needs to know except in cases where something is going wrong, I think this could be `logger.debug()`.

packages/cli/lib/configs.js
```diff
@@ -0,0 +1,53 @@
+const request = require('request-promise-native');
```
gcorne, line 1: I think it makes sense to move this file into `cli-lib` and have it live under `api`. Maybe name it something like `github.js` since it is being used to make HTTTP API request to GitHub? If we do keep in `packages/cli`, we'll need to add an explicit dep on `request-promise-native`.

packages/cli/lib/configs.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return null;
+}
+
+/**
+ * @param {String} filePath - path where config file is stored
+ * @param {String} repoName - name of the github repository
+ * @returns {Buffer|Null} Zip data buffer
+ */
+async function downloadConfig(repoName, filePath) {
```
gcorne, line 32: While this works, it does seem like a pretty indirect way to get a JSON file with information on the different samples.

packages/cli/lib/configs.js
```diff
@@ -0,0 +1,53 @@
...
+  try {
+    const configData = await fetchRawConfig(repoName, filePath);
+    if (!configData) return;
+    const { download_url: downloadUrl, name } = configData;
+    logger.log(`Fetching ${name}...`);
+    const config = await request.get(downloadUrl, {
+      encoding: null,
```
gcorne, line 39: Could do `json: true` here and avoid going from Buffer -> string -> JSON below.

packages/cli/commands/create.js
```diff
@@ -200,6 +211,62 @@ exports.handler = async options => {
...
+        if (overwrite) {
+          const removingSpinner = ora(`Removing existing ${filePath} folder`);
+          removingSpinner.start();
+          fs.rmdirSync(filePath, { recursive: true });
+          removingSpinner.stop();
+        } else {
+          return;
+        }
+      }
+      const downloadSpinner = ora('Downloading API samples config');
```
gcorne, line 227: Maybe change to "Loading available API samples"?

packages/cli/commands/create.js
```diff
@@ -200,6 +211,62 @@ exports.handler = async options => {
       createTemplate(name, dest, templateType);
       break;
     }
+    case TYPES['api-sample']: {
+      const filePath = path.join(dest, name);
+      if (fs.existsSync(filePath)) {
+        const { overwrite } = await overwriteSamplePrompt(filePath);
+        if (overwrite) {
+          const removingSpinner = ora(`Removing existing ${filePath} folder`);
```
gcorne, line 219: Since we're dealing with a local file system operation, is this necessary?

packages/cli/commands/create.js
```diff
@@ -200,6 +211,62 @@ exports.handler = async options => {
       createTemplate(name, dest, templateType);
       break;
     }
+    case TYPES['api-sample']: {
+      const filePath = path.join(dest, name);
+      if (fs.existsSync(filePath)) {
+        const { overwrite } = await overwriteSamplePrompt(filePath);
```
gcorne, line 217: This is a neat idea! We should consider using it in other `hs create` flows.

packages/cli/commands/create.js
```diff
@@ -200,6 +211,62 @@ exports.handler = async options => {
...
+        );
+        return;
+      }
+      const { sampleType, sampleLanguage } = await createApiSamplePrompt(
+        samplesConfig
+      );
+      if (!sampleType || !sampleLanguage) {
+        logger.error(
+          `Currently there are no samples available, please, try again later.`
+        );
```
gcorne, line 246: Should the message here be different?

packages/cli/commands/create.js
```diff
@@ -200,6 +211,62 @@ exports.handler = async options => {
...
+        return;
+      }
+      logger.info(
+        `You've chosen ${sampleType} sample written on ${sampleLanguage} language`
+      );
+      commandTrackingContext.templateType = sampleType;
```
gcorne, line 252: Since a sample project is pretty different from a template, I don't think we should add this to tracking context.

packages/cli/commands/create.js
```diff
@@ -200,6 +211,62 @@ exports.handler = async options => {
...
+        'samples.json'
+      );
+      downloadSpinner.stop();
+      if (!samplesConfig) {
+        logger.error(
+          `Currently there are no samples available, please, try again later.`
+        );
+        return;
+      }
+      const { sampleType, sampleLanguage } = await createApiSamplePrompt(
```
gcorne, line 240: Since at this stage, we only are offering a single node sample project. I think we should remove the prompts or remove `api-sample` from `hs create --help`.

gcorne (commented):
Great to see this. Are we planning to add additional API sample projects before merging this? If not, I think we should remove the prompt flow.

I also think we should add some info to the README on the need to create a developer account and an app in order to obtain a client id and secret.

packages/cli-lib/projects.js
```diff
@@ -44,15 +45,24 @@ async function fetchReleaseData(repoName, tag = '') {
 }
 
 /**
+ * @param {String} repoName - Name of GitHub repository to download.
  * @param {String} tag - Git tag to fetch for. If omitted latest will be fetched.
+ * @param {String} repositoryType - type of content
  * @returns {Buffer|Null} Zip data buffer
  */
-async function downloadProject(repoName, tag = '') {
+async function downloadProject(repoName, tag = '', repositoryType = 'release') {
```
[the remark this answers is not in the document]
asalauyeu-hubspot: agree, seems it'd be better name

packages/cli/commands/create.js
asalauyeu-hubspot: We're going to add new languages to existing sample app, and create new sample apps shortly, so that's why we want to leave prompts. So, will remove api-sample from help temporarily.

packages/cli/lib/configs.js
asalauyeu-hubspot: Yeah, agree. The reason we put thi to cli is that this logic is very specific (downloading config file and parsing it to json), not sure, whether it will be useful somewhere else except using it for our sample apps. But agree, it's not something that directly related to cli logic. Maybe it'd be better to move to cli-lib, since we don't want to add any unnecessary deps to cli

asalauyeu-hubspot: @gcorne maybe it will be ok to move these methods to projects.js, since they do similar logic to methods in this module (interact with github api, do something with downloaded data and return it) 
Or, here is a better idea, we can actually create something like github.js, and move methods that interact with github api to this module (fetchReleaseData for example, since currently projects.js has some mixed logic, some methods are fetching data, some methods handle the data locally)

packages/cli/commands/create.js
asalauyeu-hubspot: removed api-sample from hs create help

packages/cli/lib/configs.js
asalauyeu-hubspot: Nice point. Created DEFAULT_USER_AGENT_HEADERS constant, and shared it to projects and github.js

packages/cli/lib/configs.js
asalauyeu-hubspot: Simplified the logic of getting configs to 1 method, and made it more abstract for getting json files.

gcorne (approved):

PR #434, opened 2021-01-26
Make webpack autoupload plugin compatible with webpack 5

anthmatic:
## Description and Context
In Webpack 5, a bunch of data was removed from the `afterEmit` callback that this plugin relied on:
- the `emitted` boolean property
- the `existsAt` property

This update removes the dependency on those values and builds the `filepath` manually.

This does require `.getPath()` which is only in webpack >= v4.40. I'm curious if anyone has concerns about supporting earlier versions?

### DIFF
--- a/packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
+++ b/packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
@@ -14,7 +14,7 @@ const {
   setLogLevel,
   setLogger,
 } = require('@hubspot/cli-lib/logger');
-
+const path = require('path');
 setLogLevel(LOG_LEVEL.LOG);
 loadConfig();
 checkAndWarnGitInclusion();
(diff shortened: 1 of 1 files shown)


packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
```diff
@@ -35,34 +35,29 @@ class HubSpotAutoUploadPlugin {
     setLogger(webpackLogger);
     compiler.hooks.afterEmit.tapPromise(pluginName, async compilation => {
       Object.keys(compilation.assets).forEach(filename => {
-        const asset = compilation.assets[filename];
-        // assets with `emitted = false` haven't changed since the previous compilation
-        if (asset.emitted) {
```
anthmatic, line 40: In my testing, the only assets that were included in a compilation were assets that get written. I'm curious if I'm missing something else here, otherwise it seems safe to remove this.

cc @gcorne since you worked on v1 of this

gcorne: I think that this was intended to catch situations where the file had the same contents as the previous compilation to avoid unnecessary uploads.

packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
```diff
@@ -35,34 +35,29 @@ class HubSpotAutoUploadPlugin {
...
-          const filepath = asset.existsAt;
-
-          if (!this.autoupload || !isAllowedExtension(filepath)) {
-            return;
-          }
+        const outputPath = compilation.getPath(compilation.compiler.outputPath);
```
gcorne, line 38: Is this backwards compatible?

packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
anthmatic: I'll give this another test

packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
anthmatic: Yep, works on 4 and 5

Also, FWIW, no compatibility warnings for the `getPath()` method here https://webpack.js.org/api/compilation-object/#getpath

packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js
anthmatic: Did some more testing on this. It can't seem to reproduce a case where there is an `asset` in `compilation.assets` with `emitted: false`. Only files that were written are included in `assets`.

It's worth noting that in my test project (https://github.com/HubSpot/cms-event-registration) these files bundled for the two main entry points were updated after any change, even if it was unrelated. So those did upload even if the file content wasn't different.

miketalley (commented):
Code looks good and it works in testing with webpack versions 4 and 5!

The only thing was the upload of all assets output by the webpack build when a single file changes (as we discussed). LMK how I can help with testing this more if needed!

anthmatic:
After some testing (thanks @miketalley!) I reworked the logic here to make sure that we're only uploading emitted assets and also ensure that the first compile uploads all files consistently. I updated the PR description as well.

FYI I'm planning to get this out in a new release, but from this new repo https://github.com/HubSpot/webpack-cms-plugins

PR #438, opened 2021-01-28
Added --log-output option to console log payload body for local testing

miketalley:
## Description and Context
The function execution return value is not logged anywhere for easy viewing when running the `hs functions test` command. This PR adds the `--log-output` option that can be used to log this output. The default value is `false` for this option to mimic production.

## Screenshots
image

## Who to Notify
@TheWebTech - As discussed in the comments of the getting started doc.

### DIFF
--- a/packages/cli/commands/functions/test.js
+++ b/packages/cli/commands/functions/test.js
@@ -70,6 +70,11 @@ exports.builder = yargs => {
     type: 'boolean',
     default: true,
   });
+  yargs.option('log-output', {
+    describe: 'output the response body from the serverless function execution',
+    type: 'boolean',
+    default: false,
+  });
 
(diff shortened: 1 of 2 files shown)


packages/cli/commands/functions/test.js
```diff
@@ -70,6 +70,11 @@ exports.builder = yargs => {
     type: 'boolean',
     default: true,
   });
+  yargs.option('log-output', {
+    describe: 'output the response body from the serverless function execution',
```
drewjenkins, line 74: Should we add a note that this shouldn't be used in production because it may leak PII?

miketalley: Good call, I will add this.

PR #439, opened 2021-01-28
Handled exception if log does not contain error value

miketalley:
## Description and Context
I ran into an issue where the logging of an error encountered an error if there is no `log.error` value. This PR short circuits and does not try to log `log.error` values if there is no `log.error` property.

## Screenshots
Before
image

After
image

### DIFF
--- a/packages/cli-lib/lib/logs.js
+++ b/packages/cli-lib/lib/logs.js
@@ -11,9 +11,7 @@ const LOG_STATUS_COLORS = {
 };
 
 const errorHandler = (log, options) => {
-  return `${formatLogHeader(log, options)}${
-    options.compact ? '' : `\n${formatError(log)}`
-  }`;
+  return `${formatLogHeader(log, options)}${formatError(log, options)}`;
 };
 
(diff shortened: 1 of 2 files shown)


packages/cli-lib/lib/logs.js
```diff
@@ -5,32 +5,36 @@ const { logger, Styles } = require('../logger');
...
+  UNHANDLED_ERROR: errorHandler,
+  HANDLED_ERROR: errorHandler,
   SUCCESS: (log, options) => {
     return `${formatLogHeader(log, options)}${
       options.compact ? '' : `\n${log.log}`
     }`;
   },
 };
 
+const formatError = log => {
```
drewjenkins, line 30: Would it be good to put a try catch around the base error handler? Just to ensure we aren't breaking the customer

miketalley: There is one in the base that should catch https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lib/logs.js#L62-L66

PR #441, opened 2021-02-04
Serverless empathy fixes

miketalley:
## Description and Context
This fixes several issues that were found during the empathy session on 2/3

_Local Serverless Testing_
- Rename `hs functions test` to `hs functions server` @abelb
- Additional logging when dependencies are being installed @phalaphone 
- Errors during spawned `npm install` process will be logged and interrupt the process @brandenrodgers 
- Status will only be set on `res` If a `statusCode` is provided
- Fixed error when testing locally with no `.env` file, but having specified secrets in `serverless.json`

_Serverless Deploy_
- Additional handling if no `cdnUrl` is provided @bkrainer @gkemp94

### DIFF
--- a/packages/cli/commands/functions.js
+++ b/packages/cli/commands/functions.js
@@ -5,7 +5,7 @@ const {
 } = require('../lib/commonOpts');
 const list = require('./functions/list');
 const deploy = require('./functions/deploy');
-const offline = require('./functions/offline');
+const server = require('./functions/server');
 
 exports.command = 'functions';
 exports.describe = 'Commands for working with functions';
@@ -21,7 +21,7 @@ exports.builder = yargs => {
       aliases: 'ls',
     })
     .command(deploy)
-    .command(offline)
+    .command(server)
     .demandCommand(1, '');
 
   return yargs;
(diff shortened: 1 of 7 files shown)


packages/cli/commands/functions/deploy.js
```diff
@@ -62,6 +62,13 @@ const loadAndValidateOptions = async options => {
 };
 
 const logBuildOutput = async resp => {
+  const { cdnUrl } = resp;
+
+  if (!cdnUrl) {
+    logger.debug('Unable to display build output. No cdnUrl was provided.');
```
anthmatic, line 68: Nit: `cdnUrl` seems like an implementation detail that we don't need to pass along

miketalley: good point, I will make this message more generic

packages/cli/commands/functions.js
```diff
@@ -5,7 +5,7 @@ const {
 } = require('../lib/commonOpts');
 const list = require('./functions/list');
 const deploy = require('./functions/deploy');
-const test = require('./functions/test');
+const server = require('./functions/server');
```
kemmerle, line 8: +1 for this change; it's much more intuitive.

kemmerle (approved):
@miketalley, these changes make sense to me. I think that they make the experience spinning up a server much more developer-friendly--I appreciate that we've added more logging when installing dependencies and that we've streamlined our error handling.

anthmatic (approved):
Changes look good. Also tested and seems to work as expected.

PR #442, opened 2021-02-12
Handle international characters in remote HS paths

lars-erik:
## Description and Context
Using for instance `hs remove` with international characters like 'æøå' fails. (`[ERROR] The delete of "Æøå" in account nnnnnnn was bad.`)  
The `fileMapper.deleteFile` method now properly uses encodeURIComponent.  
I haven't browsed the rest of the methods, but likely all of them should encode paths.

The initial commit was a silly attempt breaking all tests. Second commit is likely the right solution. (Did not find any "remove" tests) 😇

## Screenshots
```
❯ hs remove Æoå
[ERROR] Deleting "Æøå" from account nnnnnnn failed
[ERROR] The delete of "Æøå" in account nnnnnnn was bad.
```
patched: 
```
❯ yarn hs remove Æoå
yarn run v1.22.10
$ "C:\Users\lars-\source\repos\open source\hubspot-cli\node_modules\.bin\hs" remove Æøå
Deleted "Æøå" from account nnnnnnn
Done in 2.22s.
```

## Who to Notify
🤷‍♂️

### DIFF
--- a/packages/cli/commands/remove.js
+++ b/packages/cli/commands/remove.js
@@ -20,7 +20,6 @@ const {
 const { logDebugInfo } = require('../lib/debugInfo');
 const { validateAccount } = require('../lib/validation');
 const { trackCommandUsage } = require('../lib/usageTracking');
-const { convertToUnixPath } = require('@hubspot/cli-lib/path');
 
 exports.command = 'remove <path>';
 exports.describe = 'Delete a file or folder from HubSpot';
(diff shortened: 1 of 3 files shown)


packages/cli/commands/remove.js
```diff
@@ -40,7 +41,8 @@ exports.handler = async options => {
   trackCommandUsage('remove', {}, accountId);
 
   try {
-    await deleteFile(accountId, hsPath);
+    let unixPath = convertToUnixPath(hsPath);
```
gcorne, line 44: The expectation here is that the path the user provides will be in HubSpot format so can we remove this?

gcorne:
Thanks for fixing this. If we can remove the change to `remove.js`, we can get this in.

lars-erik:
Done. :)

lars-erik:
🙏🎉

PR #443, opened 2021-02-12
Add new vs code endpoint

anthmatic:
## Description and Context
To support the VS Code extension, a new analytics endpoint has been created. This PR adds that endpoint to the existing usage tracking function.

### DIFF
--- a/packages/cli-lib/api/fileMapper.js
+++ b/packages/cli-lib/api/fileMapper.js
@@ -181,11 +181,23 @@ async function trackUsage(eventName, eventClass, meta = {}, accountId) {
     eventClass,
     meta,
   };
-  const endpointMapping = {
-    'cli-interaction': 'cms-cli-usage',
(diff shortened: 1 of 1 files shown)


packages/cli-lib/api/fileMapper.js
```diff
@@ -181,7 +181,12 @@ async function trackUsage(eventName, eventClass, meta = {}, accountId) {
     eventClass,
     meta,
   };
-  const path = `${FILE_MAPPER_API_PATH}/cms-cli-usage`;
+  const endpointMapping = {
+    'cli-interaction': 'cms-cli-usage',
+    'vscode-extension-interaction': 'vscode-extension-usage',
```
gcorne, line 186: Maybe create an enum for the possible `eventName` values?

anthmatic: Thanks @gcorne. I updated this if you don't mind taking another look I'd appreciate it

PR #449, opened 2021-03-02
Add marketplace asset validation

brandenrodgers:
## Description and Context

Initial work to port our marketplace asset validation (for themes, specifically) to the CLI. The main goal here was to make the validations as modular as possible so that it is easy to add new ones.

This is in a pretty rough state because it is my first time contributing to the repo. Let me know if I should rename/move any files or functions. I also did not spend much time thinking about the name of the command, so I'm sure that could use some love.

## Screenshots

Here's an example output with dummy values put into the validators.
image

## TODO

- [ ] Put more thought into command naming and console logging
- [ ] Verify that these validations are correct
## Who to Notify

@KatzMitch @gcorne @miketalley @anthmatic @drewjenkins


KatzMitch:
🤑  this is sick, thank you for working on this!
I think the command should have `marketplace` in the name somewhere as there are other (lower requirement) validations we can run on themes (i.e. the design manager validations)

packages/cli-lib/lib/walk.js
```diff
@@ -38,36 +44,52 @@ const generateFilePromise = (dir, file) => {
...
-      processFiles(files).then(filesData => {
-        resolve(filesData.reduce(filesDataReducer, []));
+      fileProcessor(files).then(filesData => {
+        resolve(filesDataReducer(filesData));
       });
     });
   });
 }
 
+function getDirectoryFiles(dir) {
```
brandenrodgers, line 76: I updated this file per feedback I received [here](https://github.com/HubSpot/hubspot-cli/commit/7f1048dc081bc645e02cd21079842cf9b4d7e84e#r47510550). Tbh I'm not crazy about my solution, but my goal was to create two separate "deeply walk" and "shallow walk" functions for grabbing files in a folder that share logic. Let me know if you can think of a better way to handle this!

packages/cli/commands/validation/themeValidators/themeConfigValidator.js
```diff
@@ -0,0 +1,31 @@
+const { walk } = require('@hubspot/cli-lib/lib/walk');
+const { ERROR_SEVERITY } = require('../validationErrorUtils');
+
+const VALIDATOR_NAME = 'ThemeConfigValidator';
+const THEME_JSON_REGEX = new RegExp(/theme\.json+$/);
+
+//TODO branden do we need this validation? Should we also look at the fields in
+// the theme.json file?
```
anthmatic, line 8: If we can generate field json schemas that would make it pretty trivial to validate theme.json

packages/cli/commands/validation/themeValidators/templateLimitValidator.js
```diff
@@ -0,0 +1,27 @@
...
+const { ERROR_SEVERITY } = require('../validationErrorUtils');
+
+const VALIDATOR_NAME = 'TemplateLimitValidator';
+const TEMPLATE_LIMIT = 50;
+
+// TODO branden are we checking for all files here? OR should we check for templates specifically
+
+// Validate that the theme does not contain more than TEMPLATE_LIMIT templates
+function templateLimitValidator(absoluteThemePath) {
+  return getDirectoryFiles(absoluteThemePath).then(files => {
```
anthmatic, line 11: Since templates could be nested at any depth, I think we need to walk here and filter out HTML files (we might even need to filter out partials)

packages/cli-lib/lib/walk.js
anthmatic: More of a design question, but would make sense at all to walk the whole theme up front when the command is run and each validator could work off that file tree?

brandenrodgers: Yeah I considered that, but decided against it b/c for the config validator I need to make sure that the `theme.json` file exists at the top level. Right now `walk` returns a flat list of all files so I wouldn't be able to determine whether the `theme.json` file was in the right place or not.

That being said, I'm guessing the `ThemeConfigValidator` is going to be the only validator that _only_ cares about top level files, and maybe running the walk command up front will still be an overall optimization.

packages/cli/commands/validation/themeValidators/templateLimitValidator.js
brandenrodgers: Oh good catch. I put this in the wrong place after testing. I meant to put it in the `ThemeConfigValidator`.

brandenrodgers: > we might even need to filter out partials

@KatzMitch should we be filtering out partials? Also I'm currently just checking for total file count (not specifically looking for templates) and I think that's wrong.

packages/cli/commands/validation/validationErrorUtils.js
```diff
@@ -0,0 +1,28 @@
+const { logger } = require('@hubspot/cli-lib/logger');
+
+//TODO branden I'm sure this should live somewhere else, but not sure where
```
brandenrodgers, line 3: Should this type of util file live somewhere else?

packages/cli/commands/validation/themeValidators/templateLimitValidator.js
KatzMitch: Yeah it's 50 top-level templates, excludes partials

packages/cli-lib/lib/walk.js
```diff
@@ -7,7 +7,7 @@ const STAT_TYPES = {
   DIRECTORY: 'dir',
 };
 
-const generateFilePromise = (dir, file) => {
+const generateFilePromise = (dir, file, { shallow = false } = {}) => {
```
gcorne, line 10: I think I mentioned this elsewhere but I am not super fond of `walk` having a shallow mode since it is designed to be for walking a tree and Node.js has `readdir` built in. Perhaps we could have a separate utility for reading a single directory that shares lower-level functionality with `walk`?

packages/cli-lib/lib/walk.js
gcorne: If we make the assumption that the directory being scanned is the root folder of a theme, one possible way of handling this would be to look for a file with a specific path in the flat listing (e.g. `files.filter(filepath => filepath === 'theme.json')`).

packages/cli/commands/validation/validateTheme.js
```diff
@@ -0,0 +1,90 @@
...
+  trackCommandUsage('validate', {}, accountId);
+
+  applyValidators(themeValidators, absoluteSrcPath).then(errors => {
+    if (errors.length) {
+      logValidatorErrors(errors);
+    } else {
+      logger.success(`Theme is valid \n`);
+    }
+  });
+};
```
gcorne, line 76: One suggestion is to consider how this might work in tooling. For example, what is needed to fail an automated build? For this reason, I think we should `process.exit(1)` if the errors are fatal. We may also want to consider how to produce machine readable output like `eslint` and other tools produce.

packages/cli/lib/validators/marketplaceValidators/templateLimitValidator.js
```diff
@@ -0,0 +1,26 @@
+const { walk } = require('@hubspot/cli-lib/lib/walk');
+const { ERROR_SEVERITY } = require('../constants');
+
+const VALIDATOR_NAME = 'TemplateLimitValidator';
+const TEMPLATE_LIMIT = 50;
+const TEMPLATE_REGEX = new RegExp(/.html+$/);
+
+// Validate that the theme does not contain more than TEMPLATE_LIMIT templates
+function templateLimitValidator(absoluteThemePath) {
```
gcorne, line 9: I realize that these are WIP and you mention correctness so maybe they are planned, but I'd recommend adding unit tests for the validators.

packages/cli-lib/lib/walk.js
brandenrodgers: I took a shot at this by creating a `read.js` file that handles reading the files shallowly in a directory. "Walk" shares some of the logic in it. Is that what you had in mind? There are quite a few different ways that this could be split so I'm open to other suggestions!

gcorne: Looks legit. We are probably long overdue to add some unit tests covering `walk` (sorry). It has gotten a lot more complex that when it was first written.

packages/cli-lib/lib/read.js
```diff
@@ -0,0 +1,67 @@
+const fs = require('fs');
+const path = require('path');
+
+const STAT_TYPES = {
+  FILE: 'file',
+  SYMBOLIC_LINK: 'symlink',
+  DIRECTORY: 'dir',
+};
+
+const generateFilePromise = (dir, file) => {
```
gcorne, line 10: Maybe rename `getFileInfoAsync` to make the purpose clearer?

packages/cli-lib/lib/read.js
```diff
@@ -0,0 +1,67 @@
...
+        type = STAT_TYPES.DIRECTORY;
+      } else if (stats.isFile()) {
+        type = STAT_TYPES.FILE;
+      }
+      resolve({ filepath, type });
+    });
+  });
+};
+
+const formatFilesData = filesData => {
```
gcorne, line 30: Maybe rename and include `flatten` or `flattenWithoutSymlinks`?

packages/cli-lib/lib/read.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }, []);
+};
+
+async function read(dir) {
+  const processFiles = files =>
+    Promise.all(files.map(file => generateFilePromise(dir, file)));
+
+  return new Promise((resolve, reject) => {
+    fs.readdir(dir, (error, files) => {
```
gcorne, line 51: Now that the minimum node version is 10, I think we could use https://nodejs.org/docs/latest-v10.x/api/fs.html#fs_fs_promises_api if we wanted. That said, if we are going to do more surgery, we should add test coverage.

packages/cli-lib/lib/walk.js
```diff
@@ -1,68 +1,32 @@
...
-            type: STAT_TYPES.DIRECTORY,
-            files,
-          });
-        });
-      } else if (stats.isFile()) {
-        resolve({
-          filepath,
-          type: STAT_TYPES.FILE,
+const generateRecursiveFilePromise = (dir, file) => {
+  return generateFilePromise(dir, file).then(fileData => {
```
gcorne, line 6: Seeing `fileData` makes me crave TypeScript. :smile:

packages/cli-lib/lib/read.js
brandenrodgers: Done in b12426c

packages/cli-lib/lib/read.js
brandenrodgers: Good call, done in b12426c

packages/cli-lib/lib/read.js
brandenrodgers: updated in b12426c, but I still need to add tests for `read` and `write`. I'm not quite sure what the best way to mock `fs` is. I'll look into it though.

gcorne: Maybe some inspiration in https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/__mocks__/fs-extra.js?

packages/cli/lib/validators/marketplaceValidators/templateLimitValidator.js
brandenrodgers: I added tests for the validators in 1a1eeec . I'm still not 100% sure that they're validating the right criteria, but I'll confirm before merging.

packages/cli/commands/validation/themeValidators/templateLimitValidator.js
brandenrodgers: @gcorne @anthmatic is there a good way to differentiate between templates and partials?

packages/cli/commands/validation/validateTheme.js
brandenrodgers: Resolving b/c I added an exit if any errors are fatal, and I also added a json output option. The format of the json output  is:
```json
{
  "success": false,
  "results": {
    "ThemeConfigValidator": [
      {"validator":"ThemeConfigValidator","error":"Invalid json in theme.json file","result":"FATAL"}
    ],
    "TemplateLimitValidator": [
      {"validator":"TemplateLimitValidator","result":"SUCCESS"}
    ]
  }
}
```

packages/cli/commands/validation/themeValidators/templateLimitValidator.js
anthmatic: Nothing that exists now, I think this would require reading the file and parsing out the HTML comment (annotation) at the top.

packages/cli-lib/lib/walk.js
brandenrodgers: Updated the code to only walk the files one time at the start of validation in 7da9192

packages/cli/commands/validate.js
```diff
@@ -0,0 +1,19 @@
...
+  addAccountOptions,
+  addOverwriteOptions,
+} = require('../lib/commonOpts');
+const validateTheme = require('./validation/validateTheme');
+
+exports.command = 'validate';
+exports.describe = 'Commands for working with validation';
+
+exports.builder = yargs => {
+  addOverwriteOptions(yargs, true);
```
gcorne, line 12: This should be removed since the command doesn't have a need to overwrite like `hs fetch` does.

packages/cli/commands/validate.js
```diff
@@ -0,0 +1,19 @@
+const {
+  addConfigOptions,
+  addAccountOptions,
+  addOverwriteOptions,
+} = require('../lib/commonOpts');
+const validateTheme = require('./validation/validateTheme');
```
gcorne, line 6: Might be worth renaming `validation` directory to `validate` to match the pattern used for the other commands where the directory matches the top-level command.

packages/cli/lib/__tests__/validators/themeConfigValdiator.js
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs');
+const themeConfigValidator = require('../../validators/marketplaceValidators/themeConfigValidator');
+const { VALIDATION_RESULT } = require('../../validators/constants');
```
gcorne, line 3: Instead of having the tests live in `lib`, I'd move to inside `validators`. It is typical for jest tests to be adjacent to code that the tests cover.

packages/cli/lib/validators/marketplaceValidators/themeConfigValidator.js
```diff
@@ -0,0 +1,52 @@
...
+const { VALIDATION_RESULT } = require('../constants');
+
+const VALIDATOR_NAME = 'ThemeConfigValidator';
+const THEME_JSON_REGEX = new RegExp(/^theme\.json$/);
+
+// Validate that the theme contains a theme.json file
+function themeConfigValidator(absoluteThemePath, files) {
+  let validationErrors = [];
+  const themeJSONFile = files.find(fileName => {
+    return THEME_JSON_REGEX.test(fileName);
```
gcorne, line 12: Instead of a regex, could this be `fileName === 'theme.json'`?

packages/cli/lib/validators/marketplaceValidators/templateLimitValidator.js
```diff
@@ -0,0 +1,26 @@
+const { VALIDATION_RESULT } = require('../constants');
+
+const VALIDATOR_NAME = 'TemplateLimitValidator';
+const TEMPLATE_LIMIT = 50;
+const TEMPLATE_REGEX = new RegExp(/.html+$/);
```
gcorne, line 5: Will this match `module.html`?

brandenrodgers: Yeah that's a good catch, updated in ee4a6dd

packages/cli/lib/validators/marketplaceValidators/themeConfigValidator.js
brandenrodgers: Yep! Much simpler. Done in ee4a6dd

packages/cli/lib/__tests__/validators/themeConfigValdiator.js
brandenrodgers: Moved in ee4a6dd

packages/cli/commands/validate.js
brandenrodgers: Ah I didn't notice that pattern, updated in ee4a6dd 👍

packages/cli/commands/validate.js
brandenrodgers: Done in ee4a6dd

brandenrodgers:
Thanks for taking another pass @gcorne! I think I addressed all of your feedback. I also made some other fairly significant changes (in ee4a6dd).
- The `templateLimitValidator` was unintentionally including partials, global partials, and module.html files. I added a `templates.js` util file so we can filter out everything that isn't a template. I had to use regex to parse the annotations in the html files because I couldn't think of a better way to do it. I left comments explaining what they do, but I realize that it adds complexity to use them so I'm open to other suggestions there.
- I introduced a bug to `walk` which was causing it to only shallowly read files. I fixed it and [simplified the string of callbacks](https://github.com/HubSpot/hubspot-cli/pull/449/commits/ee4a6dd4804e98b982e9c15fc4188ed45cb38a1c#diff-99cbb3e05abbe8746e3467716cf53808d7fe45636535ce9f92aa22d948565431R28) (same with read)

KatzMitch:
Belated follow up here:
I think for the JSON object format, we should have some unique enum per error message so they can be i18n'ed
It'd also be great if we can point to the specific file (and line number) whenever possible

anthmatic:
I may be doing something wrong, but when I run this against the theme boilerplate project the output is blank: 
image

KatzMitch:
Also suggestion for the command name: `mkpl-validate`. That should make it clear that it's marketplace validation (stricter set of requirements) and not the standard DM validation

brandenrodgers:
> I think for the JSON object format, we should have some unique enum per error message so they can be i18n'ed

@KatzMitch that's a great point. I'm going to hold off on adding this in my first pass though because @miketalley is working on a POC for [adding language yaml files to this project](https://github.com/HubSpot/hubspot-cli/pull/476). I can add the validation copy to that once we've settled on a pattern there.

> I may be doing something wrong, but when I run this against the theme boilerplate project the output is blank:

@anthmatic I think you need to run it with the `--marketplace` option for the two marketplace validators to run. Right now running `hs validate theme` without that flag will do nothing. Eventually my hope is for that to run the standard Design Manager validation, but I haven't added support for it yet. This kind of fits in with Mitchell's commend about updating the naming. I'm not in love with the current naming either so I'll run it by design.

KatzMitch:
Hey @brandenrodgers - what more needs to happen before getting this PR merged in?

packages/cli/lib/validators/logValidatorResults.js
```diff
@@ -0,0 +1,51 @@
...
+      }
+      acc[result.validatorKey] = [];
+    }
+    const {
+      validatorName: __omit1, // eslint-disable-line no-unused-vars
+      validatorKey: __omit2, // eslint-disable-line no-unused-vars
+      ...resultWithOmittedValues
+    } = result;
+    acc[result.validatorKey].push(resultWithOmittedValues);
+    return acc;
```
PaulJKim, line 19: My Javascript is not the strongest so I apologize if I'm missing something obvious here. Where does `resultWithOmittedValues` come from and what does that data look like?

packages/cli/lib/validators/logValidatorResults.js
```diff
@@ -0,0 +1,51 @@
+const { logger } = require('@hubspot/cli-lib/logger');
+const { VALIDATION_RESULT } = require('./constants');
+
+function logResultsAsJson(results) {
+  let success = true;
+  const resultObj = results.reduce((acc, result) => {
```
PaulJKim, line 6: If I'm following correctly, it seems like we're taking the `results` array, applying a bit of logic to each element, and returning another array as the accumulator. Just curious if there's a reason to use `reduce` rather than `map`

packages/cli/lib/validators/logValidatorResults.js
```diff
@@ -0,0 +1,51 @@
...
+    if (!acc[result.validator]) {
+      if (success && result.result !== VALIDATION_RESULT.SUCCESS) {
+        success = false;
+      }
+      acc[result.validatorKey] = [];
+    }
+    const {
+      validatorName: __omit1, // eslint-disable-line no-unused-vars
```
PaulJKim, line 14: again, sorry for my lack of javascript knowledge here, but what is `__omit1` and `__omit2`

packages/cli/lib/validators/logValidatorResults.js
brandenrodgers: Yeah, I'm using reduce here because I'm actually building a nested object that has a slightly different shape than the input list. I might be able to use map, but reduce felt more natural. It looks kind of like this:
```js
// Input
[
  { validatorKey: 'themeConfig', result: 'FATAL', error: 'blah1' },
  { validatorKey: 'themeConfig', result: 'WARNING', error: 'blah2' },
  { validatorKey: 'templateLimit', result: 'FATAL', error: 'blah3' },
]

// Output
{
  themeConfig: [ { result: 'FATAL', error: 'blah1' },  { result: 'WARNING', error: 'blah2' } ],
  templateLimit: [ { result: 'FATAL', error: 'blah3' } ],
}
```

packages/cli/lib/validators/logValidatorResults.js
brandenrodgers: So this is a way to remove fields from an object. It's basically doing the same thing as [_.omit()](https://underscorejs.org/#omit). I'm destructuring off the fields that I want to omit, then using `...rest` to get the resulting object w/out those fields. `__omit1` and `__omit2` are just names I used to try to make it more clear that these variables aren't being used anywhere.

packages/cli/lib/validators/logValidatorResults.js
brandenrodgers: I think I mostly explained where `resultWithOmittedValues` is coming from in my answer above, but I'll share what the shape of it looks like. All I'm doing here is omitting the `validatorKey` and `validatorName` fields from the object before we output the json because I don't think either of those fields are useful for the BE or any other consuming system. I'm only using them for error grouping and console logging purposes.
```js
// result
{
  validatorKey: 'templateLimit',
  validatorName: 'Template limit',
  key: 'templateLimit.limitExceeded',
  result: 'FATAL',
  error: 'blah blah blah'
}

// resultWithOmittedValues
{
  key: 'templateLimit.limitExceeded',
  result: 'FATAL',
  error: 'blah blah blah'
}
```

packages/cli/lib/validators/logValidatorResults.js
```diff
@@ -0,0 +1,51 @@
...
+      ...resultWithOmittedValues
+    } = result;
+    acc[result.validatorKey].push(resultWithOmittedValues);
+    return acc;
+  }, {});
+  const result = {
+    success,
+    results: resultObj,
+  };
+  return logger.log(JSON.stringify(result));
```
brandenrodgers, line 25: Here's an example of what this json output looks like:
```json
{"success":false,"results":{"themeConfig":[{"error":"The theme.json file is missing a \"label\" field","result":"FATAL","key":"themeConfig.missingLabel","meta":{"file":"/Users/brodgers/src/cms-devex-super-repo/hubspot-cli/tmp/theme.json"}}],"templateLimit":[{"result":"SUCCESS"}]}}
```

packages/cli/lib/validators/logValidatorResults.js
PaulJKim: Oh gotcha, we're consolidating all of the results based on their respective validation types

packages/cli/lib/validators/logValidatorResults.js
PaulJKim: cool! thanks for that explanation! Makes sense, we can throw away the ValidatorKey since that's what we're consolidating everything on and we don't need to know the ValidatorName anymore at this point

brandenrodgers:
I'm going to merge this as a hidden command. There is still work to do (add more tests, create more validators, etc), but I will address it in follow up PR's just so this one doesn't grow to be too large.

PR #454, opened 2021-03-16
Adding initial support for project deploys

miketalley:
## Description and Context
https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/285

This PR adds the command `hs project deploy <projectFolderPath>` and adds a `--staging` flag to the `hs fetch` command that will add the param `environmentId=2` to the request (default is `1`).

## Note
- Right now if you deploy and then fetch the `--staging` assets in the project folder, there will be an error output for every file that is live in the project folder. I spoke to @liamrharwood and he let me know there is a known BE issue dealing with fetching a folder as opposed to a file from the `content/filemapper/v1/download/${folderPath}` endpoint. Hopefully this will no longer occur when that bug is fixed, as it looks like we are passing the `environmentId: 2` properly to that endpoint when fetching.

## Who to Notify
@liamrharwood

### DIFF
--- a/packages/cli-lib/api/fileMapper.js
+++ b/packages/cli-lib/api/fileMapper.js
@@ -263,7 +263,6 @@ async function getDirectoryContentsByPath(portalId, path) {
  * @returns {Promise}
  */
 async function deployProject(portalId, projectPath) {
-  console.log('projectPath: ', projectPath);
   return http.post(portalId, {
     uri: `${FILE_MAPPER_API_PATH}/deploy/${encodeURIComponent(projectPath)}`,
   });
--- a/packages/cli-lib/api/fileMapper.js
+++ b/packages/cli-lib/api/fileMapper.js
@@ -255,9 +255,24 @@ async function getDirectoryContentsByPath(portalId, path) {
   });
 }
 
+/**
+ * Deploy project
+ *
+ * @async
+ * @param {string} projectPath
+ * @returns {Promise}
+ */
+async function deployProject(portalId, projectPath) {
+  console.log('projectPath: ', projectPath);
+  return http.post(portalId, {
+    uri: `${FILE_MAPPER_API_PATH}/deploy/${encodeURIComponent(projectPath)}`,
+  });
+}
+
 module.exports = {
   deleteFile,
   deleteFolder,
+  deployProject,
   download,
   downloadDefault,
   fetchFileStream,
--- a/packages/cli-lib/fileMapper.js
+++ b/packages/cli-lib/fileMapper.js
@@ -390,7 +390,7 @@ async function downloadFile(input) {
  * @returns {Promise<FileMapperNode}
  */
 async function fetchFolderFromApi(input) {
-  const { accountId, src, mode } = input;
+  const { accountId, src, mode, options } = input;
   const { isRoot, isFolder, isHubspot } = getTypeDataFromPath(src);
   if (!isFolder) {
     throw new Error(`Invalid request for folder: "${src}"`);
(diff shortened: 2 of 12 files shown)


packages/cli-lib/fileMapper.js
```diff
@@ -91,11 +91,18 @@ function useApiBuffer(mode) {
...
 /**
- * @param {Mode} mode
+ * Determines API param based on mode an options
+ *
+ * @typedef {Object} APIOptions
+ * @property {Mode} mode
+ * @property {Object} options
  */
-function getFileMapperApiQueryFromMode(mode) {
+function getFileMapperQueryValues({ mode, options = {} }) {
```
anthmatic, line 100: 👍

packages/cli/commands/fetch.js
```diff
@@ -82,5 +82,13 @@ exports.builder = yargs => {
     type: 'string',
   });
 
+  yargs.options({
+    staging: {
+      describe: 'retrieve staged changes for project',
```
brandenrodgers, line 87: Nit, but I think `retrieve` should be uppercase

packages/cli/commands/projects.js
```diff
@@ -0,0 +1,19 @@
+const {
+  addConfigOptions,
+  addAccountOptions,
+  addOverwriteOptions,
+} = require('../lib/commonOpts');
+const deploy = require('./projects/deploy');
+
+exports.command = 'projects';
+exports.describe = false; //'Commands for working with projects';
```
brandenrodgers, line 9: I'm new to yargs. Is this what's causing the projects command to be hidden?

packages/cli/commands/projects/deploy.js
```diff
@@ -0,0 +1,89 @@
...
+  logger.debug(`Deploying project at path: ${projectPath}`);
+
+  try {
+    const deployResp = await deployProject(accountId, projectPath);
+
+    if (deployResp.error) {
+      logger.error(`Deploy error: ${deployResp.error.message}`);
+    }
+
+    logger.success(
```
brandenrodgers, line 55: Will we show the success message here even if there is an error?

packages/cli-lib/lib/__tests__/uploadFolder.js
```diff
@@ -49,7 +49,7 @@ describe('uploadFolder', () => {
 
       uploadedFilesInOrder.forEach((file, index) => {
         expect(upload).nthCalledWith(index + 1, accountId, file, file, {
-          qs: { buffer: false },
+          qs: { buffer: false, environmentId: 1 },
```
brandenrodgers, line 52: Will the `environmentId` field be ignored in accounts that do not have Projects access?

brandenrodgers (approved):
Just a few questions, but this lgtm 👍

packages/cli-lib/api/fileMapper.js
```diff
@@ -255,9 +255,23 @@ async function getDirectoryContentsByPath(portalId, path) {
...
 }
 
+/**
+ * Deploy project
+ *
+ * @async
+ * @param {string} projectPath
+ * @returns {Promise}
+ */
+async function deployProject(portalId, projectPath) {
```
drewjenkins, line 265: Nit:  This could be someting like

```
const deployProject = async (portalId, projectPath) => http.post(portalId, {
  uri: `${FILE_MAPPER_API_PATH}/deploy/${encodeURIComponent(projectPath)}`,
});
```

Though not a big deal if you prefer your style

packages/cli/commands/fetch.js
```diff
@@ -82,5 +82,13 @@ exports.builder = yargs => {
     type: 'string',
   });
 
+  yargs.options({
```
drewjenkins, line 85: Is this something we want visible right away, or should it be hidden?

packages/cli/commands/projects.js
drewjenkins: Yes

drewjenkins (approved):

packages/cli/commands/projects.js
miketalley: Yes, if you return `false`, nothing will show in the `--help` section for the command.

packages/cli-lib/lib/__tests__/uploadFolder.js
miketalley: It will default to `1` if the `--staging` option is not tacked onto the command.

packages/cli-lib/api/fileMapper.js
miketalley: I actually prefer that syntax as well, I was just matching the current pattern used throughout the file.

packages/cli/commands/fetch.js
miketalley: Good catch, going to hide this.

PR #457, opened 2021-03-26
remove to limit notifications to the most relevant

ksvirkou-hubspot:
remove to limit notifications to the most relevant

### DIFF
--- a/packages/cli-lib/projects.js
+++ b/packages/cli-lib/projects.js
@@ -71,7 +71,6 @@ async function downloadProject(
       encoding: null,
       headers: { ...DEFAULT_USER_AGENT_HEADERS },
     });
-    logger.log('Completed project fetch.');
     return zip;
   } catch (err) {
     logger.error('An error occured fetching the project source.');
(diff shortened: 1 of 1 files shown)


tpeuneu-hubspot:
I noticed that this is a public RP with a private link in the description

tpeuneu-hubspot (approved):

packages/cli-lib/projects.js
```diff
@@ -71,7 +71,6 @@ async function downloadProject(
       encoding: null,
       headers: { ...DEFAULT_USER_AGENT_HEADERS },
     });
-    logger.log('Completed project fetch.');
```
anthmatic, line 74: It might be worth it to keep these and switch to the debug level (`logger.debug`) to aid in troubleshooting in the future

ksvirkou-hubspot: I can't do it because I switch level on my local machine but it prints debug messages

gcorne (approved):

PR #458, opened 2021-03-26
copy .env.template to .env if it exists

ksvirkou-hubspot:
copy .env.template to .env if it exists after downloading a sample app

### DIFF
--- a/packages/cli/commands/create.js
+++ b/packages/cli/commands/create.js
@@ -260,6 +260,9 @@ exports.handler = async options => {
         }
       );
       if (created) {
+        if (fs.existsSync(`${filePath}/.env.template`)) {
+          fs.copySync(`${filePath}/.env.template`, `${filePath}/.env`);
+        }
         logger.success(
           `Please, follow ${filePath}/README.md to find out how to run the sample`
         );


anthmatic (approved):

packages/cli/commands/create.js
```diff
@@ -260,6 +260,9 @@ exports.handler = async options => {
         }
       );
       if (created) {
+        if (fs.existsSync(`${filePath}/.env.template`)) {
+          fs.copySync(`${filePath}/.env.template`, `${filePath}/.env`);
+        }
```
gcorne, line 265: One thing that I am not sure I understand is why the boilerplates should not be adjusted to have `.env`.

atanasiuk-hubspot: Hi, @gcorne, adjusting `.env` to the boilerplates is a good idea, but we have some concerns about excluding it from `.gitignore` since:
- it needs to be gitignored during our development
- user will most likely add it to .gitignore again

So, wondering, how do you see adjusting `.env` to the boilerplates?

gcorne: Alright, I see why using a template was chosen. I think copying from a template is okay, although I wonder if moving the file would be better.

gcorne (approved):

PR #460, opened 2021-03-30
Oauth URL broken and runtime error with logErrorInstance

miketalley:
## Description and Context
- `buildAuthUrl` was not obtaining the correct `accountId` because it was looking for `portalId` instead this resulted in `undefined` being used in the URL instead of the `portalId`
- The import of `logErrorInstance` was wrong and causing a runtime error

```
[ERROR] A TypeError has occurred. logErrorInstance is not a function
[DEBUG] Error: TypeError: logErrorInstance is not a function
    at addOauthToAccountConfig (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli-lib/oauth.js:50:5)
    at authenticateWithOauth (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli/lib/oauth.js:98:3)
    at processTicksAndRejections (node:internal/process/task_queues:94:5)
    at async Object.oauthConfigCreationFlow [as oauth2] (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli/commands/init.js:58:3)
    at async Object.exports.handler (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli/commands/init.js:108:27) {
  [stack]: 'TypeError: logErrorInstance is not a function\n' +
    '    at addOauthToAccountConfig (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli-lib/oauth.js:50:5)\n' +
    '    at authenticateWithOauth (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli/lib/oauth.js:98:3)\n' +
    '    at processTicksAndRejections (node:internal/process/task_queues:94:5)\n' +
    '    at async Object.oauthConfigCreationFlow [as oauth2] (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli/commands/init.js:58:3)\n' +
    '    at async Object.exports.handler (/Users/mtalley/.config/yarn/global/node_modules/@hubspot/cli/commands/init.js:108:27)',
  [message]: 'logErrorInstance is not a function'
```

## Who to Notify
@alnoorp

### DIFF
--- a/packages/cli/lib/oauth.js
+++ b/packages/cli/lib/oauth.js
@@ -13,9 +13,9 @@ const redirectUri = `http://localhost:${PORT}/oauth-callback`;
 
 const buildAuthUrl = oauthManager => {
   return (
-    `${getHubSpotWebsiteOrigin(
-      oauthManager.env
-    )}/oauth/${oauthManager.portalId || oauthManager.accountId}/authorize` +
+    `${getHubSpotWebsiteOrigin(oauthManager.env)}/oauth/${
+      oauthManager.accountId
+    }/authorize` +
     `?client_id=${encodeURIComponent(oauthManager.clientId)}` + // app's client ID
     `&scope=${encodeURIComponent(oauthManager.scopes.join(' '))}` + // scopes being requested by the app
     `&redirect_uri=${encodeURIComponent(redirectUri)}` // where to send the user after the consent page
--- a/packages/cli/lib/oauth.js
+++ b/packages/cli/lib/oauth.js
@@ -79,7 +79,8 @@ const authorize = async oauthManager => {
   });
 };
 
-const setupOauth = (accountId, accountConfig) => {
+const setupOauth = accountConfig => {
+  const accountId = parseInt(accountConfig.portalId, 10);
   const config = getAccountConfig(accountId) || {};
   return new OAuth2Manager(
     {
@@ -91,8 +92,7 @@ const setupOauth = (accountId, accountConfig) => {
 };
 
 const authenticateWithOauth = async configData => {
-  const accountId = parseInt(configData.portalId, 10);
-  const oauthManager = setupOauth(accountId, configData);
+  const oauthManager = setupOauth(configData);
   logger.log('Authorizing');
   await authorize(oauthManager);
   addOauthToAccountConfig(oauthManager);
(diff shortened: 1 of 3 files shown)


packages/cli/lib/oauth.js
```diff
@@ -95,7 +95,7 @@ const authenticateWithOauth = async configData => {
   const oauthManager = setupOauth(accountId, configData);
   logger.log('Authorizing');
   await authorize(oauthManager);
-  addOauthToAccountConfig(accountId, oauthManager);
```
drewjenkins, line 98: Should we still be doing the portalId parse on line 94?

brandenrodgers:
Code lgtm 👍 

When running hs init locally, I noticed we still have this issue `[SUCCESS] The config file "/Users/brodgers/src/cms-devex-super-repo/hubspot-cli/hubspot.config.yml" was created using your personal access key for account undefined.` Is this `undefined` related to the portalId vs. accountId thing?

packages/cli/lib/oauth.js
miketalley: Good catch. Yeah we still need to do that. I moved it into `setupOauth` because that is where it's needed and it doesn't make sense to do the parsing in `authenticateWithOauth` anymore.

miketalley:
> Code lgtm 👍
> 
> When running hs init locally, I noticed we still have this issue `[SUCCESS] The config file "/Users/brodgers/src/cms-devex-super-repo/hubspot-cli/hubspot.config.yml" was created using your personal access key for account undefined.` Is this `undefined` related to the portalId vs. accountId thing?

Ah yes, that is a separate issue that I have a fix for in another branch. I'll pull that into this branch as it is related.

gcorne:
It would be good to have some test coverage if there is time since this is not the most common code path now that personal access keys are the recommended auth flow.

miketalley:
> It would be good to have some test coverage if there is time since this is not the most common code path now that personal access keys are the recommended auth flow.

Thank you for mentioning this! I actually brought this up in a meeting yesterday and didn't add it. I'll work on adding this.

miketalley:
I'm going to merge this to unblock @alnoorp and add the tests async.

PR #461, opened 2021-04-01
Init success message now outputs correct authType

miketalley:
## Description and Context
This was previously hard-coded to output "personal access key" even when setting up an "oauth2" auth.

### DIFF
--- a/packages/cli/commands/init.js
+++ b/packages/cli/commands/init.js
@@ -109,7 +109,7 @@ exports.handler = async options => {
     const path = getConfigPath();
 
     logger.success(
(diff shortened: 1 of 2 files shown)


packages/cli/commands/init.js
```diff
@@ -109,7 +109,7 @@ exports.handler = async options => {
     const path = getConfigPath();
 
     logger.success(
-      `The config file "${path}" was created using your personal access key for account ${name ||
+      `The config file "${path}" was created using ${authType} for account ${name ||
```
gcorne, line 112: Should we surround with double quotes since the `authType` is a token not a friendly name?

gcorne (approved):

Issue #27, opened 2019-12-17
Add `hs config` command for making changes to the config


gcorne:
In a fashion somewhat like `git config` and building on `hs init`, we should support making changes to the config via a `hs config` command.

corymartin:
Via @TheWebTech 

> passing thought: would be cool if CLI had a command to quickly change default portal.
>
>I have like 6 portals in my config file and everytime I switch portals to make my life easier I just change that default instead of specifying --portal every time
>
> Bonus points would be to have a way to list the nicknames in the config file, so to make this change command easier to use.

miketalley:
Two very good suggestions, I will add them when working on this!

TheWebTech:
The bonus list nicknames feature should also visually denote the currently selected nickname. (just thought of that as I currently swap portals, and forgot which I was currently defaulted to.)

PR #464, opened 2021-04-05
Added hs config commands

miketalley:
## Description and Context
#27 Finally getting around to this

Adds six new commands
- `hs config set default-account [accountNameOrId]` - selects portal to use as `defaultPortal` from a list or by specifying `accountNameOrId`
- `hs config set default-mode [mode]` - selects mode to use as `defaultMode` in config, also supports specifying `mode` as argument
- `hs config set http-timeout <timeout>` - sets `httpTimeout` in config
- `hs config set allow-usage-tracking` - selects value for `allowUsageTracking` in config
- `hs accounts list` (alias `hs accounts ls`) - lists portal `name`s, `portalId`s, and `authType`s in table format as well as `defaultPortal`
- `hs accounts rename <currentNameOrId> <newName>` - allows naming/renaming a portal `name` in the config (will also update `defaultPortal` if that is the one that is renamed) 

## Screenshots

`hs config set default-account`
image

`hs config set default-mode`
image

`hs config set http-timeout`
image

`hs config set allow-usage-tracking`
image

`hs accounts list`
image

`hs accounts rename`
image

### Notes:
- Fixed a bug where if `defaultPortal` was a number `getAccountId` was returning `null`
- Fixed table output to be borderless see https://www.npmjs.com/package/table#user-content-borderless-table


miketalley:
@abelb do you have any suggestions on this?

packages/cli-lib/lib/config.js
```diff
@@ -539,6 +543,39 @@ const updateDefaultAccount = defaultAccount => {
...
+ */
+const renameAccount = async (currentName, newName) => {
+  if (!newName) {
+    throw new Error('Please provide a new name for the portal');
+  }
+
+  if (!currentName) {
+    throw new Error(
+      'Please provide the portal name or id that should be renamed'
+    );
```
gcorne, line 557: These should be updated to use "account" instead of "portal."

packages/cli/commands/config.js
```diff
@@ -0,0 +1,28 @@
...
+} = require('../lib/commonOpts');
+const list = require('./config/list');
+const setDefault = require('./config/default');
+const rename = require('./config/rename');
+
+exports.command = 'config';
+exports.describe = 'Commands for working with the config file';
+
+exports.builder = yargs => {
+  addOverwriteOptions(yargs, true);
```
gcorne, line 14: This should be removed.

packages/cli/commands/config/default.js
```diff
@@ -0,0 +1,71 @@
...
+  loadConfig(configPath, options);
+  checkAndWarnGitInclusion();
+
+  if (!(validateConfig() && (await validateAccount(options)))) {
+    process.exit(1);
+  }
+};
+
+exports.command = 'default';
+exports.describe = 'Change default portal used in config';
```
gcorne, line 37: Rename portal to account.

packages/cli/commands/config/default.js
```diff
@@ -0,0 +1,71 @@
...
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('functions-default', {}, accountId);
```
gcorne, line 44: The usage tracking name used needs to be changed.

packages/cli/commands/config/default.js
```diff
@@ -0,0 +1,71 @@
...
+
+  const config = getConfig();
+  const { default: newDefault } = await inquirer.prompt([
+    {
+      type: 'list',
+      look: false,
+      name: 'default',
+      pageSize: 20,
+      message: 'Select a portal to use as the defaultPortal',
```
gcorne, line 53: Change "portal" to "account."

packages/cli/commands/config/default.js
```diff
@@ -0,0 +1,71 @@
...
+  ]);
+  updateDefaultAccount(newDefault);
+
+  return logger.log(`Default portal updated to: ${newDefault}`);
+};
+
+exports.builder = yargs => {
+  addConfigOptions(yargs, true);
+  addAccountOptions(yargs, true);
+  addUseEnvironmentOptions(yargs, true);
```
gcorne, line 66: I don't think the environment or account options are relevant when working with the config file.

packages/cli/commands/config/list.js
```diff
@@ -0,0 +1,63 @@
...
+  loadConfig(configPath, options);
+  checkAndWarnGitInclusion();
+
+  if (!(validateConfig() && (await validateAccount(options)))) {
+    process.exit(1);
+  }
+};
+
+exports.command = 'list';
+exports.describe = 'List names of portals defined in config';
```
gcorne, line 34: Change "portals" to "accounts."

packages/cli/commands/config/list.js
```diff
@@ -0,0 +1,63 @@
...
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('functions-list', {}, accountId);
```
gcorne, line 41: Update command.

packages/cli/commands/config/list.js
```diff
@@ -0,0 +1,63 @@
...
+
+  const config = getConfig();
+  const portalData = config.portals.reduce((acc, portal) => {
+    acc.push([portal.name, portal.portalId, portal.authType]);
+
+    return acc;
+  }, []);
+  portalData.unshift(getTableHeader(['Name', 'Portal ID', 'Auth Type']));
```
gcorne, line 49: Rename to "Account ID"

packages/cli/commands/config/list.js
```diff
@@ -0,0 +1,63 @@
...
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('functions-list', {}, accountId);
+
+  const config = getConfig();
+  const portalData = config.portals.reduce((acc, portal) => {
```
gcorne, line 44: Instead of `reduce`, it looks like `map` is a better fit.

packages/cli-lib/lib/config.js
```diff
@@ -420,20 +420,24 @@ const getAccountId = nameOrId => {
...
+      name = suppliedValue;
+    }
+  };
+
   if (!nameOrId) {
     const defaultAccount = getConfigDefaultAccount(config);
 
     if (defaultAccount) {
-      name = defaultAccount;
+      setNameOrAccountFromSuppliedValue(defaultAccount);
```
gcorne, line 437: I think it is worth adding test coverage for this change in behavior.

gcorne (changes requested):
From a UX perspective, I wonder if we should assume that the only actions that can be taken are to interact with the accounts in the config. I say this because there are other global config options that could have corresponding config commands for updating them, which makes me wonder if we should have commands like `hs config rename-account` and `hs config default-account` or similar.

I also didn't comment on all the places where `portal` is used instead of `account` in user-facing copy.

packages/cli/commands/config/list.js
miketalley: Good call, I was doing something different inside of this where `reduce` was better, but `map` is now better. Will update

packages/cli/commands/config/rename.js
```diff
@@ -0,0 +1,65 @@
...
+exports.command = 'rename <portalName> <newName>';
+exports.describe = 'Rename portal in config';
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { portalName, newName } = options;
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('functions-rename', {}, accountId);
```
anthmatic, line 41: -> `config-rename`?

packages/cli/commands/config/list.js
```diff
@@ -0,0 +1,63 @@
...
+    acc.push([portal.name, portal.portalId, portal.authType]);
```
anthmatic, line 45: Wondering if `authType` should be an enum?

packages/cli-lib/lib/config.js
```diff
@@ -539,6 +543,39 @@ const updateDefaultAccount = defaultAccount => {
   writeConfig();
 };
 
+/**
+ * @throws {Error}
+ */
+const renameAccount = async (currentName, newName) => {
+  if (!newName) {
+    throw new Error('Please provide a new name for the portal');
+  }
```
anthmatic, line 552: Not sure if I'm doing something wrong on my end but I wasn't able to hit this, I never got past yargs' argument check: `Not enough non-option arguments: got 1, need at least 2`

miketalley: Good call, I think we can remove this because yargs catches it.

packages/cli/commands/config/list.js
miketalley: We could definitely make it an `enum` though it's not elsewhere in the code. I think doing this may be out of scope for this PR though.

Is there something that you're worried about with not using an `enum` here?

packages/cli-lib/lib/config.js
miketalley: Good call. Added a test for this case.

packages/cli/commands/config.js
```diff
@@ -0,0 +1,23 @@
...
+
+exports.builder = yargs => {
+  addConfigOptions(yargs, true);
+  addAccountOptions(yargs, true);
+
+  yargs
+    .command({
+      ...list,
+      aliases: 'ls',
+    })
```
gcorne, line 17: Instead of adding the alias here, can we add to the command in `config/list`?

brandenrodgers:
When running `hs config default`, should we have special handling if there is only one available account to choose from?
image
In this scenario I only have one account so maybe it could instead show something like "You only have one account configured. Add more accounts to your hubspot.config.yml to see more options here". My phrasing there probably isn't the best, but I'm trying to think of ways to help guide the user to their intended result.

brandenrodgers:
I'm seeing an issue in the `hs config rename` command where it's not correctly logging the original account name:
image

packages/cli/commands/config/rename.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+};
+
+exports.command = 'rename <accountName> <newName>';
+exports.describe = 'Rename account in config';
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { portalName, newName } = options;
```
brandenrodgers, line 37: Looks like this should be `accountName` now. That will likely fix the issue I mentioned [here](https://github.com/HubSpot/hubspot-cli/pull/464#issuecomment-820577671).

packages/cli-lib/lib/config.js
```diff
@@ -539,6 +543,30 @@ const updateDefaultAccount = defaultAccount => {
...
+/**
+ * @throws {Error}
+ */
+const renameAccount = async (currentName, newName) => {
+  const accountId = getAccountId(currentName);
+  const accountConfigToRename = getAccountConfig(accountId);
+  const defaultAccount = getConfigDefaultAccount();
+
+  if (!accountConfigToRename) {
+    throw new Error(`Cannot find account with identifier ${currentName}`);
```
brandenrodgers, line 555: I wasn't able to hit this case in my testing. It was updating my account name regardless of what I entered for the currentName.

packages/cli/commands/config.js
miketalley: If we list aliases of subcommands in the parent command we don't have to sift through each subcommand to find every alias and avoid collisions. It also matches the current pattern for the `hs list` (`hs ls`) and `hs functions list` (`hs functions ls`).

packages/cli/commands/config/rename.js
miketalley: Fixed, good catch!

packages/cli-lib/lib/config.js
miketalley: This looks to have been fixed by a recent change.

packages/cli/commands/config/list.js
anthmatic: Ah, sorry I was thinking that we could validate that a correct authType exists but might not be a big deal

packages/cli/commands/config/list.js
```diff
@@ -0,0 +1,63 @@
...
+
+    return acc;
+  }, []);
+  portalData.unshift(getTableHeader(['Name', 'Portal ID', 'Auth Type']));
+
+  logger.log(getTableContents(portalData));
+  logger.log('Default Portal: ', config.defaultPortal);
```
anthmatic: I'm not sure if this is the best place for this, but it would be helpful to print out the path of the config file, which might be helpful for folks to know which config the CLI is actually loading

erincouse:
@gcorne can you give me an example of the global config commands that would conflict with these?

gcorne:
> @gcorne can you give me an example of the global config commands that would conflict with these?

Since `hs config` is being introduced here, there isn't a conflict. It is more of a design issue in that the structure of the commands don't match the structure of the configuration. The configuration consists of two sections: 1) global config options and 2) config for each account that the CLI has been set up to access.

Global configuration options:

1. httpTimeout - controls how long the CLI waits for network requests to the backend to complete. This is an advanced feature.
2. allowUserTracking - controls whether or not CLI usage activity is being tracked
3. defaultPortal - default account used when running commands without passing `--acount` option
4. defaultMode - determines whether files will automatically be overwritten or not when running `hs fetch`

Account configuration:

1. Name
2. Authentication type

jasminetudor:
@miketalley we are thinking...
**Great:**
`hs config list
``hs config rename
`
**but we would change:**

`hs config rename <currentNameOrId> <newName> 
`to 
`hs config set   -- default= <accountId>
`
to stick with the format of:
```
hs 
	command	noun (feature you need to use) 
	subcommand	verb (what you want to do w it)
	flags 	parameter (limits or specifics)

```

miketalley:
> @miketalley we are thinking...
> **Great:**
> `hs config list `
> `hs config rename `
> 
> **but we would change:**
> 
> `hs config rename <currentNameOrId> <newName> `
> to
> `hs config set --default= <accountId> `
> 
> to stick with the format of:
> 
> ```
> hs 
> 	command	noun (feature you need to use) 
> 	subcommand	verb (what you want to do w it)
> 	flags 	parameter (limits or specifics)
> ```

To clarify -- `hs config list` and `hs config rename` should remain as-is in the current PR and `hs config default` should be renamed to `hs config set` with 0 positional arguments and 1 option (`default`) that can be used to specify the new `defaultName` value?

For example:
`hs config set --default=accountNameToSetAsDefault` -- updates `defaultPortal` in config to `accountNameToSetAsDefault`
`hs config set` -- display error? display help? -- this is the part that seems a bit strange (see below)

Having 0 positional arguments for a command is a bit strange to me. I don't think that there are currently any other commands in the CLI that do not function unless an `--option` is passed. If `hs config default` were to be renamed to `hs config set default <newDefaultValue>` that would fit the current command paradigm better I think.

What are other people's thoughts?

miketalley:
After discussion with @erincouse and @jasminetudor on the `hs config default` command we decided to use `hs config set default [newDefaultAccountNameOrId]`.

It can be used to select from a list by running `hs config set default` or a `name` or `portalId` value can be specified as a positional argument -- `hs config set default MyAccountName`.

packages/cli/commands/config/list.js
miketalley: Good call, I updated the output from this command to include the config path and made the accounts list a bit nicer too
image

packages/serverless-dev-runtime/lib/server.js
```diff
@@ -11,6 +11,7 @@ const { validateInputs } = require('./validation');
 const { getValidatedFunctionData } = require('./data');
 const { setupRoutes, updateRoutePaths } = require('./routes');
 const { createTemporaryFunction, cleanupArtifacts } = require('./files');
+// TODO - Find a way to replace this with @hubspot/cli-lib/lib/table without causing a circular dependency
```
gcorne, line 14: It looks like `table.js` is in `packages/cli` not `@hubspot/cli-lib`. Is that what causes the circular dependency?

miketalley: Thanks @gcorne, I guess I missed this. I moved table into `cli-lib` and removed it from both `cli` and `serverless-dev-runtime`. Both the cli and runtime work great. I think this is a more appropriate place for it anyway.

gcorne:
Maybe it is just me, but I struggle with the expectations when it comes to `hs config list`. I expect it to list the current config values not just the list of accounts. I also feel like "default" needs a qualifier because the config also stores other defaults like the default mode used when running `hs fetch`.

miketalley:
> Maybe it is just me, but I struggle with the expectations when it comes to `hs config list`. I expect it to list the current config values not just the list of accounts. I also feel like "default" needs a qualifier because the config also stores other defaults like the default mode used when running `hs fetch`.

We can definitely expand the output to include `defaultMode`, `httpTimeout` and `allowUsageTracking` (not sure if I missed any others). Are you thinking we should also make the output friendly to pipe to other commands? I'm interested to know which use cases you'd like to cover.

I think changing the `hs config set default` command to `hs config set default-account` would be a good idea to avoid ambiguity. Adding `hs config set default-mode`, `hs config set http-timeout`, and `hs config set allow-usage-tracking` would also be good to completely flesh out the `hs config set` subcommands.

gcorne:
I think I am stuck on `hs config list` and `hs config rename` being operations on accounts in a config not the config itself. It makes me wonder if `hs accounts` might be a clearer top-level command for the account operations.

miketalley:
> I think I am stuck on `hs config list` and `hs config rename` being operations on accounts in a config not the config itself. It makes me wonder if `hs accounts` might be a clearer top-level command for the account operations.

I'd be fine with having this functionality nested as `hs accounts` instead. How do you see `hs config` working, then?

gcorne:
> How do you see hs config working, then?

I'd imagine that it would be used for the "global" config options that are not related to a specific account like `defaultMode`.

erincouse:
Small comment: I think `hs accounts rename` makes more sense as `hs config rename` (but understand that that may add complexity down the line, so not super hung up on it). Other is a general content nit: There's a lot of interchangeable use of `portal` and `account`, and it should really always be `account` as much as possible. It causes a lot of internal and external confusion because we aren't consistent with our own language.

abelb:
One idea is to maybe do this?
`hs config accounts rename <currentNameOrId> <newName>`
`hs config accounts list `

It makes the commands longer but it nests them underneath config and renaming them AND the list are all data that lives in config. 

Thoughts @jasminetudor @erincouse @miketalley?

> Small comment: I think `hs accounts rename` makes more sense as `hs config rename` (but understand that that may add complexity down the line, so not super hung up on it).

abelb:
Otherwise @miketalley everything looks good to me!

gcorne:
> Small comment: I think hs accounts rename makes more sense as hs config rename (but understand that that may add complexity down the line, so not super hung up on it).

@erincouse can you elaborate on why? Is your expectation that `hs accounts rename` would make changes to the name of the account inside HubSpot instead of the name that the developer chose when setting it up using `hs init` or `hs auth`? I think from an ergonomics perspective that `hs accounts rename` makes sense to me because all the other commands use the name as the value for `--account`. I also think we should consider deprecating `hs auth` in favor of `hs accounts add` since I think that is clearer. It will also make documenting how a developer can juggle multiple accounts (such as DEV, QA, PROD) via the CLI more straightforward.

gcorne:
In addition to `hs accounts add`, I could see `hs accounts set-default` and `hs accounts remove` being valuable to round things out.

miketalley:
I think there are two ways to organize the commands that are being thought about...

1. Commands that modify the `hubspot.config.yml` file are nested as `hs config <command>`
2. Commands that modify root level config values within `hubspot.config.yml` are `hs config <command>` and commands that modify the `portals` array values are `hs accounts <command>`

I can see good reasons for each. Having the `rename` command as `hs accounts rename` makes more sense to me because it doesn't leave any question about what is being renamed. It also leaves us the open possibility of having other `rename` commands to rename something else in the config in the future and nesting those separately. I can't think of a drawback to having multiple commands that can make changes to the config file values. 🤔 

That said, I think @gcorne's idea (and as @erincouse, @abelb, @jasminetudor, and I have discussed separately) of deprecating `hs auth` in favor of `hs accounts add`. I think that work should be done in a separate PR -- as well as potentially adding `hs accounts remove`.

PR #466, opened 2021-04-07
Moving hubspot-cli-tests repo into hubspot-cli repo

miketalley:
## Description and Context
@drewjenkins had previously created tests for the cli in the [hubspot-cli-tests repo](https://github.com/HubSpot/hubspot-cli-tests). This PR moves those tests into this repo with minor changes to allow them to run.


tests/README.md
```diff
@@ -0,0 +1,67 @@
+# HubSpot CLI E2E tests
+
+To assist in adding test coverage to the CLI, this project attempts to test `hs.*` commands as if they were being used by an end-user.
```
brandenrodgers, line 3: Can probably remove the "attempts to" portion of this

tests/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Getting Started
+
+The main test execution is kicked of via the `run-tests` script. `run-tests` itself is a CLI, so to see the available options, you can `run-tests --help`. This was done so that the tests can be installed against any given version of teh CLI and tested against it. In addition, it was done to keep the test logic separated from the CLI library
```
brandenrodgers, line 7: A few typos in here. "kicked off*", "version of the* CLI", and missing period at the end.

tests/README.md
```diff
@@ -0,0 +1,67 @@
...
+2. In `hubspot-cli`, `yarn link hubspot-cli-tests`
+3. To run it, `cd hubspot-cli`, `./node_modules/hubspot-cli-tests/run-tests`
+
+## Why Jasmine
+
+You may be wondering why we are using Jasmine, and not Jest. The reason is because Jest does not provide an API for executing the tests within code. We need to do this since we are wrapping the tests within a CLI.
+
+## Portals
+
+A couple of portals have been set up specifically for testing. They are QA Portal 105786502 and Prod Portal 9289088. If you add any test data to the portals for testing, ensure you add it to the above two.
```
brandenrodgers, line 60: I think we're using "account" instead of "portal"

tests/spec/tests/init.spec.js
```diff
@@ -0,0 +1,79 @@
...
+const rimraf = require('rimraf');
+const { existsSync, readFileSync } = require('fs');
+const yaml = require('js-yaml');
+
+describe('hs init', () => {
+  const { cli, config } = global;
+
+  it('should print the correct output for personal access key init', async () => {
+    rimraf.sync('hubspot.config.yml');
+    expect(existsSync('hubspot.config.yml')).toBe(false);
```
brandenrodgers, line 12: I don't have a great solution for this, but if either of these init tests fail, will that result in all the other tests also failing (since we're deleting the `hubspot.config.yaml` file)? Thinking about it a little more, maybe that's OK since it's such a core command?

tests/spec/tests/init.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+    expect(existsSync('hubspot.config.yml')).toBe(true);
+
+    expect(
+      yaml.load(readFileSync('hubspot.config.yml', 'utf8')).portals[0]
+        .personalAccessKey
+    ).toEqual(config.personalAccessKey);
+  });
+
+  if (
```
brandenrodgers, line 26: So we only conditionally run this test? Should we have some sort of console output to let the user know that it was skipped?

tests/spec/helpers/cmd.js
```diff
@@ -0,0 +1,169 @@
...
+  // Appending the process to the promise, in order to
+  // add additional parameters or behavior (such as IPC communication)
+  promise.attachedProcess = childProcess;
+
+  return promise;
+}
+
+module.exports = {
+  createProcess,
+  create: processPath => {
```
brandenrodgers, line 158: Could cut down on LOC here with `create: processPath => ({execute: (..args) => ...}),`

tests/run-tests
```diff
@@ -0,0 +1,84 @@
...
+    testRunner.addReporter(reporter);
+    testRunner.onComplete(function() {
+      rimraf.sync('hubspot.config.yml');
+    });
+    const cliPath = path.join(process.cwd(), '../packages/cli/bin/hs');
+    const cli = cmd.create(cliPath, '.');
+    global.cli = cli;
+    global.config = getTestConfig();
+
+    rimraf.sync('hubspot.config.yml');
```
brandenrodgers, line 73: This could totally be an unnecessary abstraction (so feel free to ignore), but it looks like we call `rimraf.sync('hubspot.config.yml');` pretty frequently. Should we have some sort of `removeConfig()` helper? Seems like we might also benefit from `readConfig` and `configExists` helpers.

brandenrodgers:
I left a few comments for some small changes, but this is super cool! Is the plan to introduce the pattern in this PR, and then later on we can expand test coverage?

Also, how will we be running these? Is it something that we should be running before we merge anything OR can we set them up to automatically run on code changes?

tests/spec/tests/init.spec.js
miketalley: I ended up breaking this file up into `init-personal-access-key.spec.js` and `init-oauth2.spec.js`. Also the tests assume that you have valid values for both auth flows.

I think although this is a bit more cumbersome, it will better to allow all tests to run.

miketalley:
> I left a few comments for some small changes, but this is super cool! Is the plan to introduce the pattern in this PR, and then later on we can expand test coverage?
> 
> Also, how will we be running these? Is it something that we should be running before we merge anything OR can we set them up to automatically run on code changes?

This is mostly @drewjenkins doing, so he deserves most of the credit here!

I got involved after we noticed a regression in the `oauth2` init flow ([Fixed in #460](https://github.com/HubSpot/hubspot-cli/pull/460)) and more specifically [@gcorne's comment suggesting adding test coverage](https://github.com/HubSpot/hubspot-cli/pull/460#issuecomment-811282848) after which I added `oauth2` tests to `hubspot-cli-tests` in [PR #1](https://github.com/HubSpot/hubspot-cli-tests/pull/1).

We had mentioned moving the `hubspot-cli-tests` into this repo, so that is where this PR comes in. My goal for this initial PR is to move as much of the code from `hubspot-cli-tests` as-is into this repo, and then subsequent PRs can add to or improve the current functionality.

I've had to make some modifications to get the tests to run in this repo.

tests/README.md
```diff
@@ -0,0 +1,67 @@
+# HubSpot CLI E2E tests
```
gcorne, line 1: Can we rename this directory `acceptance-tests` so that it is clearer?

tests/README.md
```diff
@@ -0,0 +1,67 @@
...
+run-tests --personalAccessKey="AiRiNGU2Y***************m1wLi2s8k2UlMYHEX"
+```
+
+4. If you need to override any environment variables at a test variable, you can do that via the `setLocalTestOverrides` function available in `env.js`
+
+The priority is Local Test Overrides > CLI Arg Overrides > Environment Variables
+
+### Running Locally
+
+This repository should be added to the `hubspot-cli` repo, root folder, as a dependency. You can accomplish this locally by utilizing yarn linking.
```
gcorne, line 48: This looks to be outdated.

tests/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. In `hubspot-cli-tests`, `yarn link`
+2. In `hubspot-cli`, `yarn link hubspot-cli-tests`
+3. To run it, `cd hubspot-cli`, `./node_modules/hubspot-cli-tests/run-tests`
```
gcorne, line 52: Can we add a script to the top-level `package.json` that is used to run the tests?

tests/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "hubspot-cli-tests",
+  "version": "1.0.0",
+  "description": "",
+  "main": "run-tests.js",
+  "scripts": {
+    "test": "node run-tests",
+    "lint": "eslint .",
+    "prettier": "prettier --write './*.{js,json}'"
+  },
```
gcorne, line 10: Linting and prettier should be handled by the top-level package. We also should consider incorporating with `lerna` so that stuff "just works" when working locally.

tests/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "hubspot-cli-tests",
```
gcorne, line 2: Can we mark `private: true` to avoid accidentally publishing to NPM?

packages/acceptance-tests/README.md
```diff
@@ -0,0 +1,64 @@
+# HubSpot CLI Acceptance Tests
```
gcorne, line 1: FWIW, it is possible to configure `lerna` to find packages outside `packages` which might be a good move for `acceptance-tests` since it is not public and has a different purpose than the packages in `packages/*`.

package.json
```diff
@@ -23,6 +23,7 @@
     "publish-release": "yarn lerna publish --conventional-graduate",
     "publish-prerelease": "yarn lerna publish prerelease --preid beta --dist-tag next",
     "test": "jest",
+    "test-cli": "node ./packages/acceptance-tests/run-tests",
```
gcorne, line 26: Might want to use [lerna run](https://github.com/lerna/lerna/tree/main/commands/run#readme) here.

packages/acceptance-tests/README.md
miketalley: I did [set the package as private](https://github.com/HubSpot/hubspot-cli/pull/466/files#diff-96d7c7c218525b55bef08e7648ed7b79f8e161db3c88763cbff609f9935a2208R5) so it won't be published. Do we want to also move the package elsewhere? If so, where do you want it moved?

gcorne: I think it would be better at the top-level as `acceptance-tests` and added to the `packages` section of the [lerna config](https://github.com/lerna/lerna#lernajson) as separate entry without a glob.

gcorne: This would have the added benefit that `jest` would be scoped to `packages` and the jasmine usage would be specific to the acceptance-tests.

tests/spec/tests/init.spec.js
miketalley: I moved the `rimraf` to the `beforeAll` method, but the rest of the tests will all still fail if the actual `hs init` command test fails. I think this is fine as that is the core of what we are testing here.

packages/acceptance-tests/README.md
miketalley: I moved the `acceptance-tests` folder out of the `packages` directory and restored the jest config package settings to look at `packages/*` again.

.eslintrc.js
```diff
@@ -17,7 +17,11 @@ module.exports = {
 
   overrides: [
     {
-      files: ['**/__tests__/**/*.js', '**/__mocks__/**/*.js'],
+      files: [
+        '**/__tests__/**/*.js',
+        '**/__mocks__/**/*.js',
+        'acceptance-tests/tests/**/*.js',
```
gcorne, line 23: Since these overrides are for jest, is this what we want?

.github/workflows/build.yml
```diff
@@ -5,11 +5,21 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v1
+    - name: Failure Slack Report
+      if: always()
+      uses: ravsamhq/notify-slack-action@master
+      with:
+        status: ${{ job.status }}
+        notify_when: 'failure'
```
gcorne, line 13: Will this notify when running in branches?

acceptance-tests/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "run-tests": "./run-tests"
+  },
+  "bugs": {
+    "url": "https://github.com/HubSpot/hubspot-cli/issues"
+  },
+  "devDependencies": {
+    "concat-stream": "^2.0.0",
+    "cross-spawn": "^7.0.3",
+    "dotenv": "^8.2.0",
+    "eslint": "^7.23.0",
```
gcorne, line 31: This should be covered by main `eslint` in the `package.json` at the root of the repo so should be removed.

packages/cli-lib/lib/config.js
```diff
@@ -364,7 +364,7 @@ const isTrackingAllowed = () => {
 
 const getAndLoadConfigIfNeeded = (options = {}) => {
   if (!_config) {
-    loadConfig(null, {
```
gcorne, line 367: Are we sure this change will not cause a regression?

packages/cli/commands/init.js
```diff
@@ -101,15 +103,18 @@ exports.handler = async options => {
 
   trackAuthAction('init', authType, TRACKING_STATUS.STARTED);
 
+  if (c) {
+    setConfigPath(path.join(process.cwd(), c));
```
gcorne, line 107: Instead of calling `setConfigPath()` here, what if `createEmptyConfigFile` accepted a path? I think we also should use the `getCwd()` helper instead of `process.cwd()` so that we're consistent when it comes to how we're resolving CWD.

packages/cli/commands/init.js
```diff
@@ -82,7 +84,7 @@ exports.command = 'init';
 exports.describe = `initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} for a HubSpot account`;
 
 exports.handler = async options => {
-  const { auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value } = options;
+  const { auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value, c } = options;
   const configPath = getConfigPath();
```
gcorne, line 88: What if the path was passed in to `getConfigPath()` like we're doing elsewhere instead of adding `_configPath` check to `getConfigPath()`?

gcorne (commented):
Looking good. Left another round of feedback. One thing that I think should be done is to add  unit test coverage for any changes to the `lib/config` since that code is getting hard to follow and is central to everything.

.github/workflows/build.yml
miketalley: Updated the if block so it should only run on `master` (https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#determining-when-to-use-contexts).

packages/cli/commands/init.js
miketalley: If we pass the path to `getConfigPath`, the error `The config file '${configPath}' already exists.` would get logged and the process would exit.

I think it makes sense to look for `_configPath` in `getConfigPath` so that non-standard configs are taken into account. I added a failing test to demonstrate the current behavior, which seems incorrect to me.

image

gcorne: Makes sense. Perhaps a check against the file existing is the way to go via `fs.existsSync()` instead of the code making the assumption that getting the path is looking for a file under the hood.

miketalley: I'm not seeing where I'm making that assumption. Does getting the set path need to know if there is a file or not?

I don't think we need to add that file check into `getConfigPath`.  It seems to work well without being tied to the file existing.

gcorne: The conditional on line 94 is intended to check if there is a config file at the path and if so to exit with an error. This was working because behind the scenes `getConfigPath()` was looking for the existence of a file instead of the code being written as an explicit check. The new logic would be more accurate and consistent if it was checking for the existence of the file in all cases.

.eslintrc.js
miketalley: I moved this into a specific override for jasmine

acceptance-tests/package.json
miketalley: Removed

packages/cli-lib/lib/config.js
miketalley: I changed this back just in case -- I didn't see any regressions in my testing though.

packages/cli/commands/init.js
miketalley: Got it -- I added a check for `configFileExists()` prior to returning the `_configPath`

packages/cli/commands/init.js
miketalley: I removed this code.

miketalley:
> Looking good. Left another round of feedback. One thing that I think should be done is to add unit test coverage for any changes to the `lib/config` since that code is getting hard to follow and is central to everything.

I added tests for `getConfigPath` and `createBlankConfigFile`.

packages/cli/commands/init.js
```diff
@@ -91,7 +91,7 @@ exports.handler = async options => {
   });
   const env = options.qa ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
 
-  if (configPath) {
+  if ((!c && configPath) || (c && configPath === c)) {
```
gcorne, line 94: Since the intention here is to exit the command early in cases where a config file already exists, could we switch this to `fs.existsSync(configPath)`?

packages/cli/commands/init.js
```diff
@@ -82,16 +84,16 @@ exports.command = 'init';
 exports.describe = `initialize ${DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME} for a HubSpot account`;
 
 exports.handler = async options => {
-  const { auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value } = options;
-  const configPath = getConfigPath();
+  const { auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value, c } = options;
+  const configPath = (c && `${getCwd()}/${c}`) || getConfigPath();
```
gcorne, line 88: I think that ``${getCwd()}/${c}`` should be using `path.join()` so that it works as expected in windows.

gcorne (approved):
One change to make sure things work in windows otherwise LGTM. 🧪 🎉

PR #475, opened 2021-04-22
Implement new approach for custom object schemas behind a flag

gcorne:
## Description and Context

This PR starts the process of implementing a more robust approach for working with custom object schemas that incorporates property groups and associations.

It introduces a top-level config flag `useCustomObjectHubfile` that is used to determine whether to use the old or new approach.

## TODO

- [x] Implement flow for getting a schema as a file
- [ ] Adjust backend to provide more useful response so we can better log successes
- [ ] Automatically detect the shape of the file when determining what logic branch to use

## Who to Notify
@chrisbaldauf @drewjenkins @brandenrodgers

### DIFF
--- a/packages/cli-lib/http.js
+++ b/packages/cli-lib/http.js
@@ -1,6 +1,9 @@
+const path = require('path');
 const request = require('request');
 const requestPN = require('request-promise-native');
 const fs = require('fs-extra');
+const contentDisposition = require('content-disposition');
+
 const { getAccountConfig } = require('./lib/config');
 const { getRequestOptions } = require('./http/requestOptions');
 const { accessTokenForPersonalAccessKey } = require('./personalAccessKey');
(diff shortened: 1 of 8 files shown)


packages/cli-lib/http.js
```diff
@@ -147,9 +150,22 @@ const createGetRequestStream = ({ contentType }) => async (
         },
         json: false,
       });
-      req.on('error', reject);
+      req.on('error', error => {
```
drewjenkins, line 153: Curious on this change.  aren't the two ways equivalent?

packages/cli-lib/lib/config.js
```diff
@@ -142,9 +142,9 @@ const getOrderedConfig = unorderedConfig => {
     ...(defaultPortal && { defaultPortal }),
     defaultMode,
     httpTimeout,
-    allowsUsageTracking,
-    portals: portals.map(getOrderedAccount),
+    allowUsageTracking,
```
drewjenkins, line 145: 👍

packages/cli-lib/http.js
gcorne: I think this is a left over from some debugging. I'll go back to the terser and more direct version.

PR #476, opened 2021-04-23
Convert in-line string usage to a centralized location (Commands)

miketalley:
## Description and Context
Use a centralized location for storing strings instead of keeping them in-line. This will allow us to more easily make changes to the CLI text outputs and give us the ability to add translations.

This PR adds the `i18n` method (which we can rename if necessary) that takes a dot notation string that is used to access the corresponding value within `/lang/en.lyaml` using the dot notation properties provided. The second parameter is an options object that can be provided property values that will be interpolated into the string obtained from `en.lyaml` using `{{ variableName }}` syntax.

The `locale` is obtained from node using `Intl.DateTimeFormat().resolvedOptions().locale` and is then `.split('-')[0]` to identify the proper translation file for the `locale` being used when running the command.

### Scope of this PR
- Add translations for all CLI commands
- Add translations for prompts
- Add translations for commonOpts
- Adds helpers for styling text within `en.lyaml` (we can add more as necessary)
  - `bold` -> `{{#bold}}text to make bold{{/bold}}`
  - `yellow` -> `{{#yellow}}text to make yellow{{/yellow}}`

### Usage Examples
en.lyaml example content
```
en:
  cli:
    commands:
      init:
        description: "My custom description! Foo, {{ barValue }}, baz! Some {{#bold}}bold{{/bold}} text!"
```

i18n Usage
```
// Note: the "en" is not necessary because the locale will be automatically determined from the node locale
const myDescription = i18n('cli.commands.init.description', {
  data: {
    barValue: 'bar'
  }
});

console.log(myDescription); // My custom description! Foo, bar, baz! Some bold text! (with the word "bold" in bold)
```

i18n Missing Value Error Handling
```
const myPropValue = i18n('cli.commands.init.somePropThatDoesNotExist.');

console.log(myPropValue); // [Missing language data]: en.cli.commands.init.somePropThatDoesNotExist
```

I18n Invalid Prop Error Handling
```
const myDescription2 = i18n('cli.commands.invalidCommand.describe');

console.log(myDescription2); // Throws error: [ERROR] Unable to access language property: en.cli.commands.invalidCommand.describe. Failed to access prop "invalidCommand".
```

### TODO (subsequent PRs)
- Expand coverage to other libs and other packages
- De-dupe strings in `en.lyaml`
  - Do we want to move all `errors` out of individual commands and into a shared `errors` object?
- Set up translations and PR flow for getting new translations easily added to this repo
- Research package recognition (`cli`, `cli-lib`, `create-cms-project`, `serverless-dev-runtime`, `webpack-cms-plugins`) to `lang.js` to automatically obtain this and allow using separate `en.lyaml` file for each package  

### Who to Notify
@gcorne @erincouse


packages/cli-lib/lib/lang.js
```diff
@@ -0,0 +1,75 @@
...
+  const lookupProps = lookupDotNotation.split('.');
+  const missingTextData = `${MISSING_LANGUAGE_DATA_PREFIX}: ${lookupDotNotation}`;
+  let textValue = languageObj;
+
+  try {
+    lookupProps.forEach(prop => {
+      textValue = textValue[prop];
+    });
+  } catch (e) {
+    logger.error(`Unable to access language property: ${lookupDotNotation}`);
```
brandenrodgers, line 38: Could we be more specific here about which prop we failed on?

packages/cli-lib/lib/lang.js
```diff
@@ -0,0 +1,75 @@
...
+    return missingTextData;
+  }
+
+  if (!textValue) {
```
brandenrodgers, line 42: This is probably never going to happen, but this check will fail if the value is an empty string. idk if it's worth doing a typeof check here instead, but I figured I'd still bring it up.

packages/cli-lib/lib/lang.js
```diff
@@ -0,0 +1,75 @@
...
+
+  return shouldInterpolate
+    ? getInterpolatedValue(textValue, options.data)
+    : textValue;
+};
+
+loadLanguageFromYaml();
+
+module.exports = {
+  i18n,
```
brandenrodgers, line 74: This might be a dumb question, but is i18n a well known term?

packages/cli/commands/init.js
```diff
@@ -82,7 +83,9 @@ const CONFIG_CREATION_FLOWS = {
 };
 
 exports.command = 'init';
```
brandenrodgers, line 85: Would it make sense to translate the command itself?

brandenrodgers:
This is really cool! I left a few minor comments, but otherwise this lgtm! 🙌

packages/cli-lib/lib/lang.js
miketalley: I think this is okay. I can't think of a use case for having an empty string value in the `en.lyaml` file. 🤔

packages/cli-lib/lib/lang.js
miketalley: Good idea. I'll add this.

packages/cli-lib/lib/lang.js
miketalley: I think it's a commonly accepted abbreviation for [internationalization and localization](https://en.wikipedia.org/wiki/Internationalization_and_localization).

packages/cli/commands/init.js
miketalley: I'm not sure the translations will work because we are using abbreviations in some cases. I'm not sure that `init`, short for `initialize`, has an equivalent translation in all languages. 🤔

packages/cli-lib/lib/config.js
```diff
@@ -301,7 +301,7 @@ const parseConfig = configSource => {
     return { parsed, error };
   }
   try {
-    parsed = yaml.safeLoad(configSource);
+    parsed = yaml.load(configSource);
```
gcorne, line 304: Why the move against `safe*`? I think it is best to be restrictive when it comes to serialization / deserialization of yaml since it can be used for more than just data if opened up.

packages/cli-lib/package.json
```diff
@@ -16,8 +16,9 @@
     "extract-zip": "^1.6.7",
     "findup-sync": "^3.0.0",
     "fs-extra": "^8.1.0",
+    "handlebars": "^4.7.7",
```
gcorne, line 19: Are we sure we want to introduce handlebars into the project?

packages/cli-lib/lib/lang.js
```diff
@@ -0,0 +1,79 @@
...
+const i18n = (lookupDotNotation, options = {}) => {
+  if (typeof lookupDotNotation !== 'string') {
+    throw new Error(
+      `i18n must be passed a string value for lookupDotNotation, received ${typeof lookupDotNotation}`
+    );
+  }
+
+  const textValue = getTextValue(lookupDotNotation);
+  const shouldInterpolate =
+    options.data && !textValue.startsWith(MISSING_LANGUAGE_DATA_PREFIX);
```
gcorne, line 68: I think `tokens` might be a clearer name than `data`.

packages/cli-lib/package.json
brandenrodgers: Could be worth exploring implementing our own simple version of `handlebars.compile` to avoid the extra package. Here's a relatively simple example of what that could look like https://gomakethings.com/a-vanilla-js-alternative-to-handlebarsjs-and-mustachejs/

packages/cli-lib/lib/config.js
miketalley: I was getting a deprecation warning, so switched it over.

packages/cli-lib/lib/lang.js
miketalley: Thanks for pointing this out, I think `data` is too generic here. I thought I was using the property name that i18n was using, but now I can't verify that. I'm not sure why I nested values within `data` -- I think it may be better and more consistent with i18n to remove the nesting of the values under the `data` property all together.

packages/cli/commands/init.js
brandenrodgers: Very good points. I agree 👍

packages/cli-lib/package.json
brandenrodgers: > One reason handlebars (or similar) is interesting to me is because we've started creating more complex strings that include things like bold or colored text and hyperlinks.

+1 to this. I think it could be valuable to have the translations support bold, colored, links, etc... but I don't think it would be too difficult to spin up our own version of compile that includes that functionality. I didn't dig too deep into the custom implementation though so I could be wrong about that.

brandenrodgers:
The code lgtm 👍 

I'll run this locally and test out a bunch of the commands.

packages/cli/lib/commonOpts.js
```diff
@@ -6,56 +6,59 @@ const {
...
-    ? `read from ${modes}`
-    : write
-    ? `write to ${modes}`
-    : `${modes}`;
 
   return yargs.option('mode', {
     alias: 'm',
-    describe: help,
+    describe: i18n(
+      `${i18nKey}.options.modes.describe${
```
brandenrodgers, line 42: missing a period here after "describe".

Side note: the error logging for missing keys is very useful! I think I noticed a bug in it though. Here's the error I got that helped me find the missing period:
```
[ERROR] Unable to access language property: en.cli.lib.commonOpts.options.modes.describewrite. Failed to access prop "modes".
```

I think it should say: `failed to access prop "describewrite"`

packages/cli/commands/customObject.js
```diff
@@ -18,22 +20,24 @@ exports.builder = yargs => {
...
-  );
-  console.warn(
-    chalk.reset.yellow(
-      'The Custom Object CLI is currently in beta and is subject to change.'
+      `***************************** ${i18n(`${i18nKey}.warning`)} ${i18n(
+        `${i18nKey}.warning`
+      )} ${i18n(`${i18nKey}.warning`)} ****************************`
     )
   );
+  console.warn(chalk.reset.yellow(i18n(`${i18nKey}.betaMessage`)));
```
anthmatic, line 28: Not a blocker, but I do wonder if we should register a handlebars helper for color or yellow or something?

packages/cli/lib/commonOpts.js
miketalley: Thanks for testing and point this out!

It looks like there were two issues...
1. The missing period after `describe`
2. The prop in `en.lyaml` was `mode` not `modes`, which led to the other issue.

brandenrodgers: No prob! And it sounds like you can ignore my second point about the missing prop message since "modes" was the right thing for it to log.

packages/cli/commands/customObject.js
miketalley: Good idea. I added this, updated the usages of `chalk.reset.yellow` to using `{{#yellow}}<content>{{/yellow}}` in the `en.lyaml` and moved the custom helpers into their own js file.

packages/cli-lib/package.json
miketalley: It definitely makes things easier to be able to use the custom handlers that can be registered with handlebars. There are two currently, and I can see us making more. It allows offloading a lot of the string customization to the `en.lyaml` file, which is nice IMO.

I'm all for looking into writing our own code to achieve this, and the nice thing is we can update the `lang.js` file and it should propagate to all the `i18n` usages. I think it may be better to use handlebars initially and save this for a subsequent PR.

Any objections to going this route? @gcorne @brandenrodgers @anthmatic

miketalley: This has been resolved in https://github.com/HubSpot/hubspot-cli/pull/611

PR #479, opened 2021-04-29
Upload .JSON files last

### LINKED ISSUE #129: Upload json files that depend on other files last
Files like `themes.json` and `fields.json` should be uploaded after all other files.

anthmatic:
## Description and Context
Sorts out all JSON files except for files in modules and uploads them last

Fixes #129

### DIFF
--- a/packages/cli-lib/lib/uploadFolder.js
+++ b/packages/cli-lib/lib/uploadFolder.js
@@ -28,6 +28,7 @@ function getFilesByType(files) {
   const cssAndJsFiles = [];
   const otherFiles = [];
   const templateFiles = [];
+  const jsonFiles = [];
 
   files.forEach(file => {
     const parts = splitLocalPath(file);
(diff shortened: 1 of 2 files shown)


packages/cli-lib/lib/uploadFolder.js
```diff
@@ -40,12 +41,14 @@ function getFilesByType(files) {
       cssAndJsFiles.push(file);
     } else if (extension === 'html') {
       templateFiles.push(file);
+    } else if (extension === 'json') {
```
gcorne, line 44: I suspect that we want to be more specific here and match only `theme.json`, `fields.json` that is not inside `*.module` and `serverless.json`.

gcorne: On second thought, maybe that is covered by how module files are already uploaded first. Might be good to add test coverage of this either way.

anthmatic: Exactly, and I can't see a reason to treat `*.functions` files differently.

Adjusted test coverage to include json files

brandenrodgers (approved):

Issue #450, opened 2021-03-02
Better message when network error occurs


anthmatic:
## Description and Context
If a network issue occurs while running `hs fetch`, an error is shown for the specific file that fails but it still shows a success message: 
![image](https://user-images.githubusercontent.com/9009552/109702052-c0b40700-7b61-11eb-80ff-7033a80142f3.png)

The "Success" message might cause some confusion, leading a developer to think that the fetch did successfully download all requested files

## Expected behavior
We might want to show an error message and prompt the user to retry the fetch command

PR #481, opened 2021-04-29
Add error messaging for if a file fails to download

drewjenkins:
## Description and Context
Fixes https://github.com/HubSpot/hubspot-cli/issues/450

This was a bit tricky to repro, but I was able to do so by manually failing the request by manipulating my wifi via a script like

```js
const { exec } = require('child_process');

exec('rm -rf temp/images', () => {
  console.log('removed');

  console.log('starting fetch');
  exec(
    './node_modules/.bin/hs fetch modules temp/images',
    (err, stdout, stderr) => {
      console.log(stdout, stderr);
      console.log('completed fetch');
      exec('networksetup -setairportpower Wi-Fi on', () => {
        console.log('wifi turned on');
      });
    }
  );

  setTimeout(() => {
    console.log('turning off wifi');
    exec('networksetup -setairportpower Wi-Fi off', () => {
      console.log('wifi turned off');
    });
  }, 1300);
});
```

With this fix, we will now show an error message instead of a success message if any of the files fails to download

### DIFF
--- a/packages/cli-lib/fileMapper.js
+++ b/packages/cli-lib/fileMapper.js
@@ -312,10 +312,10 @@ async function writeFileMapperNode(input, node, filepath) {
   if (!node.folder) {
     try {
       await fetchAndWriteFileStream(input, node.path, filepath);
+      return;
     } catch (err) {
-      // Logging handled by handler
+      return false;
     }
-    return;
   }
   try {
     await fs.ensureDir(filepath);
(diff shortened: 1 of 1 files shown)


brandenrodgers (approved):
This lgtm 👍  Just two questions

Does `logFileSystemErrorInstance` not log the errors (in writeFileMapperNode)?
Also, could you post a screenshot of what the failed state looks like now?

PR #482, opened 2021-04-29
Added validation error for non-lowercased authType values

### LINKED ISSUE #83: Make portal authType value case insensitive
When manually modifying the config, it is easy for there to be confusion between the casing of the keys and values. Both `apikey` and `usertoken` are set up such that the key=>value pair for the secret uses camelCase and the value for the auth type is lowercase. If we make the value for the `authType` case insensitive it makes the configs a bit more forgiving. We should also validate the `authType` as part of config validation so that developers know when they have made a mistake when manually editing the config file.

cc @bgoodies

miketalley:
## Description and Context
Fixes #83 

When manually editing the config file and using non-lowercased values for `authType`, it is unclear why things are not working. Only an error:
```
[ERROR] The accountId 8245999 is missing authentication configuration
```
is displayed, which is not very helpful for pinpointing the problem.

This PR adds validation to check for this and display a more helpful error message.

Do we want to allow invalid values (non-lowercased) to be used also? I think having the more helpful error may be sufficient. What are your thoughts @gcorne ?

## Screenshots
image

### DIFF
--- a/packages/cli/lib/validation.js
+++ b/packages/cli/lib/validation.js
@@ -4,6 +4,12 @@ const {
   loadConfigFromEnvironment,
   Mode,
 } = require('@hubspot/cli-lib');
+const { getConfigPath } = require('@hubspot/cli-lib/lib/config');
+const {
+  API_KEY_AUTH_METHOD,
+  OAUTH_AUTH_METHOD,
+  PERSONAL_ACCESS_KEY_AUTH_METHOD,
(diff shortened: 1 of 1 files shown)


drewjenkins (approved):
👍 I was going back and forth on if we should just coerce it for them, but I think CLIs tend to enforce things strictly, so lgtm

brandenrodgers (approved):
I agree with Andrew 👍

Issue #244, opened 2020-06-26
Expose theme preview route when watching a theme folder


anthmatic:
When running `hs watch <theme> <dest>` we could display the theme previewer URL so that developers have a quick way to see preview. At this point, I think it would only apply for themes but still would be pretty useful.

ex:
```shell
Uploaded file "../event.functions/membership.js" to "theme_folder/event.functions/membership.js"
Completed uploading files in /Users/apizzurro/projects/project/src to theme_folder in 12345

To preview this theme visit:
https://app.hubspot.com/theme-previewer/<portal>/edit/<theme_folder>
```

@gcorne @williamspiro I'm curious what you think

gcorne:
I think adding URLs to the messaging of various commands is +💯  from me. Whenever possible, I think we should do it without requiring a special option/flag.

brandenrodgers:
Related: https://github.com/HubSpot/hubspot-cli/issues/355

drewjenkins:
Re-opening. Edge case was found.  Also just now seeing I didn't address Gregory's feedback.  

See https://hubspot.slack.com/archives/C01H99J2MNE/p1622817567111200 for details on edge case

chiragchadha1:
Going to close this one out. It's been open for almost 5 years since the last activity and we haven't heard about the edge case since. Feel free to re-open if you're still hitting issues with the preview URL on theme watches. Thanks!

PR #483, opened 2021-04-29
Alert user of Theme Preview path after they've updated a file within a theme

drewjenkins:
Fixes https://github.com/HubSpot/hubspot-cli/issues/244

## Description and Context
We want to give users a link to the theme preview when they update it

### DIFF
--- a/packages/cli-lib/lib/watch.js
+++ b/packages/cli-lib/lib/watch.js
@@ -3,6 +3,7 @@ const chokidar = require('chokidar');
 const { default: PQueue } = require('p-queue');
 
 const { logger } = require('../logger');
+const debounce = require('debounce');
 const {
   ApiErrorContext,
   logApiErrorInstance,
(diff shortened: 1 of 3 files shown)


packages/cli-lib/lib/watch.js
```diff
@@ -129,6 +152,7 @@ function watch(accountId, src, dest, { mode, remove, disableInitial, notify }) {
       mode,
     });
     triggerNotify(notify, 'Added', filePath, uploadPromise);
+    notifyOfThemePreview(filePath, accountId);
```
anthmatic, line 155: Does this notify for every file upload? If so, I think it might be better to just show the message once

drewjenkins: I debounce it for that reason.  I couldn't think of a great way to know if all files have finished uploading or not.  But since it is debounced for 1 second, it should only show once in most cases

drewjenkins: Having just tried it on a whole folder though, it is saying the message before the uploads are all done, will look further into it

anthmatic: I just stumbled on this https://github.com/SBoudrias/Inquirer.js#bottom-bar---inquireruibottombar which might be an interested way to show this

drewjenkins: Oh very cool.  I checked it out, but it seems to do about the same thing with our current setup.  I _think_ we'd have to swap over to their ui.log for all of the messages, which I wasn't sure if we'd want to do since we have our own logger already.

Either way I realized I was being dumb and can check queue.size and make sure it is 0 before displaying a message.  That + the debounce seems to work great

brandenrodgers:
Should we provide this behavior on upload too?

packages/cli-lib/lib/watch.js
```diff
@@ -16,11 +17,35 @@ const { upload, deleteFile } = require('../api/fileMapper');
...
+  if (pathParts.length < 2) return;
+  const themeFolder = pathParts[pathParts.length - 2];
+
+  const baseUrl = getHubSpotWebsiteOrigin(
+    getEnv() === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD
+  );
+
+  logger.log(`
+  To preview this theme, visit:
+  ${baseUrl}/theme-previewer/${accountId}/edit/${themeFolder}
```
brandenrodgers, line 44: Can we utilize some of the utils in links.js for this url?

drewjenkins: I dont think so, none of those have a concept of a dynamic URL.  They only insert portal id and base url

packages/cli-lib/lib/watch.js
```diff
@@ -33,10 +47,12 @@ function uploadFile(accountId, file, dest, options) {
 
   logger.debug('Attempting to upload file "%s" to "%s"', file, dest);
   const apiOptions = getFileMapperQueryValues(options);
+
   return queue.add(() => {
     return upload(accountId, file, dest, apiOptions)
       .then(() => {
         logger.log(`Uploaded file ${file} to ${dest}`);
+        notifyOfThemePreview(file, accountId);
```
gcorne, line 55: Is there a reason we're doing this in the watch library instead of the watch command? I ask because the original conception of the library package was that it could be used in other scenarios beyond just `@hubspot/cli`.

I also wonder if outputting as part of initial command output perhaps after the the uploading of the folder if `--initial-upload` is passed.

PR #488, opened 2021-04-30
Change console error message to reflect correct property "defaultAccount" to "defaultPortal"

kieranja:
## Description and Context
CLI messaging refers to defaultAccount - but the YAML property is actually defaultPortal


gcorne:
Thanks!

PR #490, opened 2021-04-30
 Add ability to retrieve and tail app function logs

gcorne:
## Description and Context

This PR adds support for accessing serverless app function logs through expanding the `hs logs` command. The new options will remain undocumented for now.

An example of the format is `hs logs --appFunction=weatherCard --app=/hello-world-app`

## TODO

- [ ] See if `yargs` supports having option work as both a positional and option. If so, add `--endpoint` so that there is symmetry
- [x] Add some unit test coverage

## Who to Notify
@miketalley

### DIFF
--- a/packages/cli/commands/logs.js
+++ b/packages/cli/commands/logs.js
@@ -1,5 +1,3 @@
-const readline = require('readline');
-const ora = require('ora');
 const {
   addAccountOptions,
   addConfigOptions,
@@ -25,33 +23,8 @@ const {
   getFunctionLogs,
   getLatestFunctionLog,
 } = require('@hubspot/cli-lib/api/results');
-const { base64EncodeString } = require('@hubspot/cli-lib/lib/encoding');
 const { validateAccount } = require('../lib/validation');
-
-const TAIL_DELAY = 5000;
-
-const makeSpinner = (functionPath, accountId) => {
-  return ora(
-    `Waiting for log entries for '${functionPath}' on account '${accountId}'.\n`
-  );
-};
-
-const makeTailCall = (accountId, functionId) => {
-  return async after => {
-    const latestLog = await getFunctionLogs(accountId, functionId, { after });
-    return latestLog;
-  };
-};
-
-const handleKeypressToExit = exit => {
-  readline.emitKeypressEvents(process.stdin);
-  process.stdin.setRawMode(true);
-  process.stdin.on('keypress', (str, key) => {
-    if (key && ((key.ctrl && key.name == 'c') || key.name === 'escape')) {
-      exit();
-    }
-  });
-};
+const { tailLogs } = require('../lib/serverlessLogs');
 
 const loadAndValidateOptions = async options => {
   setLogLevel(options);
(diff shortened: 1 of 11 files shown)


packages/cli/commands/logs.js
```diff
@@ -164,13 +92,74 @@ exports.handler = async options => {
...
+    const fetchLatest = () =>
+      getLatestAppFunctionLogs(accountId, functionName, appPath);
+
+    await tailLogs({
+      accountId,
+      compact,
+      spinner,
+      tailCall,
+      fetchLatest,
+    });
```
gcorne, line 115: One thing that I am not sure about is whether the indirection and passing in of functions is easy enough to follow.

packages/cli/commands/logs.js
```diff
@@ -142,16 +62,24 @@ exports.handler = async options => {
...
+    const spinner = ora(
+      `Waiting for log entries for '${functionPath}' on account '${accountId}'.\n`
+    );
+    const tailCall = after => getFunctionLogs(accountId, functionId, { after });
+    const fetchLatest = () => getLatestFunctionLog(accountId, functionId);
     await tailLogs({
-      functionId: functionResp.id,
-      functionPath,
       accountId,
-      accountName: options.portal,
```
miketalley, line 153: Do we not need this?

packages/cli/commands/logs.js
miketalley: I think it's clear from the function name.

packages/cli/commands/logs.js
```diff
@@ -164,13 +92,74 @@ exports.handler = async options => {
...
+exports.describe = 'get logs for a function';
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { latest, functionName } = options;
+
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('logs', { latest }, accountId);
```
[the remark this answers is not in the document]
miketalley: What about adding `appPath` and `functionName` to the tracking object along with `latest`?

packages/cli/lib/serverlessLogs.js
```diff
@@ -0,0 +1,71 @@
...
+  handleKeypressToExit(() => {
+    spinner.stop();
+    process.exit();
+  });
+  tail(initialAfter);
+};
+
+module.exports = {
+  tailLogs,
+};
```
miketalley, line 71: 👍

packages/cli/commands/logs.js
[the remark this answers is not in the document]
gcorne: > What about adding appPath and functionName to the tracking object along with latest?

I think the backend needs to support any event attributes that we add.

packages/cli/commands/logs.js
gcorne: If we do want to show the account name instead of the id, we do need to incorporate something like this in the new code. That said, I think the way to handle that is to get the account config using the id so that it works consistently even in cases where the `defaultPortal` is used. We also should be consistent across all commands in terms of how we refer to an account in the output.

gcorne:
@miketalley I made an adjustment to fix situations where a function or endpoint hasn't been executed so there weren't any logs. This also improves situations where a fetching the latest log fails due to some intermittent error.

PR #491, opened 2021-05-03
Update prompts.js

fndicu:
Added a space after the link to access API Key. The previous link was 404ing

### DIFF
--- a/packages/cli/lib/prompts.js
+++ b/packages/cli/lib/prompts.js
@@ -90,7 +90,7 @@ const ACCOUNT_NAME = {
 const ACCOUNT_API_KEY = {
   name: 'apiKey',
   message:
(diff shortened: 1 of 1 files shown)


atanasiuk-hubspot (approved):

packages/cli/lib/prompts.js
```diff
@@ -90,7 +90,7 @@ const ACCOUNT_NAME = {
 const ACCOUNT_API_KEY = {
   name: 'apiKey',
   message:
-    'Enter the API key for your account (found at https://app.hubspot.com/l/api-key):',
+    'Enter the API key for your account (found at https://app.hubspot.com/l/api-key ):',
```
gcorne, line 93: I wonder if there is a better way to fix this. Adding a space to the parentheses feels a bit off. @erincouse maybe you have a suggestion for how to reword?

PR #499, opened 2021-05-11
Set output width to 80% of terminal width

drewjenkins:
## Description and Context
Before the width of the output was hardcoded to a width that was pretty short.  It made some of our descriptions hard to read because there was a lot of text wrapping.  This makes the width dynamic to the width of your terminal.

## Screenshots
Before
image

After
image

## TODO
Does 80% seem good? Should we go wider?

### DIFF
--- a/packages/cli/bin/cli.js
+++ b/packages/cli/bin/cli.js
@@ -50,7 +50,6 @@ const getTerminalWidth = () => {
 };
 
 const argv = yargs
-
   .usage('Tools for working with HubSpot')
   .middleware([setLogLevel])
   .exitProcess(false)
--- a/packages/cli/bin/cli.js
+++ b/packages/cli/bin/cli.js
@@ -88,6 +88,7 @@ const argv = yargs
   .recommendCommands()
   .demandCommand(1, '')
   .completion()
+  .wrap(yargs.terminalWidth() * 0.8)
   .strict().argv;
 
 if (argv.help) {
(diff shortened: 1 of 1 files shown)


packages/cli/bin/cli.js
```diff
@@ -88,6 +88,7 @@ const argv = yargs
   .recommendCommands()
   .demandCommand(1, '')
   .completion()
+  .wrap(yargs.terminalWidth() * 0.8)
```
gcorne, line 91: @drewjenkins Instead of always making it 80%, I wonder if we should approach this sort of like response web design where we have breakpoints so that when a terminal is narrow we use the full-width and only switch to a percentage (or larger fixed width) at larger widths.

drewjenkins: Good call, fixed

packages/cli/bin/cli.js
```diff
@@ -41,6 +41,14 @@ notifier.notify({
   message: pkg.name === '@hubspot/cms-cli' ? CLI_UPGRADE_MESSAGE : null,
 });
 
+const getTerminalWidth = () => {
+  const width = yargs.terminalWidth();
+
+  if (width >= 100) return width * 0.9;
```
drewjenkins, line 47: 0.8 ended up seeming like a lot of whitespace, so I bumped it to .9

Also, the way the terminal's width is measured is in columns, that's why the width is so low

brandenrodgers (approved):
lgtm! 👍

PR #500, opened 2021-05-13
Initial README

drewjenkins:
## Description and Context
Starting to provide some documentation for cli-lib.  Starting with a high level README


TheWebTech (approved):
It looks good to me.

packages/cli-lib/README.md
```diff
@@ -0,0 +1,71 @@
...
+```js
+const {
+  getDirectoryContentsByPath,
+} = require('@hubspot/cli-lib/api/fileMapper');
+
+const accountId = YOUR_ACCOUNT_ID;
+
+getDirectoryContentsByPath(accountId, '/').then(response => {
+  console.log(response);
+});
```
gcorne, line 41: I think that the example should illustrate how to load the config since unless something has changed this will not work without the config being loaded first.

See https://github.com/HubSpot/hubspot-cli/blob/master/packages/webpack-cms-plugins/HubSpotAutoUploadPlugin.js#L18-L19

packages/cli-lib/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Framework-agnostic Integrations
+
+HubSpot's cli-lib can integrate with various task runners and bundlers such as Gulp, Grunt, and Webpack.
+
+Internally, the HubSpot cli-lib is consumed by the HubSpot CLI, the official HubSpot Webpack plugin, and the official HubSpot serverless development runtime.
+
+## Functionality
+
+The full scope of the cli-lib is quite large. It provides the ability to manage authentication, custom objects, the file manager, files within the Design Manager, serverless functions, HubDB, and more.
```
gcorne, line 13: I appreciate the idea of starting high-level and exploring the use cases, but I wonder if Instead of leaving the functionality open ended could we be more thorough in describing the different features and perhaps go as far as outlining the public interface (e.g. the modules + functions) so that we can be clearer when it comes to what the key pieces are for developers.

packages/cli-lib/examples/slack/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
gcorne, line 1: I think this is handled automatically by the top-level `.gitignore`

packages/cli-lib/examples/file-upload/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
gcorne, line 1: Should we move the examples to the top-level of `packages` so that the version is managed by `lerna` and linking works as expected?

packages/cli-lib/examples/slack/README.md
```diff
@@ -0,0 +1,30 @@
+## Slack Integration Example
```
gcorne, line 1: This is a pretty neat example, although I wonder if we would want to recommend doing this via a separate step in a GitHub action since it will be less fragile. We also could incorporate into https://github.com/HubSpot/hubspot-cms-deploy-action to make it easy for folks.

packages/cli-lib/README.md
drewjenkins: Yeah that's the next step.  This was just the initial overview step

drewjenkins: Before documenting the public interface, though, we need to figure out what we want in the public interface

packages/cli-lib/examples/file-upload/package.json
drewjenkins: Yeah I can definitely do that.  I can do something like

examples
--> cli-lib
-----> *

packages/cli-lib/examples/slack/.gitignore
drewjenkins: I had it in there in case someone wants to copy the code as a boilerplate, I can take it out thought

packages/cli-lib/examples/slack/README.md
drewjenkins: Yeah that's why I mentioned in the overview below it would be likely be invoked from a CI/CD platform such as Jenkins or Github.  Are you suggesting I build the example off of the deploy github action?

packages/cli-lib/examples/file-upload/package.json
gcorne: Proposed structure of nesting under top-level `examples` sounds +:100: to me.

packages/cli-lib/examples/slack/README.md
gcorne: I was thinking that we add Slack to the https://github.com/HubSpot/hubspot-cms-deploy-action with a mechanism to configure it, if possible. I felt like that while cool—this example was suggesting a less-than-ideal solution.

drewjenkins: I can write up something to add it to the deploy action

PR #509, opened 2021-05-21
Add readme for getAndLoadConfigIfNeeded

drewjenkins:
Add documentation for getAndLoadConfigIfNeeded


docs/cli-lib/api/getAndLoadConfigIfNeeded.md
```diff
@@ -0,0 +1,35 @@
+## `getAndLoadConfigIfNeeded(options)`
```
drewjenkins, line 1: You can view the formatted version [here](https://github.com/HubSpot/hubspot-cli/blob/23fc2afeca0ff8f2dabe8a32761c950537914494/docs/cli-lib/api/getAndLoadConfigIfNeeded.md)

docs/cli-lib/api/getAndLoadConfigIfNeeded.md
```diff
@@ -0,0 +1,35 @@
...
+
+[cli-lib/lib/config.js](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lib/config.js)
```
drewjenkins, line 3: What do folks think about this? At first, I had the link go directly to the function, but the problem with that is if we add code to that file or move anything around, it's going to change the line number that the function starts at, but the permalink will still point at the old line number.   I doubt we would remember (or want to) update every permalink for the file whenever we change it.

This seemed like a decent middle ground.  Linking to the file, so we'd only need to update permalinks if we add/remove/move the function from the file it was in to a different file

docs/cli-lib/api/getAndLoadConfigIfNeeded.md
drewjenkins, line 1: The formatting was heavily inspired by [Redux](https://github.com/reduxjs/redux/blob/master/docs/api/applyMiddleware.md#applymiddlewaremiddleware)

brandenrodgers:
This lgtm 👍  I mentioned this in Slack, but this is a good place to get some design input before merging.

Issue #507, opened 2021-05-20
Wrong message about "initial-upload"


SylvainTirreau:
## Description and Context

When I write this command: `$ hs watch --account=PREFPROD prefhub-prefprod/_Site_v2 _Site_v2 --initial-upload` I've got this message:

```
[INFO] The "watch" command no longer uploads the watched directory when started. The directory "prefhub-prefprod/_Site_v2" was not uploaded.
[INFO] To upload the directory run "hs upload" beforehand or add the "--initial-upload" option when running "hs watch".
```
Version 3.0.5

### Steps to reproduce
Just typing: `$ hs watch --account=[account] source dest --initial-upload`

## Expected behavior
As I write "--initial-upload", I don't have to see this message: "add the "--initial-upload" option when running "hs watch".".

PR #512, opened 2021-05-27
Don't tell users about initial-upload if they are already using initial upload

drewjenkins:
## Description and Context
https://github.com/HubSpot/hubspot-cli/issues/507


packages/cli/commands/watch.js
```diff
@@ -77,7 +77,7 @@ exports.handler = async options => {
     logger.info(
       'Passing the "--disable-initial" option is no longer necessary. Running "hs watch" no longer uploads the watched directory by default.'
     );
-  } else {
+  } else if (!initialUpload) {
```
brandenrodgers, line 80: This is also removing the message about the watch command. Should we continue to show that message even when the initial upload flag is passed?

Issue #364, opened 2020-10-17
add membership template type to hs create template


TheWebTech:
## Description and Context
membership templates currently can't be generated through the CLI. Would be a nice quality of life change.

## Expected behavior
Membership template types would display when running hs create template command. Would generate templates similar to the boilerplate.

chiragchadha1:
Thanks for the suggestion @TheWebTech. Going to close this one out. It's been open for 5 years with no further demand for membership template scaffolding via the CLI. Feel free to re-open if this is still useful for your workflow. Thanks!

PR #514, opened 2021-05-27
Cleanup create.js

drewjenkins:
## Description and Context
Started working on https://github.com/HubSpot/hubspot-cli/issues/364 and decided to break out the code a bit first.

## TODO
As part of this, I want to write some tests to cover the changes

## Who to Notify
@HubSpot/cms-developer-experience

### DIFF
--- a/acceptance-tests/run-tests
+++ b/acceptance-tests/run-tests
@@ -51,6 +51,11 @@ setArgsOverrides(
       default: false,
       description: 'Set this if you are using a app.hubspotqa.com site',
     })
+    .option('headless', {
+      type: 'boolean',
+      default: false,
+      description: 'Run the tests in headless mode',
+    })
     .option('debug', {
       type: 'boolean',
       default: false,
--- a/acceptance-tests/tests/helpers/cmd.js
+++ b/acceptance-tests/tests/helpers/cmd.js
@@ -54,11 +54,14 @@ function executeWithInput(processPath, args = [], inputs = [], opts = {}) {
     inputs = [];
   }
 
-  // if (global.config.qa) {
-  //   args.push('--qa');
-  // }
+  if (global.config.qa) {
+    args.push('--qa');
+  }
+  if (global.config.headless) {
+    opts.env = { BROWSER: 'none' };
+  }
 
-  const { env = null, timeout = 500, maxTimeout = 10000 } = opts;
+  const { env = opts.env, timeout = 500, maxTimeout = 10000 } = opts;
   const childProcess = createProcess(processPath, args, env);
   childProcess.stdin.setEncoding('utf-8');
 
@@ -127,7 +130,10 @@ function executeWithInput(processPath, args = [], inputs = [], opts = {}) {
         clearTimeout(currentInputTimeout);
         inputs = [];
       }
-      reject(err.toString());
+
+      // For some reason, when we use 'ora', it is throwing this error,
+      // so for now we won't reject.
+      console.error(err.toString());
     });
 
     childProcess.on('error', reject);
--- a/acceptance-tests/tests/helpers/cmd.js
+++ b/acceptance-tests/tests/helpers/cmd.js
@@ -133,7 +133,18 @@ function executeWithInput(processPath, args = [], inputs = [], opts = {}) {
 
       // For some reason, when we use 'ora', it is throwing this error,
       // so for now we won't reject.
-      console.error(err.toString());
+      const blacklistedStrings = ['Loading available API samples'];
+
+      const error = err.toString();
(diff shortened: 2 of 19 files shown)


brandenrodgers:
Code lgtm! I think this is a good abstraction.

I'm going to test locally before giving the 👍  though just b/c a lot has moved around.

brandenrodgers:
This seems to be happening on the current version of the CLI too, but create isn't putting the assets in the correct destination folder. If it's an easy fix, might be worth doing as a part of this work (not a blocker though).

I set `./tmp/createdAssets` as my destination folder, but it just put the theme in my CWD instead
image

brandenrodgers (approved):
Looks good! I mentioned an issue with the destination of the assets that are created, but I think it's working the same as the current version of the CLI so feel free to merge.

acceptance-tests/run-tests
```diff
@@ -51,6 +51,11 @@ setArgsOverrides(
       default: false,
       description: 'Set this if you are using a app.hubspotqa.com site',
     })
+    .option('headless', {
```
brandenrodgers, line 54: 🎉

acceptance-tests/tests/helpers/cmd.js
```diff
@@ -54,11 +54,14 @@ function executeWithInput(processPath, args = [], inputs = [], opts = {}) {
     inputs = [];
   }
 
-  // if (global.config.qa) {
-  //   args.push('--qa');
-  // }
+  if (global.config.qa) {
```
brandenrodgers, line 57: Any idea why this was previously commented out?

acceptance-tests/tests/helpers/cmd.js
```diff
@@ -127,7 +130,10 @@ function executeWithInput(processPath, args = [], inputs = [], opts = {}) {
         clearTimeout(currentInputTimeout);
         inputs = [];
       }
-      reject(err.toString());
+
+      // For some reason, when we use 'ora', it is throwing this error,
+      // so for now we won't reject.
+      console.error(err.toString());
```
brandenrodgers, line 136: Does this have an impact on any other tests for commands that aren't using ora?

drewjenkins: Not that I could tell, I'll do some extra testing

drewjenkins: Ok from what I could tell everything still works great

acceptance-tests/tests/helpers/cmd.js
brandenrodgers: Oh, I just found [this in the readme](https://github.com/HubSpot/hubspot-cli/blob/master/acceptance-tests/README.md#getting-started) and it seems relevant.

> Note that if you are testing against a QA portal, not a PROD one, you'll need to add the --qa flag when running the script. There is still an outstanding issue with this because we attempt to add the --qa flag to all hs commands, however it is not available for all commands.

acceptance-tests/tests/helpers/cmd.js
drewjenkins: I ended up changing it to use a blacklisted string approach, just in case.  More granular that way

acceptance-tests/tests/helpers/cmd.js
drewjenkins: Oh that makes sense.  Good call.  I think I was the one that wrote that commend 😅 I'll recomment it out

packages/cli/commands/create.js
```diff
@@ -1,145 +1,41 @@
...
+ * @export {(Data) => Boolean=} validate - If provided, return true if it passes validation, false otherwise.
+ *                                         If not provided, validation will automatically succeed
+ *
+ * The Data object contains
+ * {
+ *   assetType: String - Type of the asset (e.g. api-sample, react-app, template)
+ *   name:      String - Filename of the asset
+ *   dest:      String - The path specified by the user on where to create the asset
+ *   options:   Object - The options object passed to the command by Yargs
+ * }
```
anthmatic, line 21: 👏 👏 👏

packages/cli/lib/oauth.js
```diff
@@ -31,7 +31,9 @@ const handleServerOnProcessEnd = server => {
 };
 
 const authorize = async oauthManager => {
-  open(buildAuthUrl(oauthManager), { url: true });
+  if (process.env.BROWSER !== 'none') {
+    open(buildAuthUrl(oauthManager), { url: true });
+  }
```
anthmatic, line 37: Not entireley sure if it makes sense to, but I'm wondering if we should log the output here as we do with PAKs

anthmatic (approved):

PR #518, opened 2021-05-28
Granular template checks for marketplace theme validation

brandenrodgers:
## Description and Context

Using this PR as the base just to split up the work into separate PR's to make it more reviewable.

This adds more accurate template checking to the theme validation command. Now we're correctly checking for label and screenshot annotations depending on the template type. 

## Screenshots

image

## TODO

- [x] Add unit tests for the new template checks
- [ ] This still needs more testing. I ran it against a marketplace theme and it seems to work, but there are a lot of edge cases that I could be missing.
## Who to Notify

cc/ @KatzMitch

### DIFF
--- a/packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js
+++ b/packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js
@@ -10,6 +10,7 @@ const BaseValidator = require('../BaseValidator');
 const { logger } = require('@hubspot/cli-lib/logger');
 
 const TEMPLATE_LIMIT = 50;
+const TEMPLATE_COUNT_IGNORE_LIST = ['global_partial', 'none'];
 const VALIDATIONS_BY_TYPE = {
   page: { allowed: true, label: true, screenshot: true },
   starter_landing_pages: { allowed: false },
@@ -102,8 +103,8 @@ class TemplateValidator extends BaseValidator {
             ANNOTATION_KEYS.screenshotPath
           );
 
-          // Exclude global partials and templates with type of none in count
-          if (!['global_partial', 'none'].includes(templateType)) {
+          // Ignore global partials and templates with type of none in count
+          if (!TEMPLATE_COUNT_IGNORE_LIST.includes(templateType)) {
             templateCount++;
           }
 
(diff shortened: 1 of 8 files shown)


packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js
```diff
@@ -4,11 +4,53 @@ const {
...
   getFileAnnotations,
-  isTemplate,
+  isCodedFile,
 } = require('@hubspot/cli-lib/templates');
 const BaseValidator = require('../BaseValidator');
+const { logger } = require('@hubspot/cli-lib/logger');
 
 const TEMPLATE_LIMIT = 50;
+const TEMPLATE_COUNT_IGNORE_LIST = ['global_partial', 'none'];
+const VALIDATIONS_BY_TYPE = {
```
drewjenkins, line 14: Very cool

packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js
```diff
@@ -35,45 +82,85 @@ class TemplateValidator extends BaseValidator {
...
+              )}`
+            );
+          }
+        } else {
+          logger.debug(
+            `No template type found for ${path.relative(
+              absoluteThemePath,
+              filePath
+            )}`
+          );
```
anthmatic, line 150: Should these also be validation errors?

brandenrodgers: I'm not sure, but maybe @KatzMitch will know. In this scenario the html file will have to have a `.html` extension, `isAvailableforNewContent !== false`, and no templateType annotation. 

somewhat related, I'm currently doing nothing for html files that have no annotations at all and I'm not sure if I should be showing some sort of validation message for those.

KatzMitch: Hmm since it's just specific to the marketplace, I think we should require that all HTML files have annotations and throw an error if they don't (cc: @ckconant )

ckconant: Yeah, I agree with @KatzMitch that all html files should have annotations. The only file that has limited annotations is the `base.html` layout in [boilerplate](https://github.com/HubSpot/cms-theme-boilerplate/blob/main/src/templates/layouts/base.html#L2), which only has `templateType: none` annotated.

brandenrodgers: OK, I'll add an additional error type for this 👍

brandenrodgers: added in a8db57d

PR #521, opened 2021-06-08
Fixing bugs in the WIP theme marketplace validation command

brandenrodgers:
## Description and Context

This fixes three bugs:
- Very long outputs were getting truncated. This is b/c I was using console.log to output the json result. I switched over to `process.stdout.write` which seems to work better.
- It's possible to hit the rate limit and we were not handling that gracefully. I added a new error type for when the validate request fails. It's called `failedDepFetch` (open to better naming and/or error handling suggestions here).
- We were incorrectly validating against files in the `_locales` folders for modules. I'm now ignoring files in those folders.

## Who to Notify

cc/ @KatzMitch

### DIFF
--- a/packages/cli-lib/modules.js
+++ b/packages/cli-lib/modules.js
@@ -47,7 +47,7 @@ const isModuleFolder = pathInput => {
  * @returns {boolean}
  * @throws {TypeError}
  */
-const isModuleFolderChild = pathInput => {
+const isModuleFolderChild = (pathInput, ignoreLocales = false) => {
   throwInvalidPathInput(pathInput);
   let pathParts = [];
   if (pathInput.isLocal) {
@@ -58,6 +58,10 @@ const isModuleFolderChild = pathInput => {
   const { length } = pathParts;
   // Not a child path?
   if (length <= 1) return false;
+  // Check if we should ignore this file
+  if (ignoreLocales && pathParts.find(part => part === '_locales')) {
+    return false;
+  }
   // Check if any parent folders are module folders.
   return pathParts
     .slice(0, length - 1)
--- a/packages/cli/lib/validators/marketplaceValidators/theme/DependencyValidator.js
+++ b/packages/cli/lib/validators/marketplaceValidators/theme/DependencyValidator.js
@@ -89,10 +89,13 @@ class DependencyValidator extends BaseValidator {
   isExternalDep(file, relativeDepPath) {
     // Get dir of file that references the dep
     const { dir } = path.parse(file);
+    console.log('1: ', dir);
     // Use dir to get the dep's absolute path
     const absoluteDepPath = path.resolve(dir, relativeDepPath);
+    console.log('2: ', absoluteDepPath);
     // Get relative path to dep using theme absolute path and dep absolute path
     const relativePath = this.getRelativePath(absoluteDepPath);
+    console.log('3: ', relativePath);
     // Check that dep is not within the theme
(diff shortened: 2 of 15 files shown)


packages/cli/lib/validators/marketplaceValidators/theme/DependencyValidator.js
```diff
@@ -46,45 +51,65 @@ class DependencyValidator extends BaseValidator {
...
-            return { file, deps: {} };
+            return { file, deps };
           }
-          const validation = await validateHubl(accountId, source);
-          const deps = getDepsFromHublValidationObject(validation);
-          deps.RENDERING_ERROR_PATHS = this.getPathsFromRenderingErrors(
-            validation
+          const file_deps = await fetchDependencies(accountId, source).catch(
+            err => {
+              this.failedToFetchDependencies(err, file, validationErrors);
```
drewjenkins, line 74: We don't need to do it now, but if we are getting rate limited, usually something like a `rate-limit-expiration` type of property is in the headers.  We should probably write a fetch wrapper that will wait and try again when that happens, instead of just failing.  It could try like twice or however many times, and then fail

packages/cli/lib/validators/marketplaceValidators/theme/DependencyValidator.js
```diff
@@ -46,45 +51,65 @@ class DependencyValidator extends BaseValidator {
...
           return { file, deps };
         })
     );
   }
 
-  isExternalDep(absoluteThemePath, file, relativeDepPath) {
+  isExternalDep(file, relativeDepPath) {
     // Get dir of file that references the dep
     const { dir } = path.parse(file);
+    console.log('1: ', dir);
```
drewjenkins, line 92: guessing these need removed

packages/cli-lib/modules.js
```diff
@@ -58,6 +58,10 @@ const isModuleFolderChild = pathInput => {
   const { length } = pathParts;
   // Not a child path?
   if (length <= 1) return false;
+  // Check if we should ignore this file
+  if (ignoreLocales && pathParts.find(part => part === '_locales')) {
```
drewjenkins, line 62: Are there other folders/files we should be ignoring?

packages/cli/lib/validators/marketplaceValidators/theme/DependencyValidator.js
brandenrodgers: yep 🤦  lol

packages/cli-lib/modules.js
brandenrodgers: I asked Mitchell about that yesterday, and it sounds like this is the only one we should ignore.

packages/cli/lib/validators/marketplaceValidators/theme/DependencyValidator.js
brandenrodgers: Yeah that's a good point. The reason we updated the endpoint instead was to prevent users from exploiting the validation endpoint. We have a new endpoint now that shouldn't run into the rate limit issue (b/c this endpoint will ignore rate limits), and it only returns the deps.

gcorne: Haven't looked closely at the code, another option may br to use the queuing approach using `p-queue` that is used elsewhere (upload, watch, and fetch) to limit concurrency.

PR #525, opened 2021-06-15
updated Autoupload plugin README to support Webpack 5 and fix --env vars

emptypockets-dev:
## Description and Context
The setup for the latest version of `copy-webpack-plugin` requires paths to live inside of a `patterns` array. Documentation here: https://www.npmjs.com/package/copy-webpack-plugin

Also, passing environment vars to webpack doesn't work with the dot notation. The script should be `webpack --watch --env account 123 --env autoupload`. Documentation here: https://webpack.js.org/guides/environment-variables/

As a side note, the Hubspot.config.yml file always provides the account so that part isn't really needed (might be better to manage that all in one place—can the autoupload option be moved there?) 

## Who to Notify
@TheWebTech

### DIFF
--- a/packages/webpack-cms-plugins/README.md
+++ b/packages/webpack-cms-plugins/README.md
@@ -48,13 +48,15 @@ module.exports = ({ account, autoupload }) => ({
       src: 'dist',
       dest: 'my-project',
     }),
-    new CopyWebpackPlugin([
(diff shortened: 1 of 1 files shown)


brandenrodgers (approved):
LGTM, thanks for updating this 👍

PR #526, opened 2021-06-15
Adding hs create project command

miketalley:
## Description and Context
#65 
Adding a basic `hs create project [name] [dest]` command to keep parity with upcoming project creation in DesignManager. This just creates a project folder and `hsproject.json` file.

## Screenshots
image

Prompt for project `label` (will be set in `hsproject.json`) -- this defaults to the [name] argument
image

Prompt for project `description` (will be set in `hsproject.json`)
image

Resulting `hsproject.json`
image
 

## Who to Notify
@rudifer @abelb


packages/cli/commands/create/api-sample.js
```diff
@@ -1,7 +1,5 @@
-const {
-  createApiSamplePrompt,
-  overwriteSamplePrompt,
-} = require('../../lib/createApiSamplePrompt');
+const { createApiSamplePrompt } = require('../../lib/createApiSamplePrompt');
+const { folderOverwritePrompt } = require('../../lib/createApiSamplePrompt');
```
brandenrodgers, line 2: Looks like this lives in `../../lib/prompts` now.

packages/cli-lib/projects.js
```diff
@@ -191,8 +194,39 @@ async function createProject(dest, type, repoName, sourceDir, options = {}) {
...
 
+const createProjectConfig = (folderPath, configOptions = {}) => {
+  const configPath = path.join(folderPath, 'hsproject.json');
+
+  logger.debug(`Creating config file ${configPath}`);
+  return fs.writeFileSync(
+    configPath,
+    JSON.stringify(
+      {
+        label: configOptions.label || 'New project',
```
brandenrodgers, line 205: Is the label field required when creating a new project? Similarly, do we allow one account to have two projects with the same label in it? I'm wondering if we should remove the default label here.

brandenrodgers:
I think we require that projects live at the top level of the file system. Should we add validation to the `dest` arg to make sure it isn't attempting to create a project in a nested folder?

packages/cli-lib/projects.js
miketalley: I added both `label` and `description` with the expectation that they would be available for projects soon. I believe the unique identifier is different from the `label`, but I could be mistaken. I think @mattcoley may have context on this.

miketalley:
> I think we require that projects live at the top level of the file system. Should we add validation to the `dest` arg to make sure it isn't attempting to create a project in a nested folder?

I believe the backend has validation for this. Good catch though, I will test it out to make sure we are displaying the error.

packages/cli/commands/create/api-sample.js
miketalley: Fixed

PR #528, opened 2021-06-18
Polling exits when encountering an error

miketalley:
## Description and Context
Polling continues even when encountering an error

## Screenshots
image

## Who to Notify
@mattcoley @gcorne

### DIFF
--- a/packages/cli/lib/serverlessLogs.js
+++ b/packages/cli/lib/serverlessLogs.js
@@ -54,6 +54,7 @@ const tailLogs = async ({
       latestLog = await tailCall(after);
       nextAfter = latestLog.paging.next.after;
     } catch (e) {
+      logger.info('Stopped polling due to error.');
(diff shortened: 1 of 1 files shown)


packages/cli/lib/serverlessLogs.js
```diff
@@ -63,6 +63,7 @@ const tailLogs = async ({
           })
         );
       }
+      process.exit();
```
gcorne, line 66: Should we provide some feedback to the user that we have cancelled the polling?

miketalley: I added a `logger.info` message "Stopped polling due to error", and the error should also be logged.

brandenrodgers (approved):
lgtm 👍

mattcoley:
Any luck in merging this in? We got paged again for a similar issue.

PR #529, opened 2021-06-28
Marketplace theme validation command: using new dep endpoint

brandenrodgers:
## Description and Context

Switching over the marketplace theme validation command to use the new dependency endpoint. This returns all the deps in one single list. It also keeps the relative paths relative, which is important for our checks.

## Who to Notify

cc/ @KatzMitch

### DIFF
--- a/packages/cli/lib/validators/__tests__/DependencyValidator.js
+++ b/packages/cli/lib/validators/__tests__/DependencyValidator.js
@@ -8,22 +8,10 @@ const { THEME_PATH } = require('./validatorTestUtils');
 jest.mock('fs-extra');
 jest.mock('@hubspot/cli-lib/api/marketplace');
 
-const getMockValidationResult = customPath => {
+const getMockDependencyResult = (customPaths = []) => {
   const result = {
-    renderingErrors: [
-      {
-        category: 'MODULE_NOT_FOUND',
-        categoryErrors: { path: './relative/file-1.js' },
-      },
-    ],
-    allDependencies: {
-      EXTERNAL_DEPENDENCIES: ['0'],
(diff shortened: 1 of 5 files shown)


packages/cli/lib/validators/marketplaceValidators/theme/ModuleValidator.js
```diff
@@ -30,12 +30,12 @@ class ModuleValidator extends BaseValidator {
       MISSING_LABEL: {
         key: 'missingLabel',
         getCopy: ({ filePath }) =>
-          `Missing required field for ${filePath}. The meta.json file is missing the "label" field`,
+          `Missing required property for ${filePath}. The meta.json file is missing the "label" property`,
```
brandenrodgers, line 33: Updated all references of "field" to "property" for every error that was not referring to module fields.

cc/ @TheWebTech

KatzMitch:
LGTM! Thanks!

packages/cli/lib/validators/__tests__/DependencyValidator.js
```diff
@@ -8,22 +8,10 @@ const { THEME_PATH } = require('./validatorTestUtils');
 jest.mock('fs-extra');
 jest.mock('@hubspot/cli-lib/api/marketplace');
 
-const getMockValidationResult = customPath => {
+const getMockDependencyResult = (customPaths = []) => {
   const result = {
-    renderingErrors: [
```
drewjenkins, line 13: Does this mean no more rendering errors? Only dep checks?

KatzMitch: Yep - everything is lumped into the one response field on our end

miketalley (approved):

PR #530, opened 2021-07-01
Use full file path in context, not relative

KatzMitch:
## Description and Context
This PR updates the CLI response object's `context.filePath` field to be the absolute path in the file system, not the relative path to the theme root

## Screenshots
n/a

## TODO
I need to test this in our system - I think this will work and use the relative path to the working directory in Docker, but I do need to make sure it doesn't use the absolute path on the machine running this on our BE

### DIFF
--- a/packages/cli/lib/validators/marketplaceValidators/BaseValidator.js
+++ b/packages/cli/lib/validators/marketplaceValidators/BaseValidator.js
@@ -31,9 +31,8 @@ class BaseValidator {
   }
 
   getError(errorObj, file, extraContext = {}) {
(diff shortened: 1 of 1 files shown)


brandenrodgers:
FWIW the relative path is actually "relative to the theme root folder". The thinking behind it is that the FE can use the relative path to locate the theme files in the DFS. Using absolute path means that the path returned in the context will change depending on where the theme gets stored in the docker container.

KatzMitch:
Yeah so given our logs I think that's ok, and we store files in the root directory of the docker container, and I'm fine with that being a requirement on our system to make the FE works easier

PR #531, opened 2021-07-20
Surface warnings for upload/watch

brandenrodgers:
## Description and Context

This surfaces the warnings that the BE is returning on upload. Some errors are non-fatal which means that the upload/watch will still succeed. In this scenario we want to log the warnings to the user as `[WARNING]` messages.

## Screenshots

image

## TODO

- [x] Test this to make sure it works as expected
## Who to Notify

cc/ @bkrainer

### DIFF
--- a/packages/cli-lib/errorHandlers/apiErrors.js
+++ b/packages/cli-lib/errorHandlers/apiErrors.js
@@ -200,6 +200,22 @@ function logApiUploadErrorInstance(error, context) {
   logApiErrorInstance(error, context);
 }
 
+/**
+ * Logs a message for a warning resulting from filemapper API upload.
+ *
+ * @param {Object} resp
+ */
+function logApiUploadWarnings(resp, context) {
+  if (resp && resp.warnings && resp.warnings.length) {
+    resp.warnings.forEach(warning => {
+      if (warning.message) {
+        logger.warn(warning.message);
+      }
+      debugErrorAndContext(warning, context);
+    });
+  }
+}
+
 async function verifyAccessKeyAndUserAccess(accountId, scopeGroup) {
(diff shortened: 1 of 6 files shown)


brandenrodgers:
After feedback I've updated this to include the file's name in the output
image

packages/cli-lib/lib/uploadFolder.js
```diff
@@ -82,7 +83,14 @@ async function uploadFolder(accountId, src, dest, options) {
     return async () => {
       logger.debug('Attempting to upload file "%s" to "%s"', file, destPath);
       try {
-        await upload(accountId, file, destPath, apiOptions);
+        await upload(accountId, file, destPath, apiOptions).then(resp => {
```
anthmatic, line 86: Could you assign `resp` instead of using `then()`

packages/cli-lib/errorHandlers/apiErrors.js
```diff
@@ -200,6 +201,21 @@ function logApiUploadErrorInstance(error, context) {
...
 }
 
+/**
+ * Logs a message for a warning resulting from filemapper API upload.
+ *
+ * @param {Object} resp
+ */
+function logApiUploadWarnings(resp, context) {
+  if (resp && resp.warnings && resp.warnings.length) {
+    const { base } = path.parse(resp.path);
```
anthmatic, line 211: nit: is it cleaner to call `path.basename()` here?

brandenrodgers: Yep, updated in 747b3ac

brandenrodgers:
This is very outdated. Closing it

PR #534, opened 2021-08-06
Add template utils

anthmatic:
## Description and Context
Expands existing template utils to help determine which templateType a file is and also implements a method to get annotations from source.

Adding these to support https://github.com/HubSpot/hubspot-cms-vscode/pull/117


packages/cli-lib/lib/constants.js
```diff
@@ -118,6 +118,49 @@ const PROJECT_TEMPLATE_TYPES = {
   },
 };
 
+const TEMPLATE_TYPES = {
```
brandenrodgers, line 121: Not necessarily a blocker, but it would be nice if this was something that the BE could return to us so that we could keep a single source of truth. I don't imagine that it will change frequently, but it'll be yet another area where we have to manually keep data in sync.

packages/cli-lib/templates.js
```diff
@@ -27,6 +31,16 @@ const getFileAnnotations = filePath => {
   }
 };
 
+const getAnnotationsFromSource = source => {
```
brandenrodgers, line 34: This looks very similar to `getFileAnnotations`. Could you update `getFileAnnotations` to use this new `getAnnotationsFromSource` func? Just to DRY things up a little.

packages/cli-lib/templates.js
```diff
@@ -39,9 +53,24 @@ const getAnnotationValue = (annotations, key) => {
  */
 const isCodedFile = filePath => TEMPLATE_EXTENSION_REGEX.test(filePath);
 
+/*
+ * Returns true if:
+ * filename is module.html, inside of *.module folder
+ */
+const isModuleHTMLFile = filePath => MODULE_HTML_EXTENSION_REGEX.test(filePath);
```
brandenrodgers, line 60: Should these module utils live in the `templates.js` util file?

packages/cli-lib/templates.js
miketalley: Agreed, maybe `getFileAnnotations` can use `getAnnotationsFromSource`?

packages/cli-lib/templates.js
anthmatic: that's a good point

packages/cli-lib/templates.js
anthmatic: Ah, good call. Will update

packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js
```diff
@@ -99,24 +100,23 @@ class TemplateValidator extends BaseValidator {
 
     files.forEach(file => {
       if (isCodedFile(file)) {
-        const annotations = getFileAnnotations(file);
+        let data;
+        try {
+          data = fs.readFileSync(file, 'utf8');
```
anthmatic, line 105: Moves the file reading step from cli-lib to cli which I don't think has any negative impact

packages/cli-lib/templates.js
```diff
@@ -15,24 +13,18 @@ const ANNOTATION_KEYS = {
...
   const match = annotations.match(valueRegex);
   return match ? match[1].trim() : null;
 };
 
+const buildAnnotationValueGetter = source => {
+  const match = source.match(ANNOTATIONS_REGEX);
+  const annotation = match && match[1] ? match[1] : '';
+  return annotationKey => getAnnotationValue(annotation, annotationKey);
+};
+
```
anthmatic, line 27: Works as expected here https://github.com/HubSpot/hubspot-cms-vscode/pull/117/files#diff-dfcd8afcb372a75a1eb1bdc0af57d36a5e0a727dbca99350779abb21333a298fR65

and also tested the `marketplace-validate` command out and that works as expected

packages/cli-lib/modules.js
```diff
@@ -4,6 +4,11 @@ const { getCwd, getExt, splitHubSpotPath, splitLocalPath } = require('./path');
 const { walk } = require('./lib/walk');
 const { MODULE_EXTENSION } = require('./lib/constants');
 
+// Matches files named module.html
+const MODULE_HTML_EXTENSION_REGEX = new RegExp(/(\.module\/module\.html)/);
```
brandenrodgers, line 8: Nit, but I'm pretty sure this would also match if there was something after the `.html` extension. For example, `my-module.module/module.htmlanythingelse`

packages/cli-lib/templates.js
brandenrodgers: Cool! This pattern looks cleaner 🙌

brandenrodgers (approved):
Just one regex nit, but otherwise lgtm!

PR #536, opened 2021-08-13
Sandbox creation

miketalley:
## Description and Context
Adds `hs sandbox create [name]` command. Name can be provided to the command as a positional argument or the user will be prompted if no name is specified.

### Screenshots

This is now `hs sandbox create`
image

image

### Additional Fixes
- Fixes "Cannot read property 'name' of undefined" error when running `hs auth` with an invalid personal access key

image

## Who to Notify
@kimunz @cbaldauf @fndicu @gcornelius @mhazlewood @brodgers


gcorne:
One thing that I am not super comfortable with adding `hs create sandbox` is that this breaks the existing semantics of `hs create`. All of the various subcommands of `hs create` create things on the user's local computer — either single files or more complex things like a theme from the theme boilerplate. This command is very different in that it makes a change to a user's HubSpot account.

brandenrodgers:
>  All of the various subcommands of hs create create things on the user's local computer 

I hadn't thought of it until you mentioned it, but that's a good call out. The current mental model of the "create" command is that it generates the boilerplate files for certain assets. It is a much different pattern to be using it to initialize a new HubSpot sandbox account. This probably isn't the greatest suggestion, but creating and connecting the CLI to a sandbox account feels like an action that is closer to what `hs init` currently does.

miketalley:
I've changed this command to be `hs sandbox create` as that seems to fit better with the current pattern of commands.

PR #537, opened 2021-08-16
Project support

anthmatic:
## Description and Context
Updated project support:

`projects` renamed to `project`

- [x] hs project init
- [x] hs project upload

Changes to `hsproject.json`:
```json
{
  "name": "weather_app",
  "srcDir": "src"
}
```

NOTE: I've purposely left some TODOs that I'll address later, I'd just like to get the basics in so that we can get a solid feedback loop in.


packages/cli/commands/projects/init.js
```diff
@@ -0,0 +1,118 @@
...
+  logger.log('Use `hs project --help` to learn more about the command.');
+  logger.log('-------------------------------------------------------------');
+};
+
+exports.builder = yargs => {
+  yargs.positional('path', {
+    describe: 'Path to a project folder',
+    type: 'string',
+  });
+  // TODO:
```
anthmatic, line 94: I was previously accounting for these to be passed as options. Not exactly sure how that should work with the new config file. Thinking of keeping these and getting some UX feedback.

packages/cli/commands/projects.js
```diff
@@ -4,16 +4,21 @@ const {
...
+exports.command = 'project';
 exports.describe = false; //'Commands for working with projects';
 
 exports.builder = yargs => {
   addOverwriteOptions(yargs, true);
   addConfigOptions(yargs, true);
   addAccountOptions(yargs, true);
 
+  // TODO: deploy must be updated
   yargs.command(deploy).demandCommand(1, '');
```
anthmatic, line 19: Not currently used in the flow. Will need to be adjusted eventually, but will mostly stay the same so keeping this for now.

packages/cli/commands/projects/init.js
```diff
@@ -0,0 +1,118 @@
+const {
+  addAccountOptions,
+  addConfigOptions,
+  setLogLevel,
+  getAccountId,
+  addUseEnvironmentOptions,
+} = require('../../lib/commonOpts');
+// const { trackCommandUsage } = require('../../lib/usageTracking');
```
miketalley, line 8: Should we add some tracking?

packages/cli-lib/api/dfs.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @async
+ * @returns {Promise}
+ */
+async function fetchProjects(portalId) {
+  return http.get(portalId, {
+    uri: DEVELOPER_FILE_SYSTEM_API_PATH,
+  });
+}
+// TODO: paging?
```
miketalley, line 17: Did we want to add paging, or is this a reminder for later?

packages/cli/commands/projects/init.js
```diff
@@ -0,0 +1,118 @@
...
+exports.command = 'init [path]';
+exports.describe = false;
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { path: projectPath } = options;
+  const accountId = getAccountId(options);
+
+  // TODO:
```
miketalley, line 47: Should we enable this now?

packages/cli/commands/projects/init.js
```diff
@@ -0,0 +1,118 @@
...
+  logger.log('1. hs project upload');
+  logger.log(
+    '   Upload your project files to HubSpot. Upload action adds your files to a build.'
+  );
+  logger.log();
+  logger.log('2. View your changes on the preview build url');
+  logger.log();
+  logger.log('Use `hs project --help` to learn more about the command.');
+  logger.log('-------------------------------------------------------------');
+};
```
miketalley, line 87: 👍

brandenrodgers:
Pending answers to Mike's questions this all lgtm! This is a solid first pass 👌

miketalley (commented):
LGTM -- Just wanted to check if we should address the TODOs now or if they are meant to be left for later.

packages/cli-lib/api/dfs.js
anthmatic: We'll need to add paging but I'll remove the TODO for now

anthmatic:
> if we should address the TODOs now or if they are meant to be left for later

These were mostly reminders for me to come back to later or enable things like tracking when we really need it but in the interest of keeping things cleaner I enabled tracking and removed most of the TODOs for now.

PR #540, opened 2021-08-18
Function upload messaging

miketalley:
## Description and Context
Adds messaging when uploading functions that `hs functions deploy` needs to be run.

## Screenshots

Single .function folder
image

Multiple .function folders
image

Single file
image

## Who to Notify
@bkrainer @abelb

### DIFF
--- a/packages/cli-lib/lib/regex.js
+++ b/packages/cli-lib/lib/regex.js
@@ -1,7 +1,9 @@
 const API_KEY_REGEX = /^([a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12})$/i;
+const FUNCTION_FOLDER_REGEX = /\w*.functions/;
 const STRING_WITH_NO_SPACES_REGEX = /^\S*$/;
 
 module.exports = {
   API_KEY_REGEX,
+  FUNCTION_FOLDER_REGEX,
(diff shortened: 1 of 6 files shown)


brandenrodgers:
Correct me if I'm wrong about this, but one nit I have with the UX for this is that there's no scenario where we wouldn't show this message after an upload. Is it odd that after every single upload we'll show this warning to tell the user that their function hasn't been deployed? To me, a warning sounds like something has gone wrong or I did something wrong. Could we potentially use a more informative/guiding tone for this message? Like a "cool you uploaded changes, you may want to run the "deploy" command next" type of message.

miketalley:
> Correct me if I'm wrong about this, but one nit I have with the UX for this is that there's no scenario where we wouldn't show this message after an upload. Is it odd that after every single upload we'll show this warning to tell the user that their function hasn't been deployed? To me, a warning sounds like something has gone wrong or I did something wrong. Could we potentially use a more informative/guiding tone for this message? Like a "cool you uploaded changes, you may want to run the "deploy" command next" type of message.

The message only shows for files uploaded within a `.functions` folder. I've also changed the message to use `logger.info` instead of `logger.warn` as this is actually an info message and not a warning.

I think it's not intrusive enough when being shown after all function folder uploads. It does take into account bulk uploads and will only show once at the end of the upload, instead of after each file.

@abelb may have some input on this.

PR #542, opened 2021-08-20
Add build status to project upload

anthmatic:
## Description and Context
After successful project upload, polls for build status

## Screenshots

Spinners while project dependencies are building
image

Success message
image

General failure
image

Failure of a single dependency build
image


packages/cli/lib/projects.js
```diff
@@ -4,10 +4,15 @@ const path = require('path');
 const chalk = require('chalk');
 const findup = require('findup-sync');
 const { prompt } = require('inquirer');
+const Spinnies = require('spinnies');
```
miketalley, line 7: We were using `ora` for loading spinners elsewhere in the CLI (example https://github.com/HubSpot/hubspot-cli/blob/9da73e88b7c0d4d37993f1cc3a6a53d588c8d36f/packages/cli/commands/logs.js#L78). Is there a reason to have both or should we use `ora` here also?

miketalley (commented):
Looks good, just one question about adding the spinnies dependency.

packages/cli/lib/projects.js
```diff
@@ -14,6 +14,27 @@ const {
...
+  [BUILD_STATUS.FAILURE]: 'failed to build',
+  [BUILD_STATUS.SUCCESS]: 'built successfully',
+};
+
+const isBuildComplete = build => {
+  return (
+    build.status === BUILD_STATUS.SUCCESS ||
+    build.status === BUILD_STATUS.FAILURE
+  );
+};
```
anthmatic, line 36: 😀

packages/cli/lib/projects.js
anthmatic: Yeah, that's exactly it and I agree

miketalley (approved):
Looks good after testing and digging in a bit more.

PR #546, opened 2021-09-03
[Marketplace] Support theme section validation

ckconant:
## Description and Context
Adds section support for marketplace theme validation. The update validates that
- There are no more than 50 sections
- The section annotation contains a **description** property
- The section annotation contains a **label** property
- The section annotation contains a **screenshotPath** property

Please let me know if there is anything that I can change as this is my first time contributing to the repo.

## Screenshots
Below is a screenshot displaying the validation when testing with the `sample-section.html` file.
Screen Shot 2021-09-03 at 1 58 37 PM

## Who to Notify
@anthmatic @brandenrodgers @KatzMitch @jevenson @jasonnrosa @TanyaScales


KatzMitch (approved):
So I definitely shouldn't be the final review here but LGTM! Thanks for knocking this out!😋🥳

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
```diff
@@ -0,0 +1,116 @@
...
+          }
+          if (!screenshotPath) {
+            validationErrors.push(
+              this.getError(this.errors.MISSING_SCREENSHOT_PATH, file)
+            );
+          }
+        }
+      } else {
+        validationErrors.push(
+          this.getError(this.errors.MISSING_TEMPLATE_TYPE, file)
```
jevenson, line 95: `MISSING_TEMPLATE_TYPE` doesn't appear to be defined in the errors object.

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
```diff
@@ -0,0 +1,116 @@
...
+          annotations,
+          ANNOTATION_KEYS.isAvailableForNewContent
+        );
+
+        if (templateType !== 'section') {
+          return;
+        }
+        sectionCount++;
+
+        if (isAvailableForNewContent !== 'false') {
```
anthmatic, line 66: Just confirming: sections with `isAvailableForNewContent === false` do not could towards the limit?

packages/cli/defaults/sample-section.html
```diff
@@ -0,0 +1,14 @@
...
+  {% dnd_column %}
+    {% dnd_row %}
+      {% dnd_module
+        path='@hubspot/rich_text',
+        html='<h1>Hello world</h1>'
+      %}
+      {% end_dnd_module %}
+    {% end_dnd_row %}
+  {% end_dnd_column %}
+{% end_dnd_section %}
```
anthmatic, line 14: NIT: missing new line

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
```diff
@@ -0,0 +1,116 @@
...
+      MISSING_DESCRIPTION: {
+        key: 'missingDescription',
+        getCopy: ({ filePath }) =>
+          `Missing required property for ${filePath}. The section is missing the "description" property`,
+      },
+    };
+  }
+
+  // Validates:
+  // - All sections have a 'description' annotation
```
anthmatic, line 40: NIT: quotes

packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js
```diff
@@ -48,6 +48,7 @@ const VALIDATIONS_BY_TYPE = {
   proposal: { allowed: false },
   blog_listing: { allowed: true, label: true, screenshot: true },
   blog_post: { allowed: true, label: true, screenshot: true },
+  section: { allowed: true, label: false, screenshot: false },
```
anthmatic, line 51: Is this needed since sections have their own validator?

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
brandenrodgers: This else check isn't needed since we will abort [here](https://github.com/HubSpot/hubspot-cli/pull/546/files#diff-547055959147bfda4ac86ee037e1eeb2d499ec437041df4b9196d541f799fd7eR61) if the template type is undefined. Also, [this else check](https://github.com/HubSpot/hubspot-cli/blob/ee1e9e802d1a8423d88adab3c31b8f56c4ad2bbe/packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js#L154-L158) in the template validator should already catch this case where the section doesn't have a template type type.

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
ckconant: Correct. I am open to discussion on this one, but I feel that we can allow developers to use sections as building blocks without offering them to customers.
@KatzMitch curious your thoughts here, and how you're approaching the internal validation.

packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js
ckconant: Nice catch. I ended up adding a `TEMPLATE_IGNORE_LIST` that ignores any templateType that we want to ignore. Currently we're just ignoring `section`.

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
ckconant: Thanks, I removed the else check.

packages/cli/defaults/sample-section.html
ckconant: Thanks, resolved in https://github.com/HubSpot/hubspot-cli/pull/546/commits/c95e30d9d7a7e260f4d0af1b8c2efc6a77be658e

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
ckconant: Thanks, resolved in https://github.com/HubSpot/hubspot-cli/pull/546/commits/c95e30d9d7a7e260f4d0af1b8c2efc6a77be658e

packages/cli/lib/validators/marketplaceValidators/theme/SectionValidator.js
KatzMitch: I vote we keep it the same as templates - looks like it's 50 total, not 50 avail for new content (https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js#L124-L125)
Also do we need to update the ignored type list in that file?

ckconant: > I vote we keep it the same as templates - looks like it's 50 total, not 50 avail for new content

Sounds good! I'll account for that.

> Also do we need to update the ignored type list in that file?

Yep! Accounting for that - https://github.com/HubSpot/hubspot-cli/blob/marketplace/section-support/packages/cli/lib/validators/marketplaceValidators/theme/TemplateValidator.js#L12

ckconant: @KatzMitch @anthmatic updates to account for non-available sections in https://github.com/HubSpot/hubspot-cli/pull/546/commits/a36ec9f9cf6d78a3519e10c4e3c41ed3e7cb074f

anthmatic (approved):
Looks great!

Issue #180, opened 2020-04-27
Replace `request` library with another http library


gcorne:
The `request` library is deprecated so we need to replace it. This is a significant change to `@hubspot/cms-lib` so I think we'll bump to a new major version.

PR #547, opened 2021-09-07
Upload progress POC

anthmatic:

### DIFF
--- a/packages/cli-lib/api/dfs.js
+++ b/packages/cli-lib/api/dfs.js
@@ -1,5 +1,6 @@
 const http = require('../http');
 const fs = require('fs');
+const cliProgress = require('cli-progress');
 
 const PROJECTS_API_PATH = 'dfs/v1/projects';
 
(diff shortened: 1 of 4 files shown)


packages/cli-lib/api/dfs.js
```diff
@@ -40,11 +41,20 @@ async function createProject(portalId, name) {
...
+
+  progressBar.start(size, 0);
+
   return http.post(accountId, {
     uri: `${PROJECTS_API_PATH}/upload/${projectName}`,
     timeout: 60000,
     formData: {
-      file: fs.createReadStream(projectFile),
+      file: fs.createReadStream(projectFile).on('data', chunk => {
+        progressBar.update((bytes += chunk.length));
```
gcorne, line 56: I think that you would want to track the progress of the POST request but I am not sure that the deprecated `request` npm module supports it based on discussion in https://github.com/request/request/issues/1450. Might be time to think about what it would take to start [shifting to another http library](https://github.com/HubSpot/hubspot-cli/issues/180). It looks like Axios [supports progress](https://github.com/axios/axios/blob/d99d5faac29899eba68ce671e6b3cbc9832e9ad8/examples/upload/index.html#L30-L32).

anthmatic: Yeah, listening on the readable steam was the closest I could get with request.js. I did attempt to implement axios, thinking the same thing, but then I discovered that the `onUploadProgress` event is only supported in browsers.

image

Overall, the support for tracking POST/upload status in node is pretty poor from what I can tell. It does seem to be more straightforward with s3-style APIs.

PR #551, opened 2021-09-09
[POC] Test upload via node htttps module

anthmatic:


anthmatic:
@gcorne I put together a quick POC to test uploading out with the standard `https` node module. The end result (to the user) is very similar to #547.

Unless I'm doing something wrong, I'm only seeing one `drain` event with the full file size followed by latency waiting for the BE response

PR #555, opened 2021-09-14
Adjust Oauth refresh cutoff

anthmatic:
## Description and Context
Updates the Oauth strategy to refresh a closer to the token expiration time


miketalley (approved):
This makes sense to me. Looks like we will be refreshing things more often with the [reduction mentioned](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lib/models/OAuth2Manager.js#L74) and using a 5 minute buffer now.

PR #556, opened 2021-09-15
Fixing buildAnnotationValueGetter util

brandenrodgers:
## Description and Context

The annotation util was [updated here](https://github.com/HubSpot/hubspot-cli/pull/534), and a [PR here](https://github.com/HubSpot/hubspot-cli/pull/546) that used the old version of the util was merged.

I also updated the util to include readFileSync so it's a little DRY-er

## TODO

- [ ] Need to test this still

### DIFF
--- a/packages/cli-lib/templates.js
+++ b/packages/cli-lib/templates.js
@@ -1,5 +1,5 @@
 const fs = require('fs');
-const { logger } = require('@hubspot/cli-lib/logger');
+const { logger } = require('./logger');
 
 // Matches the .html file extension, excluding module.html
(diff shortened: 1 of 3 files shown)


anthmatic (approved):
Looks like there's a dependency issue but otherwise looks good and worked as expect while testing

brandenrodgers:
@drewjenkins you can test this out by running the theme validation commands. for example `hs theme marketplace-validate <path to local theme>`

miketalley (approved):
Code looks good, I don't have a lot of context on what this is doing though.

PR #558, opened 2021-09-15
Fix spinner issue on project upload

anthmatic:
## Description and Context
Fixes an issue where we continually try to update the status of a spinner even after it's been removed

## Screenshots
image

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -173,6 +173,10 @@ const pollBuildStatus = async (accountId, name, buildId) => {
 
       if (Object.keys(spinnies.spinners).length) {
         subbuildStatuses.forEach(subBuild => {
+          if (!spinnies.pick(subBuild.buildName)) {
+            return;
+          }
+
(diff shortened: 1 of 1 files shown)


packages/cli/lib/projects.js
```diff
@@ -173,6 +173,10 @@ const pollBuildStatus = async (accountId, name, buildId) => {
 
       if (Object.keys(spinnies.spinners).length) {
```
brandenrodgers, line 174: I was poking around the spinnies npm page (I'm new to that package), and I saw that there's a `hasActiveSpinners()` util. Could that replace the `Object.keys(spinnies.spinners).length` check?

brandenrodgers (approved):
I had one question related to the spinnies lib, but otherwise this lgtm 👍

miketalley (approved):

packages/cli/lib/projects.js
anthmatic: Good catch! I'll try swapping that out.

PR #561, opened 2021-09-21
Project deploy changes

anthmatic:
## Description and Context

- [x]  Update `hs project deploy` for new deploy service
  - [x] add new endpoints
  - [x] poll for deployed status


packages/cli/lib/projects.js
```diff
@@ -232,11 +232,102 @@ const pollBuildStatus = async (accountId, name, buildId) => {
   });
 };
 
+const pollDeployStatus = async (accountId, name, deployId) => {
```
anthmatic, line 235: This is very similar to the build status polling logic. Will definitely DRY this up- time permitting, I'll do that in this PR

brandenrodgers:
I haven't dug into this yet, but I wanted to mention that IMO it's OK to get this deploy work out without handling the " Update hs project upload to be aware of deployments" work. If you want you could handle that in a separate PR.

miketalley (approved):
Code looks good!

anthmatic:
Example of the manual deploy flow:
image

Since the auto-deploy side of things is still in flight I'm going to get this in so that folks can test.

PR #562, opened 2021-09-22
Prevent spinnies error on upload if buildPolling fails

anthmatic:
## Description and Context
When polling was sharing a catch block with the upload function, which was causing issues where an inactive "spinner" was incorrectly updated. 

image

### DIFF
--- a/packages/cli/commands/project/deploy.js
+++ b/packages/cli/commands/project/deploy.js
@@ -62,22 +62,27 @@ exports.handler = async options => {
     logger.error('No latest build ID was found.');
     return;
   };
-  const deployedBuildId = buildId || (await getBuildId());
-  let deployId;
 
   try {
+    const deployedBuildId = buildId || (await getBuildId());
+
     const deployResp = await deployProject(
       accountId,
       projectConfig.name,
       deployedBuildId
     );
 
-    deployId = deployResp.deployId;
-
     if (deployResp.error) {
       logger.error(`Deploy error: ${deployResp.error.message}`);
       return;
     }
+
+    await pollDeployStatus(
+      accountId,
+      projectConfig.name,
+      deployResp.deployId,
+      deployedBuildId
+    );
   } catch (e) {
     if (e.statusCode === 400) {
       logger.error(e.error.message);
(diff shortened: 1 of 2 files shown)


packages/cli/commands/project/deploy.js
```diff
@@ -94,6 +89,17 @@ exports.handler = async options => {
...
+
+  try {
+    await pollDeployStatus(
+      accountId,
+      projectConfig.name,
+      deployId,
+      deployedBuildId
+    );
+  } catch (err) {
+    logger.error(err);
```
brandenrodgers, line 101: Could you elaborate on why moving this here fixes the problem?

Also, what type of errors can exist here? Are they api-related errors, or internal JS errors? If it's the latter I think we should switch this to be `logger.debug()` so we don't fill the console with internal js errors.

anthmatic: Good call on error types, I'll address that. 

Also, after looking at this again don't think think that this change is relevant here. It is in upload.js, so I'll add a comment there

packages/cli/commands/project/upload.js
```diff
@@ -57,19 +57,22 @@ const uploadProjectFiles = async (accountId, projectName, filePath) => {
...
       text: `Uploaded ${chalk.bold(projectName)} project files to ${chalk.bold(
         accountId
       )}`,
     });
 
     logger.debug(
-      `Project "${projectName}" uploaded and build #${upload.buildId} created`
+      `Project "${projectName}" uploaded and build #${buildId} created`
     );
-    await pollBuildStatus(accountId, projectName, upload.buildId);
```
anthmatic, line 72: If this fails, the `spinnies.fail()` within the catch statement may throw an error because the spinner was already removed [above](https://github.com/HubSpot/hubspot-cli/pull/562/files/7706201da43b916d927b9ffbfb025b4b0364aaf8#diff-1989a521c805539a7ae5e60fa3f7b0f79d83106bbb225045cf33f831abf97757L63)

brandenrodgers (approved):
Seems reasonable to me 👍

PR #563, opened 2021-09-23
Check if config file can be exposed via git during hs init

anthmatic:
## Description and Context
 `hs init` does not perform the check to make sure that the `hubspot.config.yml` file is not committed to git repos

## Screenshots
image

### DIFF
--- a/packages/cli/commands/init.js
+++ b/packages/cli/commands/init.js
@@ -1,5 +1,6 @@
 const path = require('path');
 const fs = require('fs-extra');
+const { checkAndWarnGitInclusion } = require('@hubspot/cli-lib');
 const {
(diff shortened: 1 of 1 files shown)


anthmatic:
My only concern is that this message might be a little jarring for a user that's `init`-ing for the first time. Curious if folks share that concern?

gcorne:
I actually wonder if we should check before the auth flow and not allow the user to continue. We could even provide more concrete steps for how to create the config in the $HOME (at least for Linux/MacOS).

brandenrodgers:
Closing this because we've decided to automatically create or update the gitignore file on init https://github.com/HubSpot/hubspot-cli/pull/720

PR #564, opened 2021-09-23
Roll hs init step into hs project upload

anthmatic:
## Description and Context
Now as part of the `hs project upload` command, we:
- make sure that a config exists, and if not- prompt the user to run `hs project init`
- check to see if the project exists in the account and if not, prompt the user to create it

When no config is found:
```
[ERROR] Project config not found. Try running 'hs project init' first.
```

When config is found but project doesn't exist in account:
```
? The project test12 does not exist in 123. Would you like to create it? (Y/n)
```

## TODO
The next step here is to rename and update the existing `hs project init` function to `hs project create` and build out that command

### DIFF
--- a/packages/cli/commands/project/upload.js
+++ b/packages/cli/commands/project/upload.js
@@ -101,10 +101,12 @@ exports.handler = async options => {
 
   trackCommandUsage('project-upload', { projectPath }, accountId);
 
-  const cwd = projectPath ? path.resolve(getCwd(), projectPath) : getCwd();
-  const projectConfig = await getProjectConfig(cwd);
+  const projectDir = projectPath
+    ? path.resolve(getCwd(), projectPath)
+    : getCwd();
+  const projectConfig = await getProjectConfig(projectDir);
 
-  validateProjectConfig(projectConfig);
+  validateProjectConfig(projectConfig, projectDir);
 
   await ensureProjectExists(accountId, projectConfig.name);
 
@@ -134,8 +136,10 @@ exports.handler = async options => {
 
   archive.pipe(output);
 
-  archive.directory(path.resolve(cwd, projectConfig.srcDir), false, file =>
-    shouldIgnoreFile(file.name) ? false : file
+  archive.directory(
+    path.resolve(projectDir, projectConfig.srcDir),
+    false,
+    file => (shouldIgnoreFile(file.name) ? false : file)
   );
 
   archive.finalize();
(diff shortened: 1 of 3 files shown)


packages/cli/lib/projects.js
```diff
@@ -103,14 +112,45 @@ const validateProjectConfig = projectConfig => {
...
     );
     process.exit(1);
   }
 };
 
+const ensureProjectExists = async (accountId, projectName) => {
+  try {
+    await fetchProject(accountId, projectName);
+  } catch (err) {
+    if (err.statusCode === 404) {
```
brandenrodgers, line 127: This may not be worth supporting, but what if we get a 500 or some other internal error code?

packages/cli/commands/project/upload.js
```diff
@@ -100,10 +102,14 @@ exports.handler = async options => {
...
-  const projectConfig = await getProjectConfig(cwd);
+  const projectDir = projectPath
+    ? path.resolve(getCwd(), projectPath)
+    : getCwd();
+  const projectConfig = await getProjectConfig(projectDir);
+
+  validateProjectConfig(projectConfig, projectDir);
 
-  validateProjectConfig(projectConfig);
+  await ensureProjectExists(accountId, projectConfig.name);
```
brandenrodgers, line 112: Are we sure that accountId exists here? I was seeing an issue when testing this (sometime last week) where it was logging `The project <Project name> does not exist in undefined. Would you like to create it?`

brandenrodgers:
I had two questions, but otherwise this looks good to me 👌

packages/cli/lib/projects.js
anthmatic: My intent was for https://github.com/HubSpot/hubspot-cli/pull/564/files/370500c158ffbe7d34e32fdfef2c1ebdff146f1f#diff-fa4fa15b34bd2259fc2c24ec990fdbaa6492e6030d940700895318b399058174R150 to handle any other error codes that might come through

packages/cli/commands/project/upload.js
anthmatic: So yeah, I believe accountId should always exist here. If it doesn't https://github.com/HubSpot/hubspot-cli/pull/564/files#diff-1989a521c805539a7ae5e60fa3f7b0f79d83106bbb225045cf33f831abf97757R98 should be ending the process

PR #565, opened 2021-09-24
Exit 2 on FATAL marketplace validations

KatzMitch:
## Description and Context
This PR updates the exit code to `2` (instead of `1`) when the CLI has FATAL validation warnings. This will allow us to determine on the BE caller whether we are exiting because of a validation failure or because of an internal error of some other sort

## Screenshots
Before
Screen Shot 2021-09-24 at 10 40 35 AM

After
Screen Shot 2021-09-24 at 10 39 53 AM

## TODO
n/a

## Who to Notify
n/a

### DIFF
--- a/packages/cli/commands/marketplaceValidate/validateTheme.js
+++ b/packages/cli/commands/marketplaceValidate/validateTheme.js
@@ -75,7 +75,7 @@ exports.handler = async options => {
         .flat()
         .some(result => result.result === VALIDATION_RESULT.FATAL)
     ) {
-      process.exit(1);
+      process.exit(2);
     }
   });
 };


gcorne:
We should do the same in https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/lint.js#L60. I also think it would be valuable to start using named constants for the exit codes to help with understandability but that is probably a good followup for someone working day-to-day on the CLI.

brandenrodgers:
> that is probably a good followup for someone working day-to-day on the CLI.

I can follow this PR up with something like this 👍 

@KatzMitch this is OK to merge

PR #567, opened 2021-09-29
Project create command

anthmatic:
## Screenshots
Create project with no template:
image

Creating project from template: 
image


brandenrodgers (approved):
Tested this locally and it looks great! IMO a huge improvement over the init command.

PR #568, opened 2021-10-01
Make project upload aware of autodeploy

anthmatic:

### DIFF
--- a/packages/cli/commands/project/upload.js
+++ b/packages/cli/commands/project/upload.js
@@ -32,6 +32,7 @@ const {
   validateProjectConfig,
   pollBuildStatus,
   ensureProjectExists,
+  pollDeployStatus,
 } = require('../../lib/projects');
 
 const loadAndValidateOptions = async options => {
@@ -96,11 +97,7 @@ const uploadProjectFiles = async (accountId, projectName, filePath) => {
     );
   }
 
-  try {
-    await pollBuildStatus(accountId, projectName, buildId);
-  } catch (err) {
-    logger.log(err);
-  }
+  return { buildId };
 };
 
 exports.handler = async options => {
@@ -130,7 +127,25 @@ exports.handler = async options => {
   output.on('close', async function() {
     logger.debug(`Project files compressed: ${archive.pointer()} bytes`);
 
-    await uploadProjectFiles(accountId, projectConfig.name, tempFile.name);
+    const { buildId } = await uploadProjectFiles(
+      accountId,
+      projectConfig.name,
+      tempFile.name
+    );
+
+    const {
+      autoDeployEnabled,
+      deployStatusTaskLocator,
+    } = await pollBuildStatus(accountId, projectConfig.name, buildId);
+
+    if (autoDeployEnabled && deployStatusTaskLocator.id) {
+      await pollDeployStatus(
+        accountId,
+        projectConfig.name,
+        deployStatusTaskLocator.id,
+        buildId
+      );
+    }
 
     try {
       tempFile.removeCallback();
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -211,7 +211,13 @@ const pollBuildStatus = async (accountId, name, buildId) => {
       const buildStatus = await getBuildStatus(accountId, name, buildId).catch(
         reject
       );
-      const { status, subbuildStatuses } = buildStatus;
+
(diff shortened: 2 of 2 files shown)


packages/cli/lib/projects.js
```diff
@@ -243,7 +249,7 @@ const pollBuildStatus = async (accountId, name, buildId) => {
         });
       }
 
-      if (isBuildComplete(buildStatus)) {
+      if (isBuildComplete(buildStatus) && spinnies.hasActiveSpinners()) {
```
brandenrodgers, line 252: Why do we care if there are active spinners here?

packages/cli/lib/projects.js
```diff
@@ -268,6 +274,13 @@ const pollBuildStatus = async (accountId, name, buildId) => {
             }
           });
         }
+
+        if (!autoDeployEnabled) {
+          resolve(buildStatus);
+        }
+      }
+
+      if (autoDeployEnabled && deployStatusTaskLocator) {
```
brandenrodgers, line 283: What happens if `deployStatusTaskLocator` does not exist? Should we treat that as a BE internal error?

packages/cli/lib/projects.js
anthmatic: Oof, I think this is left over from testing something else out 👍

packages/cli/lib/projects.js
anthmatic: Yeah, I'll check for a valid `deployStatusTaskLocator`. Working on the messaging/error state part of this now

packages/cli/lib/strings.js
```diff
@@ -0,0 +1,22 @@
...
+      `Deploying ${chalk.bold(
+        name
+      )}\n\nFound ${numOfComponents} components in this project ...\n`,
+    SUCCESS: name => `Deployed ${chalk.bold(name)}`,
+  },
+};
+
+module.exports = {
+  PROJECTS,
+};
```
anthmatic, line 22: @brandenrodgers @miketalley I wanted to centralize some strings to make upkeep easier. Before I go too much further does this seem too weird? Open to suggestions!

packages/cli/lib/projects.js
```diff
@@ -188,183 +204,109 @@ const showWelcomeMessage = (projectName, accountId) => {
...
-const pollBuildStatus = async (accountId, name, buildId) => {
-  const buildStatus = await getBuildStatus(accountId, name, buildId);
-  const spinnies = new Spinnies();
-
-  logger.log();
-  logger.log(`Building ${chalk.bold(name)}`);
-  logger.log();
-  logger.log(`Found ${buildStatus.subbuildStatuses.length} deployables ...`);
-  logger.log();
+const makeGetTaskStatus = taskType => {
```
anthmatic, line 207: Build and Deploy status are very similar, so I unified the logic here

packages/cli/lib/projects.js
```diff
@@ -188,183 +206,110 @@ const showWelcomeMessage = (projectName, accountId) => {
...
-  for (let subBuild of buildStatus.subbuildStatuses) {
-    spinnies.add(subBuild.buildName, {
-      text: `${chalk.bold(subBuild.buildName)} #${buildId} ${
-        PROJECT_BUILD_STATUS_TEXT[PROJECT_BUILD_STATUS.ENQUEUED]
-      }`,
+  return async (accountId, taskName, taskId) => {
+    const spinnies = new Spinnies({
+      succeedColor: 'white',
+      failColor: 'white',
+      // TODO: spacing is a bit off when customizing these
```
miketalley, line 230: Do we want to address this?

packages/cli/commands/project/upload.js
```diff
@@ -130,7 +129,64 @@ exports.handler = async options => {
...
+      status,
+      subbuildStatuses,
+    } = await pollBuildStatus(accountId, projectConfig.name, buildId);
+
+    if (status === 'FAILURE') {
+      const failedSubbuilds = subbuildStatuses.filter(
+        subbuild => subbuild.status === 'FAILURE'
+      );
+
+      logger.log('-'.repeat(50));
```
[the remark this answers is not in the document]
anthmatic: I had a similar thought about this and started implementing `log.line()` but I also want to push forward on string centralization or some sort (possibly #476) and I could see this being tied to that as well

PR #569, opened 2021-10-01
Support route-based log fetching

miketalley:
## Description and Context
Updating the endpoints for fetching logs and also when fetching using `appPath`.

## Who to Notify
@bkrainer @gcorne

### DIFF
--- a/packages/cli/commands/logs.js
+++ b/packages/cli/commands/logs.js
@@ -14,13 +14,8 @@ const {
   checkAndWarnGitInclusion,
 } = require('@hubspot/cli-lib');
 const { logger } = require('@hubspot/cli-lib/logger');
-const {
-  logServerlessFunctionApiErrorInstance,
-  ApiErrorContext,
-} = require('@hubspot/cli-lib/errorHandlers');
 const { outputLogs } = require('@hubspot/cli-lib/lib/logs');
 const {
-  getFunctionByPath,
   getAppFunctionLogs,
   getLatestAppFunctionLogs,
 } = require('@hubspot/cli-lib/api/functions');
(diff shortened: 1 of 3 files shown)


packages/cli/commands/logs.js
```diff
@@ -111,6 +93,7 @@ const endpointLog = async (accountId, options) => {
 
 const appFunctionLog = async (accountId, options) => {
   const { latest, follow, compact, functionName, appPath } = options;
+  const projectName = appPath.split('/')[0];
```
gcorne, line 96: I am not sure about this from a UX perspective. Maybe worth discussing internally how this should work but one idea would be to automatically detect that we're in a project and get the `projectName` via `hsproject.json`. There could also be value in moving towards `hs project logs` and automatically detecting an app especially in the common case where there is a single app in a project.

gcorne: To add a bit more, the prefixing of the project name only comes into play when Design Manager comes into play, which is only the case for themes.

miketalley: I've added the `--projectName` argument for now as that is the easiest solution. I think we should put some thought into how to obtain this programmatically.

packages/cli/commands/project/logs.js
```diff
@@ -0,0 +1,183 @@
...
+    compact,
+    appPath,
+    functionName,
+    projectName,
+  } = options;
+
+  let logsResp;
+
+  if (follow) {
+    const spinner = ora(
```
anthmatic, line 56: Not a deal breaker but it might be good to make the switch to `spinnies` here

PR #570, opened 2021-10-05
Standardize process exit codes

brandenrodgers:
## Description and Context

Following up this PR here: https://github.com/HubSpot/hubspot-cli/pull/565

We would like to have three different exit codes:
- `0`: A successful run
- `1`: There was a config problem or an internal error
- `2`: There are warnings or validation issues

I'm adding some utils for us to use around the repo to make sure that we're using the exit codes consistently. There are a lot of places that will need to be updated, so I haven't done that yet. I'm showing three different patterns that we could use. I don't feel super strongly about any of them, but I figured I'd check in with everyone before picking one of them.

Here's some more context. This is how the [ESLint CLI](https://eslint.org/docs/user-guide/command-line-interface) handles their error codes.

## TODO

- [x] Pick a pattern, and update all the commands to use it
## Who to Notify

cc/ @gcorne


gcorne:
My preference is the enum with process.exit() because I think it makes it clearer that execution will stop at that point.

anthmatic:
> My preference is the enum with process.exit() because I think it makes it clearer that execution will stop at that point. 

I second this

anthmatic (approved):
Looks good

packages/cli/lib/exitCodes.js
```diff
@@ -0,0 +1,14 @@
+/*
+ * 0: Successful run
+ * 1: Config problem or internal error
+ * 2: Warnings or validation issues
+ */
+const EXIT_CODES = {
```
miketalley, line 6: Would it make sense to put this under `/cli/lib/enums/exitCodes.js`?

brandenrodgers: I can do that 👍

PR #575, opened 2021-10-20
Show account name instead of ID

anthmatic:
## Description and Context
We'd like to expose the `name` of the account you're operating on when possible. There are a number of places we could get this from but I'm curious if folks thing this is a good approach.

For example:
https://github.com/HubSpot/hubspot-cli/blob/9a4b776076423ab43c7d59ee2f3c0b115a372e98/packages/cli/commands/remove.js#L38
could just be
```js
const { accountId } = getAccountDetails(options);
```


packages/cli-lib/lib/config.js
```diff
@@ -452,12 +452,28 @@ const getAccountId = nameOrId => {
...
+ */
+const getAccountId = nameOrId => {
+  const { portalId } = getAccount(nameOrId);
+
+  if (portalId) {
+    return portalId;
+  }
+  return null;
+};
+
```
anthmatic, line 476: Wouldn't mind a gut-check on this. Too messy?

brandenrodgers:
Code lgtm and I'm a fan of this new pattern 👍 

I saw [this PR ](https://github.com/HubSpot/hubspot-cli/pull/592/files) for adding a `getAccountDescription` util. Is the idea to use the new util in here?

anthmatic:
@brandenrodgers I was going to bring this up in our sync, but in [that PR](https://github.com/HubSpot/hubspot-cli/pull/592/files) does the same thing as this PR, just a different option. I'll add some notes to that PR but my hope is to merge that and close this PR out

PR #582, opened 2021-10-28
Add list builds command

anthmatic:
## Description and Context

- Adds `hs project list-builds`
- Displays a table view of builds to provide an overview of each build
- Marks the currently deployed build
- Shows 10 builds by default, supports the `--limit` flag
- Has simple paging (hoping to get feedback on this)

## Screenshots

image

## TODO

-Add links to project details UI

### DIFF
--- a/packages/cli/commands/project/deploy.js
+++ b/packages/cli/commands/project/deploy.js
@@ -19,7 +19,11 @@ const {
 const { logger } = require('@hubspot/cli-lib/logger');
 const { deployProject, fetchProject } = require('@hubspot/cli-lib/api/dfs');
 const { validateAccount } = require('../../lib/validation');
-const { getProjectConfig, pollDeployStatus } = require('../../lib/projects');
+const {
+  getProjectConfig,
+  pollDeployStatus,
+  validateProjectConfig,
+} = require('../../lib/projects');
 
 const loadAndValidateOptions = async options => {
   setLogLevel(options);
@@ -44,7 +48,9 @@ exports.handler = async options => {
 
   trackCommandUsage('project-deploy', { projectPath }, accountId);
 
-  const { projectConfig } = await getProjectConfig(projectPath);
+  const { projectConfig, projectDir } = await getProjectConfig(projectPath);
+
+  validateProjectConfig(projectConfig, projectDir);
 
   logger.debug(`Deploying project at path: ${projectPath}`);
 
--- a/packages/cli/commands/project/listBuilds.js
+++ b/packages/cli/commands/project/listBuilds.js
@@ -69,13 +69,19 @@ exports.handler = async options => {
       options
     );
     const currentDeploy = project.deployedBuildId;
-    logger.log(
-      `Sowing the ${results.length} most recent builds for ${project.name}. ` +
-        link(
-          'View all builds in project details.',
-          getProjectDetailUrl(projectConfig.name, accountId)
-        )
-    );
+    if (options && options.after) {
+      logger.log(
+        `Showing the next ${results.length} builds for ${project.name}`
+      );
+    } else {
+      logger.log(
+        `Showing the ${results.length} most recent builds for ${project.name}. ` +
+          link(
+            'View all builds in project details.',
+            getProjectDetailUrl(project.name, accountId)
+          )
+      );
+    }
 
     if (results.length === 0) {
       logger.log('No builds found.');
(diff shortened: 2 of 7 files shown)


packages/cli/commands/project/listBuilds.js
```diff
@@ -0,0 +1,173 @@
...
+
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { path: projectPath, limit } = options;
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('project-list-builds', { projectPath }, accountId);
+
+  const cwd = projectPath ? path.resolve(getCwd(), projectPath) : getCwd();
```
brandenrodgers, line 60: Maybe out of scope for this PR, but it looks like we could put this `const cwd = projectPath ? path.resolve(getCwd(), projectPath) : getCwd();'  into the `getProjectConfig` util because we do it almost every time we call it.

packages/cli/commands/project/listBuilds.js
```diff
@@ -0,0 +1,173 @@
...
+      await prompt({
+        name: 'more',
+        message: 'Press <enter> to load more, or ctrl+c to exit',
+      });
+      await fetchAndDisplayBuilds(project, { limit, after: paging.next.after });
+    }
+  };
+
+  try {
+    const project = await fetchProject(accountId, projectConfig.name);
```
brandenrodgers, line 136: Is it possible that this request fails? Are we handling that anywhere?

packages/cli/commands/project/listBuilds.js
```diff
@@ -0,0 +1,173 @@
...
+exports.handler = async options => {
+  loadAndValidateOptions(options);
+
+  const { path: projectPath, limit } = options;
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('project-list-builds', { projectPath }, accountId);
+
+  const cwd = projectPath ? path.resolve(getCwd(), projectPath) : getCwd();
+  const projectConfig = await getProjectConfig(cwd);
```
brandenrodgers, line 61: We should do a check for projectConfig here. I ran `hs project list-builds some/bad/path` and I get an error that says `Cannot read property 'name' of null` coming from [this line here](https://github.com/HubSpot/hubspot-cli/pull/582/files#diff-36a35fb191a893b8f5de180e0cc2bd7656d4c00c972ef65a9687a6e3f8304817R136).

Maybe we could wrap up the error handling for missing project config into the `getProjectConfig` util? So all the commands that use it will get the error handling for free.

packages/cli/commands/project/listBuilds.js
```diff
@@ -0,0 +1,173 @@
...
+      );
+      logger.log(
+        getTableContents(builds, {
+          columnDefault: {
+            paddingLeft: 3,
+          },
+        })
+      );
+    }
+    if (paging && paging.next) {
```
brandenrodgers, line 126: Sorry if this is already a known issue, but it looks like `paging.next` exists even if I only have one build. So this prompt is showing when it shouldn't.

packages/cli/commands/project/listBuilds.js
anthmatic: Ah, that's a good call and a good idea. Will do

packages/cli/commands/project/listBuilds.js
anthmatic: Yep! This endpoint always returns an empty "last page" and I flagged with the BE folks and they have a fix in progress

packages/cli/commands/project/listBuilds.js
anthmatic: Handling this here https://github.com/HubSpot/hubspot-cli/pull/580/files#diff-fa4fa15b34bd2259fc2c24ec990fdbaa6492e6030d940700895318b399058174R64

packages/cli/commands/project/deploy.js
```diff
@@ -44,7 +48,9 @@ exports.handler = async options => {
 
   trackCommandUsage('project-deploy', { projectPath }, accountId);
 
-  const { projectConfig } = await getProjectConfig(projectPath);
+  const { projectConfig, projectDir } = await getProjectConfig(projectPath);
+
+  validateProjectConfig(projectConfig, projectDir);
```
anthmatic, line 53: Just adding this while I'm here

packages/cli/commands/project/listBuilds.js
anthmatic: I forgot we have `validateProjectConfig()` for this. Added here: https://github.com/HubSpot/hubspot-cli/pull/582/files#diff-36a35fb191a893b8f5de180e0cc2bd7656d4c00c972ef65a9687a6e3f8304817R67

brandenrodgers (approved):
Tested locally and works great!

PR #583, opened 2021-10-29
Fix bug where CLI appears to hang when build fails

anthmatic:
## Description and Context
Currently when polling for project status, when a build fails the CLI continues to look for a `deployId`, which results in the CLI appearing to hang.

This updates the check to skip the check for deploy Id on `FAILURE`


brandenrodgers (approved):
😬  I think I had initially suggested the broken logic here. Thanks for finding/fixing!

PR #585, opened 2021-11-01
Improve UX of project logs subcommand

gcorne:
## Description and Context

This PR makes the following improvements to `hs project logs`:

1. Leverage `hsproject.json` to determine the project name when command is run inside a project directory if `--projectName` is not passed
2. Remove `--overwrite` options from all `hs project` commands. The option is not relevant.
3. Improve documentation. Don't hide `--appPath` and `--projectName` from help and make clear that the function name is an "app function."

## Who to Notify
@miketalley @anthmatic @brandenrodgers

### DIFF
--- a/packages/cli/commands/project.js
+++ b/packages/cli/commands/project.js
@@ -1,8 +1,4 @@
-const {
-  addConfigOptions,
-  addAccountOptions,
-  addOverwriteOptions,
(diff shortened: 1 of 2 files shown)


brandenrodgers (approved):
lgtm 👍

PR #587, opened 2021-11-02
Adding process exits to upload command

brandenrodgers:
## Description and Context

This adds a `process.exit()` to the end of the project upload command. It will make it so consuming systems like the GH upload action can act accordingly when uploads fail.

This will exit with `process.exit(1)` when:
- Build status is "FAILURE"
- Deploy status is "FAILURE" (if autodeploy is enabled)

Everything else will exit with `process.exit(0)`

### DIFF
--- a/packages/cli/commands/project/upload.js
+++ b/packages/cli/commands/project/upload.js
@@ -122,6 +122,7 @@ exports.handler = async options => {
   const archive = archiver('zip');
 
   output.on('close', async function() {
+    let exitCode = 0;
     logger.debug(`Project files compressed: ${archive.pointer()} bytes`);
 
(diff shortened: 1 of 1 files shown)


anthmatic (approved):

anthmatic:
LGTM! I am wondering, does it make sense to have all commands explicitly exit with a code?

brandenrodgers:
> does it make sense to have all commands explicitly exit with a code?

@anthmatic are there any other ones you had in mind? I _think_ it is best practice to have all commands exit with a code (but I could be wrong about that). I'm actually attempting to help standardize the exit codes [here](https://github.com/HubSpot/hubspot-cli/pull/570).

PR #588, opened 2021-11-03
Reduce confusion around project create

anthmatic:
## Description and Context
This PR aims to reduce confusion around how `hs project config` works

- Removes `path` positional argument
- Flag options now include: `--name`, `--location`, and `--template`. All flags are required. If a flag is not set, the CLI display a prompt.
- The `--name` option will inform the default value for `--location`: 
image
- If the `--location` already contains a project, a warning is displayed and the user is prompted to overwrite: 
image

- (TBD) Disallows creating a project nested within another project
  - Looks at ancestors to make sure the project will not be nested
  - Looks at descendants to make sure that project won't encapsulate an existing project

## TODO

- TBD / Check within child directories for existing `hsproject.json`


packages/cli/lib/projects.js
```diff
@@ -48,6 +49,11 @@ const PROJECT_STRINGS = {
   },
 };
 
+const containsNestedProject = async projectPath => {
+  const files = await walk(projectPath);
+  return files.find(file => path.basename(file) === 'hsproject.json');
```
brandenrodgers, line 54: we should probably create a const for `hsproject.json` if one doesn't already exist

packages/cli/lib/projects.js
```diff
@@ -65,82 +71,76 @@ const getProjectConfig = async projectPath => {
...
+    };
   } catch (e) {
     logger.error('Could not read from project config');
   }
 };
 
-const createProjectConfig = async (projectPath, projectName) => {
-  const projectConfig = await getProjectConfig(projectPath);
-  const projectConfigPath = path.join(projectPath, 'hsproject.json');
+const createProjectConfig = async (projectPath, projectName, template) => {
```
brandenrodgers, line 89: Could you elaborate on the different cases being handled here? It's a little tricky to follow

packages/cli/commands/project/create.js
```diff
@@ -32,42 +31,93 @@ const loadAndValidateOptions = async options => {
...
+        }
+        return true;
+      },
+    },
+    {
+      name: 'template',
+      message: () => {
+        return options.template &&
+          !PROJECT_TEMPLATES.find(t => t.name === options.template)
+          ? `[--template] Could not find template ${options.template}. Please choose an available template.`
```
miketalley, line 73: Should we output a list of available templates here?

miketalley (commented):
Nice work on this! I have a couple of questions, otherwise looks good!

packages/cli/commands/project/create.js
anthmatic: Yep, this `rawlist` is actually shown with all available templates. Choices are added here: https://github.com/HubSpot/hubspot-cli/pull/588/files#diff-846ab0ff6b94aadce7bb7339adf043c6f81967013a7dfba6dd5e5ee83a055985R85

packages/cli/lib/projects.js
anthmatic: I'm actually going to rework this now since we're going to allow developers to move forward even if a project exists so I'll make sure this is clearer.

There are 3 cases here:
- Project already exists at `--location`
- Project exists in a parent of `--location`
- Project exists in a child folder of `--location`

The last "child folder" check can be pretty slow if there are a lot of files and since we're allowing devs to override, I'm thinking it make sense to remove this case.

anthmatic: Took a pass at updating this FYI

brandenrodgers (approved):
Tested locally and this is working great! 🎉

PR #590, opened 2021-11-04
Fix bug while tailing logs for new function

anthmatic:
## Description and Context
An unexpected API error is displayed if you run `hs logs <route> --followw` on a function that has no log data / previous executions

### DIFF
--- a/packages/cli/commands/logs.js
+++ b/packages/cli/commands/logs.js
@@ -63,7 +63,7 @@ const endpointLog = async (accountId, options) => {
       getFunctionLogs(accountId, functionPath, { after });
     const fetchLatest = () => {
       try {
-        getLatestFunctionLog(accountId, functionPath);
+        return getLatestFunctionLog(accountId, functionPath);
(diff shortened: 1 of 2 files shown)


packages/cli/commands/logs.js
```diff
@@ -63,7 +63,7 @@ const endpointLog = async (accountId, options) => {
       getFunctionLogs(accountId, functionPath, { after });
     const fetchLatest = () => {
       try {
-        getLatestFunctionLog(accountId, functionPath);
+        return getLatestFunctionLog(accountId, functionPath);
```
brandenrodgers, line 66: Seems reasonable, do we need to return in the catch as well?

anthmatic: Doubled checked with @miketalley on this: I believe the intent is for polling to continue and `handleLogsError()` simply logs an error as a side effect.

brandenrodgers (approved):
👍

PR #591, opened 2021-11-09
Update project logs

anthmatic:

### DIFF
--- a/packages/cli/commands/project/logs.js
+++ b/packages/cli/commands/project/logs.js
@@ -56,28 +56,29 @@ const appFunctionLog = async (accountId, options) => {
 
   let logsResp;
 
+  const tailCall = after =>
+    getProjectAppFunctionLogs(accountId, functionName, projectName, appPath, {
+      after,
+    });
+  const fetchLatest = () => {
+    try {
+      return getLatestProjectAppFunctionLog(
+        accountId,
+        functionName,
+        projectName,
+        appPath
+      );
+    } catch (e) {
+      handleLogsError(e, accountId, projectName, appPath, functionName);
+    }
+  };
+
   if (follow) {
     const spinnies = new Spinnies();
 
     spinnies.add('tailLogs', {
       text: `Waiting for log entries for '${functionName}' on account '${accountId}'.\n`,
     });
-    const tailCall = after =>
-      getProjectAppFunctionLogs(accountId, functionName, projectName, appPath, {
-        after,
-      });
-    const fetchLatest = () => {
-      try {
-        return getLatestProjectAppFunctionLog(
-          accountId,
-          functionName,
-          projectName,
-          appPath
-        );
-      } catch (e) {
-        handleLogsError(e, accountId, projectName, appPath, functionName);
-      }
-    };
 
     await tailLogs({
       accountId,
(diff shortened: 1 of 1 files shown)


markhazlewood:
Couple comments/questions about the params:

* `--functionName` - Propose we shorten this to just `--function`. In general including "name" in the label for something is redundant unless there's another common way to identify it. In this case I think `function` is clear enough.
* `--appPath` - What exactly is this a path to? Is this still using the DFS tree path of the `app.json` file? If so, I'm not sure that's the best identifier since in the current workflow we don't surface the DFS file tree. Can we just provide a unique name and use `--app` instead?

anthmatic:
Changing to `--function` makes sense to me.

`--appPath` refers to the path to the app within your project. Using https://github.com/HubSpot/getting-started-project-template as an example, the value for this option would be `app`

packages/cli/commands/logs.js
```diff
@@ -77,7 +77,7 @@ const endpointLog = async (accountId, options) => {
 };
 
 exports.command = 'logs [endpoint]';
-exports.describe = 'get logs for a function';
```
brandenrodgers, line 80: Does this logs command only work for project functions now?

packages/cli/commands/project/logs.js
```diff
@@ -13,19 +13,25 @@ const {
...
 const { getProjectConfig } = require('../../lib/projects');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const { tailLogs } = require('../../lib/serverlessLogs');
 
 const handleLogsError = (e, accountId, projectName, appPath, functionName) => {
   if (e.statusCode === 404) {
     logger.error(
-      `No logs were found for the function name '${functionName}' in the app path '${appPath}' within the project '${projectName}' in account ${accountId}.`
+      appPath
+        ? `No logs were found for the function name '${functionName}' in the app path '${appPath}' within the project '${projectName}' in account ${accountId}.`
```
kemmerle, line 28: Is there any way to rephrase this? It's four prepositional phrases in a row, I'm getting a little lost. Perhaps we could consult Erin.

miketalley (commented):
Should we handle someone running `hs logs`? When I run this now, I get ```[ERROR] No logs were found for the function path 'undefined' in account 101867970.```

packages/cli/commands/project/logs.js
```diff
@@ -128,22 +135,22 @@ exports.handler = async options => {
...
+        alias: 'function',
+        describe: 'Serverless app function name or endpoint route',
         type: 'string',
         demandOption: true,
       },
+      appPath: {
+        describe: 'Path to app folder, relative to project',
+        type: 'string',
+      },
       projectName: {
```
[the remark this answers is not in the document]
anthmatic: Fixed on https://github.com/HubSpot/hubspot-cli/pull/591/files#diff-0caa64de5ff3fe3c960557dd2d69784fef280690b5c63e004cef47b0e4e7bdacR121

packages/cli-lib/lang/en.lyaml
```diff
@@ -441,9 +441,9 @@ en:
           logs:
             describe: "Get logs for a function within a project"
             errors:
-              appPathRequired: "You must specify an app path using the --appPath argument."
               functionNameRequired: "You must pass a function name to retrieve logs for."
-              logs: "No logs were found for the function name \"{{ functionName }}\" in the app path \"{{ appPath }}\" within the project \"{{ projectName }}\" in account {{ accountId }}."
+              noAppFunctionLogs: "No logs were found for the function name \"{{ functionName }}\" in the app path \"{{ appPath }}\" within the project \"{{ projectName }}\" in account {{ accountId }}."
+              noEndpointLogs: "No logs were found for the function name \"{{ functionName }}\" within the project \"{{ projectName }}\" in account {{ accountId }}."
```
anthmatic, line 446: @erincouse was hoping to get your feedback on these errors. They're shown when no logs are found

cc @kemmerle

brandenrodgers (approved):

PR #592, opened 2021-11-09
Add util for getting account description

anthmatic:
## Description and Context
An alternative to #575. This PR achieves the same goal of displaying `Account Name (Account ID)` but can be called anywhere that `accountId` is known and eliminates the need to pass around the `accountDescription` variable

This also introduces `cli/lib/ui` for CLI UI related utils

### DIFF
--- a/packages/cli/lib/ui.js
+++ b/packages/cli/lib/ui.js
@@ -1,3 +1,4 @@
+const chalk = require('chalk');
 const supportsHyperlinks = require('supports-hyperlinks');
 const { getAccountConfig } = require('@hubspot/cli-lib/lib/config');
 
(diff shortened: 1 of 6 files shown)


brandenrodgers (approved):

packages/cli/lib/ui.js
```diff
@@ -0,0 +1,37 @@
...
+  const account = getAccountConfig(accountId);
+  return chalk.bold(
+    account.name ? `${account.name} (${account.portalId})` : account.portalId
+  );
+};
+
+module.exports = {
+  link,
+  getAccountDescription,
+};
```
anthmatic, line 37: @brandenrodgers I requested another review because I thought it might make sense to pull these "ui" related utils together in the CLI package but wanted to get your 👍 first

brandenrodgers: I like it 👍

brandenrodgers (approved):

PR #595, opened 2021-11-10
DRY up the command config validation

brandenrodgers:
## Description and Context

DRY-ing up the command config validation that (almost) every command uses. We had a util for it that the commands weren't using.

## TODO

- [x] Test to make sure I haven't broken anything


anthmatic (approved):
Much cleaner! I tested out a bunch of commands (but not all) and they worked as expected

PR #599, opened 2021-11-15
Encode theme folder when generating theme preview link

### LINKED ISSUE #596: "To preview theme, visit: " should be URL encoded
### Description
The preview theme link that Hubspot generates fails if the theme name has a space in it since it isn't URL encoded.

### Steps to reproduce
`hs watch` a theme that isn't the published one and a preview link appears: `https://app.hubspot.com/theme-previewer/{account number}/edit/{theme name}` 

## Expected behavior
The theme name should be URL encoded, ie spaces converted to `%20`, etc,

anthmatic:
## Description and Context
Fixes #596 

Encodes the theme folder when generating a link to the theme previewer.

## Screenshots
image

## TODO

When testing this, I did find a flaw in the way this link is generated. If the theme is uploaded to a different folder, the preview link will not be accurate. For example

`hs upload "cool theme" theme2000` would yield a link to `/theme-previewer/1234567/edit/cool%20theme` when it should point to  `/theme-previewer/1234567/edit/theme2000`

Filed in https://github.com/HubSpot/hubspot-cli/issues/600

### DIFF
--- a/packages/cli-lib/lib/files.js
+++ b/packages/cli-lib/lib/files.js
@@ -25,7 +25,9 @@ const getThemePreviewUrl = (filePath, accountId) => {
     getEnv() === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD
   );
 
-  return `${baseUrl}/theme-previewer/${accountId}/edit/${themeName}`;
+  return `${baseUrl}/theme-previewer/${accountId}/edit/${encodeURIComponent(
+    themeName
+  )}`;
 };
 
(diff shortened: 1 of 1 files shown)


brandenrodgers:
> When testing this, I did find a flaw in the way this link is generated. If the theme is uploaded to a different folder, the preview link will not be accurate. For example

Interesting. It sounds like we're using src as the theme path. Should we use `dest` instead?

brandenrodgers (approved):
👍

PR #601, opened 2021-11-16
Fix issue where `--use-env` does not work with `hs fetch`

### LINKED ISSUE #598: `--use-env` flag fails
## Description and Context
The `--use-env` flag is supposed to allow the use of env variables instead of needing a `hubspot.config.yml` according to the docs: https://developers.hubspot.com/docs/cms/developer-reference/local-development-cli#environment-variables. The docs specifically outline its usage for GitHub Actions

>The HubSpot CLI supports the use of environment variables, this can be especially useful when creating automations like a GitHub Action.

Version: 3.0.9

### Steps to reproduce
1. Create a GitHub Action that uses `--use-env` flag with `hs` command (or just run from a shell script).

Here's an example from my `deploy.yml`

```yml
    - name: Deploy
      env:
        HUBSPOT_PORTAL_ID: ${{ secrets.HUBSPOT_PORTAL_ID }}
        HUBSPOT_PERSONAL_ACCESS_KEY: ${{ secrets.HUBSPOT_PERSONAL_ACCESS_KEY }}
      run: |
        npx hs upload dist cms-theme-dir --use-env
      shell: bash
```

### Expected behavior

Flag successfully removes need for a hubspot yml config file. It worked just the day prior to this:

Actual behavior: it errors out asking for a config file.

```
Run npx hs upload dist cms-theme-dir --use-env
  npx hs upload dist cms-theme-dir --use-env
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    HUBSPOT_PORTAL_ID: ***
    HUBSPOT_PERSONAL_ACCESS_KEY: ***
Error:  A hubspot.config.yml file could not be found
Error:  No config was found
Error: Process completed with exit code 1.
```

anthmatic:
## Description and Context
`hs fetch` was not aware of the value of `--use-env`, so it always defaulted to trying to load a config from the filesystem

fixes #598

### DIFF
--- a/packages/cli/commands/fetch.js
+++ b/packages/cli/commands/fetch.js
@@ -31,7 +31,7 @@ exports.handler = async options => {
   setLogLevel(options);
   logDebugInfo(options);
 
-  loadConfig(configPath);
+  loadConfig(configPath, options);
   checkAndWarnGitInclusion();
 
   if (


anthmatic:
@brandenrodgers I just realized that this will probably get fixed in #595. Are you ok if I get this quick fix in?

brandenrodgers (approved):
@anthmatic yeah totally 👍  let's get the quick fix in

PR #602, opened 2021-11-18
Make sure polling errors are always exposed

anthmatic:
## Description and Context
Previously polling errors were only shown when a build failed and not when an auto or manual deploy had failed. This moved the error handling logic closes to the abstract poll-er so that it always shows up.

## Screenshots

**Failed build:** 
image

**build succeeds, failed auto deploy:** 
image

**Failed manual deploy:** 
image

##To test:
-To test a failed build you can upload a malformed `theme.json` file
-To test a failed deploy you can upload a malformed scheme json file

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -39,6 +39,10 @@ const PROJECT_STRINGS = {
       } in this project ...\n`,
     SUCCESS: name => `Built ${chalk.bold(name)}`,
     FAIL: name => `Failed to build ${chalk.bold(name)}`,
+    SUBTASK_FAIL: (taskId, name) =>
+      `Build #${taskId} failed because there was a problem\nbuilding ${chalk.bold(
+        name
+      )}`,
   },
   DEPLOY: {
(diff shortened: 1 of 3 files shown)


packages/cli/lib/projects.js
```diff
@@ -330,6 +338,31 @@ const makeGetTaskStatus = taskType => {
...
+              logger.log(
+                `${statusStrings.SUBTASK_FAIL(
+                  buildId || taskId,
+                  failedSubtask.length === 1
+                    ? failedSubtask[0][statusText.SUBTASK_NAME_KEY]
+                    : failedSubtask.length + ' components'
+                )}\n`
+              );
+              logger.log('See below for a summary of errors.');
+              logger.log('-'.repeat(50));
```
brandenrodgers, line 356: Not necessary to do in this PR, but seeing your recent PR that added a "ui.js" file makes me think that things like this horizontal line could maybe go in there.

anthmatic: 100%

brandenrodgers (approved):
tested locally and lgtm!

PR #603, opened 2021-11-19
Improve UI utils

anthmatic:
## Description and Context
Adds/renames some UI related utils:
- `uiLink`: output hyperlink
- `uiLine`: output a horizontal line
- `uiAccountDescription`: output formatted account description

## Screenshots
image

### DIFF
--- a/packages/cli/lib/ui.js
+++ b/packages/cli/lib/ui.js
@@ -8,7 +8,7 @@ const { logger } = require('@hubspot/cli-lib/logger');
  *
  * @returns
  */
-const horizontalLine = () => {
+const uiLine = () => {
   logger.log('-'.repeat(50));
 };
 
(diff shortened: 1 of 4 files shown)


packages/cli/lib/ui.js
```diff
@@ -1,6 +1,16 @@
...
 const supportsHyperlinks = require('supports-hyperlinks');
 const { getAccountConfig } = require('@hubspot/cli-lib/lib/config');
+const { logger } = require('@hubspot/cli-lib/logger');
+
+/**
+ * Outputs horizontal line
+ *
+ * @returns
+ */
+const horizontalLine = () => {
```
brandenrodgers, line 11: This all looks great to me 👍 

I wonder if it's worth trying to standardize the naming of these utils a little bit to make their purpose more clear. Not a blocker though, just an idea.

brandenrodgers (approved):

packages/cli/lib/ui.js
anthmatic: Are you thinking along the lines of `uiLink` and `uiLine` or something else?

brandenrodgers: Yeah exactly, just something that lets the user know that it's a UI util.

Issue #550, opened 2021-09-08
Command `hs` conflicts with another common npm package


aminimalanimal:
## Description and Context
When trying to follow along with the tutorials, the `hs init` command wasn’t working for me. It turns out that the `hs` command was already globally taken by [http-server](https://www.npmjs.com/package/http-server).

I make it a point not to install global npm scripts generally, but both @hubspot/cli and http-server are the sorts of packages that could make sense to have together in the global namespace.

`http-server` has actually already implemented a change to remove the `hs` alias from their script by default, so upgrading it and reinstalling @hubspot/cli sufficed. However, I think that hubspot should be aware of this possibility and also adopt the approach of using a longer, more unique, less-likely-to-conflict namespace (like `hubspot`) and allow end-users to alias it with shorter commands if desired.

Cleary a change like that will need to happen over multiple versions of the package. In the meantime, I do suggest aliasing `hubspot` so that it can be used instead in instances of conflict.

brandenrodgers:
Thanks for reporting this! I think supporting `hubspot` as an alternate option is a great place to start 👍

chiragchadha1:
Going to close this one out. It's been open for 4+ years without further demand, and as you noted, http-server has since dropped their `hs` alias so the conflict is gone in practice. Feel free to re-open if you're hitting a new conflict. Thanks!

PR #604, opened 2021-11-19
Support hubspot as alternate to hs

brandenrodgers:
## Description and Context

Issue: #550

Adds `hubspot` as an alternate way to run our CLI commands.

Example:
`hs upload` will still work, but so will `hubspot upload template.html template.html`


brandenrodgers:
whoops, didn't mean to PR this one. Ignore it

PR #605, opened 2021-11-19
Support using hubspot as alternate to hs

brandenrodgers:
## Description and Context

Issue: #550

Adds `hubspot` as an alternate way to run our CLI commands.

Example:
`hs upload` will still work, but so will `hubspot upload template.html template.html`

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -34,7 +34,8 @@
   },
   "bin": {
     "hs": "./bin/hs",
-    "hscms": "./bin/hscms"
+    "hscms": "./bin/hscms",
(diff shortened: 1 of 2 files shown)


packages/cli/package.json
```diff
@@ -34,7 +34,8 @@
   },
   "bin": {
     "hs": "./bin/hs",
-    "hscms": "./bin/hscms"
+    "hscms": "./bin/hscms",
+    "hubspot": "./bin/hubspot"
```
anthmatic, line 38: do we also need to create this file in https://github.com/HubSpot/hubspot-cli/tree/master/packages/cli/bin?

PR #608, opened 2021-11-23
Change failed status icon to exclamation mark

kemmerle:
## Description and Context
As @anthmatic noted, in the original plans for a failed project build or deploy, we wanted to use :exclamation: as a status icon instead of a red :x:. 

However, when we attempted to add ❗ as the status icon, the character spacing was off. I've done some research on resizing Unicode characters in a CLI environment, and I don't believe there's any easy way to adjust it. 

As a solution, I've added a bold **!** as a failed status icon for a similar effect. If we would rather go with the Unicode character, however, I'm all ears for alternate approaches :) 

## Screenshots

With changes:

Screen Shot 2021-11-23 at 2 04 09 PM

## Who to Notify
 @brodgers @anthmatic

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -271,6 +271,7 @@ const makeGetTaskStatus = taskType => {
     const spinnies = new Spinnies({
       succeedColor: 'white',
       failColor: 'white',
+      failPrefix: '\033[1m!\033[0m',
     });
 
     spinnies.add('overallTaskStatus', { text: 'Beginning' });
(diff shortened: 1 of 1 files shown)


packages/cli/lib/projects.js
```diff
@@ -271,6 +271,7 @@ const makeGetTaskStatus = taskType => {
     const spinnies = new Spinnies({
       succeedColor: 'white',
       failColor: 'white',
+      failPrefix: '\033[1m!\033[0m',
```
anthmatic, line 274: Nice! In the CLI, we've use `chalk` a lot for formatting text output. I suspect what you have here will work just fine across different terminals, but it might be worth confirming that

brandenrodgers: Yeah +1 to using chalk here if possible, just for consistency.

brandenrodgers (approved):
👌

PR #611, opened 2021-11-30
Replace handlebars dep with custom code

miketalley:
## Description and Context
After [discussing the use of the `handlebars` dependency](https://github.com/HubSpot/hubspot-cli/pull/476#discussion_r744257905) for [string centralization](https://github.com/HubSpot/hubspot-cli/pull/476), we decided to look into writing custom code to handle the delimiters within the `en.lyaml` file instead of adding a new dependency.

This is a branch of `testing/yaml-for-strings` to make review of custom handlebars replacement code easier.

## TODO
- [x] Create regex to check for {{ interpolation delimiters }} and replace with interpolation data or `''` if not provided (similar functionality to `handlebars` here)
- [x] Create regex that checks for {{ #helper }}...{{ /helper }} and runs helper function on `...` content between helper identifiers
- [x] Add custom helpers previously used in handlebarsCustomHelpers.js
  - [x] `bold`
  - [x] `yellow`
- [x] Clean up code and break up into separate methods and files

## Who to Notify
@brandenrodgers @gcorne @anthmatic @kemmerle @bmatto


yarn.lock
```diff
@@ -4309,7 +4309,7 @@ growly@^1.3.0:
   resolved "https://registry.yarnpkg.com/growly/-/growly-1.3.0.tgz#f10748cbe76af964b7c96c93c6bcc28af120c081"
   integrity sha1-8QdIy+dq+WS3yWyTxrzCivEgwIE=
 
-handlebars@^4.7.6, handlebars@^4.7.7:
+handlebars@^4.7.6:
```
miketalley, line 4312: 😂

packages/cli-lib/lib/interpolation.js
```diff
@@ -0,0 +1,139 @@
...
+
+const isHelperIdentifier = identifier => {
+  return (
+    identifier.startsWith(delimiters.helpers.start) ||
+    identifier.startsWith(delimiters.helpers.end)
+  );
+};
+
+const generateReplaceFn = (matchedText, startIndex, replacementString) => {
+  return theString => {
```
kemmerle, line 21: Could we replace this variable with something more descriptive?

packages/cli-lib/lib/interpolation.js
```diff
@@ -0,0 +1,139 @@
...
+    const newString = `${theString.slice(0, startIndex)}${replacementString ||
+      ''}${theString.slice(startIndex + matchedText.length)}`;
+    return newString;
+  };
+};
+
+/**
+ * Interpolate a string with data
```
kemmerle, line 29: I appreciate the detailed notes for these functions. I was going to suggest perhaps combining `interpolation` and `compileHelper`, since they're so similar, but honestly, I think that keeping them separate helps with readability.

packages/cli-lib/lib/text.js
```diff
@@ -8,6 +10,20 @@ const commaSeparatedValues = (arr, conjunction = 'and', ifempty = '') => {
   return arr.join(', ');
 };
 
+/**
+ * These helper methods are used to modify text output within the CLI. They
+ * should all take in a string value and output a modified string value.a
```
kemmerle, line 15: Typo here with ending `a`.

kemmerle (approved):
@miketalley, I reviewed the changes, and they make sense to me. I only had one minor suggestion to change a variable name.

packages/cli-lib/lib/text.js
```diff
@@ -8,6 +10,20 @@ const commaSeparatedValues = (arr, conjunction = 'and', ifempty = '') => {
...
+ */
+const helpers = {
```
brandenrodgers, line 17: This might be better off living in the interpolation file. I can imagine someone adding extra options to this object and using it outside of the interpolation context. Every new key added here results in another call to `compileHelper` in the interpolation file.

packages/cli-lib/lib/interpolation.js
```diff
@@ -0,0 +1,139 @@
...
+
+  return compiledString;
+};
+
+const compileHelpers = stringValue => {
+  return Object.keys(helpers).reduce((currentStringValue, helperIdentifier) => {
+    return compileHelper(
+      currentStringValue,
+      helperIdentifier,
+      helpers[helperIdentifier]
```
brandenrodgers, line 117: not a blocker, but the `compileHelper` func also has access to `helpers` so there's no need to pass this in as an arg.

brandenrodgers:
I left a few small comments, but overall this approach lgtm! 👍

packages/cli-lib/lib/text.js
miketalley: I originally had this in a separate `interpolationHelpers` file, but after looking at it, I believe this could be used more generally to modify any other text in the CLI.

packages/cli-lib/lib/interpolation.js
miketalley: True, this is my functional programming instincts kicking in here 😂

miketalley: I'm thinking leaving this one as-is because it doesn't hurt anything and makes the function more functional and self-contained.

PR #614, opened 2021-12-02
Replace ora with spinnies for loading indicators

kemmerle:
## Description and Context
This PR stemmed from a discussion @miketalley and I had, while he was helping me onboard with the CLI. We decided to use the `spinnies` package for loading indicators for project commands, and Mike suggested that I replace `ora` with `spinnies` throughout the CLI. 

## TODO
There is still a use of `ora` in`app/deploy.js`. After some discussion on Slack, it was suggested that we retire this command, so I didn't think it made sense to update it.


packages/cli/commands/functions/deploy.js
```diff
@@ -68,21 +62,22 @@ exports.handler = async options => {
...
+    spinnies.remove('buildAndDeployStatus');
+    spinnies.stopAll();
     await outputBuildLog(successResp.cdnUrl);
     logger.success(
       `Built and deployed bundle from package.json for ${functionPath} on account ${accountId} in ${buildTimeSeconds}s.`
     );
   } catch (e) {
-    spinner && spinner.stop && spinner.stop();
+    spinnies && spinnies.remove && spinnies.remove('buildAndDeployStatus');
+    spinnies && spinnies.stopAll && spinnies.stopAll();
```
brandenrodgers, line 80: Are these checks necessary (I assume you were just copying the existing pattern)? We might be able to remove them.

kemmerle: Initializing the spinnies outside the try...catch block makes the checks unnecessary, thanks for the callout 👍 

A note: I was initially trying to remove the `spinnies` in a ...finally block, but that doesn't remove the spinnies in time for the error message in the catch block to correctly display. Ergo, we need to remove and stop spinnies in both blocks.

packages/cli/commands/create/api-sample.js
```diff
@@ -31,13 +31,17 @@ module.exports = {
         return;
       }
     }
-    const downloadSpinner = ora('Loading available API samples');
-    downloadSpinner.start();
+    const spinnies = new Spinnies();
+    spinnies.add('fetchingSampleAppsStatus', {
+      text: 'Loading available API samples',
```
brandenrodgers, line 36: If you merge this branch with master, you should be able to make use of the new lyaml file that @miketalley added! We should get into the habit of adding any new copy to that file.

brandenrodgers:
Is this replacing every instance of Ora? If so, can we remove the ora dep?

kemmerle:
@brandenrodgers, there is still a use of `ora` in the `app deploy` command, but I have another PR out to delete that command altogether. I believe as long as I merge that PR first, we should be safe to remove the ora dependency entirely 👍

packages/cli/commands/functions/deploy.js
```diff
@@ -71,26 +77,22 @@ exports.handler = async options => {
...
-        accountId,
-        functionPath,
-      })
-    ).start();
     const buildId = await buildPackage(accountId, functionPath);
     const successResp = await pollBuildStatus(accountId, buildId);
     const buildTimeSeconds = (successResp.buildTime / 1000).toFixed(2);
-    spinner.stop();
+    spinnies.remove('buildAndDeployStatus');
+    spinnies.stopAll();
```
anthmatic, line 84: I could be mistaken but I don't think you need `stopAll()` if the spinner has been removed

anthmatic: By no means a blocker, but the other thing I'm wondering is whether we should use the `succeed()` or `fail()` methods that spinnies provides to potentially improve the UX. Haven't given much thought to that, but figured I'd mention it now that we have new functionality available

kemmerle: @anthmatic, I had to do a bit of research to uncover this workflow, but if I simply use the `remove()` method without `stopAll()`, the CLI command hangs forever. The `stopAll()` command allows the command to finish in the absence of `succeed()` or `fail()`. 

I wasn't sure whether we wanted to continue to use `logger` for messaging, but I love your suggestion to use `succeed()` or `fail()` in place of the clunkier `remove()` and `stopAll()` methods. Hopping on it now.

kemmerle: @anthmatic, coming back with an update on this.

I tried to replace our `logger.success` and `logger.error` messages with `spinnies.succeed` and `spinnies.fail`, but I ran into some issues. 

In the `api-sample` command, the initial message `Loading available API samples` is visible for two steps where the customer first chooses the API sample app and second the app's language. We need to stop the spinnie before the customer chooses this, but there's no way to stop a spinnie without removing it entirely (please tell me if I'm wrong, but I did read through the docs pretty extensively and looked through sample code). 

In the `functions deploy` command, while using spinnies, I always hit the `else` command [here](https://github.com/HubSpot/hubspot-cli/blob/1c181ac74129869a377caea1f299990e6f96ce49/packages/cli/commands/functions/deploy.js#L111), even when I succeed in deploying the function. 

Having experimented with this for a bit now, I would suggest that we keep the code as is.

anthmatic (approved):

brandenrodgers:
@kemmerle what's the status of this? Should we keep pushing on it?

PR #616, opened 2021-12-02
Retire app deploy command

kemmerle:
## Description and Context
In working on #614, we realized that we could retire the `hs app deploy` command, since it's been superseded by projects work.

## Who to Notify

@gcorne


packages/cli/commands/app/deploy.js
```diff
@@ -1,110 +0,0 @@
...
-const {
-  getAccountId,
-  addUseEnvironmentOptions,
-} = require('../../lib/commonOpts');
-const { trackCommandUsage } = require('../../lib/usageTracking');
-const { outputBuildLog } = require('../../lib/serverlessLogs');
-const { loadAndValidateOptions } = require('../../lib/validation');
-const { i18n } = require('@hubspot/cli-lib/lib/lang');
-
-const i18nKey = 'cli.commands.app.subcommands.deploy';
```
anthmatic, line 20: I think we'll also need to remove the translations for this command too

brandenrodgers:
I think you can remove [this file](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/app.js) too, since the only "app" command was "deploy". I'd also double check to make sure nothing else in the project is referencing these deleted files.

brandenrodgers (approved):

PR #621, opened 2021-12-13
Add projects watch

anthmatic:
## Description and Context
Adds `hs project watch` which watches for file changes in a project. When changes are detected, it creates a staging build and queues up any additional changes. When 2 seconds elapses without any chances, the build is kicked off.

When a user hits `ctrl+c`, and there is an active staging build, the CLI will attempt to cancel the build to clean up.

## Screenshots

image
image


gcorne:
Maybe this already in the works, but you should be able to use `process.on()` to listen for signals from the system. I think `process.on('SIGINT', handler)` should fire when `CTL+C` is issued.

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+
+  const { path: projectPath } = options;
+  const accountId = getAccountId(options);
+
+  trackCommandUsage('project-watch', { projectPath }, accountId);
+
+  const { projectConfig, projectDir } = await getProjectConfig(projectPath);
+
+  validateProjectConfig(projectConfig, projectDir);
+
```
gcorne, line 214: I think we should automatically stage a build when the command is started so that we're ready to go before any change has been made.

packages/cli-lib/api/dfs.js
```diff
@@ -153,6 +153,78 @@ async function fetchProjectSettings(portalId, projectName) {
...
+ *
+ * @async
+ * @param {string} projectName
+ * @returns {Promise}
+ */
+async function provisionBuild(portalId, projectName) {
+  return http.post(portalId, {
+    uri: `${PROJECTS_API_PATH}/${encodeURIComponent(
+      projectName
+    )}/provision-build`,
```
gcorne, line 167: I think that this needs a longer timeout since provisioning can take some time.

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+    logger.debug('Attempting to create a new build');
+    const { buildId } = await provisionBuild(accountId, projectName);
+    return buildId;
+  } catch (err) {
+    if (err.error.subCategory === 'PipelineErrors.PROJECT_LOCKED') {
+      logger.error('Project is locked, cannot create new build');
+    } else if (err.error.subCategory === 'PipelineErrors.MISSING_PROJECT') {
+      logger.error(`Project ${projectName} does not exist.`);
+    } else {
+      logApiErrorInstance(err, new ApiErrorContext({ accountId, projectName }));
```
gcorne, line 194: If we get here, I think we should cancel the build so that it ends up being unlocked.

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+
+    try {
+      await queueBuild(accountId, projectName, currentBuild.get());
+      logger.debug(i18n(`${i18nKey}.debug.buildStarted`, { projectName }));
+    } catch (err) {
+      logApiErrorInstance(err, new ApiErrorContext({ accountId, projectName }));
+      return;
+    }
+
+    await pollBuildStatus(accountId, projectName, currentBuild.get());
```
gcorne, line 130: Are we planning to also incorporate the deploy status?

gcorne (commented):
It is working!

packages/cli/commands/project/watch.js
anthmatic: I'll do this and also make sure we unlock on SIGINT

anthmatic:
Thanks for the 👀 @gcorne

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+    if (queue.isPaused) {
+      standbyeQueue.push({
+        filePath,
+        remotePath,
+      });
+      return;
+    }
+    await queueFileUpload(accountId, projectConfig.name, filePath, remotePath);
+  });
+  watcher.on('change', async filePath => {
```
brandenrodgers, line 237: Does change trigger on file delete?

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+const {
+  getProjectConfig,
+  validateProjectConfig,
+  pollBuildStatus,
+} = require('../../lib/projects');
+const { i18n } = require('@hubspot/cli-lib/lib/lang');
+
+const i18nKey = 'cli.commands.project.subcommands.watch';
+
+const loadAndValidateOptions = async options => {
```
brandenrodgers, line 40: No need to define this. You should be able to import from [here](https://github.com/HubSpot/hubspot-cli/blob/bc84343fa5b2ab522a4bf3533141a0d3bbcaec25/packages/cli/lib/validation.js#L29)

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+    return buildId;
+  } catch (err) {
+    if (err.error.subCategory === 'PipelineErrors.PROJECT_LOCKED') {
+      logger.error('Project is locked, cannot create new build');
+    } else if (err.error.subCategory === 'PipelineErrors.MISSING_PROJECT') {
+      logger.error(`Project ${projectName} does not exist.`);
+    } else {
+      logApiErrorInstance(err, new ApiErrorContext({ accountId, projectName }));
+    }
+    process.exit(1);
```
brandenrodgers, line 196: This is somewhat of a newer thing, but we have constants for process exit types now ([here](https://github.com/HubSpot/hubspot-cli/blob/bc84343fa5b2ab522a4bf3533141a0d3bbcaec25/packages/cli/lib/enums/exitCodes.js))

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+  concurrency: 10,
+});
+const standbyeQueue = [];
+let timer;
+
+const processStandByQueue = async (accountId, projectName) => {
+  await currentBuild.update(accountId, projectName);
+  queue.addAll(
+    standbyeQueue.map(({ filePath, remotePath }) => {
+      return async () => {
```
brandenrodgers, line 62: You might be able to DRY things up a little between this chunk of code and [this one](https://github.com/HubSpot/hubspot-cli/pull/621/files#diff-058f6fead69eab728e1c40ba17719a33d3cac89af8ef94de43e6cf127e90824fR164).

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,268 @@
...
+
+  const watcher = chokidar.watch(path.join(projectDir, projectConfig.srcDir), {
+    ignoreInitial: true,
+    ignored: file => shouldIgnoreFile(file),
+  });
+
+  watcher.on('ready', () => {
+    logger.log(i18n(`${i18nKey}.logs.watching`, { projectPath }));
+  });
+  watcher.on('add', async filePath => {
```
brandenrodgers, line 223: The first param here can be an array of strings. `add` and `change` look identical so I _think_ this can just be `watcher.on(['add', 'change']...`

packages/cli/commands/project/watch.js
anthmatic: BE does not support this but I did create an internal issue for this

packages/cli/commands/project/watch.js
anthmatic: Is this documented somewhere? I don't see it in [the docs](https://github.com/paulmillr/chokidar) and in testing, that didn't work.

packages/cli/commands/project/watch.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (isAutoDeployEnabled && deployStatusTaskLocator) {
+    await pollDeployStatus(
+      accountId,
+      projectName,
+      deployStatusTaskLocator.id,
+      buildId
+    );
+  }
+};
```
anthmatic, line 37: The polling status function uses a lot of CLI related stuff that doesn't exist in CLI-LIB (spinnies, for instance) so this is really just an effort to avoid adding new deps to CLI-LIB

packages/cli-lib/projectsWatch.js
```diff
@@ -0,0 +1,199 @@
...
+  debounceQueueBuild(accountId, projectName);
+};
+
+const createNewStagingBuild = async (accountId, projectName) => {
+  currentBuildId = await createNewBuild(accountId, projectName);
+  logger.log(
+    i18n(`${i18nKey}.logs.createNewBuild`, { buildId: currentBuildId })
+  );
+
+  bindSigIntHandler(accountId, projectName);
```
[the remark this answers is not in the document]
anthmatic: The reason I call this in `createNewStagingBuild()` is because `spinnies` [removes all event listeners](https://github.com/jcarpanelli/spinnies/blob/90e33ee6384eab6d4783a506301ab3dbaee24b00/index.js#L191) from `process`.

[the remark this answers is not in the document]
anthmatic: But that's a good point about the CLI context, I can probably move that over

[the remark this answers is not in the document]
gcorne: Oh, that spinnies code looks problematic since it calls `process.exit()`. Won't that get called ahead of any cleanup code we bind to `SIGINT`?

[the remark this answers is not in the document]
anthmatic: Yeah. Originally I was only setting the listener up on watcher.ready() but `spinnies` was removing my listener so I opted to add the listener upon new build creation.

If `spinnies` is active, it means that there the staging build has been queued so if a user hits `ctr + c`, the process will exit (`spinnies` handler). At that point, there's nothing to clean up. Once `spinnies` returns, a new build is created and the SIGINT handler is added again.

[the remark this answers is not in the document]
gcorne: Seems pretty fragile. I also found https://github.com/jcarpanelli/spinnies/issues/30 but the project hasn't seen a release since 2019 and no activity since 2020.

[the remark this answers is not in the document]
anthmatic: Yeah, at this point I think we should come up with an alternative for this library. For the sake of having a working POC, I can take another pass at hardening this using spinnies for now. Would you object to that?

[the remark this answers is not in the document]
gcorne: Totally fine with revisiting later.

packages/cli/commands/project/watch.js
brandenrodgers: Huh, yeah looks like you're right. At this point I forget where I initially saw it, but maybe I was mistakingly looking at the docs for one of the other `watcher` funcs.

PR #632, opened 2022-01-19
Support Projects-compatible theme boilerplate

brandenrodgers:
## Description and Context

Relevant cms-theme-boilerplate branch: https://github.com/HubSpot/cms-theme-boilerplate/pull/404

This updates the `hs create website-theme` command to dynamically download the Projects-compatible version of the theme boilerplate repo if the destination folder is within a Project.

I also added a `getIsInProject` util because I think that will be useful to have.

## Who to Notify

cc/ @jasonnrosa

### DIFF
--- a/packages/cli-lib/projects.js
+++ b/packages/cli-lib/projects.js
@@ -57,13 +57,14 @@ async function fetchReleaseData(repoName, tag = '') {
 async function downloadProject(
   repoName,
   tag = '',
-  releaseType = GITHUB_RELEASE_TYPES.RELEASE
+  releaseType = GITHUB_RELEASE_TYPES.RELEASE,
+  ref
 ) {
   try {
     let zipUrl;
     if (releaseType === GITHUB_RELEASE_TYPES.REPOSITORY) {
       logger.log(`Fetching ${releaseType} with name ${repoName}...`);
-      zipUrl = `https://api.github.com/repos/HubSpot/${repoName}/zipball`;
+      zipUrl = `https://api.github.com/repos/HubSpot/${repoName}/zipball/${ref}`;
     } else {
       const releaseData = await fetchReleaseData(repoName, tag);
       if (!releaseData) return;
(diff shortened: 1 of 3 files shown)


packages/cli-lib/projects.js
```diff
@@ -57,13 +57,14 @@ async function fetchReleaseData(repoName, tag = '') {
...
-  releaseType = GITHUB_RELEASE_TYPES.RELEASE
+  releaseType = GITHUB_RELEASE_TYPES.RELEASE,
+  ref
 ) {
   try {
     let zipUrl;
     if (releaseType === GITHUB_RELEASE_TYPES.REPOSITORY) {
       logger.log(`Fetching ${releaseType} with name ${repoName}...`);
-      zipUrl = `https://api.github.com/repos/HubSpot/${repoName}/zipball`;
+      zipUrl = `https://api.github.com/repos/HubSpot/${repoName}/zipball/${ref}`;
```
brandenrodgers, line 67: The github api supports passing in a branch name to specify which branch to download (it defaults to "main" or "master" I think). That means we won't be using any sort of published release when we download the project-compatible version of the boilerplate. Whatever is currently in that branch will get downloaded.

The alternative route is to create a separate repo for the project-compatible version of the boilerplate, with its own release. IMO downloading the branch should be okay, but I figured I should still mention the alternative.

packages/cli/lib/projects.js
```diff
@@ -69,14 +69,24 @@ const writeProjectConfig = (configPath, config) => {
   }
 };
 
-const getProjectConfig = async _dir => {
+const getIsInProject = async _dir => {
```
anthmatic, line 72: 👍

packages/cli-lib/projects.js
anthmatic: I think we should account for `ref = undefined` when building this string. Otherwise, you get a 404 running `hs create api-sample`.

Otherwise, I think this approach works!

brandenrodgers: Good point, updated in fe42874

anthmatic (approved):
Code lgtm and tested old v. new flow, works good!

packages/cli-lib/projects.js
brandenrodgers: I was looking more into this, and realized there's another option. We could tag the latest commit in the boilerplate branch (w/out linking it to a new release), and then the CLI would download that specific tagged commit. I still think what we have in place now is totally fine since it will be easier to iterate on the boilerplate project. Maybe down the road we could update the flow to use a tagged commit if we're running into issues.

PR #633, opened 2022-01-21
Make project command help visible

anthmatic:
## Description and Context
Updates so that `hs project --help` shows the help for subcommands but the top level command is still not visible in `hs --help`


brandenrodgers (approved):
👌

PR #637, opened 2022-01-28
Project watch: watch for deleted directories

anthmatic:
## Description and Context
`project watch` was previous only watching for directories. This updates the watcher to include directories and hit the same delete endpoint.

### DIFF
--- a/packages/cli-lib/projectsWatch.js
+++ b/packages/cli-lib/projectsWatch.js
@@ -101,7 +101,7 @@ const queueFileUpload = async (
     try {
       if (action === 'upload') {
         await uploadFileToBuild(accountId, projectName, filePath, remotePath);
-      } else if (action === 'delete') {
+      } else if (action === 'deleteFile' || action === 'deleteFolder') {
         await deleteFileFromBuild(accountId, projectName, remotePath);
       }
       logger.log(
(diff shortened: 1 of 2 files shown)


packages/cli-lib/projectsWatch.js
```diff
@@ -186,7 +186,16 @@ const createWatcher = async (
       projectConfig.name,
       projectSourceDir,
       filePath,
-      'delete'
+      'deleteFile'
+    );
+  });
+  watcher.on('unlinkDir', async filePath => {
```
brandenrodgers, line 192: kind of a nit, but should this arg be `folderPath`?

packages/cli-lib/projectsWatch.js
```diff
@@ -186,7 +186,16 @@ const createWatcher = async (
...
+    addFile(
```
brandenrodgers, line 193: Not a blocker, but this func supports folders and also the "delete" action now. I wonder if we should change the name of it to be more generic.

brandenrodgers (approved):
Neither of my comments are blockers

anthmatic:
Thanks @brandenrodgers, renamed a couple things based on your feedback

PR #638, opened 2022-01-28
Add hubapi override

KatzMitch:
## Description and Context
We want to be able to override which HubAPI domain to hit via environment variable. Fall back on the existing logic if it's unset

### DIFF
--- a/packages/cli-lib/lib/urls.js
+++ b/packages/cli-lib/lib/urls.js
@@ -13,7 +13,7 @@ const getHubSpotWebsiteOrigin = env => {
 };
 
 const getHubSpotApiOrigin = (env, useLocalHost) => {
-  var domain = process.env.HUBAPI_DOMAIN_OVERRIDE;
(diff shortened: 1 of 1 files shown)


packages/cli-lib/lib/urls.js
```diff
@@ -13,9 +13,12 @@ const getHubSpotWebsiteOrigin = env => {
 };
 
 const getHubSpotApiOrigin = (env, useLocalHost) => {
-  return `https://${useLocalHost ? 'local' : 'api'}.hubapi${getEnvUrlString(
-    env
-  )}.com`;
+  var domain = process.env.HUBAPI_DOMAIN_OVERRIDE;
```
brandenrodgers, line 16: nit, but this supports es6 syntax so you can use `let` here.

brandenrodgers (approved):
Just one small suggestion, but otherwise lgtm!

PR #639, opened 2022-01-28
Support uploading  project files on watch

brandenrodgers:
## Description and Context

The `hs project watch` command currently requires a project to exist in HubSpot AND for it to contain at least one build. This PR updates the command to automatically create the project and upload all project files if it detects that the project doesn't exist in HubSpot yet.

I DRYed things up a bit by moving the upload util into a shared place. 

## Screenshots

image

## TODO

- [ ] Get feedback on UX/copy from @markhazlewood

## Who to Notify

cc/ @gcorne

### DIFF
--- a/packages/cli/commands/project/watch.js
+++ b/packages/cli/commands/project/watch.js
@@ -101,7 +101,6 @@ exports.handler = async options => {
 
   // Upload all files if no build exists for this project yet
   if (!results || !results.length) {
-    logger.log(i18n(`${i18nKey}.logs.initialUploadRequired`));
     await handleProjectUpload(
       accountId,
       projectConfig,
--- a/packages/cli/commands/project/watch.js
+++ b/packages/cli/commands/project/watch.js
@@ -1,6 +1,5 @@
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
 const { createWatcher } = require('@hubspot/cli-lib/projectsWatch');
-const { cancelStagedBuild } = require('@hubspot/cli-lib/api/dfs');
 const {
   logApiErrorInstance,
   ApiErrorContext,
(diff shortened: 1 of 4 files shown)


packages/cli/commands/project/watch.js
```diff
@@ -76,13 +81,36 @@ exports.handler = async options => {
...
+
+  // Upload all files if no build exists for this project yet
+  if (!results || !results.length) {
+    logger.log(i18n(`${i18nKey}.logs.initialUploadRequired`));
+    await handleProjectUpload(
+      accountId,
+      projectConfig,
+      projectDir,
+      startWatching
+    );
```
anthmatic, line 110: Should we also handle build & deploy status here as well?

brandenrodgers: Yeah I wasn't sure about that. My thinking for omitting it was to get to the "watching for file changes" output quicker. The watch command doesn't necessarily need the initial build to be completed before starting up. It just needs at least one build to exists. That being said, I don't feel too strongly about it. Maybe we can loop @markhazlewood in for some design input.

Mark the **TLDR** is:
This PR adds new functionality to the `hs project watch` command. Now it will detect that a project has never been uploaded to HubSpot, and will automatically upload all of the project's files before starting up the "watch" functionality. I'm re-using the same logic that we use for the `hs project upload` command, but I left out the part where the CLI shows you a visual of the components building and deploying (I explain my thinking for doing that above). Do you have any thoughts on it?

brandenrodgers: @anthmatic are you ok with me getting this in without the initial upload build/deploy logging? Should be easy enough to add later on if we decide that we want it. I just want to move this along to fill the gap.

I think I may do a follow up PR to move around some of the project util files anyways. We have some naming conflicts going on in the repo where we use the word "project" for two different things (github repos and also developer projects).

brandenrodgers: Here's how it currently looks:
Feb-09-2022 16-22-24

anthmatic: Yep! This looks good otherwise!

PR #644, opened 2022-02-15
Fixing project upload when there are no components

brandenrodgers:
## Description and Context

There was an issue with the hs project upload` command where it would hang on the `Building...` step when there are no components in the project to be built.

It was hanging because of this check [here](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/projects.js#L383)

`deployStatusTaskLocator`  is null here, even though isAutoDeployEnabled  is true. I added another check to make sure that the list of sub tasks has length.


anthmatic (approved):
👍  I think there's a case to be made for the BE failing the upload if it's empty but thanks for patching this!

PR #648, opened 2022-03-11
Project download command

brandenrodgers:
## Description and Context

This adds support for the `hs project download [name] [dest]` command.

The goal of this command is to unblock users who are not using source control to store their project files, but still want to get a project's files onto another computer.

This command will also create the `hsproject.json` config file, and it will place all of the project files into a `src` dir.
For example, if this is what you see in Projects UI
```js
my-app/
my-theme/
```
This will be the resulting structure after running the download command
```js
hsproject.json
src/
  my-app/
  my-theme/
```

### Standard successful download
image

### Example where I try to download a project into an existing project 
image

## Who to Notify

cc/ @m-roll @gcorne @kemmerle


m-roll:
I think it would also be good to optionally support a build number so previous builds can be fetched - does it just default to the latest build number?

brandenrodgers:
**Updated output:**
image

packages/cli-lib/lang/en.lyaml
```diff
@@ -2,706 +2,726 @@ en:
   cli:
     commands:
       accounts:
-        describe: "Commands for working with accounts."
+        describe: 'Commands for working with accounts.'
```
kemmerle, line 5: It seems like prettier replaced all double quotation marks with singles--was this intentional?

kemmerle (approved):
I pulled down the branch to test out the changes, and I was able to download the latest build of my project very easily. I think this is a great command to add. 

The code LGTM--very clear and streamlined.

PR #649, opened 2022-03-14
Marketplace module validation

jsines:
## Description and Context

- Adds a new command to validate module submissions for the marketplace using the CLI. Command can be ran with `hs module marketplace-validate <module path>`. Module validation is a subset of theme validation, in particular: 

```
ModuleValidator.MISSING_META_JSON
ModuleValidator.INVALID_META_JSON
ModuleValidator.MISSING_LABEL
ModuleValidator.MISSING_ICON
```
have been carried over under `SpecificModuleValidator` and 
```
DependencyValidator.FAILED_TO_FETCH_DEPS
DependencyValidator.EXTERNAL_DEPENDENCY
DependencyValidator.ABSOLUTE_DEPENDENCY_PATH
```
have been carried over under `ModuleDependencyValidator`, along with a new `ModuleDependencyValidator.ASSET_CO_LOCATION` to validate that attached CSS/JS/Assets are in the same folder as the module.


brandenrodgers:
This is looking great so far! 🎉  I'm still in the process of looking through the code, but I can answer a few of the questions

> Is there a more preferred naming convention for these? 

It's mostly up to personal preference. If I had to nit pick, this would be my suggestion:
- ModuleValidator -> ThemeModuleValidator (validates many modules for the theme)
- SpecificModuleValidator -> ModuleValidator (validates one module)
- ModuleDependencyValidator -> keep this one as is (validates module deps)
- DependencyValidator -> ThemeDependencyValidator (validates theme deps)

But you can ignore this if you like it better how it currently is.

> In some places I reused functions from the other validators ([here](https://github.com/HubSpot/hubspot-cli/blob/6b085c8eddd45c635a1a92fb65488d6d578d36a9/packages/cli/lib/validators/marketplaceValidators/module/SpecificModuleValidator.js#L36) and [here](https://github.com/HubSpot/hubspot-cli/blob/6b085c8eddd45c635a1a92fb65488d6d578d36a9/packages/cli/lib/validators/marketplaceValidators/module/ModuleDependencyValidator.js#L51)), do these look fine to you? Curious if you think they'll cause any problems as written or if they'll work OK.

I _think_ the dependency validator stuff should all work as expected in ModuleDependencyValidator. I'm not sure about the code in SpecificModuleValidator though. It looks like a lot of the code in there is set up to handle multiple modules. I haven't tested it out, but we can probably remove `getUniqueModulesFromFiles` since we should already know the path to the module that we're trying to validate (`this._absolutePath`).

Somewhat related, it looks like there's a lot of shared logic between ModuleValidator + SpecificModuleValidator and also between ModuleDependencyValidator + DependencyValidator. I'm actually OK with having repeated code in there for a first pass of this. We can look to DRY things up after we're happy with how it looks.

> Not totally certain on how to implement ModuleDependencyValidator.ASSET_CO_LOCATION...

This sounds like a good place to loop @KatzMitch in for some help. I'm not sure what this check is looking for. I can help nail down a solution once we verify though.

KatzMitch:
Ooh thanks for the tag here, I missed this original PR coming in. Thanks for the work on it! 🎉  super exciting stuff to see come together

Sorry I was super unclear with the co-location check 😬 

Some context here: Themes are inherently packaged together it's a collection of files all within one folder. Meaning for the zipping and delivery process we just need to point our code at one folder and say zip it up and we know we'll get all relevant files.

However, a standalone module could have attached CSS/JS/other assets that live _anywhere_ in the design manager. We want to avoid leaking that messiness into any shopper's portal and enforce that all attached assets are colocated in the same folder as the module. So for example: `/SomeFolder/my_widget.module` *should* be allowed to reference `/SomeFolder/some_style.css` as attached CSS but *should not* be allowed to have `/ATotallyDifferentFolder/WithASubFolder/some_other_style.css` as attached CSS

Is that clearer?

jsines:
I like your rename suggestions better, I'll hold off on switching them to avoid confusing myself though haha

> we can probably remove getUniqueModulesFromFiles since we should already know the path to the module that we're trying to validate (this._absolutePath).

I considered this but it actually transforms the array of file paths into a mapping of the module's files like:
```
{
  '<filepath>/accordion.module': {
    'fields.json': '<filepath>/accordion.module/fields.json',
    'meta.json': '<filepath>/accordion.module/meta.json',
    ...
    'module.js': '<filepath>/accordion.module/module.js'
  }
}
```
which makes it much easier to check for `meta.json` and such. I think I could remove the `Object.keys().forEach()` and be fine since there should only be one per module.

>So for example: `/SomeFolder/my_widget.module` should be allowed to reference `/SomeFolder/some_style.css` as attached CSS but should not be allowed to have `/ATotallyDifferentFolder/WithASubFolder/some_other_style.css` as attached CSS

This makes sense! Still going to need some help on implementation I think. My initial thought was we could just check the dependency data we get from [here](https://github.com/HubSpot/hubspot-cli/blob/6b085c8eddd45c635a1a92fb65488d6d578d36a9/packages/cli/lib/validators/marketplaceValidators/module/ModuleDependencyValidator.js#L92), which looks like:
```
[
  {
    file: '<filepath>/accordion.module/module.css',
    deps: []
  },
  ... (repeat for each file in the module)
]
```
but I adding a dependency to the accordion module and couldn't seem to get it to show up in the deps array. For example I added `{% include './../test_folder/test.css' %}` at the top of `module.html` and the corresponding `deps` array remained empty. It looks like the implementation of this calls `fetchDependencies(accountId, source)` which is an API call to `product/marketplace/v2/template/render` so no idea how this works on the backend. If you had an example of a setup that would trigger this error that'd be awesome.

KatzMitch:
Hmm we may need to spin up a new `/module/render` endpoint to support this - added https://git.hubteam.com/HubSpot/ProductMarketplaceV2/issues/232 to create it

jsines:
Hiya folks, jumping back into this and wanted to clear up a little confusion I'm having:

Is this command (`hs module marketplace-validate <>`) intended to be used on local modules or modules that have already been uploaded to the design manager? My original thought was the former because the API endpoint used for theme validation (`product/marketplace/v2/template/render`) took source code as a string in the request body, but the new endpoint for modules (`product/marketplace/v2/module/render/{path}?portalId={pid}`) looks like it only operates on modules already uploaded so now I'm thinking its the latter. @KatzMitch

KatzMitch:
Yeah unfortunately that endpoint needs the module to be uploaded to work. Is there a way the CLI could check if the module exists on the server and skip that validation if not (maybe throw a warning saying it needs to be uploaded first)?

jsines:
A few more things!

1. I've split up BaseValidator into AbsoluteValidator and RelativeValidator, there may be a better way to handle this to keep it DRYer but for now this makes the way the two module and theme validation processes work easier to handle.

2. Looks like when I'm hitting the endpoint `product/marketplace/v2/module/render/{path}?portalId={pid}` I'm getting "This oauth-token (<token>) does not have proper permissions! (requires all of [marketplace-provider-access])". Is this an issue with how the endpoint is set up or am I doing something incorrect on my end that you see?

3. Is there an existing endpoint to access the module's `meta.json` in the design manager? When I was working with it expecting the files to be local this was no issue but if they're uploaded there needs to be some way to access them in order for the `ModuleValidator` to run its validation process.

KatzMitch:
Chatted offline about #2

jsines:
@brandenrodgers Any idea here? 

> 3. Is there an existing endpoint to access the module's `meta.json` in the design manager? When I was working with it expecting the files to be local this was no issue but if they're uploaded there needs to be some way to access them in order for the `ModuleValidator` to run its validation process.

jsines:
@KatzMitch Could you elaborate a little more on the distinction between `ModuleDependencyValidator.EXTERNAL_DEPENDENCY` and `ModuleDependencyValidator.ASSET_CO_LOCATION`? I believe I understand how `ASSET_CO_LOCATION` is meant to work but now that I've begun implementation not totally sure how `EXTERNAL_DEPENDENCY` is meant to behave differently. 

Also if you've got an answer to the above ^ at all would help a bunch, not sure if you just didn't see it last time or don't know 😅

KatzMitch:
I think they're effectively the same validation (or could be built that way). We just want to make sure that any file referenced by the module (either in attached CSS/JS, linked files, or imported into the HTML/CSS/JS code using HubL) lives in the same folder as the module, and is referenced by relative path

As long as that requirement is fulfilled, whether it's 1 validator or 2 all works!

The module's meta.json contents are kind of scattered across the design manager:
Screen Shot 2022-04-14 at 3 13 47 PM

jsines:
This is now ready for review 🎉 feel free to tag anyone else to add for review requests

TanyaScales:
Added a couple other folks familiar with the codebase for review - thanks @jsines!

packages/cli/lib/validators/marketplaceValidators/module/ModuleDependencyValidator.js
```diff
@@ -0,0 +1,102 @@
...
+    );
+  }
+
+  async getAllDependenciesByPath(relativePath, accountId, validationErrors) {
+    let deps = [];
+    const file_deps = await fetchModuleDependencies(
+      accountId,
+      relativePath
+    ).catch(err => {
+      console.log(err);
```
brandenrodgers, line 51: Is this log intentionally here? If so, the typical pattern in the CLI is to use `logger.debug` for logging out errors like this. Alternatively, you could define a lyaml string for this scenario with some sort of explanation.

packages/cli/lib/validators/marketplaceValidators/module/ModuleValidator.js
```diff
@@ -0,0 +1,103 @@
...
+    );
+    validationErrors.push(
+      this.getError(this.errors.FAILED_TO_FETCH_META_JSON, relativePath)
+    );
+  }
+
+  async getModuleMetaByPath(relativePath, accountId, validationErrors) {
+    const moduleMeta = await fetchModuleMeta(accountId, relativePath).catch(
+      err => {
+        console.log(err);
```
brandenrodgers, line 51: Same ["log" suggestion](https://github.com/HubSpot/hubspot-cli/pull/649/files#r853312889) here

packages/cli/lib/validators/constants.js
```diff
@@ -4,11 +4,13 @@ const SUCCESS = 'SUCCESS';
 
 const VALIDATION_RESULT = { WARNING, FATAL, SUCCESS };
 const VALIDATOR_KEYS = {
-  dependency: 'dependency',
-  module: 'module',
+  themeDependency: 'themeDependency',
```
brandenrodgers, line 7: This may have already happened, but I _think_ there are some BE systems that make use of these keys. Just double checking that they've updated their services in response to this.

packages/cli/lib/validators/marketplaceValidators/module/ModuleDependencyValidator.js
jsines: Good catch, accidentally left those in from testing 👀 

Don't think we need to `logger.debug`, if these errors trip then they'll show up as validation errors.

packages/cli-lib/lang/en.lyaml
```diff
@@ -607,6 +607,21 @@ en:
...
+            errors:
+              invalidPath: "The path \"{{ path }}\" is not a path to a folder"
+            logs:
+              validatingModule: "Validating module \"{{ path }}\" \n"
+            options:
+              json:
+                describe: "Output raw json data"
+            positionals:
+              src:
+                describe: "Path to the local module, relative to your current working directory."
```
brandenrodgers, line 624: Is this `src` referring to the path on the local file system OR the path to the module in the Design Manager? It looks like we're making use of it in our [request to fetch the module's deps](https://github.com/HubSpot/hubspot-cli/pull/649/files#diff-98b83e9a5bde3137a78d32f0fd35d4aff65b825a813931c50ecd77a7dd050dcaR28), so I'm guessing that it's the src in the Design Manager. Should we update this copy to reflect that?

jsines: Correct, forgot to fix these strings when changing paths to be DM-focused. Fixed 👍

packages/cli/lib/validators/marketplaceValidators/AbsoluteValidator.js
```diff
@@ -2,23 +2,23 @@ const path = require('path');
 
 const { VALIDATION_RESULT } = require('../constants');
 
-class BaseValidator {
+class AbsoluteValidator {
```
brandenrodgers, line 5: This works for me 👍 . One other potential solution is to pass something like "useRelativePath" as a config option to the constructor for the `BaseValidator`. Definitely not a blocker though

brandenrodgers (approved):
This lgtm! 🎉 

The only other thing that's worth mentioning is that we have the ability to introduce new commands in a "hidden" fashion. All it really means is that they don't show up when the user runs "hs --help". It gives us the ability to keep external usage low while we test internally. If the timing of the release isn't urgent, I'd suggest going that route to make sure we can uncover any bugs.

You can do it by changing [this line](https://github.com/HubSpot/hubspot-cli/pull/649/files#diff-337792fe9f893dda5c8adf9a158724d97692695bde627b4816d0d590bcefd934R8) to:
```js
exports.describe = false; // i18n(`${i18nKey}.describe`);
```

jsines:
> we have the ability to introduce new commands in a "hidden" fashion

Sounds good!  Had to comment out two other lines too to make linter happy. Should I make an issue to un-hide it once module marketplace goes live? Or somewhere else to note to remove that in the future?

Issue #444, opened 2021-02-16
The hubspot cli tool should output the browser it tries to open


allan-simon:
## Description and Context

When you use the CLI tool in a dockerized context or a on remote server , the cli tool has not directly access to the "main" browser of the user,  in that case, like for example what is done with heroku cli, it would be good if hubspot cli would also output the link in the console so that one can manually open it

anthmatic:
@allan-simon is this in regards to the page that is opened when you run `hs auth`?

allan-simon:
yes sorry I forgot to precise it was `hs auth` , right now I "hack" around it, by adding a console.log before the `open()`

gcorne:
I am going to go ahead and take care of this.

allan-simon:
thanks a lot !

allan-simon:
awesome

PR #654, opened 2022-03-24
Updates to project upload and watch commands

brandenrodgers:
## Description and Context

This makes a handful of small UX improvements to the project upload and watch commands.

### Upload
- Added a link to Projects UI for the newly created build
- Indented the subbuild statuses so hierarchy is more clear
- Added some lines to make it more closely match the initial mocks

### Watch
- Properly clean up staged build in more scenarios (including when the user closes the terminal tab)
- Show the user a prompt to quit the watch process
- Remove the word "staging" from all output because it is confusing to the users.

### Updated watch output
image

### Upload (building)
Screen Shot 2022-03-24 at 5 44 55 PM

### Upload (deploying)
Screen Shot 2022-03-24 at 5 45 10 PM

### Upload (done)
Screen Shot 2022-03-24 at 5 45 18 PM

## Who to Notify

cc/ @gcorne @markhazlewood @kemmerle @m-roll

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -468,7 +468,8 @@ en:
                 succeed: "Uploaded {{#bold}}{{ projectName }}{{/bold}} project files to {{ accountIdentifier }}"
             logs:
               buildSucceeded: "Build #{{ buildId }} succeeded\n"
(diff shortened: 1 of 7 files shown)


kemmerle (approved):
I pulled down the branch to test locally, and everything is working as expected! The code is also streamlined :shipit:

gcorne:
For links, I wonder if we should use the URL so that users can copy and paste. I think that will be more accessible in a range of terminals/shells. It also is useful when the command is being run in a headless environment. Something similar came up  came up back in the day with personal access keys (see #444).

PR #656, opened 2022-03-25
Support initial upload for project watch

brandenrodgers:
## Description and Context

Adding the `--initial-upload` option for the `hs project watch` command.

Now there are two scenarios where the project watch command will run an upload before starting the watch.
1. There are no builds in the project yet
2. The user sets initial upload to true

The next step is to decide if this should default to true or false, but this should be a reasonable first step.

## Screenshots

image

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -503,6 +503,12 @@ en:
               uploadFailed: "Failed to upload file \"{{ filePath }}\" to \"{{ remotePath }}\""
               deleteFileFailed: "Failed to delete file \"{{ remotePath }}\""
               deleteFolderFailed: "Failed to delete folder \"{{ remotePath }}\""
+            positionals:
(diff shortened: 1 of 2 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -621,10 +627,14 @@ en:
           destinationRequired: "A destination directory needs to be passed"
           invalidPath: "The \"{{ path }}\" is not a path to a directory"
         options:
-          disableInitial: "Disable the initial upload when watching a directory (default)"
-          initialUpload: "Upload directory before watching for updates"
-          notify: "Log to specified file when a watch task is triggered and after workers have gone idle. Ex. --notify path/to/file"
-          remove: "Will cause watch to delete files in your HubSpot account that are not found locally."
+          disableInitial:
```
brandenrodgers, line 630: These weren't working. The path wasn't right

PR #663, opened 2022-04-06
bump shelljs version to fix high severity vulnerabilities

calesennett:
## Description and Context
Bumps `shelljs` in CLI package to version `0.8.5` to fix high severity vulnerability, [Improper Privilege Management in shelljs](https://github.com/advisories/GHSA-4rq4-32rv-6wp6).

CC: @miketalley @anthmatic


gcorne:
Much appreciated! Turns out we can go even further and remove the code using `shelljs`, which is no longer being used so I filed #664.

PR #665, opened 2022-04-08
Improve file upload failure handling

### LINKED ISSUE #626: Return a non-zero exit code (or a flag for that option) when files fail to update on `hs upload` command
## Description and Context

Could a non-zero exit code be returned when a file doesn't upload on the `hs upload` command due to an error? We run this command with a GitHub deploy action and saw that the deployment was showing to be successful even though some module files didn't update due to errors.

## Expected behavior

I would expect to receive a non-zero exit code to alert that there was an error with the upload, as of right now you won't know until you notice some of your modules didn't update.

## Screenshots

Screen Shot 2022-01-07 at 10 03 05 AM

gcorne:
## Description and Context

This PR treats any failure to upload a file as a failure of the `hs upload` command and uses the `WARNING` exit code so that scripts and other automated system detect the failures.

Fixes #626 

One thing that I wonder about is if we should also incorporate warnings as proposed in #531 — not in as command failures but into the output that is logged. In general, I think there is an opportunity to improve the output.

## Screenshots
Before:
image

After:
image

## Who to Notify
@anthmatic @miketalley @brandenrodgers @stephenandersondev @SteffenMahler

### DIFF
--- a/packages/cli-lib/lib/uploadFolder.js
+++ b/packages/cli-lib/lib/uploadFolder.js
@@ -155,9 +155,8 @@ async function uploadFolder(accountId, src, dest, options) {
 }
 
 function hasUploadErrors(results) {
-  return (
-    results.filter(result => result.resultType === FileUploadResultType.FAILURE)
-      .length > 0
+  return results.some(
+    result => result.resultType === FileUploadResultType.FAILURE
   );
 }
 
--- a/packages/cli-lib/lib/uploadFolder.js
+++ b/packages/cli-lib/lib/uploadFolder.js
@@ -19,6 +19,11 @@ const {
   isFatalError,
 } = require('../errorHandlers');
 
+const FileUploadResultType = {
+  SUCCESS: 'SUCCESS',
+  FAILURE: 'FAILURE',
+};
+
 const queue = new PQueue({
   concurrency: 10,
 });
@@ -112,13 +117,18 @@ async function uploadFolder(accountId, src, dest, options) {
     await queue.addAll(filesToUpload.map(uploadFile));
   }
 
-  return queue.addAll(
+  const results = await queue.addAll(
     failures.map(({ file, destPath }) => {
       return async () => {
         logger.debug('Retrying to upload file "%s" to "%s"', file, destPath);
         try {
           await upload(accountId, file, destPath, apiOptions);
           logger.log('Uploaded file "%s" to "%s"', file, destPath);
+          return {
+            resultType: FileUploadResultType.SUCCESS,
+            error: null,
+            file,
+          };
         } catch (error) {
           logger.error('Uploading file "%s" to "%s" failed', file, destPath);
           if (isFatalError(error)) {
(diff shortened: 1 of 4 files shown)


packages/cli/commands/upload.js
```diff
@@ -162,13 +163,22 @@ exports.handler = async options => {
...
       mode,
     })
-      .then(() => {
-        logger.success(
-          i18n(`${i18nKey}.success.uploadComplete`, {
-            dest,
-          })
-        );
-        logThemePreview(src, accountId);
+      .then(results => {
```
gcorne, line 166: I was tempted to switch to `async`/`await` to reduce some of the nesting.

packages/cli/commands/upload.js
```diff
@@ -55,7 +55,7 @@ exports.handler = async options => {
   await loadAndValidateOptions(options);
 
   if (!validateMode(options)) {
-    process.exit(EXIT_CODES.ERROR);
+    process.exit(EXIT_CODES.WARNING);
```
gcorne, line 58: Seeing `WARNING` as an exit code value feels kind of off since these really are errors — they just aren't internal. We also should be mindful of exit codes that `yargs` produces so that there is consistency, especially in its own validation of user input.

brandenrodgers: Based off of the current description for exit code, we have `EXIT_CODES.ERROR` defined as
"Config problem or internal error". Should we treat invalid user input as a config issue (i.e. keep this as an error)?

Warnings are defined as `Warnings or validation issues`, which I'm realizing is pretty vague. I suppose this could also fall into that category too.

brandenrodgers (approved):
Just one question about exit codes, but otherwise this all lgtm 👍

packages/cli/commands/upload.js
gcorne: Looked at #570 and I _think_ this is correct, but I am not really sure. The HubSpot CLI is more broad in terms of purpose than `eslint` so maybe we need to be clearer on when different codes should be used. I could also try and add more nuance into capturing why a given upload failed so that we're treating 400s due to validation errors differently than other failure states.

packages/cli-lib/lib/uploadFolder.js
```diff
@@ -132,12 +142,27 @@ async function uploadFolder(accountId, src, dest, options) {
...
         }
       };
     })
   );
+  return results;
+}
+
+function hasUploadErrors(results) {
+  return (
+    results.filter(result => result.resultType === FileUploadResultType.FAILURE)
```
stephenandersondev, line 159: You could probably just use `.some` here so you don't have to check for length

stephenandersondev (commented):
Thanks for doing this @gcorne, this is going to be super helpful for us folks that use GH actions with hubspot-cli

packages/cli-lib/lib/uploadFolder.js
gcorne: Switched over in af6fb96

PR #666, opened 2022-04-08
Prompt for src and dest when running upload/watch

anthmatic:
## Description and Context
Currently, when running the `hs watch` and `hs upload` commands, the positional arguments for `src` and `dest` are required. This PR makes 2 changes to that command:

1. Allow users to pass these values as positional arguments or flags. For example, the same command can be run as 
`hs upload dir1 dir2` or `hs upload --src=dir1 --dest=dir2`

2. If no values are passed, the user is prompted for them. The default values for each prompt are:
- `src`: default to cwd
- `dest`: default to current folder name

## Examples
New command description. Note the optional `[]` brackets.
```
hs upload [--src] [--dest]
```
As positionals
```
hs upload src.html dest.html
```
As flags
```
hs upload --src=src.html --dest=dest.html
```

## Screenshots
image
Note: `--src` and `--dest` are not listed as options here


packages/cli/commands/upload.js
```diff
@@ -34,7 +35,7 @@ const { i18n } = require('@hubspot/cli-lib/lib/lang');
 const i18nKey = 'cli.commands.upload';
 const { EXIT_CODES } = require('../lib/enums/exitCodes');
 
-exports.command = 'upload <src> <dest>';
+exports.command = 'upload [--src] [--dest]';
```
anthmatic, line 38: Yargs allows you to pass the name of a positional arg when running the command so we don't need to (also, can't?) add explicit options here.

packages/cli-lib/lang/en.lyaml
```diff
@@ -761,6 +761,12 @@ en:
           enterName: "Enter a name to use for the sandbox: "
           errors:
             invalidName: "You entered an invalid name. Please try again."
+        uploadPrompt:
+          enterSrc: "[--src] Enter the path that you want to upload: "
```
brandenrodgers, line 765: Might be worth getting some quick input on the copy from @markhazlewood

brandenrodgers (approved):
Just one comment about copy, but otherwise this all lgtm!

PR #672, opened 2022-04-13
Sandbox cli bugfixes

adamawang:
## Description and Context

Bug fixes for the `hs sandbox create` commands. This resolves an issue where incorrect messages would be printed upon failures, and a fix for a missing translation item. 

## Screenshots

When any command fails, an INFO log gets printed incorrectly.

Before:
Screen Shot 2022-04-11 at 12 29 44 PM

After:
Screen Shot 2022-04-13 at 12 41 49 PM

Fix success text missing translation 

Before:
Screen Shot 2022-04-11 at 12 30 35 PM

After:
Screen Shot 2022-04-13 at 12 44 36 PM

- Fix logs that print after sandbox create failure (success and info logs should not print here)
- Fix logs that print after trying to create a sandbox without the proper scopes (success and info logs shouldn't print here either)
- Fix logs that print when hitting the sandboxes limit (success and info logs should not print here either)

Note: these all were fixed by one change where we now throw the error instead of logging it. It was running through the success messages because nothing was thrown before. 

Before:
Screen Shot 2022-04-13 at 12 47 19 PM
Screen Shot 2022-04-11 at 12 32 31 PM
Screen Shot 2022-04-11 at 12 33 17 PM

After:
Screen Shot 2022-04-13 at 12 41 49 PM
Screen Shot 2022-04-13 at 12 48 50 PM

### DIFF
--- a/packages/cli-lib/sandboxes.js
+++ b/packages/cli-lib/sandboxes.js
@@ -1,4 +1,6 @@
+const { EXIT_CODES } = require('../cli/lib/enums/exitCodes');
 const { createSandbox: _createSandbox } = require('./api/sandbox-hubs');
+const { logger } = require('./logger');
 
 /**
  * Creates a new Sandbox portal instance.
@@ -11,7 +13,8 @@ async function createSandbox(accountId, name) {
   try {
     resp = await _createSandbox(accountId, name);
   } catch (err) {
-    throw err.error;
+    logger.error(err.error.message);
+    process.exit(EXIT_CODES.ERROR);
   }
 
   return {
--- a/packages/cli-lib/sandboxes.js
+++ b/packages/cli-lib/sandboxes.js
@@ -1,5 +1,4 @@
 const { createSandbox: _createSandbox } = require('./api/sandbox-hubs');
-const { logger } = require('./logger');
 
 /**
  * Creates a new Sandbox portal instance.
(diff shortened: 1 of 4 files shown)


packages/cli-lib/sandboxes.js
```diff
@@ -12,7 +11,7 @@ async function createSandbox(accountId, name) {
   try {
     resp = await _createSandbox(accountId, name);
   } catch (err) {
-    logger.error(err.error.message);
+    throw err.error;
```
adamawang, line 14: I removed the logger here since it was double printing error messages with the throw here. Didn't see the logger doing anything other than console logging, but let me know if you'd like that back in there

brandenrodgers: Is your intent to exit the process when this happens? If so, you can exit more gracefully via `process.exit(EXIT_CODES.ERROR)`

adamawang: Got it - thanks for that suggestion! Added back the logger and replaced the throw with the process.exit

brandenrodgers (approved):
👌

adamawang:
@brandenrodgers added some additional descriptions so sandbox commands show when running `hs` and `hs sandbox`. 

Screenshots attached:
Screen Shot 2022-04-13 at 4 53 24 PM

Screen Shot 2022-04-13 at 4 58 35 PM

brandenrodgers (approved):
lgtm!

PR #673, opened 2022-04-13
reworking the config and accounts commands

brandenrodgers:
## Description and Context

Updating the behavior of the `hs config` and `hs accounts` commands. Notable changes:
- `hs config set default-account [newDefaultAccount]` was moved to `hs accounts set-default [--default=newDefaultAccount]`
- All of the positionals in `hs config set` have been converted into options, with prompts as a fallback.

## Screenshots

`hs config set` before:
image
`hs config set` after:
image
`hs config set` example usage:
image
Alternatively...
image
---

Switching default account before:
image
Alternatively...
image
Switching default account after:
image
Alternatively...
image
--

## Who to Notify

cc/ @markhazlewood @kemmerle

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -22,20 +22,20 @@ en:
                 describe: "New name for account."
             success:
               renamed: "Account \"{{ name }}\" renamed to \"{{ newName }}\""
-          setDefault:
-            describe: "Change default account"
+          use:
+            describe: "Change default account in config"
             errors:
-              accountNotFound: "The account \"{{ specifiedAccount }}\" was not found in {{ configPath }}."
+              accountNotFound: "The account \"{{ specifiedAccount }}\" could not be found in {{ configPath }}"
             examples:
               default: "Select account to use as the default from a list"
               idBased: "Set the default account to the account in the config with accountId equal to \"1234567\""
               nameBased: "Set the default account to the account in the config with name equal to \"MyAccount\""
-            options:
-              default:
+            positionals:
+              account:
                 describe: "Account name or id to use as the default"
             promptMessage: "Select an account to use as the default"
             success:
-              defaultAccountUpdated: "Default account updated to: \"{{ accountName }}\"."
+              defaultAccountUpdated: "Default account updated to \"{{ accountName }}\""
       auth:
         describe: "Configure authentication for a HubSpot account. Supported authentication protocols are {{ supportedProtocols }}."
         errors:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -22,6 +22,20 @@ en:
                 describe: "New name for account."
             success:
               renamed: "Account \"{{ name }}\" renamed to \"{{ newName }}\""
+          setDefault:
+            describe: "Change default account"
+            errors:
+              accountNotFound: "The account \"{{ specifiedAccount }}\" was not found in {{ configPath }}."
+            examples:
+              default: "Select account to use as the default from a list"
+              idBased: "Set the default account to the account in the config with accountId equal to \"1234567\""
+              nameBased: "Set the default account to the account in the config with name equal to \"MyAccount\""
+            options:
+              default:
+                describe: "Account name or id to use as the default"
+            promptMessage: "Select an account to use as the default"
+            success:
+              defaultAccountUpdated: "Default account updated to: \"{{ accountName }}\"."
       auth:
         describe: "Configure authentication for a HubSpot account. Supported authentication protocols are {{ supportedProtocols }}."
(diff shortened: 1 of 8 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -22,6 +22,20 @@ en:
                 describe: "New name for account."
             success:
               renamed: "Account \"{{ name }}\" renamed to \"{{ newName }}\""
+          setDefault:
+            describe: "Change default account"
+            errors:
+              accountNotFound: "The account \"{{ specifiedAccount }}\" was not found in {{ configPath }}."
+            examples:
+              default: "Select account to use as the default from a list"
+              idBased: "Set the default account to the account in the config with accountId equal to \"1234567\""
```
kemmerle, line 31: Will the accountId `1234567` potentially be confusing to any customers?

packages/cli-lib/lang/en.lyaml
```diff
@@ -38,55 +52,27 @@ en:
         describe: "Commands for working with the config file."
         subcommands:
           set:
-            describe: "Commands for working with the config file."
-            subcommands:
+            describe: "Commands for modifying the CLI configuration"
+            promptMessage: "Select a config option to update"
+            examples:
+              default: "This is an example of something"
```
kemmerle, line 58: I'm a little confused by "This is an example of something." What's the use case here?

packages/cli-lib/lang/en.lyaml
brandenrodgers: It's what was there before. Do you have an idea for a better example accountId?

packages/cli-lib/lang/en.lyaml
brandenrodgers: LOL, I had this in as a placeholder and forgot to go back and update it. It should say something along the lines of "Opens a prompt to edit config items".

packages/cli/commands/config/set.js
```diff
@@ -1,25 +1,93 @@
...
+
+  process.exit(EXIT_CODES.SUCCESS);
+};
+
+exports.builder = yargs => {
   yargs
-    .command(defaultAccount)
-    .command(defaultMode)
-    .command(httpTimeout)
-    .command(allowUsageTracking)
```
anthmatic, line 21: General q: are you planning to bump the major version in the next release since we're removing commands here?

brandenrodgers: We talked offline, but the tldr is that I probably won't bump a major version here because this command sees very low usage. Also, it very likely is not being utilized in any automation flows.

gcorne: I think we need to bump the major version for the `hs upload` command change.

packages/cli-lib/lang/en.lyaml
kemmerle: Ah, if that was the example before, then I think it's good to go!

kemmerle (approved):
I looked over the code, and it all LGTM. I also tested out the new commands locally, and everything is working as expected 👍

packages/cli/commands/config/set.js
```diff
@@ -1,25 +1,93 @@
...
+        { name: 'Allow usage tracking', value: { allowUsageTracking: '' } },
+        { name: 'HTTP timeout', value: { httpTimeout: '' } },
+      ],
+    },
+  ]);
+
+  return mode;
+};
+
+const handleConfigUpdate = async (accountId, options) => {
```
kemmerle, line 35: This is the only function that I'm questioning a bit; it seems a bit off to be checking that the options aren't `undefined`, but I also can't come up with any way to refactor 🤷‍♀️ Just thought I would leave a quick note.

packages/cli/commands/config/set.js
brandenrodgers: Are you referencing [this update](https://github.com/HubSpot/hubspot-cli/pull/666)? If so, I believe Anthony plans to make some updates to that PR to make it a non-breaking change.

anthmatic:
I really like how `hs accounts use` almost shifts the mental model a bit

packages/cli/commands/config/set.js
gcorne: Sorry for not being more specific. I was thinking about #665. My thought process was that since it changes the error handling of the upload command that it warranted bumping the major version. For most CLI users, the process of upgrading is the same since it always involves running a command. The difference between major, minor, and patch only really comes into play when one of the CLI NPM packages is referenced in `package.json` or is a dep of another package.

packages/cli/commands/config/set.js
brandenrodgers: Yeah I have to do this because some of them are booleans. So I can't just check for `!<option>`

PR #674, opened 2022-04-20
Add prompt when users forget `hs project` command

kemmerle:
## Description and Context
We would like to add a warning, when a customer attempts to run a command (like `hs watch`) on a project folder to remind them to run the correct project-specific command (like `hs project watch`). 

## Screenshots
Current (`hs upload` succeeds with a project folder)

Screen Shot 2022-04-20 at 9 30 36 AM

With these changes (`hs upload` errors out with a project folder)

Screen Shot 2022-04-20 at 9 29 44 AM

## TODO

1) I would love some feedback on the copy for the error. 

2) I think we'll just need to thoroughly test all the commands to ensure that we're not accidentally blocking any commands that we should run. 

## Who to Notify

@brandenrodgers

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -2,7 +2,7 @@ en:
   cli:
     commands:
       generalErrors:
-        srcIsProject: "\"{{ src }}\" is a project folder. Did you mean \"hs project {{command}}\"?"
+        srcIsProject: "\"{{ src }}\" is in a project folder. Did you mean \"hs project {{command}}\"?"
       accounts:
         describe: "Commands for working with accounts."
         subcommands:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -2,7 +2,7 @@ en:
   cli:
     commands:
       generalErrors:
-        srcIsProject: "\"{{ src }}\" is a project folder. Please use a command with the \"hs project\" prefix."
+        srcIsProject: "\"{{ src }}\" is a project folder. Did you mean \"hs project {{command}}\"?"
       accounts:
         describe: "Commands for working with accounts."
         subcommands:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -1,6 +1,8 @@
 en:
   cli:
     commands:
+      generalErrors:
+        srcIsProject: "\"{{ src }}\" is a project folder. Please use a command with the \"hs project\" prefix."
       accounts:
         describe: "Commands for working with accounts."
         subcommands:
--- a/packages/cli/bin/cli.js
+++ b/packages/cli/bin/cli.js
@@ -82,10 +82,7 @@ const argv = yargs
     type: 'boolean',
   })
   .check(argv => {
-    if (
-      argv._.length === 1 &&
-      (argv._.join(' ') === 'upload' || argv._.join(' ') === 'watch')
-    ) {
+    if (argv._.length === 1 && ['upload', 'watch'].includes(argv._[0])) {
       if (getIsInProject(argv.src)) {
         logger.error(
           i18n(`${i18nKey}.srcIsProject`, {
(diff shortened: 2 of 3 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -1,6 +1,8 @@
 en:
   cli:
     commands:
+      generalErrors:
+        srcIsProject: "\"{{ src }}\" is a project folder. Please use a command with the \"hs project\" prefix."
```
brandenrodgers, line 5: Copy suggestion: `"\"{{ src }}\" is a project folder. Did you mean \"hs project upload\"?"`

The "upload" portion would vary depending on which command the user ran

packages/cli/bin/cli.js
```diff
@@ -74,6 +81,17 @@ const argv = yargs
     describe: 'set log level to debug',
     type: 'boolean',
   })
+  .check(argv => {
+    if (getIsInProject(argv.src)) {
+      throw new Error(
```
brandenrodgers, line 86: The current pattern for logging errors is to use `logger.error` and then an explicit `process.exit` with an error code if you want to exit.

packages/cli/bin/cli.js
```diff
@@ -74,6 +81,17 @@ const argv = yargs
     describe: 'set log level to debug',
     type: 'boolean',
   })
+  .check(argv => {
+    if (getIsInProject(argv.src)) {
```
brandenrodgers, line 85: Some project commands support a `path` argument, which means that they can run from outside of a project folder. Example `hs project upload ./path/to/my/project`

I don't think this check will catch that use case. Does `argv` also include info about the options passed in?

packages/cli/bin/cli.js
```diff
@@ -74,6 +81,17 @@ const argv = yargs
     describe: 'set log level to debug',
     type: 'boolean',
   })
+  .check(argv => {
```
brandenrodgers, line 84: This check will run for every command. Is there a way to determine what the command name is in here? I'm thinking we could have some sort of filter that will let us only run this check for certain commands like "upload" and "watch". That way we don't run it on irrelevant commands.

kemmerle: @brandenrodgers, working on incorporating your feedback now. Do you have a specific list of commands we'd like to run the check for? I'm also going to Slack you, but wanted to leave this here for posterity.

kemmerle: Resolved: We've decided to check only the "upload" and "watch" commands for now.

packages/cli/bin/cli.js
```diff
@@ -82,12 +82,21 @@ const argv = yargs
...
   .check(argv => {
-    if (getIsInProject(argv.src)) {
-      throw new Error(
-        i18n(`${i18nKey}.srcIsProject`, {
-          src: argv.src,
-        })
-      );
+    if (
+      argv._.length === 1 &&
+      (argv._.join(' ') === 'upload' || argv._.join(' ') === 'watch')
```
brandenrodgers, line 87: Is `join` necessary here? I think this could be shortened to `['upload', 'watch'].includes(argv._[0])`

packages/cli-lib/lang/en.lyaml
```diff
@@ -2,7 +2,7 @@ en:
   cli:
     commands:
       generalErrors:
-        srcIsProject: "\"{{ src }}\" is a project folder. Please use a command with the \"hs project\" prefix."
+        srcIsProject: "\"{{ src }}\" is a project folder. Did you mean \"hs project {{command}}\"?"
```
brandenrodgers, line 5: I was playing around with this a little bit, and I think we could improve the wording a bit more. The `src` folder won't always be a project folder. The `getIsInProject` util will also return true for a folder that's nested inside of a project.

Maybe something like this is better?
`The path "\"{{ src }}\" is in a project folder. Did you mean \"hs project {{command}}\"?"`
Or maybe just adding "in" is enough?
`"\"{{ src }}\" is in a project folder. Did you mean \"hs project {{command}}\"?"`

packages/cli/bin/cli.js
kemmerle: Discussed this yesterday in the 1-1, and we're going to hold off on covering the scenario with the `path` argument for now, since it seems much less common.

packages/cli-lib/lang/en.lyaml
kemmerle: I like Option 2, a bit simpler and still gets the point across.

brandenrodgers (approved):
Looks good!

PR #675, opened 2022-04-20
UX improvements to "project logs" command

brandenrodgers:
## Description and Context

We've gotten some early feedback about the `hs project logs` command not being very intuitive. This PR attempts to make it easier to use by making it more interactive. Rather than require a user to input all of the args, we now fall back to a prompt flow if any required args are missing.

Changes:
- The required options are now `--app` (app name), `--project` (project name), and `--function` (function name)
- We fall back to prompts if any of the required options are missing
- Users can now exit "follow" mode by pressing "Q"
- Added a link to Hubspot Private Apps UI
- Explicitly state which account, project, app, and function we are fetching logs for

## Screenshots
**Showcasing the prompt**

image

**With all required options + following**
image

**Missing one required option + showing the 'no logs' message**
image

**Example usage**
Apr-20-2022 12-48-55

## TODO

There are some known issues/limitations with this:
- [x] Right now this requires you to run it from within the project that you want to view app logs for. It doesn't support any sort of `projectPath` option. I want to look into how I could make this a little better. Maybe I could fetch the list of available projects from HubSpot, but then show the CWD project as the default option?
- [x] To get the `appPath` (a required param for the logs endpoint), I'm reading through the project's files and looking for the `app.json` files so I can match the `--app` (app name) option to the matching project component. Then I can derive the relative path to the component from the project src. The problem with doing it this way is that I'm relying on the user's local filesystem matching what's deployed in HubSpot. An example of where this can break down is if the user changes the app's folder path locally without uploading the changes.
- [ ] This currently routes to the home screen of Private Apps UI. It would be cool if I could route to the logs page for the specific app. I think that would require me being able to fetch the app's ID.


brandenrodgers:
In [e0ce12e](https://github.com/HubSpot/hubspot-cli/pull/675/commits/e0ce12e58f19ead67c91613b07d2991e9c002f26) I updated the code that generates the list of apps in a project. Rather than parsing the user's local files, I'm now hitting the projects endpoint and using the data from that to get the list of available apps.

I also updated the command to be able to run outside of the context of a project. So you should be able to run this from anywhere in your filesystem (not just from within a project). If you are in a project, it will automatically enter the project name as the default answer to the first prompt question.

kemmerle (commented):
I looked through the code, and it generally looks good--I do think I ran into an issue that's either a bug or that needs more explicit error handling. 

Just a question: the serverless function will only have logs if it's been deployed, correct? Because I was encountering an issue where I couldn't fetch logs for the `crm-cards` function in the `sample-project-4-18` project in my dev portal (Link [here](https://app.hubspotqa.com/developer-projects/101140939/project/sample-project-4-18/build/1) for quick reference). I kept getting a 404 error here, but the function does exist in my account, so I was a bit confused. Could there be a clearer error message here?

kemmerle (approved):
Ignore my above comment, I forgot to add `--` in front of the `--function` flag 🤦‍♀️ 

Code is good to go!

PR #676, opened 2022-04-20
Open projects page from CLI

adamawang:
## Description and Context

- Adds a new command `hs projects open` that will open the projects page in a browser window for the default account
  - `const url = ${websiteOrigin}/developer-projects/${accountId};`
- User can specify QA for testing 
- User can use the `--account=` option to specify which account to open projects for

## Screenshots

Screen Shot 2022-04-20 at 3 00 55 PM

## Who to Notify

@brandenrodgers

### DIFF
--- a/packages/cli/commands/project/open.js
+++ b/packages/cli/commands/project/open.js
@@ -12,13 +12,14 @@ const { logger } = require('@hubspot/cli-lib/logger');
 const {
   getProjectConfig,
   getProjectDetailUrl,
-  verifyProjectExists,
+  ensureProjectExists,
 } = require('../../lib/projects');
 const { projectNamePrompt } = require('../../lib/prompts/projectNamePrompt');
+const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 
 const i18nKey = 'cli.commands.project.subcommands.open';
 
-exports.command = 'open';
+exports.command = 'open [--project]';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.handler = async options => {
@@ -27,25 +28,28 @@ exports.handler = async options => {
   const accountId = getAccountId(options);
   const { project } = options;
   const { projectConfig } = await getProjectConfig();
+
   let projectName = project;
+
   if (projectName) {
-    const projectExists = await verifyProjectExists(accountId, projectName);
+    const projectExists = await ensureProjectExists(accountId, projectName, {
+      allowCreate: false,
+    });
+
     if (!projectExists) {
-      return;
+      process.exit(EXIT_CODES.ERROR);
     }
   } else if (!projectName && projectConfig) {
     projectName = projectConfig.name;
   } else if (!projectName && !projectConfig) {
     const namePrompt = await projectNamePrompt(accountId, projectConfig);
-    const projectExists = await verifyProjectExists(
-      accountId,
-      namePrompt.projectName
-    );
-    if (!projectExists) {
-      return;
+
+    if (!namePrompt.projectName) {
+      process.exit(EXIT_CODES.ERROR);
     }
     projectName = namePrompt.projectName;
   }
+
   const url = getProjectDetailUrl(projectName, accountId);
   open(url, { url: true });
   logger.success(i18n(`${i18nKey}.success`, { projectName }));
--- a/packages/cli/commands/project/open.js
+++ b/packages/cli/commands/project/open.js
@@ -1,5 +1,4 @@
 const open = require('open');
-const { ENVIRONMENTS } = require('@hubspot/cli-lib/lib/constants');
 const {
   addAccountOptions,
   addConfigOptions,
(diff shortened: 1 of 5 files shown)


brandenrodgers:
I think you could take this one step further by linking to a specific project in projects UI

We have a util for generating the details page URL: [getProjectDetailUrl](https://github.com/HubSpot/hubspot-cli/blob/7ef17e48b31d6cc9bff3a0f7ab7857df4b82f14c/packages/cli/lib/projects.js#L207)

You'd just need to support some sort of "--project" option and ideally a prompt fallback. Alternatively we could keep it as `hs project open` and then look up the project name (project name can be found in `projectConfig` [in this example](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/project/upload.js#L36)). The second option would just require the user to run the command from within a project, which is probably fine.

adamawang:
For the first option (if we're not in a project directory), would it be reliable to use `hs list` to find the project directories available? and use that list to prompt the user to choose one?

Maybe we could have both options where you'd get a list to choose from if you're not in a project directory, but if you are, then open that one

packages/cli/commands/project/open.js
```diff
@@ -0,0 +1,70 @@
...
+exports.describe = i18n(`${i18nKey}.describe`);
+
+exports.handler = async options => {
+  await loadAndValidateOptions(options);
+
+  const accountId = getAccountId(options);
+  const { project } = options;
+  const { projectConfig } = await getProjectConfig();
+  let projectName = project;
+  if (projectName) {
```
adamawang, line 31: If a `--project` option is provided, check if project exists based on the default account, or a provided account using the `--account` option. This can be used in any directory with a hubspot.config.yml file.

packages/cli/commands/project/open.js
```diff
@@ -0,0 +1,70 @@
...
+    const projectExists = await verifyProjectExists(accountId, projectName);
+    if (!projectExists) {
+      return;
+    }
+  } else if (!projectName && projectConfig) {
```
adamawang, line 36: If no project option is provided and a project config exists in the directory you're in, use that as the name.

packages/cli/commands/project/open.js
```diff
@@ -0,0 +1,70 @@
...
+    projectName = projectConfig.name;
+  } else if (!projectName && !projectConfig) {
```
adamawang, line 38: If no project option is provided and no project config exists, prompt the user for a name. Verify that a project exists with the name and default account, or the provided account using the `--account` option.

packages/cli/lib/projects.js
```diff
@@ -162,6 +162,26 @@ const validateProjectConfig = (projectConfig, projectDir) => {
   }
 };
 
+const verifyProjectExists = async (accountId, projectName) => {
```
adamawang, line 165: I created a new util here because `ensureProjectExists` wasn't able to satisfy my project verification use case

adamawang (commented):
Some comments below on all possible scenarios

packages/cli/lib/projects.js
brandenrodgers: I think you should be able to use `ensureProjectExists`. There's an arg `allowCreate: false` that forces the util to skip the "create project" flow when the project doesn't exist. 

You may just need to update the `ensureProjectExists` util to return a boolean, which IMO it should be doing anyways. If it's not too much of a lift, I'd prefer to only have one util for checking a project's existence.

packages/cli/commands/project/open.js
```diff
@@ -0,0 +1,70 @@
...
+const {
+  getProjectConfig,
+  getProjectDetailUrl,
+  verifyProjectExists,
+} = require('../../lib/projects');
+const { projectNamePrompt } = require('../../lib/prompts/projectNamePrompt');
+
+const i18nKey = 'cli.commands.project.subcommands.open';
+
+exports.command = 'open';
```
brandenrodgers, line 21: Can this be updated to `open [--project]` to help make the project option more discoverable?

packages/cli/commands/project/open.js
```diff
@@ -0,0 +1,70 @@
...
+  await loadAndValidateOptions(options);
+
+  const accountId = getAccountId(options);
+  const { project } = options;
+  const { projectConfig } = await getProjectConfig();
+  let projectName = project;
+  if (projectName) {
+    const projectExists = await verifyProjectExists(accountId, projectName);
+    if (!projectExists) {
+      return;
```
brandenrodgers, line 34: Instead of just returning, I think it would be better to log out some message along the lines of "The project <project> could not be found in the account <accountName>", and then exiting via `process.exit(EXIT_CODES.ERROR)`

packages/cli/commands/project/open.js
```diff
@@ -0,0 +1,70 @@
...
+      namePrompt.projectName
+    );
+    if (!projectExists) {
+      return;
+    }
+    projectName = namePrompt.projectName;
+  }
+  const url = getProjectDetailUrl(projectName, accountId);
+  open(url, { url: true });
+  logger.success(i18n(`${i18nKey}.success`, { projectName }));
```
brandenrodgers, line 51: We should make sure to exit with `process.exit` here too

brandenrodgers:
The fallbacks of the `--project` option are slightly different than how I have them set up in my [PR here.](https://github.com/HubSpot/hubspot-cli/pull/675).

TBH I'm personally fine with keeping this one as-is and seeing how it feels. It may be useful to play around with a few different flows. Just to point out the differences...

**This PR**
- if `--project` is passed, use that
- if not, fall back to CWD project and use that
- if no CWD project, fall back to a prompt

**The other PR**
- if `--project` is passed, use that
- if not, fall back to a prompt, and automatically enter the CWD project as the default option

I'll share with the team to get some more thoughts on it.

packages/cli/lib/projects.js
adamawang: Definitely no lift there, wanted to make sure I didn't change any behaviors of existing utils but will get this addressed

packages/cli/commands/project/open.js
adamawang: We get a log message if its a missing project from the `verifyProjectExists` util (will change to `ensureProjectExists`), didn't want to double up on messaging here. But good call on the process.exit, will add those in

brandenrodgers (approved):
This lgtm 👍 

I wasn't able to pull the code down to test though b/c it looks like this branch is on a forked version of the project.

adamawang:
Ah yeah, I wasn't able to figure out how to push directly to this repo... do I have to be added to some user list for that to work?

brandenrodgers (approved):

erincouse:
Hi @brandenrodgers I'm so sorry I'm so late on this. But now that it's merged (sorry)... I like the second option. I think it makes sense with our overall prompt strategy? We should probably start documenting some general guidelines for them that we can follow by default in the future.

if --project is passed, use that
if not, fall back to a prompt, and automatically enter the CWD project as the default option

PR #680, opened 2022-04-27
Fix bugs in HS link in `hs project upload`

kemmerle:
## Description and Context
We currently provide a link to the build page, when a user enters the `hs project upload` command. But some terminals don't support hyperlinks, and our condition was not falling back to show the correct url. 

## Screenshots

With changes:

`hs project upload` in Hyper shell (doesn't support hyperlinks):

Screen Shot 2022-04-27 at 4 49 36 PM

`hs project upload` in ITerm2 shell (does support hyperlinks):

Screen Shot 2022-04-27 at 4 50 59 PM

## TODO
@brandenrodgers, if you could test these changes out in different terminals, that would be much appreciated. I tested out the terminals listed in [this article](https://www.macobserver.com/news/mac-terminal-app-alternatives/) along with the VSCode terminal, but I'm not sure whether I'm missing a common terminal. 

I also was not sure which checks to delete from the new `supportHyperlinks` file. I imagine that the check for `netlify` is the best candidate, since I can't imagine a workflow where users would be using the hubspot cli from netlify.  

## Who to Notify
@brandenrodgers


package.json
```diff
@@ -40,6 +40,8 @@
     }
   },
   "dependencies": {
-    "jest": "^26.6.3"
+    "has-flag": "^4.0.0",
```
kemmerle, line 43: I wasn't sure whether I need to add dependencies in this package.json file, or whether adding them to the cli folder's package.json is sufficient. I would love to discuss this in more detail in our 1:1 tomorrow, so I can have a better understanding of this in the future.

packages/cli/lib/supportHyperlinks.js
```diff
@@ -0,0 +1,104 @@
...
+
+  if (stream && !stream.isTTY) {
+    return false;
+  }
+
+  if (process.platform === 'win32') {
+    return false;
+  }
+
+  if ('NETLIFY' in env) {
```
kemmerle, line 61: It seems to me that we could remove the checks for netlify and teamcity with little impact on users, but I wanted to doublecheck.

packages/cli/lib/ui.js
```diff
@@ -31,7 +31,9 @@ const uiLink = (linkText, url, options = {}) => {
     ].join('');
     return options.useColor ? chalk.cyan(result) : result;
   } else {
-    return options.fallback ? `${linkText}: ${url}` : linkText;
+    return options.useColor
```
brandenrodgers, line 34: I think we should get rid of this `options.useColor` option. We should use color in every terminal that supports it.

package.json
brandenrodgers: I don't think we need to add this as a dep. Ideally we can reduce the overall number of deps in the project. This dep is a single small function that we could easily wrap up into `supportHyperlinks`. https://github.com/sindresorhus/has-flag/blob/main/index.js

packages/cli/lib/supportHyperlinks.js
```diff
@@ -0,0 +1,104 @@
+'use strict';
```
brandenrodgers, line 1: If we can, I think it's also worthwhile to look into removing `supports-color` as a dep and wrapping that functionality into this file (or maybe pull it into its own file). It would also be cool if we had some `getTerminalUISupport` util function that aggregated these color + hyperlinks utils and returned something along the lines of:
```js
{
 hyperlinks: true,
 color: false
}
```
Then all of the UI utils in [ui.js](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/ui.js) can use it to determine how to render output

packages/cli/lib/supportHyperlinks.js
```diff
@@ -0,0 +1,104 @@
...
+  ) {
+    return false;
+  }
+
+  if (hasFlag('hyperlink=true') || hasFlag('hyperlink=always')) {
+    return true;
+  }
+
+  // If they specify no colors, they probably don't want hyperlinks.
+  if (!supportsColor.supportsColor(stream)) {
```
brandenrodgers, line 49: IMO we should treat "supports color" and "supports hyperlinks" separately

packages/cli/lib/supportHyperlinks.js
brandenrodgers: Yeah I agree. I think we can remove these 👍

packages/cli/lib/supportHyperlinks.js
kemmerle: I'm going to disagree a bit here to say that I think this is a reasonable check. I have, however, added another separate check for colors in the new `getTerminalUISupport` util function, which I feel is a good compromise.

packages/cli/lib/ui.js
```diff
@@ -12,6 +13,19 @@ const uiLine = () => {
...
+/**
+ * Returns an object that aggregates what the terminal supports (eg. hyperlinks and color)
+ *
+ * @returns {object}
+ */
+
+const getTerminalUISupport = () => {
+  return {
+    hyperlinks: supportsHyperlinks.stdout,
+    color: supportsColor.stdout.level >= 1,
```
kemmerle, line 25: I'm not quite sure what check to perform here. 

`supportsColor.stout` returns the following object: 

```
{ level: 3, hasBasic: true, has256: true, has16m: true }
```

Should I simply check that the level is greater than or equal to 1 (which means the terminal has basic color)? Or would it be better to check whether the `hasBasic` boolean returns true? That seems to be enough for our purposes, since we're basically just using chalk to color text, correct? I wanted to check, though, in case I'm missing something.

brandenrodgers: That makes sense to me. `hasBasic` sounds like a reasonable thing to check.

packages/cli/lib/supportHyperlinks.js
```diff
@@ -0,0 +1,96 @@
...
+  ) {
+    return false;
+  }
+
+  if (hasFlag('hyperlink=true') || hasFlag('hyperlink=always')) {
+    return true;
+  }
+
+  // If they specify no colors, they probably don't want hyperlinks.
+  if (!supportsColor.createSupportsColor(stream)) {
```
brandenrodgers, line 49: I think we should probably remove this check because we're separately checking for whether or not the terminal supports color. We shouldn't be assuming that they don't want hyperlinks just because they don't support color.

kemmerle:
@brandenrodgers, I think the PR is ready for a final (🤞 ) review. If could test out my changes in a few more terminals, that would be much appreciated!

packages/cli/lib/supportsColor.js
```diff
@@ -0,0 +1,122 @@
+const process = require('process');
+const os = require('os');
+const tty = require('tty');
+const hasFlag = require('./hasFlag');
+
+const { env } = process;
+
+function translateLevel(level) {
+  if (level === 0) {
+    return false;
```
brandenrodgers, line 10: I just noticed, but this returns either `false` or an object. Maybe we should update this function to consistently return an object?

kemmerle: I would hesitate to change this, since the function `createSupportsColor` returns `translateLevel`, and `createSupportsColor` is critical to the util. If we were to change it, I think it would require a fair bit of testing, and I'm not sure the juice is worth the squeeze.

packages/cli/bin/cli.js
```diff
@@ -82,6 +82,18 @@ const argv = yargs
     describe: 'set log level to debug',
     type: 'boolean',
   })
+  .option('noHyperlinks', {
```
kemmerle, line 85: `no-hyperlinks` and `no-colors` both errored as flags, so I changed them to camel case, and that fixed the issue.

packages/cli/lib/supportsColor.js
kemmerle: Just discussed over Slack, will change now.

brandenrodgers (approved):
🎉

PR #682, opened 2022-05-04
Log useful account commands when user auths a new account

brandenrodgers:
## Description and Context

This is an idea to help make account switching in the CLI easier to understand and more discoverable:
- After the user runs `hs auth` or `hs init` we log out some additional commands that help make account handling easier. 
- I updated the prompts + copy in the flow to (hopefully) make them easier to understand. 
- The user now has the option to automatically set the newly connected account as the default when running `ha auth`

### hs init (OLD)
image

### hs init (NEW)
image

### hs auth (OLD)
image

### hs auth (NEW) w/ setting as default
image

### hs auth (NEW) w/out setting as default
image

### DIFF
--- a/packages/cli/commands/sandbox/create.js
+++ b/packages/cli/commands/sandbox/create.js
@@ -26,56 +26,24 @@ const {
 } = require('@hubspot/cli-lib/personalAccessKey');
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 const {
+  getConfig,
   writeConfig,
   updateAccountConfig,
   getAccountConfig,
 } = require('@hubspot/cli-lib');
-const { promptUser } = require('../../lib/prompts/promptUtils');
-const { accountNameExistsInConfig } = require('@hubspot/cli-lib/lib/config');
-const { STRING_WITH_NO_SPACES_REGEX } = require('../../lib/regex');
+const {
+  enterAccountNamePrompt,
+} = require('../../lib/prompts/enterAccountNamePrompt');
+const {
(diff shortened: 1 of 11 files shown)


brandenrodgers:
### Update

I updated the spacing of this a little and added this same "feature highlight" to the `hs init` command too:

#### hs init
Screen Shot 2022-05-04 at 6 11 51 PM

#### hs auth
Screen Shot 2022-05-04 at 5 13 02 PM

brandenrodgers:
Latest iteration:
image (2)

packages/cli/commands/sandbox/create.js
```diff
@@ -85,7 +53,16 @@ const personalAccessKeyFlow = async (env, account, name) => {
     process.exit(EXIT_CODES.SUCCESS);
   }
 
-  const validName = await promptForAccountNameIfNotSet(updatedConfig, name);
+  let validName = updatedConfig.name;
```
brandenrodgers, line 56: @adamawang just giving you the heads up. I tried to DRY things up a little bit by sharing prompts between this command and the auth command. I also added the new `hs auth` logs to the "personalAccessKeyFlow" portion of this command. If you get the chance, I'd love some eyes on the changes in this file to make sure everything makes sense.

adamawang: Everything looks perfect, thanks for the heads up! 🎉

TheWebTech:
This is great. Too many folks don't know about `hs accounts use`. This should help with adoption.

PR #684, opened 2022-05-05
Removing package-lock.json and yarn.lock

brandenrodgers:
## Description and Context

Related article about why it's not necessary to include these: https://www.twilio.com/blog/lockfiles-nodejs
NPM docs explaining that the package-lock.json file is always ignored: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files

## TODO

- [ ] Is this actually an okay change? I still need to verify


joe-yeager:
Closing since this is a couple years old

PR #686, opened 2022-05-06
add hs accounts add command

brandenrodgers:
## Description and Context

Adding a `hs accounts add` command that will eventually replace `hs auth`. We've been getting feedback lately that `hs accounts add` is more intuitive than `hs auth`. For now, we will support both commands for backwards compatibility. This works the exact same way that `hs auth` works.

Note: As a part of this update, I'm making it so that `hs accounts add` no longer supports auth via api key which is a deprecated auth method that was replaced by personal access tokens.

Questions:
- Should we have a separate command for the scenario where a user wants to refresh a personal access token for a Hubspot account that is already connected to the CLI? "add" makes it sound like this command is only used to connect net-new accounts to the CLI.

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -39,6 +39,17 @@ en:
             promptMessage: "Select an account to use as the default"
             success:
               defaultAccountUpdated: "Default account updated to \"{{ accountName }}\""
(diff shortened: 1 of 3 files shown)


brandenrodgers:
deleting this. We can revisit it another time

PR #687, opened 2022-05-06
Sandboxes dev platform

adamawang:
## Description and Context

Updates for `hs sandbox create`
- The CLI auto prompts the user to generate a PAK and add to the config file. 
- The sandbox name is pulled in by default, but gives the user an option to use a custom one. 
- Adds an `accountId` option to the PAK prompt, where we forward the user to the PAK generation page rather than the account picker. 

## Screenshots

Screen Shot 2022-05-06 at 12 01 08 PM

## TODO

Have some updates to `hs init` and `hs auth` that I will put up in a separate PR. Those updates will also use the `accountId` option in the PAK prompt

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -553,6 +553,7 @@ en:
         subcommands:
           create:
             describe: "Create a sandbox account"
+            enterAccountName: "Enter a unique name to reference your account: "
             debug:
               creating: "Creating sandbox \"{{ name }}\""
             examples:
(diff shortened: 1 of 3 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -553,6 +553,7 @@ en:
         subcommands:
           create:
             describe: "Create a sandbox account"
+            enterAccountName: "Enter a unique name to reference your account: "
```
adamawang, line 556: I used the same line as the `hs init` step. It currently prompts the user for a name but sets the sandbox name as the default if the user hits enter. 

Taking Mark's [comment here](https://hubspot.slack.com/archives/C01GV708YH3/p1651516531083669?thread_ts=1651167455.710479&cid=C01GV708YH3) into consideration, should we skip that step and just set the sandbox name as the cli config name?

PR #688, opened 2022-05-06
POC for showing command shortcuts

brandenrodgers:
## Description and Context

Showing what it could look like for us to show "shortcuts" (name is WIP) to the user whenever we show them a prompt.

The idea is that it helps the users to work faster. The logic I have in place right now makes it so that we only show the shortcut to users if they answered at least one prompt question. I'll post some screenshots to show how it works.

### Example where we show them the shortcut b/c they didn't pass in any options
image

### Example where we show them the shortcut because we asked one prompt question
image

### Example where we wouldn't show the shortcut because we never showed them the prompt
image


brandenrodgers:
Deleting this. We can always revisit

PR #690, opened 2022-05-06
hs sandbox create

adamawang:
## Description and Context

Updates for `hs sandbox create`
- The CLI auto prompts the user to generate a PAK and add to the config file. 
- The sandbox name is pulled in by default, but gives the user an option to use a custom one. 
- Adds an `account` option to the PAK prompt, where we forward the user to the PAK generation page rather than the account picker. 
- Adds an error message specific to missing scopes when creating a sandbox

## Screenshots

Screen Shot 2022-05-06 at 5 06 47 PM

Screen Shot 2022-05-06 at 5 03 28 PM

## TODO

Have some updates to `hs init` and `hs auth` that I will put up in a separate PR. Those updates will also use the `accountId` option in the PAK prompt

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -553,6 +553,7 @@ en:
         subcommands:
           create:
             describe: "Create a sandbox account"
+            enterAccountName: "Enter a unique name to reference your account: "
             debug:
               creating: "Creating sandbox \"{{ name }}\""
             examples:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -564,6 +564,7 @@ en:
                 describe: "Name to use for created sandbox"
             success:
               create: "Sandbox \"{{ name }}\" with portalId \"{{ sandboxHubId }}\" created successfully."
+              configFileUpdated: "{{ configFilename }} updated with {{ authMethod }}."
       secrets:
         describe: "Manage HubSpot secrets."
         subcommands:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -565,7 +565,7 @@ en:
                 describe: "Name to use for created sandbox"
             success:
               create: "Sandbox \"{{ name }}\" with portalId \"{{ sandboxHubId }}\" created successfully."
-              configFileUpdated: "{{ configFilename }} updated with {{ authMethod }}."
+              configFileUpdated: "{{ configFilename }} updated with {{ authMethod }} for account {{ account }}."
       secrets:
         describe: "Manage HubSpot secrets."
         subcommands:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -573,9 +573,9 @@ en:
               configFileUpdated: "{{ configFilename }} updated with {{ authMethod }} for account {{ account }}."
             failure:
               scopes:
-                message: "The HubSpot CLI doesn’t have permission to create sandboxes under \"{{ accountName }}\"."
+                message: "The personal access key you provided doesn't include sandbox permissions."
                 instructions: "To update CLI permissions for \"{{ accountName }}\":
-                  \n- Visit {{ url }}, deactivate the existing personal access key, and create a new one that includes Sandboxes permissions.
+                  \n- Go to {{ url }}, deactivate the existing personal access key, and create a new one that includes Sandboxes permissions.
                   \n- Update the CLI config for this account by running `hs auth` and entering the new key.\n"
       secrets:
         describe: "Manage HubSpot secrets."
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -570,12 +570,8 @@ en:
               scopes:
                 message: "The HubSpot CLI doesn’t have permission to create sandboxes under \"{{ accountName }}\"."
                 instructions: "To update CLI permissions for \"{{ accountName }}\":
-
-                  - Visit {{ url }}, deactivate the existing personal access key, and create a new one that includes Sandboxes permissions.
-
-                  - Update the CLI config for this account by running `hs auth` and entering the new key.
-
-                "
+                  \n- Visit {{ url }}, deactivate the existing personal access key, and create a new one that includes Sandboxes permissions.
+                  \n- Update the CLI config for this account by running `hs auth` and entering the new key.\n"
       secrets:
         describe: "Manage HubSpot secrets."
         subcommands:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -565,24 +565,17 @@ en:
                 describe: "Name to use for created sandbox"
             success:
               create: "Sandbox \"{{ name }}\" with portalId \"{{ sandboxHubId }}\" created successfully."
-              prePrompt: "
-
-
-              The following step will prompt you to authenticate with sandbox \"{{ name }}\" and add it to the CLI config.
-
-              "
               configFileUpdated: "{{ configFilename }} updated with {{ authMethod }} for account {{ account }}."
             failure:
-              scopes: "
+              scopes:
+                message: "The HubSpot CLI doesn’t have permission to create sandboxes under \"{{ accountName }}\"."
+                instructions: "To update CLI permissions for \"{{ accountName }}\":
 
+                  - Visit {{ url }}, deactivate the existing personal access key, and create a new one that includes Sandboxes permissions.
 
-              Verify that the personal access key used has the Sandboxes scope by navigating to \"{{ url }}\".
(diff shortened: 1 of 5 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -553,6 +553,7 @@ en:
         subcommands:
           create:
             describe: "Create a sandbox account"
+            enterAccountName: "Enter a unique name to reference your account: "
```
adamawang, line 556: This is the same line we use in the init step. I currently have it prompt the user for a name, but defaults to the sandbox name the user just typed in in case they want it custom. 

Keeping Mark's [comment in mind here](https://hubspot.slack.com/archives/C01GV708YH3/p1651516531083669?thread_ts=1651167455.710479&cid=C01GV708YH3), do we want to skip the prompt and just set the same name by default?

packages/cli/commands/sandbox/create.js
```diff
@@ -37,18 +120,43 @@ exports.handler = async options => {
...
+          })
+        );
+        return { name, sandboxHubId };
+      }
+    );
+  } catch (err) {
+    if (
+      err.error &&
+      err.error.category &&
+      err.error.category === 'MISSING_SCOPES'
```
adamawang, line 140: Added a scope specific error here:

Screen Shot 2022-05-06 at 2 55 46 PM

packages/cli-lib/lang/en.lyaml
brandenrodgers: That would be really cool! IMO what you have now is reasonable, and we can always tack that on as an update later if we want to.

packages/cli/commands/sandbox/create.js
```diff
@@ -11,9 +12,90 @@ const { createSandbox } = require('@hubspot/cli-lib/sandboxes');
...
+            name: promptAnswer.name,
+          })
+        );
+      }
+    }
+    return validName;
+  }
+};
+
+const personalAccessKeyFlow = async (env, accountId, name) => {
```
brandenrodgers, line 73: Don't worry about handling this in this PR, but I'm noticing that we're accumulating some repetitive auth-related prompts (I have some WIP work that's also adding more of it). I may try to consolidate some of it later on.

packages/cli/commands/sandbox/create.js
```diff
@@ -37,18 +120,47 @@ exports.handler = async options => {
...
+          `
 
-  return createSandbox(accountId, sandboxName).then(
-    ({ name, sandboxHubId }) => {
-      logger.success(
-        i18n(`${i18nKey}.success.create`, {
-          name,
-          sandboxHubId,
-        })
+The following step will prompt you to authenticate with sandbox "${name}" and add it to the CLI config.
```
brandenrodgers, line 134: Can this also be added to the en.lyaml?

packages/cli/commands/sandbox/create.js
brandenrodgers: Looks like we have a util for this [here](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/errorHandlers/apiErrors.js#L24). You may be able to export that and use it here

packages/cli/commands/sandbox/create.js
```diff
@@ -37,18 +120,47 @@ exports.handler = async options => {
...
+    );
+  } catch (err) {
+    if (
+      err.error &&
+      err.error.category &&
+      err.error.category === 'MISSING_SCOPES'
+    ) {
+      const websiteOrigin = getHubSpotWebsiteOrigin(env);
+      const url = `${websiteOrigin}/personal-access-key/${accountId}`;
+      logger.info(
```
brandenrodgers, line 148: We may want to get some design eyes on this message.

packages/cli/commands/sandbox/create.js
adamawang: Ah yep! I originally had it in `cli-lib` but ended up moving it back to the cli package. Will add translations

packages/cli-lib/lang/en.lyaml
```diff
@@ -564,6 +565,24 @@ en:
...
               create: "Sandbox \"{{ name }}\" with portalId \"{{ sandboxHubId }}\" created successfully."
+              prePrompt: "
+
+
+              The following step will prompt you to authenticate with sandbox \"{{ name }}\" and add it to the CLI config.
+
+              "
+              configFileUpdated: "{{ configFilename }} updated with {{ authMethod }} for account {{ account }}."
+            failure:
+              scopes: "
```
adamawang, line 576: Not sure the best way to do this, but I added a bunch of white space here so the messaging wouldn't get bunched up/all on one line

Screen Shot 2022-05-06 at 4 22 12 PM

brandenrodgers: Nice 👍  I like it with the extra spacing. FWIW I've been using blank `logger.log` calls to add newlines. You can also use `\n` in the en.lyaml file to add newlines (there should be examples of it in there somewhere).

adamawang:
Received updated copy and direction from Mark - updated screenshots in main comment!

packages/cli/lib/prompts/personalAccessKeyPrompt.js
```diff
@@ -15,10 +15,13 @@ const i18nKey = 'cli.lib.prompts.personalAccessKeyPrompt';
  * Displays notification to user that we are about to open the browser,
  * then opens their browser to the personal-access-key shortlink
  */
-const personalAccessKeyPrompt = async ({ env } = {}) => {
+const personalAccessKeyPrompt = async ({ env, accountId } = {}) => {
```
brandenrodgers, line 18: I'm sure this is already on your radar, but just a reminder to update things accordingly after [this](https://github.com/HubSpot/hubspot-cli/pull/691) is merged.

adamawang: Good catch - updated and resolved! Waiting on Erin if she has any updates before merging, thread here: https://hubspot.slack.com/archives/C01GV708YH3/p1652106906292789?thread_ts=1651863988.648859&cid=C01GV708YH3

brandenrodgers (approved):
LGTM (pending final copy updates)

packages/cli-lib/lang/en.lyaml
erincouse: I think we should set the same name by default, 100%. Let's make everyone's lives easier.

packages/cli/commands/sandbox/create.js
erincouse: what's the message? Forgive my Github incompetence.

packages/cli/commands/sandbox/create.js
```diff
@@ -118,18 +120,37 @@ exports.handler = async options => {
...
+        `
+          Verify that the personal access key used has the Sandboxes scope by navigating to "${url}".
```
erincouse: is this the error message? I would say "The personal access key you provided doesn't give access to sandbox permissions. Deactivate the key and generate a new one with the correct permissions. After that, run `hs auth` again and enter the new key."

packages/cli/commands/sandbox/create.js
```diff
@@ -128,7 +128,11 @@ exports.handler = async options => {
           i18n(`${i18nKey}.success.create`, {
             name,
             sandboxHubId,
-          })
+          }),
+          `
+
+The following step will prompt you to authenticate with sandbox "${name}" and add it to the CLI config.
```
erincouse: are you authenticating with a sandbox, or are you authenticating the sandbox? genuine question, not sure if there's nuance needed with the language here

packages/cli-lib/lang/en.lyaml
```diff
@@ -565,7 +565,24 @@ en:
                 describe: "Name to use for created sandbox"
             success:
               create: "Sandbox \"{{ name }}\" with portalId \"{{ sandboxHubId }}\" created successfully."
+              prePrompt: "
+
+
+              The following step will prompt you to authenticate with sandbox \"{{ name }}\" and add it to the CLI config.
```
erincouse: do they need to take an action to get the the following step?

packages/cli/commands/sandbox/create.js
adamawang: Yep this is a permissions error message here - I'll get this change rolled in!

packages/cli/commands/sandbox/create.js
adamawang: This step was removed in a later commit here so we should be all set on this

packages/cli-lib/lang/en.lyaml
adamawang: Same as above - addressed in a later commit and this section has been removed

packages/cli/commands/sandbox/create.js
adamawang: After checking in with Erin, we adjusted to use `The personal access key you provided doesn't include sandbox permissions` in place of the "ERROR" line. Also adjusted `Visit` to `Go to` for the URL line.

PR #691, opened 2022-05-06
Init and auth options

adamawang:
## Description and Context

- Adds `--account` option to both `hs auth` and `hs init` which will change the PAK prompt link and forward the user directly to the PAK generation page. 
- Changes the `hs auth` `[type]` positional to an optional. Updated command help text to reflect

## Screenshots

Screen Shot 2022-05-06 at 12 38 05 PM

Screen Shot 2022-05-06 at 12 38 41 PM

Screen Shot 2022-05-06 at 12 39 34 PM

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -50,8 +50,8 @@ en:
             describe: "Authentication mechanism"
             defaultDescription: "\"{{ authMethod }}\": \nAn access token tied to a specific user account. This is the recommended way of authenticating with local development tools."
         options:
-          accountId:
-            describe: "Account id to authenticate to"
+          account:
+            describe: "HubSpot account to authenticate"
         success:
           configFileUpdated: "{{ configFilename }} updated with {{ authMethod }}."
       config:
@@ -335,8 +335,8 @@ en:
           auth:
             describe: "Specify auth method to use [\"personalaccesskey\", \"oauth2\", \"apikey\"]"
             defaultDescription: "\"{{ defaultType }}\": \nAn access token tied to a specific user account. This is the recommended way of authenticating with local development tools."
-          accountId:
-            describe: "Account id to authenticate to"
+          account:
+            describe: "HubSpot account to authenticate"
         success:
           configFileCreated: "The config file \"{{ configPath }}\" was created using \"{{ authType }}\" for account {{ account }}."
         info:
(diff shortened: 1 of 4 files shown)


packages/cli/commands/auth.js
```diff
@@ -66,13 +66,13 @@ const promptForAccountNameIfNotSet = async updatedConfig => {
   }
 };
 
-exports.command = 'auth [type]';
+exports.command = 'auth [--type] [--accountId]';
```
brandenrodgers, line 69: Can this be `--account` for consistency with the other commands?

packages/cli/commands/auth.js
```diff
@@ -158,18 +158,27 @@ exports.handler = async options => {
...
-    choices: [
-      `${PERSONAL_ACCESS_KEY_AUTH_METHOD.value}`,
-      `${OAUTH_AUTH_METHOD.value}`,
-      `${API_KEY_AUTH_METHOD.value}`,
-    ],
-    default: PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
-    defaultDescription: i18n(`${i18nKey}.positionals.type.defaultDescription`, {
-      authMethod: PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
-    }),
+  yargs.options({
```
brandenrodgers, line 161: This is a breaking change for the "type" arg. We're moving towards using options for all of our args, but we'll need to keep this backwards compatible. We can support the `--type` option, but we'll still need to support the "type" positional

packages/cli-lib/lang/en.lyaml
```diff
@@ -332,6 +335,8 @@ en:
           auth:
             describe: "Specify auth method to use [\"personalaccesskey\", \"oauth2\", \"apikey\"]"
             defaultDescription: "\"{{ defaultType }}\": \nAn access token tied to a specific user account. This is the recommended way of authenticating with local development tools."
+          accountId:
+            describe: "Account id to authenticate to"
```
brandenrodgers, line 339: Copy suggestion: "HubSpot account to authenticate"

packages/cli/commands/auth.js
adamawang: Hmm, might be too verbose to keep the [--type] option there alongside the positional in the `command` line, unless we keep it hidden until its not needed?

brandenrodgers (approved):
Works great! 🎉

PR #692, opened 2022-05-17
Add/whoami command

### LINKED ISSUE #627: Add command to display user info
## Description and Context
Create a command (`hs whoami` for example) that prints information about the current user and portal.

**Important**: Display the scopes available to the user for the associated portal. This will be useful for debugging.

kcmaxwell:
## Description and Context

This pull request adds an `hs whoami` command that prints the default account name, ID, and the scopes available to it.
It can also be called using a specified account name or ID to print information for that account.
Resolves  #627

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,9 +12,7 @@
         "acceptance-tests"
       ],
       "dependencies": {
-        "has-flag": "^5.0.1",
-        "jest": "^26.6.3",
-        "supports-color": "^9.2.2"
+        "jest": "^26.6.3"
       },
       "devDependencies": {
         "depcheck": "^1.3.1",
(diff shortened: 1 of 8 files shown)


brandenrodgers:
Thanks for submitting this @kcmaxwell! Just to give an update, I'm happy with the code you have in here but I'm running this by our team to verify the command structure and output meshes with some of the upcoming plans we have for the CLI.

brandenrodgers:
Hey @kcmaxwell I chatted with the team and have some feedback for you!

We're making a push to move all account-related commands into the `hs accounts` bucket. To give an example, in the latest release of the CLI we moved the command for switching default accounts to `hs accounts use`. With that in mind, could you update this so that a user can run it via `hs accounts info`? "info" feels like a good word for it because it enables us to add more stuff to the output of the command down the road (beyond just default account and scopes).

Let me know if you have any clarifying questions!

brandenrodgers (approved):
Looks great! Thanks for helping out 🎉

brandenrodgers:
@kcmaxwell I was testing this out one last time before merging it, and noticed there's a gap. This currently only works with the personal access key auth type. I think there's still value in releasing this even without support for the oauth auth type, but we should probably verify that `personalAccessKey` exists before making the call to `getAccessToken`. Ideally we can abort early with some sort of warning. Maybe along the lines of "This command currently only supports fetching scopes for the personal access key auth type".

brandenrodgers:
Thanks @kcmaxwell, this is good to go!

PR #693, opened 2022-05-17
Allow .hubl extension

pipriles:
## Description and Context

Allow `.hubl` extension files, used for example to store macros to be shared between HTML and CSS files.
Changed the `ALLOWED_EXTENSIONS` and `HUBL_EXTENSIONS` variables.


pipriles:
Closed because I just noticed that the Design Manager does not allow you to upload those types of files in the first place.

PR #695, opened 2022-05-17
Fixed response prop check to allow loading of multiple pages of files

### LINKED ISSUE #694: Filemanager fetch limited to 99 files
## Description and Context
When running `hs filemanager fetch --overwrite /Some/remote/path ./temp/files` and `/Some/remote/path` has a lot of files (7000+ in this case), I only end up getting 99 files and the command states `[SUCCESS] Completed fetch of folder "/Some/remote/path" to "/Users/mtalley/temp/files" from the File Manager`.

## Expected behavior
I would expect to get all 7000+ files.

## Who to Notify
@brandenrodgers @anthmatic

miketalley:
## Description and Context
Fixes #694 

## Who to Notify
@anthmatic @brandenrodgers @kemmerle

### DIFF
--- a/packages/cli-lib/fileManager.js
+++ b/packages/cli-lib/fileManager.js
@@ -134,7 +134,7 @@ async function fetchAllPagedFiles(accountId, folderId, { includeArchived }) {
     });
 
     if (totalFiles === null) {
-      totalFiles = response.total;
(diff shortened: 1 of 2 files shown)


brandenrodgers (approved):
👌  thanks for fixing this

PR #697, opened 2022-05-17
Delete sandbox placeholder message

adamawang:
## Description and Context

Add a placeholder command for `hs sandbox delete` where we display a message telling the user to go to their HubSpot portal to delete the account. 

## Screenshots

Screen Shot 2022-05-17 at 6 39 48 PM

## TODO

We will eventually replace this with an actual delete command once `LocalDevAuth` is updated to include hubTypes. Issue tracked here: https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/327

## Who to Notify

@brandenrodgers @erincouse

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -590,6 +590,8 @@ en:
             options:
               account:
                 describe: "Account name or id to delete"
+            temporaryMessage: "The CLI currently does not support deleting sandboxes.
+              \n\nTo delete a sandbox, go to https://app.hubspot.com and navigate to the Sandboxes accounts page to delete your sandbox.\n"
       secrets:
         describe: "Manage HubSpot secrets."
         subcommands:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -578,6 +578,18 @@ en:
                 instructions: "To update CLI permissions for \"{{ accountName }}\":
                   \n- Go to {{ url }}, deactivate the existing personal access key, and create a new one that includes Sandboxes permissions.
                   \n- Update the CLI config for this account by running `hs auth` and entering the new key.\n"
+          delete:
+            describe: "Delete a sandbox account"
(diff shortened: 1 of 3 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -578,6 +578,18 @@ en:
                 instructions: "To update CLI permissions for \"{{ accountName }}\":
                   \n- Go to {{ url }}, deactivate the existing personal access key, and create a new one that includes Sandboxes permissions.
                   \n- Update the CLI config for this account by running `hs auth` and entering the new key.\n"
+          delete:
+            describe: "Delete a sandbox account"
+            examples:
+              default: "Deletes the sandbox account named MySandboxAccount."
+            success:
```
brandenrodgers, line 585: Looks like the success and error keys aren't being used. Is that intentional because they'll be used later?

brandenrodgers (approved):
One thing that could potentially make this a little more useful to the user would be to link into hubspot, but this update is still a step forward so I'm okay with merging 👍

adamawang:
Without the hierarchy thing, I'm not able to generate an accurate link to the sandboxes page since the URL requires that parent hub ID... I might as well just add a link to `app.hubspot.com` to at least make it clear on what hubspot website to go to

packages/cli-lib/lang/en.lyaml
adamawang: Yeah, I had a bunch of code that I deleted as placeholder but left the translations in for now. Hopefully we'll get around to completing this soon, but totally happy to remove if we don't want dead end strings lying around

brandenrodgers: Makes sense. I'll leave the decision up to you. If you're confident that you'll update this soon-ish then I'm okay with keeping it.

adamawang: Yep! cms dev infra team just got back to me with some info on how best to start working on the auth part, will start on it next Monday

PR #699, opened 2022-05-23
Prevent max event listener warning on watch

brandenrodgers:
## Description and Context

There was an issue where running the watch command for a while would cause a "max event listener limit exceeded" warning to show (See screenshot).

## Screenshots

image


kemmerle (approved):
LGTM!

PR #701, opened 2022-06-01
Remove support for new apikey auth connections

brandenrodgers:
## Description and Context

We are deprecating support for connecting to HubSpot accounts via the `apikey` auth type. We will continue to support existing CLI configs that have api key as the authtype (for now), but we are preventing new connections to be established. Users will now have to use the recommended Personal Access Key auth type or oauth.

## TODO

- Do we want better handling for the scenario where a user runs `hs auth --type="apikey"`? For example, rather than error out (which is what's happening now), we could give messaging along the lines of "The apikey auth method is no longer supported, please use personalaccesskey instead". I didn't do this in the first pass because supporting "apikey" as a valid `type` option means that it will show up as an available choice when running `hs auth --help`. I figured it was better to reduce the discoverability of it and keep it hidden.

### DIFF
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -2,20 +2,40 @@
 
 This project tests `hs.*` commands as if they were being used by an end-user.
 
+
+
 ## Getting Started
 
 The main test execution is kicked off by running `yarn test-cli` from the root of `hubspot-cli`. This will run the `run-tests` scipe which itself is a CLI, so to see the available options, you can `run-tests --help`. This was done so that the tests can be installed against any given version of the CLI and tested against it. In addition, it was done to keep the test logic separated from the CLI library.
 
+
+
 Note that if you are testing against a QA portal, not a PROD one, you'll need to add the `--qa` flag when running the script. There is still an outstanding issue with this because we attempt to add the `--qa` flag to all `hs` commands, however it is not available for all commands.
+
+
+### Setup
+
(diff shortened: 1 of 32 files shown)


miketalley:
> Do we want better handling for the scenario where a user runs hs auth --type="apikey"? For example, rather than error out (which is what's happening now), we could give messaging along the lines of "The apikey auth method is no longer supported, please use personalaccesskey instead". I didn't do this in the first pass because supporting "apikey" as a valid type option means that it will show up as an available choice when running hs auth --help. I figured it was better to reduce the discoverability of it and keep it hidden.

What if we add that notification/warning and make the `hs auth apikey` command hidden via `exports.describe = false`? ((example of this with sandbox command)[https://github.com/HubSpot/hubspot-cli/blob/1e974c2941b38a5ac1a379cbf9cbcb3ed2c156f9/packages/cli/commands/sandbox.js#L8])

miketalley (commented):
This is great and really brings us a step closer to a more secure experience! 🥇 

I would suggest leaving the tests since we are only removing the capability if adding an api key, however, we still want to support the functionality of legacy configs and make sure that code is working.

Also, do we have any plans to update the public docs to mark this as deprecated or remove it?

brandenrodgers:
> What if we add that notification/warning and make the hs auth apikey command hidden via exports.describe = false?

@miketalley the issue with that is that the syntax is actually `hs auth type="apikey"`, where we pass apikey in through the `type` option. AFAIK there's no way to support but also hide an individual option via [choices](https://github.com/HubSpot/hubspot-cli/pull/701/files#diff-784213d66c657afc6bd87be8a0efbbbf24c3c64062eb1e3f69a416307a99fd58L188) in yargs. Definitely let me know if I'm wrong about that though!

brandenrodgers:
Thanks for the review @miketalley! I'm going to get this out into a beta release so we can test it to make sure it's all working as expected.

PR #710, opened 2022-06-24
Fix hs init --c flag

j-malt:
## Description and Context
This PR fixes a bug in the CLI where, if a config file exists and the user runs `hs init --c="another_file.config.yaml"` to create a separate config file, the newly created file will copy the contents of the existing config file with the new portal appended. For example, if there is a existing `hubspot.config.yaml` file that contains:

```yaml
portals:
  - name: MyPortal
    portalId: PORTAL_ID
    authType: personalaccesskey
    auth: 
        tokenInfo:
            accessToken: ACCESS_TOKEN
    personalAccessKey: PERSONAL_ACCESS_KEY
```

and the user runs `hs init --c="testing.config.yaml", the newly created `testing.config.yaml` file will be:
```yaml
portals:
  - name: MyPortal
    portalId: PORTAL_ID
    authType: personalaccesskey
    auth: 
        tokenInfo:
            accessToken: ACCESS_TOKEN
    personalAccessKey: PERSONAL_ACCESS_KEY
  - name: MyNewPortal
    portalId: NEW_PORTAL_ID
    authType: personalaccesskey
    auth: 
        tokenInfo:
            accessToken: NEW_ACCESS_TOKEN
    personalAccessKey: NEW_PERSONAL_ACCESS_KEY
```
Which is not the expected result. The upshot of this is that the acceptance-tests for the CLI will fail if there is already an existing config in the file system. 

This PR also updates the acceptance-tests README to make setup of the tests easier.

## Who to Notify
@brandenrodgers @miketalley

### DIFF
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -1,25 +1,19 @@
 # HubSpot CLI Acceptance Tests
 
-  
-
 This project tests `hs.*` commands as if they were being used by an end-user.
 
-  
+
 
 ## Getting Started
 
-  
-
 The main test execution is kicked off by running `yarn test-cli` from the root of `hubspot-cli`. This will run the `run-tests` scipe which itself is a CLI, so to see the available options, you can `run-tests --help`. This was done so that the tests can be installed against any given version of the CLI and tested against it. In addition, it was done to keep the test logic separated from the CLI library.
 
-  
+
 
 Note that if you are testing against a QA portal, not a PROD one, you'll need to add the `--qa` flag when running the script. There is still an outstanding issue with this because we attempt to add the `--qa` flag to all `hs` commands, however it is not available for all commands.
 
 
 ### Setup
-
-
 
 To setup these tests, first [create an app](https://developers.hubspot.com/docs/api/creating-an-app) in a [HubSpot developer account](https://app.hubspot.com/signup-hubspot/crm). Under the "Auth" tab, give your app at least one scope, and set the redirect URL to `localhost:3000/oauth-callback`, and save your changes. Also, take note of your Client ID, and Client Secret and click "Copy full URL".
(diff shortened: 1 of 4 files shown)


miketalley:
I was able to verify that the existing config entries get copied to the new config when using the `--c` flag.

I also tested this fix out with...
- `hs init` - config gets generated correctly with only the new entry
- `hs auth` - new entry gets added and existing entries are maintained in config
- `hs sandboxes create` - sandbox entry is added and existing entries are maintained in config
- `hs config set` - updated values are updated properly and existing entries are maintained

acceptance-tests/README.md
```diff
@@ -2,17 +2,37 @@
...
+
+### Setup
+
+To setup these tests, first [create an app](https://developers.hubspot.com/docs/api/creating-an-app) in a [HubSpot developer account](https://app.hubspot.com/signup-hubspot/crm). Under the "Auth" tab, give your app at least one scope, and set the redirect URL to `localhost:3000/oauth-callback`, and save your changes. Also, take note of your Client ID, and Client Secret and click "Copy full URL".
+![image](https://user-images.githubusercontent.com/48874841/175648758-b2fdb491-90f8-4573-ad9e-af7e1236548c.png)
+
+
+Next, in the navigation bar, click "Testing", and then "Create app test account". Then, paste the OAuth URL you copied earlier into your browser,  choose your newly created test account, and click "Connect App".
+
+Now, navigate back to the "Testing" page, and click your test account. Take note of the Portal ID in the URL (`https://app.hubspot.com/dashboard-library/PORTAL_ID`). Then [navigate to the following page and generate a personal access key](https://app.hubspot.com/portal-recommend/l?slug=personal-access-key) that is associated with your test account.
```
brandenrodgers, line 24: Slight nit, we're moving towards "account" over "portal" throughout the repo (there are still references to portal, but we're at least trying to use account in new code). Can this be updated to "Account ID"?

bmatto (approved):
Thanks!

PR #712, opened 2022-06-28
Add fields.js functionality

j-malt:
# Description and Context
This PR adds a new way for developers to write `fields.json` files using Javascript. Note that all commands in this PR must be run with the `--processFieldsJs` flag.

## Motivation

Currently, developing large or complex modules with the CMS will very quickly lead to large `fields.json` files that become highly repetitive and tedious to maintain.  

This PR solves this by allowing developers to write javascript files in place of json files. These javascript files allow the developer to abstract away repeated field definitions, dynamically generate new ones, and easily maintain existing ones. This gives the developer much more power in creating their `fields.json` file.

## How it works

On their local filesystem, instead of having a `modules/some-module.module/fields.json` file, the developer creates a `modules/some-module.module/fields.js` file. This `fields.js` file must have a function exported via `module.exports`. This function can take in an `options` parameter that will be an array of strings. The function must return an array of Javascript objects. Each object must fall into one of two cases: 
1. The object can be directly `JSON.stringify`'d and added to the `fields.json` or,
2. The object comes with a `.toJSON()` method that will return an object that can be `stringify`'d.

When a developer uses `hs upload --processFieldsJs` on their project and a Javascript fields file is found, the exported function is run, and the returned array is compiled into a `fields.json` that is then uploaded. The JS fields file is not uploaded. 

Each module can have either a `fields.json` or a javascript fields file, but not both. If both are present, the user will be prompted to choose which to use on upload time.

Note that by default we do not save any output locally. This can be enabled with the `--saveOutput` flag. This will save the output as `fields.output.json`, in the same folder as the JS fields file.

## Other Features
- Supports `fields.js`, `fields.cjs`, and `fields.mjs`. Note that `.mjs` files require Node 13.2.0+ to run. 
- Each `fields.js` function gets passed an array of command-line arguments that are passed in through the `--fieldOptions` flag. For instance, `hs upload --src=<src> --dest=<dest> --processFieldsJs --fieldOptions=qa` will pass `['qa']` to the exported function. This can be used to, for instance, toggle field definitions in the fields.json file.
- Handles `async` functions exported from Javascript fields files
- Adds the `hs process --src=<src> --fieldOptions=[...]` command, which will process and save fields files locally 

## Screenshots
Here is a video showing an example of uploading a brand new project to Design Manager that uses a `fields.js` file. At the end it also shows the use of --options=qa flag to toggle field definitions.
https://www.loom.com/share/fb319e2b253a40ef89683b8f9fde0380 

## TODO
- [x] Add tests
- [x] Add compilation to `hs upload` if it is targeted at specific file with --src, rather than a folder.
- [x] Add compilation to `hs watch`.
- [x] Handle async calls being made inside exported function.
- [x] Add flag for gating.
- [x] Add `hs process` command to process a projects fields.js file(s) without uploading.
- [x] Support `--saveOutput` when using `hs watch`. 
- [x] ES Module support
- [x] Better error catching/logging
#### Other features...
Both of these would belong in a separate PR, but would be nice-to-haves in the future:
- Compiled `fields.json` files are uploaded to HubSpot as read-only? 
- It would be nice to upload a "faithful" copy of the developers project (i.e., with uncompiled `fields.js`) that is stored for any `hs fetch` calls. This would keep the CMS as a source of truth for their project.


packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,113 @@
...
+    // Get the path to the folder containing the fields.js file through a stack trace so we can resolve relative paths.
+    const line = new Error().stack.split('\n')[2];
+    const callerPath = line.slice(
+      line.lastIndexOf('(') + 1,
+      line.lastIndexOf('/') + 1
+    );
+    const absoluteSrcPath = path.resolve(getCwd(), callerPath, filePath);
+    return target(absoluteSrcPath, ...args);
+  },
+};
```
j-malt, line 21: I do not like this method of the stack + proxy to resolve relative paths since it feels fragile, but I'm not sure of a better way to do this.

Maybe this isn't the place to have the loadPartials/loadJson methods in the first place, and both should be moved to a (future) library that includes some other utilities for writing fields.js files (such as Field objects).

packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,138 @@
...
+  } catch (e) {
+    handleFieldErrors(e, filePath);
+    throw e;
+  }
+  return finalPath;
+}
+
+function jsonLoader(filePath) {
+  const file = fs.readFileSync(filePath);
+  let json = JSON.parse(file);
```
miketalley, line 95: Do we want to wrap this in a try/catch to handle any JSON parsing errors?

packages/cli-lib/lib/handleFieldsJs.js
j-malt: After some IRL conversation with @miketalley, I've decided that `loadJSON` and `loadPartial` don't really fit in this PR & have removed them.

acceptance-tests/README.md
```diff
@@ -2,17 +2,37 @@
...
 ## Getting Started
 
 The main test execution is kicked off by running `yarn test-cli` from the root of `hubspot-cli`. This will run the `run-tests` scipe which itself is a CLI, so to see the available options, you can `run-tests --help`. This was done so that the tests can be installed against any given version of the CLI and tested against it. In addition, it was done to keep the test logic separated from the CLI library.
 
+
+
 Note that if you are testing against a QA portal, not a PROD one, you'll need to add the `--qa` flag when running the script. There is still an outstanding issue with this because we attempt to add the `--qa` flag to all `hs` commands, however it is not available for all commands.
 
+
+### Setup
```
anthmatic, line 16: 👍

packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,94 @@
...
+  logger.info(
+    i18n(`${i18nKey}.converting`, {
+      src: dirName + '/fields.js',
+      dest: dirName + '/fields.json',
+    })
+  );
+
+  try {
+    // If we do not clear the cache, hs watch will not work
+    delete require.cache[filePath];
```
anthmatic, line 52: What issue were you seeing with `hs watch`?

packages/cli-lib/lib/walk.js
```diff
@@ -7,6 +7,13 @@ const {
   STAT_TYPES,
 } = require('./read');
 
+const listFilesInDir = dir => {
```
brandenrodgers, line 10: Is this similar to what the read func in read.js is doing?

packages/cli-lib/lib/uploadFolder.js
```diff
@@ -28,32 +30,82 @@ const queue = new PQueue({
...
   files.forEach(file => {
     const parts = splitLocalPath(file);
     const extension = getExt(file);
-
     const moduleFolder = parts.find(part => part.endsWith('.module'));
+    const fileName = parts[parts.length - 1];
+    const options = yargs.argv.options;
     if (moduleFolder) {
-      moduleFiles.push(file);
+      //If the folder contains a fields.js, we will always overwrite the existing fields.json.
```
anthmatic, line 50: I think that we should avoid overwriting or deleting any customer code here. I think we could either ignore `fields.json` when `fields.js` is present or prompt the user for the correct file (or maybe to overwrite?)

packages/cli-lib/lib/handleFieldsJs.js
j-malt: Since the fields.js file is called with `require`, node will cache it like a module. So when the fields.js is updated and it's recompiled, when it hits the `require()` line, node returns the cached version of the fields.js file rather than the updated version. See https://nodejs.org/api/modules.html#caching

packages/cli-lib/lib/walk.js
j-malt: It's very similar, but it does it synchronously and just returns file names in the directory. I put it here because `walk()` does a very similar thing to it as well - but could move to read.js if you think it's better suited.

packages/cli-lib/lib/uploadFolder.js
j-malt: Because we write a file called `fields.json` when we compile the `fields.js`, I don't think we could ignore it. Perhaps we could rename it to `fields.original.json` or something. I do like the idea of prompting the user as well.

packages/cli-lib/lib/handleFieldsJs.js
anthmatic: Ahhh, makes sense! It might be worth rewording this comment since it _sounds like_ `hs watch` does work correctly but the cache might trip up `convertFieldsJs`

packages/cli-lib/lib/uploadFolder.js
anthmatic: For some other [build-artifacty commands](https://github.com/HubSpot/hubspot-cli/blob/7de0079d5054b8102a16ffa1698e29e27e0d83bf/packages/cli-lib/archive.js#L28) we've written files to a `tmp` directory and I'm wondering if that's the safer bet for this too. That way we can avoid making any unexpected changes in/around the customer's source files

anthmatic: Looks like a solid plan!

Just in case you haven't seen them, lots of prompt examples here https://github.com/HubSpot/hubspot-cli/tree/817283def17783d01df3b69ce159c87321add4e9/packages/cli/lib/prompts

j-malt: After talking w/ Anthony, decided to push prompting off for now, but have add tmp folder writing, as well as a `--saveOutput` command that will write any built output to a `fields.output.json` file in the same folder as the `fields.js`.

packages/cli/commands/upload.js
```diff
@@ -63,9 +63,15 @@ exports.handler = async options => {
   const uploadPromptAnswers = await uploadPrompt(options);
 
   const src = options.src || uploadPromptAnswers.src;
-  const dest = options.dest || uploadPromptAnswers.dest;
+  let dest = options.dest || uploadPromptAnswers.dest;
```
anthmatic, line 66: Maybe missing it but does `dest` need to be reassigned?

j-malt: Nope. My bad, not sure how that got changed.

packages/cli-lib/lang/en.lyaml
```diff
@@ -678,16 +678,28 @@ en:
           invalidPath: "The path \"{{ path }}\" is not a path to a file or folder"
           uploadFailed: "Uploading file \"{{ src }}\" to \"{{ dest }}\" failed"
           someFilesFailed: "One or more files failed to upload to \"{{ dest }}\" in the Design Manager"
+          fileConversionFailed: "Converting \"{{ src }}\" to \"{{dest }}\" failed"
+          jsonPartiaNotFound: "Partial \"{{ partial }}\" not found in \"{{ src }}\""
```
anthmatic, line 682: Looks like this might not be in use?

packages/cli/commands/upload.js
```diff
@@ -214,5 +279,20 @@ exports.builder = yargs => {
...
+    type: 'array',
+    default: [''],
+  });
+  yargs.option('saveOutput', {
+    describe: i18n(`${i18nKey}.positionals.options.saveOutput`),
+    type: 'boolean',
+    default: true,
+  });
+  yargs.option('processFields', {
+    describe: i18n(`${i18nKey}.positionals.options.processFields`),
```
anthmatic, line 293: Looks like this string might be missing.

On the other hand, do we want to hide this flag? Unsure of rollout plan for this, but happy to chat over Slack

packages/cli/commands/upload.js
```diff
@@ -214,5 +279,20 @@ exports.builder = yargs => {
     describe: i18n(`${i18nKey}.positionals.dest.describe`),
     type: 'string',
   });
+  yargs.option('options', {
+    describe: i18n(`${i18nKey}.positionals.options.describe`),
+    type: 'array',
+    default: [''],
+  });
+  yargs.option('saveOutput', {
+    describe: i18n(`${i18nKey}.positionals.options.saveOutput`),
```
anthmatic, line 288: I think this key needs `.describe`, similar to `processFields`

packages/cli-lib/lib/uploadFolder.js
```diff
@@ -28,6 +31,91 @@ const queue = new PQueue({
...
+  const options = yargs.argv.options;
+
+  files.forEach(file => {
+    const parts = splitLocalPath(file);
+    const extension = getExt(file);
+    const moduleFolder = parts.find(part => part.endsWith('.module'));
+    const fileName = parts[parts.length - 1];
+    const relativePath = file.replace(writeDirRegex, '');
+
+    if (fileName == 'fields.output.json') {
```
anthmatic, line 53: With the move to `tmp` dir, do we need this?

packages/cli-lib/lib/uploadFolder.js
```diff
@@ -28,6 +31,91 @@ const queue = new PQueue({
...
+      return;
+    }
+    if (moduleFolder) {
+      //If the folder contains a fields.js, we will always overwrite the existing fields.json.
```
anthmatic, line 57: ^ This still true?

packages/cli-lib/lang/en.lyaml
j-malt: 👍 Hold over from when there were json & json partial processing functions.

packages/cli/commands/upload.js
j-malt: Whoops, it should have been `positial.processFields.describe`. Fixed. And yeah, I think it's a good idea to hide these for now.

packages/cli-lib/lib/uploadFolder.js
j-malt: Not anymore 👍

packages/cli-lib/lib/uploadFolder.js
j-malt: Yes, if the user has `--saveOutput` (which is enabled by default for now) then we do write the `fields.output.json` to the same folder as the `fields.js`

packages/cli/commands/upload.js
j-malt: Forgot to respond to this before fixing but yeah dest should be getting reassigned - this is to correct for the case when a user uploading a fields.js types the dest file as fields.js instead of .json.

packages/cli-lib/lib/uploadFolder.js
anthmatic: Ahhhh right, sorry I forgot about that.

packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,116 @@
...
+  try {
+    /* Since we require() fields.js files, node will cache them.
+     * Thus during an hs watch, if the user edits a fields.js file, node will return the cached version on require() instead of re-requiring the new one.
+     * So we clean the cache before requiring so that node requires the fresh fields.js. See https://nodejs.org/api/modules.html#caching.
+     */
+    delete require.cache[filePath];
+
+    // Save CWD and then switch CWD to the project. This is so that any calls to the file system that are written relatively will resolve properly.
+    const cwd = getCwd();
+    process.chdir(dirName);
```
anthmatic, line 72: Is the risk here that `filePath` might be relative? I'm wondering if we could use `path.resolve()` where needed to make sure we're working with absolute paths and avoid changing the cwd.

j-malt: The comment is probably a bit unclear, sorry. 

By "any calls to the filesystem" I mean calls that are in the `fields.js`. So in the case when we don't change the cwd, if a dev has `fs.readFileSync('../example.json')` in their `fields.js`, then we execute the readFile and look for the `example.json` file relative to whatever the cwd of the shell is. So for a `fields.js` in a module folder, this wouldn't work unless they cd directly into that folder. 

The alternative solution to this problem was providing file reading proxy functions or doing some call-stack trickery to resolve the path of the module that they are requiring, but both of these seem scarier than just changing the process directory while we load the module.

I have a feeling there is a better way to do this built into node but I was never able to find a good solution besides chdir.

anthmatic: Ah, very good point- I hadn't considered the actual `fields.js` resolution. Seems like a reasonable way to handle this

packages/cli-lib/lib/uploadFolder.js
```diff
@@ -28,43 +29,100 @@ const queue = new PQueue({
...
+  filePaths.forEach(filePath => {
+    const fileType = getFileType(filePath);
+    const fileName = path.basename(filePath);
+    const relativePath = filePath.replace(srcDirRegex, '');
+
+    if (!processFields) {
+      filePathsByType[fileType].push(filePath);
+      return;
+    }
+    if (fileName === 'fields.output.json') return;
```
anthmatic, line 84: I wonder if we could add this to the [CLI ignore rules](https://github.com/HubSpot/hubspot-cli/blob/75800573c73b7c08e16bd9bc132b9d3dd1e6b3e0/packages/cli-lib/ignoreRules.js) to always skip these files?

packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,218 @@
...
+
+  /**
+   * Converts a fields.js file into a fields.json file, writes, and returns of fields.json
+   * @param {string} file - The path of the fields.js javascript file.
+   * @param {string} writeDir - The directory to write the file to.
+   * @returns {Promise} finalPath - Promise that returns path of the written fields.json file.
+   */
+  convertFieldsJs(writeDir) {
+    const filePath = this.filePath;
+    const options = yargs.argv.options;
```
anthmatic, line 38: instead of getting `options` directly from `yargs` can we pass it down from the handler?

packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,218 @@
...
+       * But fieldsArray _might_ not be a Promise. In order to be sure that it is, we use Promise.resolve.
+       */
+      const fieldsArray = require(filePath)(options);
+      return Promise.resolve(fieldsArray).then(fields => {
+        if (!Array.isArray(fields)) {
+          throw new SyntaxError(`${filePath} does not return an array.`);
+        }
+
+        let finalPath = path.join(writeDir, '/fields.json');
+        let json = fieldsArrayToJson(fields);
```
anthmatic, line 70: nit ^ should these be constants?

packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,218 @@
...
+        fs.outputFileSync(finalPath, json);
+        logger.info(
+          i18n(`${i18nKey}.converted`, {
+            src: dirName + '/fields.js',
+            dest: dirName + '/fields.json',
+          })
+        );
+        // Switch back to the original directory.
+        process.chdir(cwd);
```
anthmatic, line 79: If this promise is rejected, do we still need to change the dir back?

packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * Determines if file is a processable fields.js file (i.e., if it is called 'fields.js' and in a root or in a module folder)
+ */
+function isProcessableFieldsJs(rootDir, filePath) {
+  const regex = new RegExp(`^${escapeRegExp(rootDir)}`);
+  const parts = splitLocalPath(filePath);
+  const relativePath = filePath.replace(regex, '');
+  const baseName = path.basename(filePath);
+  const moduleFolder = parts.find(part => part.endsWith('.module'));
```
anthmatic, line 154: Might be useful to use https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/modules.js#L42 here

packages/cli-lib/lib/handleFieldsJs.js
j-malt: Yes. I think I'll rename `options` to `fieldOptions` as well. Current naming is confusing.

packages/cli-lib/lib/handleFieldsJs.js
j-malt: Some slack conversations were had but just to summarize the change I ended up making:

We don't want to kill the process if there is an issue with the `fields.js`. This makes it so that if `hs watch` is running and the user makes a mistake in their fields.js, the process continues. The DevEx is definitely better this way. Further, if it trips up on `hs upload`, the desired behaviour will be to log that an error occurred and continue uploading. This also mimics how `hs watch` currently runs - it will log an error and not upload, but keep watching.

So the way this is done is by adding a `resolved` prop to the FieldsJs object. Then, the process to the FieldsJs we:
1. Make a new FieldsJs object pointing to the correct file path
2. `await` the `.getOutputPathPromise()` method of the FieldsJs object.
3. We check the `resolved` prop. If this promise failed then this will be true. If it's true, don't upload (or in the case of `getFilesByType()`, don't add it to the filePath arrays).
4. If it's false, then set the `outputPath` prop of the fieldsjs object to the return value.

The reason I had to change `getFilesByType()` is so that we can stop errored `fields.js` files before they get added to the arrays, otherwise we'd need to figure out how to remove/skip the rejected Promises during the upload.

packages/cli/commands/upload.js
```diff
@@ -61,11 +67,32 @@ exports.handler = async options => {
...
+  const saveOutput = options.saveOutput;
+  let dest = options.dest || uploadPromptAnswers.dest;
 
   const absoluteSrcPath = path.resolve(getCwd(), src);
+
+  // The theme.json file must always be at the root of the project - so we look for that and determine the root path based on it.
+  const projectRoot = path.dirname(getThemeJSONPath(absoluteSrcPath));
+  const processFieldsJs =
+    isProcessableFieldsJs(projectRoot, absoluteSrcPath) && processFieldsOpt;
+  let fieldsJs;
```
bmatto, line 81: The branching associated with `processFieldsJs` is a somewhat akward. Also I find it strange that were instantiating `FieldsJS` and then mutating the `outputPath` with an object method ... why can't it just do this implicitly?

packages/cli-lib/lib/watch.js
```diff
@@ -35,7 +41,9 @@ const _notifyOfThemePreview = (filePath, accountId) => {
 };
 const notifyOfThemePreview = debounce(_notifyOfThemePreview, 1000);
 
-function uploadFile(accountId, file, dest, options) {
+async function uploadFile(accountId, file, dest, options) {
+  const src = yargs.argv.src;
```
anthmatic, line 45: I think there are still a couple of places where `yargs.argv` could be pulled from `options`

packages/cli-lib/lib/watch.js
```diff
@@ -44,33 +52,54 @@ function uploadFile(accountId, file, dest, options) {
     logger.debug(`Skipping ${file} due to an ignore rule`);
     return;
   }
+  const processFieldsJsOpt = yargs.argv.processFieldsJs;
+  const processFieldsJs =
+    isProcessableFieldsJs(src, file) && processFieldsJsOpt;
```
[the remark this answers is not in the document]
j-malt: Yeah that is a better solution. Will change this.

packages/cli-lib/lib/watch.js
```diff
@@ -42,64 +40,57 @@ const _notifyOfThemePreview = (filePath, accountId) => {
...
+      src: file,
+      dest: path.join(path.dirname(dest), 'fields.json'),
+      mode: options.mode,
+      _: ['upload'],
+    });
 
-    fieldsJs.outputPath = outputPath;
-    dest = path.join(path.dirname(dest), 'fields.json');
+    const fieldsJsProcessCode = `require('@hubspot/cli/commands/upload').handler(${parsedOptions})`;
+    spawn(fieldsJsProcessCode);
```
j-malt, line 64: Just to make it clear what is going on here and why I'm doing things this way...

[So as it stands, ES modules cannot be removed from the module cache](https://github.com/nodejs/modules/issues/307). This is not a big deal for `hs upload` because once the module is loaded in and uploaded, the process dies and the cache is cleared. But for `hs watch`, since the process persists, if we do not clear previously loaded JS field modules out, then old ones will be loaded when a user changes a file. Thus, `fields.js` files will never update while being watched.

So what happens here is that whenever we are running a watch and encounter a processable JS fields file,  we spawn a child node process and call the `hs upload` command. The nice thing here is that since child processes have their own memory, and die immediately after running, we get a clear cache every time. There is a small performance hit with this, but seems negligible. 

There are some other workarounds to this (Transform ESM to CJS with Babel, or [doing something like this](https://github.com/nodejs/modules/issues/307#issuecomment-764560656)), but both have pretty big issues. This solution (courtesy @timmfin) works really well.

packages/cli-lib/errorHandlers.js
```diff
@@ -26,4 +31,6 @@ module.exports = {
   logApiUploadErrorInstance,
   logFileSystemErrorInstance,
   logServerlessFunctionApiErrorInstance,
+  FieldErrors,
```
brandenrodgers, line 34: Nit, can we add something like `module` or `cms` somewhere in the naming of these two utils? The use cases of the CLI are expanding so I'd like to create as much clarity as possible through naming.

packages/cli/lib/prompts/uploadPrompt.js
```diff
@@ -34,6 +37,48 @@ const uploadPrompt = (promptOptions = {}) => {
   ]);
 };
 
+const fieldsJsPrompt = async (filePath, projectDir, skipFiles = []) => {
```
brandenrodgers, line 40: Should this live in its own file?

packages/cli/commands/process.js
```diff
@@ -0,0 +1,84 @@
...
+const { walk } = require('@hubspot/cli-lib/lib/walk');
+const { getThemeJSONPath } = require('@hubspot/cli-lib/lib/files');
+const { i18n } = require('@hubspot/cli-lib/lib/lang');
+const {
+  FieldsJs,
+  isProcessableFieldsJs,
+} = require('@hubspot/cli-lib/lib/handleFieldsJs');
+const i18nKey = 'cli.commands.process';
+
+exports.command = 'process';
```
brandenrodgers, line 15: I think we will want to pull in some design help to figure out where this command should live. We are moving towards creating buckets of commands since the scope of the CLI is increasing beyond the CMS. So there will be a `hs cms` bucket. I'm thinking that maybe we should put this in there.

cc/ @markhazlewood

packages/cli-lib/lib/watch.js
brandenrodgers: It may be worth adding in a short comment to briefly explain some of this since it's not immediately apparent in the code.

packages/cli-lib/lib/uploadFolder.js
```diff
@@ -111,46 +203,55 @@ async function uploadFolder(accountId, src, dest, options) {
     };
   };
 
-  // Implemented using a for lop due to async/await
+  // Implemented using a for loop due to async/await
```
brandenrodgers, line 206: for lop 😆

packages/cli/commands/watch.js
```diff
@@ -21,7 +21,7 @@ const { i18n } = require('@hubspot/cli-lib/lib/lang');
 const i18nKey = 'cli.commands.watch';
 const { EXIT_CODES } = require('../lib/enums/exitCodes');
 
-exports.command = 'watch [--src] [--dest]';
+exports.command = 'watch [--src] [--dest] [--option]';
```
brandenrodgers, line 24: Since option isn't a required option, we shouldn't need to add it here.

packages/cli/commands/upload.js
```diff
@@ -29,13 +28,20 @@ const {
...
 const i18nKey = 'cli.commands.upload';
 const { EXIT_CODES } = require('../lib/enums/exitCodes');
+const {
+  FieldsJs,
+  isProcessableFieldsJs,
+  cleanupTmpDirSync,
+} = require('@hubspot/cli-lib/lib/handleFieldsJs');
 
-exports.command = 'upload [--src] [--dest]';
+exports.command = 'upload [--src] [--dest] [--option]';
```
brandenrodgers, line 44: Same suggestion about mentioning option here. We only need to list options here that are required inputs for the command to successfully run.

miketalley (approved):
This looks good, just want to make sure that the `process` command is hidden from the `--help` menu for now.

PR #713, opened 2022-06-30
Update link to the private app logs

parthd4:
## Description and Context

Currently, when running `hs project logs --follow`, there is a link to view all of the private apps in HubSpot that goes to: `app.hubspot.com/private-apps/{portalId}`

This PR updates the link to go directly to the app's logs: `app.hubspot.com/private-apps/{portalId}/{appId}/logs/extensions` when the `appId` can be found from the `fetchDeployComponentsMetadata` endpoint.

### Before
image

### After
image

## Who to Notify

@brandenrodgers

### DIFF
--- a/packages/cli-lib/api/dfs.js
+++ b/packages/cli-lib/api/dfs.js
@@ -183,19 +183,16 @@ async function fetchProjectSettings(accountId, projectName) {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


brandenrodgers (approved):
Nice 👌

PR #714, opened 2022-07-01
Fixing the error output for project watch command

brandenrodgers:
## Description and Context

Catching the scenario where we attempt to create a staged build, but the project is connected to a github repo. We were unintentionally spitting out the error twice.

#### Before
image (5)

#### After
image


kemmerle (approved):
LGTM!

PR #716, opened 2022-07-12
hs sandbox delete

adamawang:
## Description and Context

Adds functionality to delete sandboxes using the `hs sandbox delete` command. This is the follow up PR to https://github.com/HubSpot/hubspot-cli/pull/706 where sandbox account types and parent portal IDs are saved into the config file. 

- Selecting an account to delete from the picker will pull the account from the config, verify a `parentAccountId` exists, and send delete request
- On successful delete, remove the account entry from the config file
- If the account removed was the default account, prompt user to set a new default
- As a bonus, I added a loading animation to the create sandbox step so users don't think its broken (could take a little while during create).
  - Do we maybe want to add the cli-spinner package to have better progress/loading animations? 

## Screenshots

Loom: https://www.loom.com/share/1338b9d3bcb54b17af6f45aa8caae9d2

Screen Shot 2022-07-12 at 5 51 13 PM

This only shows if the default account was the one that was deleted
Screen Shot 2022-07-12 at 5 52 19 PM

## TODO

There is an interesting edge case/bug here. Since `hs sandbox delete` uses the parent account access token to delete the sandbox, there is a chance a user uses `hs auth` to add a sandbox without adding the parent account. While we do have access to the parent portal, I don't think we have any way to fetch the PAK of the parent portal, nor generate an access token, so the delete will break. Since we know at delete time whether or not the parent is in the config or not, we could print a message saying the parent account must be in the config or something? 

## Who to Notify

@brandenrodgers @markhazlewood @miketalley

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -591,6 +591,8 @@ en:
                   \n- Update the CLI config for this account by running `hs auth` and entering the new key.\n"
           delete:
             describe: "Delete a sandbox account"
+            debug:
+              deleting: "Deleting sandbox account \"{{ account }}\""
             examples:
               default: "Deletes the sandbox account named MySandboxAccount."
             success:
@@ -849,6 +851,8 @@ en:
           enterName: "Enter a name to use for the sandbox: "
           errors:
             invalidName: "You entered an invalid name. Please try again."
+          selectAccountName: "Select the sandbox account you want to delete"
+          selectParentAccountName: "Select the account that the sandbox belongs to"
         uploadPrompt:
           enterDest: "[--dest] Enter the destination path: "
           enterSrc: "[--src] Enter the source path: "
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -602,6 +602,7 @@ en:
               delete: "Sandbox \"{{ account }}\" with portalId \"{{ sandboxHubId }}\" deleted successfully."
               configFileUpdated: "Removed account {{ account }} from {{ configFilename }}."
             objectNotFound: "Sandbox {{#bold}}{{ account }}{{/bold}} not found, it may have already been deleted."
+            noParentPortalAvailable: "This sandbox can't be deleted from the CLI because you haven't given the CLI access to its parent account. To do this, run {{#bold}}{{ command }}{{/bold}}. You can also delete the sandbox from HubSpot management tool: {{#bold}}{{ url }}{{/bold}}."
             options:
               account:
                 describe: "Account name or id to delete"
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -600,11 +600,10 @@ en:
             success:
               delete: "Sandbox \"{{ account }}\" with portalId \"{{ sandboxHubId }}\" deleted successfully."
               configFileUpdated: "Removed account {{ account }} from {{ configFilename }}."
+            objectNotFound: "Sandbox {{#bold}}{{ account }}{{/bold}} not found, it may have already been deleted."
             options:
               account:
                 describe: "Account name or id to delete"
-            temporaryMessage: "The CLI currently does not support deleting sandboxes.
-              \n\nTo delete a sandbox, go to https://app.hubspot.com and navigate to the Sandboxes accounts page to delete your sandbox.\n"
       secrets:
         describe: "Manage HubSpot secrets."
         subcommands:
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -602,7 +602,7 @@ en:
               delete: "Sandbox \"{{ account }}\" with portalId \"{{ sandboxHubId }}\" deleted successfully."
               configFileUpdated: "Removed account {{ account }} from {{ configFilename }}."
             objectNotFound: "Sandbox {{#bold}}{{ account }}{{/bold}} not found, it may have already been deleted."
-            noParentPortalAvailable: "This sandbox can't be deleted from the CLI because you haven't given the CLI access to its parent account. To do this, run {{#bold}}{{ command }}{{/bold}}. You can also delete the sandbox from HubSpot management tool: {{#bold}}{{ url }}{{/bold}}."
+            noParentPortalAvailable: "This sandbox can't be deleted from the CLI because you haven't given the CLI access to its parent account. To do this, run {{#bold}}{{ command }}{{/bold}}. You can also delete the sandbox from the HubSpot management tool: {{#bold}}{{ url }}{{/bold}}."
             options:
               account:
                 describe: "Account name or id to delete"
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -598,8 +598,6 @@ en:
             success:
               delete: "Sandbox \"{{ account }}\" with portalId \"{{ sandboxHubId }}\" deleted successfully."
               configFileUpdated: "Removed account {{ account }} from {{ configFilename }}."
-            failure:
-              delete: "Unable to delete sandbox account {{ account }}."
             options:
               account:
                 describe: "Account name or id to delete"
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -571,8 +571,6 @@ en:
         subcommands:
           create:
             describe: "Create a sandbox account"
-            debug:
-              creating: "Creating sandbox \"{{ name }}\""
             examples:
               default: "Creates a sandbox account named MySandboxAccount."
             info:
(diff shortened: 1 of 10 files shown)


packages/cli-lib/lib/config.js
```diff
@@ -359,6 +359,37 @@ const getAccountId = nameOrId => {
   return null;
 };
 
+/**
+ * @throws {Error}
+ */
+const removeAccountFromConfig = nameOrId => {
```
adamawang, line 365: Added a helper here to remove accounts from the config

packages/cli/commands/accounts/use.js
```diff
@@ -9,30 +9,11 @@ const { loadAndValidateOptions } = require('../../lib/validation');
 
 const { getAccountId } = require('../../lib/commonOpts');
 const { trackCommandUsage } = require('../../lib/usageTracking');
-const { promptUser } = require('../../lib/prompts/promptUtils');
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
+const { selectAccountFromConfig } = require('../../lib/prompts/accountsPrompt');
 
 const i18nKey = 'cli.commands.accounts.subcommands.use';
 
-const selectAccountFromConfig = async config => {
```
adamawang, line 17: Moved this to a new prompt file

packages/cli/commands/sandbox/create.js
```diff
@@ -129,6 +129,14 @@ exports.handler = async options => {
     })
   );
   let result;
+  const twirlTimer = (function() {
```
adamawang, line 132: Super basic, can probably replace this with something from cli-spinners if we want

packages/cli/commands/sandbox/delete.js
```diff
@@ -16,8 +27,60 @@ exports.describe = i18n(`${i18nKey}.describe`);
...
+
+  trackCommandUsage('sandbox-delete', {}, sandboxAccountId);
+
+  let parentAccountId;
+  for (const portal of config.portals) {
+    if (portal.portalId === sandboxAccountId) {
+      if (portal.parentAccountId) {
+        parentAccountId = portal.parentAccountId;
+      } else {
+        const parentAccountPrompt = await deleteSandboxPrompt(config, true);
```
adamawang, line 49: Backwards compatibility for accounts without parentAccountId set, but will break if user selects wrong parent portal

miketalley (commented):
Cool stuff! I think we should definitely use spinnies ([example here](https://github.com/HubSpot/hubspot-cli/blob/75800573c73b7c08e16bd9bc132b9d3dd1e6b3e0/packages/cli/lib/projects.js#L229)) for the loading state.

What would happen if someone wanted to delete a sandbox programmatically? Is it possible to specify the sandbox to delete when running the command? 

What if someone cancels out during the selection of the new default account after deleting a sandbox that is set as default?

packages/cli/commands/sandbox/delete.js
```diff
@@ -16,8 +27,60 @@ exports.describe = i18n(`${i18nKey}.describe`);
...
+  }
+
+  logger.debug(
+    i18n(`${i18nKey}.debug.deleting`, {
+      account: account || accountPrompt.account,
+    })
+  );
+
+  try {
+    await deleteSandbox(parentAccountId, sandboxAccountId);
```
miketalley, line 64: Should we add a confirmation here? The ramifications of fat-fingering the wrong account here seem pretty bad.

packages/cli/commands/sandbox/create.js
miketalley: We should definitely make use of the existing spinnies dependency here.

packages/cli/commands/sandbox/delete.js
adamawang: Ah yes, totally forgot about the confirmation step. Good call out!

packages/cli/lib/validation.js
```diff
@@ -26,14 +26,19 @@ const fs = require('fs');
...
+async function loadAndValidateOptions(options, shouldValidateAccount = true) {
   setLogLevel(options);
   logDebugInfo(options);
   const { config: configPath } = options;
   loadConfig(configPath, options);
   checkAndWarnGitInclusion(getConfigPath());
 
-  if (!(validateConfig() && (await validateAccount(options)))) {
+  let validAccount = true;
+  if (shouldValidateAccount) {
```
adamawang, line 37: Added an optional account validation prop to bypass the `validateAccount()` check if needed. This is because if the default account in the config file does not exist in the portals list (edge case), it logs and error and exits, leaving the user no option to set a new default account using commands. 

Let me know if you'd like to handle this differently, I had extracted it at first but figure adding one prop here would be cleaner

packages/cli/commands/sandbox/delete.js
```diff
@@ -2,22 +2,99 @@ const {
...
+  }
+
+  logger.debug(
+    i18n(`${i18nKey}.debug.deleting`, {
+      account: account || accountPrompt.account,
+    })
+  );
+
+  try {
+    const { confirmSandboxDeletePrompt: confirmed } = await promptUser([
```
adamawang, line 65: Adds confirmation step here:
Screen Shot 2022-07-13 at 1 12 36 PM

packages/cli/commands/sandbox/create.js
```diff
@@ -113,6 +113,10 @@ exports.handler = async options => {
   const accountId = getAccountId(options);
   const accountConfig = getAccountConfig(accountId);
   const env = options.qa ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
+  const spinnies = new Spinnies({
```
adamawang, line 116: Added spinnies:
Screen Shot 2022-07-13 at 1 09 18 PM

packages/cli/commands/accounts/use.js
```diff
@@ -58,6 +35,12 @@ exports.handler = async options => {
     newDefaultAccount = await selectAccountFromConfig(config);
   }
 
+  trackCommandUsage(
+    'accounts-use',
+    {},
+    getAccountIdFromConfig(newDefaultAccount)
```
adamawang, line 41: Updated tracking to use the new default account ID

packages/cli/commands/accounts/use.js
```diff
@@ -5,47 +5,24 @@ const {
...
-
-  return selectedDefault;
-};
-
 exports.command = 'use [--account]';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.handler = async options => {
-  await loadAndValidateOptions({ debug: options.debug });
+  await loadAndValidateOptions(options, false);
```
adamawang, line 20: Updated to use the full options object. Before, it was never receiving the `--account` flag if passed

packages/cli/commands/sandbox/delete.js
```diff
@@ -2,22 +2,122 @@ const {
...
+    const promptDefaultAccount = removeAccountFromConfig(sandboxAccountId);
+    if (promptDefaultAccount) {
+      await selectAndSetAsDefaultAccountPrompt(config);
+    }
+    process.exit(EXIT_CODES.SUCCESS);
+  } catch (err) {
+    if (
+      err.error &&
+      err.error.category === OBJECT_NOT_FOUND &&
+      err.error.subCategory === SANDBOX_NOT_FOUND
```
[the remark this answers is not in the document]
adamawang: This is if a user deletes a sandbox from the UI and it still exists in the CLI config. 

I also bypassed the account validation on delete because I was getting errors that the PAK no longer was valid (account was already deleted), which was correct but that was preventing users from removing accounts.

adamawang:
@miketalley @brandenrodgers @markhazlewood  just checking in on the TODO section above - wondering if you have any ideas on how I should handle that situation? 

Also pushed a number of changes to address questions/comments above!

brandenrodgers:
> While we do have access to the parent portal, I don't think we have any way to fetch the PAK of the parent portal, nor generate an access token, so the delete will break. Since we know at delete time whether or not the parent is in the config or not, we could print a message saying the parent account must be in the config or something?

I think that's reasonable. In the UI you delete the sandbox accounts from within the parent account. So it makes sense to me that a user would be required to have the parent account configured in the CLI in order to delete a sandbox account. I think at delete time we could log something along the lines of:

 "You cannot delete this sandbox from the CLI because you have not given the CLI access to the parent HubSpot account. Run `hs auth --account=<parentId>` if you would like to use the CLI to delete sandboxes. OR, delete the sandbox from HubSpot (and link to the sandboxes app here)"

markhazlewood42:
Some minor copy edits:

>"This sandbox can't be deleted from the CLI because you haven't given the CLI access to its parent account. To do this, run `hs auth --account=<parent account ID>`. You can also delete the sandbox from HubSpot management tool: <link>"

adamawang:
Updated messaging if parent portal ID does not exist in config when attempting delete: 
Screen Shot 2022-07-21 at 6 56 14 PM

Messaging supports QA also

adamawang:
@brandenrodgers @miketalley This should be good to go - have done thorough testing of backwards compatibility if config files do not contain sandbox type or parent account ids, and various points where users can cancel out like the default account setter. Let me know your thoughts on the latest set of commits, specifically on the account validation bypass

packages/cli-lib/lang/en.lyaml
```diff
@@ -602,6 +602,7 @@ en:
               delete: "Sandbox \"{{ account }}\" with portalId \"{{ sandboxHubId }}\" deleted successfully."
               configFileUpdated: "Removed account {{ account }} from {{ configFilename }}."
             objectNotFound: "Sandbox {{#bold}}{{ account }}{{/bold}} not found, it may have already been deleted."
+            noParentPortalAvailable: "This sandbox can't be deleted from the CLI because you haven't given the CLI access to its parent account. To do this, run {{#bold}}{{ command }}{{/bold}}. You can also delete the sandbox from HubSpot management tool: {{#bold}}{{ url }}{{/bold}}."
```
brandenrodgers, line 605: Re: "...from HubSpot management tool:" this is a nit, but should it either be "HubSpot's* management tool" or "the* HubSpot management tool"?

adamawang: Good call out, I'd probably lean more towards `from the HubSpot management tool` but I'll confirm with Mark/Erin 👍

brandenrodgers (approved):
I just had one small comment about copy, but otherwise it looks good! 🙌

PR #720, opened 2022-07-18
Automatically generate a .gitignore file to protect the hubspot config file

brandenrodgers:
## Description and Context

There are a handful of instances where users (presumably unintentionally) committed their hubspot config files into public github. This is an issue because the config files contain access keys to their HubSpot accounts. This PR updates the `hs init` command to automatically create or update a gitignore file.


kemmerle (approved):
LGTM 👍

PR #721, opened 2022-07-21
Add version param to fetch

ronyshkevych:
## Description and Context
When downloading a default asset, we want to allow developers to pass in an `assetVersion` argument like so:

```
hs fetch @hubspot/button.module --assetVersion 1
```

to download a specific version of the default asset. Without this parameter the `hs fetch` command will behave as it does currently, where it simply downloads the version of the asset your portal has

## Screenshots
Screen Shot 2022-07-21 at 4 13 49 PM

## Who to Notify
@brandenrodgers

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -218,6 +218,8 @@ en:
         options:
           staging:
             describe: "Retrieve staged changes for project"
+          assetVersion:
+            describe: "Specify what version of a default asset to fetch"
         positionals:
           dest:
             describe: "Local directory you would like the files to be placed in, relative to your current working directory"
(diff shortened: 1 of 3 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -218,6 +218,8 @@ en:
         options:
           staging:
             describe: "Retrieve staged changes for project"
+          assetVersion:
+            describe: "Specify what version of a default asset to fetch"
```
brandenrodgers, line 222: cc/ @markhazlewood for copy input here.

brandenrodgers:
The code all looks good to me 👍. We should check in with design on copy before merging though

TheWebTech:
@brandenrodgers curious, what happens if someone just fetched `@hubspot/` What version do they get?  

I would assume they'd get the version that's active in the current account but I felt this is worth asking in-case there's recursive issues.

PR #722, opened 2022-08-03
fix logging version number

ronyshkevych:
## Description and Context
When passing in 0 as a version number, JS resolves 0 as falsey and so it doesnt log the version number being specified

## Screenshots
Screen Shot 2022-08-03 at 2 23 52 PM

## TODO
N/A


brandenrodgers (approved):
👍

PR #723, opened 2022-08-03
Add debug logs, change positional to optional

adamawang:
## Description and Context

Adds debug logs for sandbox create and delete commands in failure states. Also changes sandbox create to use optional rather than positional, as per the [CLI Design Principles](https://docs.google.com/document/d/1EbFFvlLfZPseGnjn-IwfDCuy8G5Uzqw-EoGJLwWN6Ug/edit) doc

## Screenshots

Screen Shot 2022-08-03 at 5 49 20 PM

## Who to Notify

@brandenrodgers

### DIFF
--- a/packages/cli/commands/sandbox/create.js
+++ b/packages/cli/commands/sandbox/create.js
@@ -103,7 +103,7 @@ const personalAccessKeyFlow = async (env, account, name) => {
   ]);
 };
 
-exports.command = 'create [name]';
+exports.command = 'create [--name]';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.handler = async options => {
@@ -146,6 +146,9 @@ exports.handler = async options => {
       }),
     });
   } catch (err) {
+    logger.debug(i18n(`${i18nKey}.debug.error`));
+    logger.debug(err.error);
+
     spinnies.fail('sandboxCreate', {
       text: i18n(`${i18nKey}.loading.fail`, {
         sandboxName,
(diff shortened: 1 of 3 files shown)


adamawang:
Let me know if there are any additional debug logs preferred, or any best practices!

packages/cli/commands/sandbox/delete.js
```diff
@@ -117,6 +117,9 @@ exports.handler = async options => {
     }
     process.exit(EXIT_CODES.SUCCESS);
   } catch (err) {
+    logger.debug(i18n(`${i18nKey}.debug.error`));
+    logger.debug(err.error);
```
brandenrodgers, line 121: we do have a `logErrorInstance` helper function. I forget exactly how it works, but it may be worth seeing if it can work here too

adamawang: Good call! Updating those now

packages/cli/commands/sandbox/create.js
```diff
@@ -146,11 +149,14 @@ exports.handler = async options => {
       }),
     });
   } catch (err) {
+    debugErrorAndContext(err);
```
adamawang, line 152: Ended up using `debugErrorAndContext` instead of the full `logErrorInstance` as the latter was providing too much detail into the specific error here. 

For example, the first log is with `logErrorInstance` and the second is maintaining the old log that I have except with the extra `debugErrorAndContext`:

Screen Shot 2022-08-04 at 10 46 02 AM

Screen Shot 2022-08-04 at 10 45 27 AM

brandenrodgers (approved):
👍

PR #724, opened 2022-08-04
Fix for theme fetch in new lighthouse command

brandenrodgers:
## Description and Context

There's an issue where not all of the themes are being fetched correctly. Ideally we add pagination to this fetch so that the user can see all of their available themes, but in the meantime I think a passable solution is to increase the limit and to sort by most used.

## Who to Notify

cc/ @ssethia2

### DIFF
--- a/packages/cli/commands/cms/lighthouseScore.js
+++ b/packages/cli/commands/cms/lighthouseScore.js
@@ -50,7 +50,7 @@ const selectTheme = async accountId => {
       choices: async () => {
         try {
           const result = await fetchThemes(accountId, {
-            limit: 50,
+            limit: 500,
             sorting: 'MOST_USED',
           });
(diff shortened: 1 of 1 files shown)


packages/cli/commands/cms/lighthouseScore.js
```diff
@@ -49,7 +49,10 @@ const selectTheme = async accountId => {
       message: i18n(`${i18nKey}.info.promptMessage`),
       choices: async () => {
         try {
-          const result = await fetchThemes(accountId);
+          const result = await fetchThemes(accountId, {
+            limit: 50,
```
ssethia2, line 53: Default limit is 100 😆 we can just make this 500 for now, and then work the offset logic in later

ssethia2 (approved):

PR #728, opened 2022-08-11
Move tracking to ensure success before emitting event

adamawang:
## Description and Context

Sandboxes has identified some discrepancies when looking at amplitude tracking, and we want to ensure successful sandbox actions before emitting a tracking event. 

This PR addresses that and moves tracking events further down in the flow. 

## TODO

We'll need to have a broader discussion on this with the CLI team, but Sandboxes may want to start tracking sandbox sub types, eg Development sandbox vs Standard sandbox. 

## Who to Notify
@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli/commands/sandbox/create.js
+++ b/packages/cli/commands/sandbox/create.js
@@ -122,8 +122,6 @@ exports.handler = async options => {
 
   let namePrompt;
 
-  trackCommandUsage('sandbox-create', {}, accountId);
-
   if (!name) {
     namePrompt = await createSandboxPrompt();
   }
@@ -140,6 +138,8 @@ exports.handler = async options => {
     });
 
     result = await createSandbox(accountId, sandboxName);
+
+    trackCommandUsage('sandbox-create', {}, accountId);
 
     logger.log('');
     spinnies.succeed('sandboxCreate', {
--- a/packages/cli/commands/sandbox/create.js
+++ b/packages/cli/commands/sandbox/create.js
@@ -120,6 +120,8 @@ exports.handler = async options => {
     succeedColor: 'white',
   });
 
+  trackCommandUsage('sandbox-create', {}, accountId);
+
   let namePrompt;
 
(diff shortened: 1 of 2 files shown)


brandenrodgers:
The current pattern in the CLI is to track all attempted runs of a command, successful or not. It's still good information to know that a user attempted to run a command, but failed to get the result they were looking for. What if we added another event tracking call in the `catch` block, and pass in some additional to both of the calls. Maybe something like `success: true/false`?

adamawang:
Ah I see... yeah I had been following patterns from other files before and that totally makes sense, but we wanted it to match up with what we do on Sandboxes FE which is only track a sandbox create upon success. 

Not sure I follow on the catch block addition, would it be the same tracker name for "sandbox-create" with an extra param? Or a new tracking event?

brandenrodgers:
> Not sure I follow on the catch block addition, would it be the same tracker name for "sandbox-create" with an extra param? Or a new tracking event?

That's up to you. I was picturing using an extra param (i.e. `track('sandbox-create', {success: false}`), but using a new tracking event would work too. I think it's useful CLI tracking data to know every single command that users attempt to run (successful or not), so I'd like to keep that pattern if possible.

adamawang:
Yep I totally agree, and we definitely don't want to stray from established patterns. Will put up a commit soon on this, thank you on the suggestions!

packages/cli/commands/sandbox/create.js
```diff
@@ -151,6 +151,8 @@ exports.handler = async options => {
   } catch (err) {
     debugErrorAndContext(err);
 
+    trackCommandUsage('sandbox-create', { success: false }, accountId);
```
adamawang, line 154: To get to an accurate number of successful creates, we'll just subtract the `success: false` events from the grand total. Same for deletes

brandenrodgers (approved):
Thanks for updating this!

PR #729, opened 2022-08-16
Sandboxes CLI: prevent removing non-sandbox portal from config on attempted delete

shannon-lichtenwalter:
## Description and Context

We introduced the command to allow a user to delete a sandbox from the CLI (hs sandbox delete). 
Following deletion the sandbox is removed from the config regardless of if the delete failed or not.
This PR is to prevent non-sandbbox account types from being removed from the config file on attempted delete

## Screenshots

before:
Screen Shot 2022-08-16 at 3 53 22 PM

after:
Screen Shot 2022-08-16 at 3 52 49 PM

## Who to Notify

@adamawang

### DIFF
--- a/packages/cli-lib/lib/config.js
+++ b/packages/cli-lib/lib/config.js
@@ -389,6 +389,7 @@ const removeSandboxAccountFromConfig = nameOrId => {
   }
 
   writeConfig();
+
   return promptDefaultAccount;
 };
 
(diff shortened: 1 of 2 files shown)


brandenrodgers (approved):
👌

adamawang (approved):
Niiiice, good call on the helper function rename 🚀

PR #730, opened 2022-08-17
Add links to deploys

camden11:
## Description and Context
CLI now logs a link to deploys for any commands that can deploy builds (`watch`, `upload`, `deploy`)

## Screenshots
Before:
Screen Shot 2022-08-17 at 1 05 14 PM

After:
Screen Shot 2022-08-17 at 1 04 02 PM

## TODO
~We may want to update the copy for the link here. I opted not to specify a deploy # because it looks like deploys are only ever referred to by build number in the projects app, but could also tweak the `linkToHubSpot` function a bit so we can show what build the deploy is for~ Going with "deploy of build n"

## Who to Notify
@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -354,7 +354,9 @@ const makePollTaskStatusFunc = ({
     const displayId = deployedBuildId || taskId;
 
     if (linkToHubSpot) {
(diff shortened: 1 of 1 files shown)


brandenrodgers (approved):
Looks great!

PR #731, opened 2022-08-17
Updating options for some project commands

brandenrodgers:
## Description and Context

Making a handful of small changes to the options of some projects-related commands:
- Enumerating the options for `--template` in the create command
- `--limit` now shows as an option in the output of `hs project logs --help`
- Bumping the default limit in the logs command from 5 to 10


camden11 (approved):
LGTM

PR #732, opened 2022-08-22
Prevent creation of /src and hsproject.json for projects that dont exist

camden11:
Currently, `hs project download` still create a config file and empty `src` directory if a user tries to download a project that doesn't exist. This prevents that from happening, and adds an explicit error message.

See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/335

Before:
Screen Shot 2022-08-22 at 3 53 37 PM

After
Screen Shot 2022-08-22 at 3 52 43 PM

cc @kemmerle


kemmerle (approved):
Reviewed the changes and tested this locally, and LGTM 👍

brandenrodgers (approved):

PR #733, opened 2022-08-22
Encode URLs in UI Links

camden11:
In terminals that don't support actual links, we print the URL to projects when they build and deploy (this probably happens in other places too). These urls weren't being encoded properly, so projects with spaces in their names would end up with broken URLs. This fixes that. See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/330

Before:
Screen Shot 2022-08-22 at 4 29 40 PM

After:
Screen Shot 2022-08-22 at 4 29 52 PM

cc @kemmerle

### DIFF
--- a/packages/cli/lib/ui.js
+++ b/packages/cli/lib/ui.js
@@ -37,10 +37,11 @@ const getTerminalUISupport = () => {
  */
 const uiLink = (linkText, url) => {
   const terminalUISupport = getTerminalUISupport();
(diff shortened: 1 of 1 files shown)


kemmerle (approved):
Reviewed code and tested locally, and everything is working well 👍 

Thanks for including screenshots for the commands you used, very helpful!

brandenrodgers (approved):

PR #734, opened 2022-08-23
Fix check-deps issue

j-malt:
## Description and Context
There were quite a few dependency errors in #712 that I missed. This fixes them.

cc/ @miketalley @anthmatic

### DIFF
--- a/packages/cli-lib/lib/__tests__/uploadFolder.js
+++ b/packages/cli-lib/lib/__tests__/uploadFolder.js
@@ -18,10 +18,6 @@ jest.mock('../../ignoreRules');
 jest.mock('../handleFieldsJs');
 jest.mock('../../errorHandlers');
 
-/* jest
-  .spyOn(cmsFieldPrompt, 'fieldsJsPrompt')
-  .mockImplementation(filePath => [filePath, []]); */
-
 //folder/fields.js -> folder/fields.converted.js
 // We add the .converted to differentiate from a unconverted fields.json
 const defaultFieldsJsImplementation = jest.fn((src, filePath, rootWriteDir) => {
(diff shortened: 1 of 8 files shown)


packages/cli/commands/upload.js
```diff
@@ -237,6 +247,46 @@ exports.handler = async options => {
   }
 };
 
+/*
+ * Walks the src folder for files, filters them based on ignore filter.
+ * If processFieldsJs is true then will check for any JS fields conflicts (i.e., JS fields file and fields.json file) and prompt to resolve
+ */
+const getUploadableFileList = async (src, processFieldsJs) => {
```
j-malt, line 254: This was done so that we do not need to require the `cmsFieldPrompt` inside of `cli-lib`. This code was previously spread across `uploadFolder()` and `getFilesByType()` inside of `cli-lib/lib/uploadFolder.js`.

This is a *bit* slower now, because if `processFieldsJs` is true, we end up looping over the `filePaths` twice (once in this function and once in `getFilesByType()`) but it's very minor all things considered.

miketalley: What is the reason that we need to loop over things twice? Is it possible to do it once and pass that data to both places that need it?

packages/cli-lib/lib/__tests__/uploadFolder.js
```diff
@@ -11,18 +11,16 @@ const {
...
 
 jest.mock('../walk');
 jest.mock('../../api/fileMapper');
 jest.mock('../../ignoreRules');
 jest.mock('../handleFieldsJs');
 jest.mock('../../errorHandlers');
-jest.mock('@hubspot/cli/lib/prompts/cmsFieldPrompt');
 
-jest
+/* jest
```
miketalley, line 21: Should we remove this?

j-malt: Oops

packages/cli/commands/upload.js
j-malt: It is possible: we could just pull the entire `getFilesByType` function into this file and pass back the tuple to uploadFolder. I don't have a strong opinion one way or the other, so I will do it that way instead.

miketalley (approved):
Looks good, I think the tradeoff here of keeping `cli-lib` lean and having fewer dependencies is worth the small performance hit.

PR #738, opened 2022-08-24
JS Fields: Resolve promises correctly & Add tracking

j-malt:
## Description and Context
This PR fixes three things:
- When a Fields array is returned from a JS fields function, it wraps it in a `Promise.all()` and waits, so that each Promise will resolve before proceeding. Previously, this would not happen, making any Promises in the returned Array be written as a blank object.
- Makes the `process` command work with relative paths.
- Adds some tracking to the processing
- I forgot to require `logger` in `dynamicImport` in #712 

Checks:
For `node -v = v10.24.1`:
- Passes `yarn test`
- Passes `yarn test-cli`
- Passes `yarn check-deps`

cc/ @miketalley

### DIFF
--- a/packages/cli-lib/lib/handleFieldsJs.js
+++ b/packages/cli-lib/lib/handleFieldsJs.js
@@ -41,7 +41,7 @@ class FieldsJs {
    * @param {string} writeDir - The directory to write the file to.
    * @returns {Promise} finalPath - Promise that returns path of the written fields.json file.
    */
-  convertFieldsJs(writeDir) {
+  async convertFieldsJs(writeDir) {
     const filePath = this.filePath;
     const baseName = path.basename(filePath);
     const dirName = path.dirname(filePath);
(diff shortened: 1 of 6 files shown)


packages/cli-lib/lib/handleFieldsJs.js
```diff
@@ -172,8 +176,9 @@ function flattenArray(arr) {
 }
 
 //Transform fields array to JSON
-function fieldsArrayToJson(fields) {
-  fields = flattenArray(fields).map(field => {
+async function fieldsArrayToJson(fields) {
+  fields = await Promise.all(flattenArray(fields));
```
brandenrodgers, line 180: Does this require a promise? `flattenArray` looks like it's synchronous

j-malt: `flattenArray` itself is synchronous, but the array it returns can have a mix of objects and Promises. The `await Promise.all` lets us wait for every Promise in the array to resolve before moving on.

brandenrodgers: Ahh I see what you mean. Makes sense 👍

brandenrodgers (approved):
lgtm! 👍

PR #739, opened 2022-08-26
[BUGFIX] send the entire input to the getAssetVersionIdentifier function instead of just an options

dmh:
## Description and Context

It seems that there shouldn't be an `src` key in the `options` object, and without `src` the code below will throw an error.
```
function getAssetVersionIdentifier(options) {
  if (options.assetVersion !== null && options.src.startsWith('@hubspot/')) {
    return ' v' + options.assetVersion;
  }
  return '';
}
```
by default `options.src = undefined`

`startsWith` will not work with `undefined` properties and throw a js `TypeError`

***

The solution is to send the entire `input` to the `getAssetVersionIdentifier` function instead of just an `options`
```
 function getAssetVersionIdentifier(input) {
  if (input.options.assetVersion !== null && input.src.startsWith('@hubspot/')) {
    return ' v' + input.options.assetVersion;
  }
  return '';
}
```

### DIFF
--- a/packages/cli-lib/fileMapper.js
+++ b/packages/cli-lib/fileMapper.js
@@ -114,7 +114,7 @@ function getFileMapperQueryValues({ mode, options = {} }) {
  * @returns {string}
  */
  function getAssetVersionIdentifier(input) {
-  if (input.options.assetVersion !== null && input.src.startsWith('@hubspot/')) {
+  if (input.options.hasOwnProperty('assetVersion') && input.hasOwnProperty('src') && input.src.startsWith('@hubspot/')) {
     return ' v' + input.options.assetVersion;
   }
   return '';
--- a/packages/cli-lib/fileMapper.js
+++ b/packages/cli-lib/fileMapper.js
@@ -108,14 +108,14 @@ function getFileMapperQueryValues({ mode, options = {} }) {
 }
 
 /**
- * Determines version number to log based on options
+ * Determines version number to log based on input.options
  *
- * @property {Object} options
+ * @property {Object} input
  * @returns {string}
  */
-function getAssetVersionIdentifier(options) {
-  if (options.assetVersion !== null && options.src.startsWith('@hubspot/')) {
-    return ' v' + options.assetVersion;
+ function getAssetVersionIdentifier(input) {
+  if (input.options.assetVersion !== null && input.src.startsWith('@hubspot/')) {
+    return ' v' + input.options.assetVersion;
   }
   return '';
 }
@@ -393,7 +393,7 @@ async function downloadFile(input) {
     logger.success(
       'Completed fetch of file "%s"%s to "%s" from the Design Manager',
       input.src,
-      getAssetVersionIdentifier(input.options),
+      getAssetVersionIdentifier(input),
       input.dest
     );
   } catch (err) {
@@ -492,7 +492,7 @@ async function downloadFolder(input) {
       logger.success(
         'Completed fetch of folder "%s"%s to "%s" from the Design Manager',
         input.src,
-        getAssetVersionIdentifier(input.options),
+        getAssetVersionIdentifier(input),
         input.dest
       );
     } else {
--- a/packages/cli-lib/fileMapper.js
+++ b/packages/cli-lib/fileMapper.js
@@ -113,7 +113,7 @@ function getFileMapperQueryValues({ mode, options = {} }) {
  * @property {Object} input
  * @returns {string}
  */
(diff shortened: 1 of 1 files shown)


packages/cli-lib/fileMapper.js
```diff
@@ -108,14 +108,14 @@ function getFileMapperQueryValues({ mode, options = {} }) {
...
  *
- * @property {Object} options
+ * @property {Object} input
  * @returns {string}
  */
-function getAssetVersionIdentifier(options) {
-  if (options.assetVersion !== null && options.src.startsWith('@hubspot/')) {
-    return ' v' + options.assetVersion;
+ function getAssetVersionIdentifier(input) {
+  if (input.options.assetVersion !== null && input.src.startsWith('@hubspot/')) {
```
brandenrodgers, line 117: Thanks for finding and fixing this bug! I think `input.src.startsWith('@hubspot/')` should actually be `input.options.src.startsWith('@hubspot/')` since we're now passing in input rather than options.

Another solution here could be to add an additional check to the if statement to make sure that `options.src` exists before attempting to call startsWith on it (similar to how we're doing the null check on assetVersion).

dmh: @brandenrodgers 
Hello,
Thanks for the quick response and sorry to keep bothering you, hope it's not a problem. I am heavily using `cli-lib` for some of my automatizations, so would like to keep helping you, but at the same time bother with questions :)

I have a couple of questions before I change my PR. 

1. My initial confusion was with the `src` inside the `options` object. I thought `src` and `dist` is a high-level attributes and `options` are for parts like `option.overwrite` or the new one `option.assetVersion`. For now, the high-level src is still working, so I am a little confused about why we need both, maybe for future change...?

Example of two `src`:
`downloadFileOrFolder({ accountId, src(1), dest, mode, options.src(2) })`

https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/fileMapper.js#L188-L195

2. Looks like this part is also not a fully correct `options.assetVersion !== null`
The condition will be met if I, for example, omit assetVersion altogether, it will return "vundefined". So we probably need to check for undefined here as well.

***

Solution variants:

If we still need both `src` and `option.src`, I would suggest using this solution
1.
```
function getAssetVersionIdentifier(options) {
    if (options.hasOwnProperty('assetVersion') && options.hasOwnProperty('src')) {
        if (typeof options.assetVersion === 'string' && typeof options.src === 'string' && options.src.startsWith('@hubspot/')) {
            return ' v' + options.assetVersion;
        }
    }
    return ''
}
```

or if one high-level `src`
2.
```
function getAssetVersionIdentifier(input) {
    if (input.options.hasOwnProperty('assetVersion') && input.hasOwnProperty('src')) {
        if (typeof input.options.assetVersion === 'string' && typeof input.src === 'string' && input.src.startsWith('@hubspot/')) {
            return ' v' + options.assetVersion;
        }
    }
    return ''
}
```

***

please tell me if I am on the right track to solve this problem properly

Thanks

brandenrodgers: Ah @dmh thanks for clarifying! I should have looked closer at this before commenting. You're not bothering at all btw! It's exciting to see some community involvement. Sorry for the late response here. We've been busy with INBOUND.

> Looks like this part is also not a fully correct options.assetVersion !== null
The condition will be met if I, for example, omit assetVersion altogether, it will return "vundefined". So we probably need to check for undefined here as well.

That's a good find! Definitely a bug. 👍 

> My initial confusion was with the src inside the options object.

Yeah I agree that it's odd that we're passing in both `src` and also `options.src`. For some context, even though `src` and `dest` are both [positional args](https://yargs.js.org/docs/#api-reference-positionalkey-opt), they still exist on the options object because of how yargs works ([CLI command example](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/fetch.js#L25)). I'm not entirely sure why the initial decision to pass both `src` + `dest` and also `options.src` + `options.dest` was made, but my guess is that it was to allow us to modify them before passing them to the `downloadFileOrFolder` util [like we're doing here](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/fetch.js#L47). I think you're right that we should be prioritizing `input.src` over `input.options.src`. I like your # 2 solution, although I think we can do without both of the string checks in there. IMO the hasOwnProperty checks in combination with the startsWith check should be plenty.

Let me know if any of this doesn't make sense. Thanks again for helping out!

PR #742, opened 2022-09-07
Add sandbox type to project tracking commands

adamawang:
## Description and Context

Sandboxes would like to start tracking which type of sandbox users are using while uploading or deploying projects. This PR adds an extra `sandboxType` meta property to the tracking events for both `projects upload` and `projects deploy`

## Screenshots

Screen Shot 2022-09-07 at 15 11 18
Screen Shot 2022-09-07 at 15 11 24

## TODO

Sandboxes may identify more areas to add a sandbox type property, but these are all we have for now. 

## Who to Notify

@brandenrodgers @kemmerle


brandenrodgers (approved):
🚢

PR #743, opened 2022-09-08
Cleaning usage tracking metadata

brandenrodgers:
## Description and Context

Our usage tracking proxies through one of our BE services. That service has a set of hardcoded usage tracking fields that we allow. Any fields that aren't in the list will get filtered out. I'm cleaning up all of those fields that aren't ever making it into our usage tracking tool.

As a follow up to this, we will need to make some updates to the BE service:
- Support `sandboxType`
- Support `success`
- Remove `latest` because it's not used anymore. IMO it's not necessary to track this one

## Who to Notify

cc/ @adamawang

### DIFF
--- a/packages/cli-lib/fileMapper.js
+++ b/packages/cli-lib/fileMapper.js
@@ -114,7 +114,11 @@ function getFileMapperQueryValues({ mode, options = {} }) {
  * @returns {string}
  */
 function getAssetVersionIdentifier(input) {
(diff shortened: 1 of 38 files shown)


adamawang:
Should we remove the `sandboxType` params I added to project deploy and upload too?

adamawang:
I removed the places we're using any sandboxes meta as well, and will include the props we want to request with the larger projects request

PR #747, opened 2022-09-23
Hs feedback command

camden11:
This adds the feedback command + reminders for users to leave feedback.

Screen Shot 2022-09-26 at 3 25 12 PM

Screen Shot 2022-09-26 at 3 43 34 PM

### Feedback command
* `hs feedback`
* Asks user what type of feedback they want to leave, then prompts them to open the respective feedback page in the browser
* use `--general` or `--bug` to bypass prompts and open the browser directly

### Feedback reminders
* `hs project upload` - Displays every 5 builds. For this first pass, we're not tracking when users respond to the form, so we might want to make this less frequent
* `hs project create` - Added to the list of suggestions under "What's next?"

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -518,6 +518,8 @@ en:
               buildSucceededAutomaticallyDeploying: "Build #{{ buildId }} succeeded. {{#bold}}Automatically deploying{{/bold}} to {{ accountIdentifier }}\n"
               readyToGoLive: "🚀 Ready to take your project live?"
               runCommand: "Run `{{ command }}`"
+              feedbackHeader: "We'd love to hear your feedback!"
+              feedbackMessage: "How are you liking the new projects and developer tools? \n > Run `{{#yellow}}hs feedback{{/yellow}}` to let us know what you think!\n"
             options:
               forceCreate:
                 describe: "Automatically create project if it does not exist"
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -592,6 +592,7 @@ en:
             success: "Successfully opened \"{{ projectName }}\""
           feedback:
             describe: "Leave your feedback on the CLI or file a bug report"
+            success: "We opened {{ url }} in your browser."
       remove:
         describe: "Delete a file or folder from HubSpot."
         deleted: "Deleted \"{{ path }}\" from account {{ accountId }}"
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -593,6 +593,12 @@ en:
           feedback:
             describe: "Leave feedback on HubSpot projects or file a bug report"
             success: "We opened {{ url }} in your browser."
+            options:
+              bug:
+                describe: "Open Github issues in your browser to report a bug."
+              general:
+                describe: "Open the projects feedback form in your browser."
+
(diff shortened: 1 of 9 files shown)


packages/cli/commands/feedback.js
```diff
@@ -0,0 +1,34 @@
...
+exports.describe = i18n(`${i18nKey}.describe`);
+
+exports.handler = async () => {
+  const { type } = await feedbackTypePrompt();
+
+  const promptForType =
+    type === FEEDBACK_OPTIONS.BUG ? bugPrompt : generalPrompt;
+  const { shouldOpen } = await promptForType();
+
+  if (shouldOpen === 'Y') {
```
brandenrodgers, line 28: Rather than check for "Y" or "N", I believe you can make the shouldOpen prompts `type: confirm`, which should return a boolean. ([docs](https://www.npmjs.com/package/inquirer))

packages/cli/commands/feedback.js
```diff
@@ -0,0 +1,34 @@
...
+
+const i18nKey = 'cli.commands.project.subcommands.feedback';
+
+exports.command = 'feedback';
+exports.describe = i18n(`${i18nKey}.describe`);
+
+exports.handler = async () => {
+  const { type } = await feedbackTypePrompt();
+
+  const promptForType =
```
brandenrodgers, line 24: I think you can actually handle this dynamic "shouldOpen" prompt all within the initial `feedbackTypePrompt()` prompt. I linked the docs below, but check out `message`. It supports a function value where you can dynamically choose what to show.

packages/cli/commands/feedback.js
```diff
@@ -0,0 +1,34 @@
...
+  bugPrompt,
+  generalPrompt,
+} = require('../lib/prompts/feedbackPrompt');
+
+const i18nKey = 'cli.commands.project.subcommands.feedback';
+
+exports.command = 'feedback';
+exports.describe = i18n(`${i18nKey}.describe`);
+
+exports.handler = async () => {
```
brandenrodgers, line 21: One thing we could do is additionally support flags in this command (may be worth double checking this w/ design though).

It could work something like this:
`hs feedback --bug` automatically opens our GH issues repo
`hs feedback` does what you currently have coded up (prompts for which type)

packages/cli-lib/lang/en.lyaml
```diff
@@ -923,6 +937,13 @@ en:
           errors:
             invalidName: "You entered an invalid name. Please try again."
             projectDoesNotExist: "Project \"{{ projectName }}\" could not be found in \"{{ accountId }}\""
+        feedbackPrompt:
+          feedbackType:
+            message: "What type of feedback would you like to leave?"
+            bug: "[--bug] Report a bug"
+            general: "[--general] Tell us about your experience with projects"
```
brandenrodgers, line 944: Now that this is more of a general feedback command, should we still explicitly mention that the feedback should be for projects here?

camden11: Good question - this copy is from @markhazlewood so I'll double check

markhazlewood42:
Given that we can't track more specific engagement, can we change this to every 10 builds?

Also, how will this currently work when `watch` is being run? Will it just trigger on the next manual `upload`?

camden11:
@markhazlewood42 @brandenrodgers Changed interval to 10 + added the message to the watch command! Will show up within watch, you don't need to run a manual upload

PR #748, opened 2022-09-29
Add additional info to project locked error message

camden11:
This adds additional debugging info to the error that gets logged when you run upload while a watch is in progress.

image


brandenrodgers (approved):
👌

PR #753, opened 2022-10-13
Add reupload reminder to hs secrets update command

camden11:
## Description and Context
When updating secrets via the CLI, it is not obvious that you have to rebuild and deploy your projects/themes etc to get them to use the secret's new value. This adds a reminder. Final copy will likely need some work - we need something that applies to both projects and the CMS side of things

## Screenshots
Screen Shot 2022-10-13 at 2 17 21 PM

## TODO
Finalize copy

## Who to Notify
@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -686,6 +686,7 @@ en:
                 describe: "Name of the secret to be updated"
             success:
               update: "The secret \"{{ secretName }}\" was updated in the HubSpot account: {{ accountId }}"
+              updateReupload: "Make sure to `{{#yellow}}upload{{/yellow}}` all components using this secret to access its new value."
       theme:
         describe: "Commands for working with themes, including marketplace validation with the marketplace-validate subcommand."
         subcommands:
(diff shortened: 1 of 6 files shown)


packages/cli-lib/lang/en.lyaml
```diff
@@ -686,6 +686,7 @@ en:
                 describe: "Name of the secret to be updated"
             success:
               update: "The secret \"{{ secretName }}\" was updated in the HubSpot account: {{ accountId }}"
```
markhazlewood42, line 688: We should be using the account name set during CLI auth rather than its numeric ID whenever possible.

brandenrodgers: Maybe instead of "component" here we should say "serverless functions"? Component is a term that's specific to projects.

camden11: Ahh gotcha yeah, i can change that

camden11: @markhazlewood42 Looks like we're using account ID in a lot of places, probably would make sense to do that separately

camden11:
Screen Shot 2022-10-13 at 3 56 24 PM
Updated copy + added account name to all `secrets` commands

kemmerle (approved):
LGTM!

PR #755, opened 2022-10-14
Fix bug with subdeploy statuses 

kemmerle:
## Description and Context

@m-roll discovered a bug when attempting to deploy with the new subdeploys: a status was returning as "undefined" during the deploy. We discussed this via Slack. I did some digging and discovered that we were missing a constant in `PROJECT_DEPLOY_TEXT`.  

To test out this change, you'll need to ungate your portal to `Developers:DependentComponents`. 

## Screenshots

Before the change (with bug):

Screen Shot 2022-10-12 at 12 44 17 PM

After change:

Screen Shot 2022-10-14 at 1 02 37 PM

## Who to Notify

@brandenrodgers @m-roll


brandenrodgers (approved):
Thanks for fixing this!

camden11 (approved):

PR #757, opened 2022-10-17
Added access token authentication

dpritos:
## Description and Context
Since usage of `ApiKey` will be soon deprecated, I've added a custom auth type for using Access Keys from Private Apps. This is necessary to still use the hubspot cli on any CI/CD pipelines. Sample configuration would be:
```
portals:
  - name: ACCESSTOKEN
    portalId: 4444
    authType: accesstoken 
    accessToken: fakeAccessToken  

```

## TODO

- Update `init` command to fully support and configure `accesstoken` auth type
- Update necessary documentations


dpritos:
Completely understood @brandenrodgers . So for now I think we will find a work around to still use the CLI in our CI/CD pipeline with Private App access token. I'm thinking of extending the library and do customizations or directly calling the API at least for `hs upload` command. Please let me know if you have other better idea. Thank you.

brandenrodgers:
Thanks for understanding. 

> So for now I think we will find a work around to still use the CLI in our CI/CD pipeline with Private App access token. I'm thinking of extending the library and do customizations or directly calling the API at least for hs upload command.

That sounds like a very solid plan 👍. Hopefully you won't have to use this workaround for too long!

adamstoffel:
+1 I would like to voice my support for this PR. We are currently using one of our developer's private access keys as a deployment credential in our CI/CD pipeline but this is obviously not ideal. We would very much like to migrate to using a private app token in our deploy pipelines.

loganknecht:
Hello

@brandenrodgers  - It looks as though it's been about a year later for this PR
Has there been any update on changes for the `hubspot cli` to support Private App tokens?

@dpritos - What did you end up doing to support your CI/CD using the `hubspot command line tool`?

I am currently trying to identify a solution for our CI/CD to work with `hs upload` and cannot seem to get this working with the CLI and a private app. Are there any solutions or paths forward that are advocated for regarding this authentication approach?

dpritos:
@loganknecht In our CI/CD we overwrite the files with this changes in the PR after running install in the pipeline before running the upload command i.e. `HUBSPOT_PORTAL_ID=$HUBSPOT_PORTAL_ID HUBSPOT_ACCESS_TOKEN=$HUBSPOT_ACCESS_TOKEN ./node_modules/.bin/hs upload --use-env templates/test test`

loganknecht:
@dpritos - Thank you so much for providing that insight.

I don't appear to be able to get this working?

My steps are as follows
1. Generate a `Private App Token`
    - https://developers.hubspot.com/docs/api/private-apps
1. Install Hubspot CLI
    - `npm install -g @hubspot/cli`
1. Set Portal ID
    - `export HUBSPOT_PORTAL_ID="REDACTED"`
1. Set `Private App Token` as `HUBSPOT_ACCESS_TOKEN`
    - `export HUBSPOT_ACCESS_TOKEN="PRIVATE_APP_TOKEN"
1. Perform Upload
    -  `hs upload --use-env build REDACTED`

However, when I do this I receive
```
[ERROR] Error while retrieving new access token: Failed to sign refresh token because refresh token was malformed and could not be decoded. Generate a new refresh token and try again.
```

Are you using a `Private App Token` to authenticate? If not, how are you achieving your authentication solution?

brandenrodgers:
Hey all, apologies for the late response here. We haven't been able to give this the attention it deserves. Supporting private app access tokens as a valid CLI auth method is something that's still on our radar, but we're not ready to commit to supporting it quite yet. There are in-app flows and mental models that we'd like to work through before exposing it to more users.

In the meantime, if using private app access token auth is important to you I think following @dpritos advice is your best bet for now. Keep in mind that you'll need to use this branched version of the CLI for it to work though. The published version of the CLI won't support `HUBSPOT_ACCESS_TOKEN`.

PR #758, opened 2022-10-20
Refactor theme validation

kemmerle:
## Description and Context
We are refactoring theme validation to make use of new endpoints rather than built-in JavaScript validators. 

For Jeff: 

With the old `hs theme marketplace-validate <src>` command, `<src>` referred to a path to a theme on the customer's local machine. With the new command, which has an identical format, `<src>` refers to the path to a theme in the Design Manager. Customers will now need to upload their theme to Design Manager in order to validate it with this command.  

Final edit (for now): I'm experiencing some intermittent issues with the API requests timing out. @ssethia2 is going to work to address this by early next week. 

Testing steps:

1) Create a theme in the Design Manager, or use the `hs create website-theme` to create a theme locally and the `hs upload` command to upload it to Design Manager.

2) Pull down this branch, and run the `yarn hs theme marketplace-validate website-theme`, where `website-theme` is the path to your theme in Design Manager. 

3) The easiest way to purposefully fail a test is to delete the `theme.json` file from a theme locally and then upload it to Design Manager. You could also just validate a random folder. 

## Screenshots
Preliminary appearance of failing and passing validations:

Screen Shot 2022-10-20 at 3 42 33 PM

## TODO

- [x] Delete old validators from repo
- [x] Match module validation design more closely when displaying results

## Who to Notify
@brandenrodgers @camden11, this is ready for a code review.


packages/cli-lib/api/marketplaceValidation.js
```diff
@@ -0,0 +1,45 @@
+const http = require('../http');
+
+const VALIDATION_API_BASE = 'quality-engine/v1/validation';
+
+/**
+ * @async
+ * @param {number} accountId
+ * @returns {Promise}
+ */
+async function requestValidation(accountId, body = {}) {
```
bmatto, line 10: Do these need to be async if there is no usage of await within them?

packages/cli-lib/api/marketplaceValidation.js
```diff
@@ -0,0 +1,45 @@
+const http = require('../http');
+
+const VALIDATION_API_BASE = 'quality-engine/v1/validation';
```
bmatto, line 3: OOC are these validation endpoints specific to themes?

kemmerle: Yes, I believe so. @ssethia2, could you confirm?

packages/cli-lib/api/marketplaceValidation.js
kemmerle: Let me look that up, good question.

packages/cli-lib/api/marketplaceValidation.js
kemmerle: They are specific to themes, confirmed today.

bmatto: Should that be reflected in the naming - these read like generic marketplace validation endpoints.

packages/cli/commands/theme/marketplace-validate.js
```diff
@@ -35,67 +31,99 @@ exports.handler = async options => {
...
+    const checkValidationStatus = async () => {
+      let validationStatus;
+      if (requestResult) {
+        validationStatus = await getValidationStatus(accountId, {
+          validationId: requestResult,
+        });
+      }
+
+      if (validationStatus && validationStatus === 'REQUESTED') {
+        await new Promise(resolve => setTimeout(resolve, 2000));
```
camden11, line 70: `2000` would be good as a constant at the top of the file, like `SLEEP_TIME` or something

camden11 (commented):
Code-wise looks good, just one nit. And when I ran it everything worked as expected! Saw @brandenrodgers's slack comment about the `src` error message, def agree that we should fix that

packages/cli-lib/api/marketplaceValidation.js
camden11: you can remove async here!

packages/cli/commands/theme/marketplace-validate.js
```diff
@@ -35,67 +32,116 @@ exports.handler = async options => {
...
+  }
+
+  if (validationResults.errors.length) {
+    const { errors } = validationResults;
+
+    errors.forEach(err => {
+      logger.error(`${err.context}`);
+
+      if (err.failureReasonType === 'DOWNLOAD_EMPTY') {
+        logger.log(
```
brandenrodgers, line 100: I think this extra log is causing us to log a duplicate error in the case where the theme is not found. Example:
image

brandenrodgers: Maybe just rendering the top "No asset exists at path..." log is enough?

kemmerle: I can remove that now 👍

brandenrodgers (approved):
LGTM! 👌

PR #759, opened 2022-10-21
Handle errors from disconnecting CLI from UI

camden11:
## Description and Context
This is the CLI side of this PR https://git.hubteam.com/HubSpot/developer-projects-ui/pull/341
Modifies watch command error handling so CLI processes cancelled via the UI don't give users error messages, since its behavior we support.

This occurs when
* User exits watch from the UI, then modifies a file
* User exits watch from the UI, then kills CLI

## Screenshots
Modifying a file - Before (doesn't exit)
Screen Shot 2022-10-25 at 12 10 51 PM

Modifying a file - After (exits)
Screen Shot 2022-10-25 at 12 10 11 PM


markhazlewood42:
🚢

PR #762, opened 2022-11-01
Add warnings to theme validation

kemmerle:
## Description and Context
@ssethia2 discovered this morning that we weren't displaying warnings during theme validation or file names/line numbers. This PR adds both. 

## Screenshots

Current validation results:

Screen Shot 2022-11-01 at 12 57 34 PM

Validation results with changes: 

Screen Shot 2022-11-01 at 12 52 52 PM

## Who to Notify

@ssethia2 @brandenrodgers @camden11

### DIFF
--- a/packages/cli/commands/theme/marketplace-validate.js
+++ b/packages/cli/commands/theme/marketplace-validate.js
@@ -100,6 +100,21 @@ exports.handler = async options => {
   }
 
   const displayResults = checks => {
+    const displayFileInfo = (file, line) => {
+      file &&
+        logger.log(
+          i18n(`${i18nKey}.results.warnings.file`, {
+            file,
+          })
+        );
+      line &&
+        logger.log(
+          i18n(`${i18nKey}.results.warnings.lineNumber`, {
+            line,
+          })
+        );
+    };
+
     if (checks) {
       const { status, results } = checks;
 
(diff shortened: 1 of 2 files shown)


packages/cli/commands/theme/marketplace-validate.js
```diff
@@ -100,20 +100,54 @@ exports.handler = async options => {
   }
 
   const displayResults = checks => {
+    const displayFileInfo = (file, line) => {
+      file &&
```
brandenrodgers, line 104: Nit, can we use if statements here? This `&&` pattern is more common for react rendering, and is slightly confusing in this context.

camden11 (approved):
agree on the `&&`, otherwise looks good

packages/cli/commands/theme/marketplace-validate.js
```diff
@@ -100,20 +100,54 @@ exports.handler = async options => {
...
       }
 
       if (status === 'PASS') {
         logger.success(i18n(`${i18nKey}.results.noErrors`));
+
+        const warningValidations = results.filter(
+          test => test.status === 'WARN'
+        );
+
+        warningValidations.forEach(val => {
```
brandenrodgers, line 147: You could cut down on lines of code a little here by using an if check for `val.status === 'warn'` in the `forEach`

PR #763, opened 2022-11-09
Adding deleteAccount, deleteConfigFile and returning config from methods (VSCode Extension Updates)

miketalley:
## Description and Context
To coordinate with https://github.com/HubSpot/hubspot-cms-vscode/pull/134, there were some changes made to the config section to allow functionality there.
- Added `deleteAccount` method - deletes a single specified account from the `hubspot.config.yml` file
- Added `deleteConfigFile` method - deletes the `hubspot.config.yml` file
- Return `getConfig` from `loadConfig` and `loadConfigFromFile` methods instead of `undefined`.

## Who to Notify
@anthmatic @TanyaScales @jsines @brandenrodgers @kemmerle

### DIFF
--- a/packages/cli-lib/lib/config.js
+++ b/packages/cli-lib/lib/config.js
@@ -581,6 +581,11 @@ const deleteAccount = async accountName => {
 
   setConfig({
     ...config,
+    defaultPortal:
+      config.defaultPortal === accountName ||
+      config.defaultPortal === accountIdToDelete
+        ? null
+        : config.defaultPortal,
     portals: accounts.filter(account => account.portalId !== accountIdToDelete),
   });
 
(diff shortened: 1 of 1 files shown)


packages/cli-lib/lib/config.js
```diff
@@ -602,6 +630,10 @@ const deleteEmptyConfigFile = () => {
   );
 };
 
+const deleteConfigFile = () => {
```
brandenrodgers, line 633: Should `deleteEmptyConfigFile` call this new delete function? Just to DRY things up a little

brandenrodgers (approved):
👍

packages/cli-lib/lib/config.js
miketalley: `deleteConfigFile` is more destructive in that it will delete a config file that is not empty, however, `deleteEmptyConfigFile` will only delete an empty config file.

PR #764, opened 2022-11-10
JS Fields Updates

anthmatic:
## Description and Context
Documents field js commands

`hs process` is now `hs cms convert-fields`

and the supported flags for `hs watch` and `hs upload` have changed from `--processFieldsJs` to `--convertFields`

## Screenshots
`hs cms convert-fields`
image

`hs watch/upload --convertFields`
image

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -786,6 +786,7 @@ en:
           initialUpload: "To upload the directory run \"hs upload\" beforehand or add the \"--initial-upload\" option when running \"hs watch\"."
           notUploaded: "The \"hs watch\" command no longer uploads the watched directory when started. The directory \"{{ path }}\" was not uploaded."
       process:
+        describe: "Processes a specific javascript fields file"
         positionals:
           src:
             describe: Path to JS Fields file or directory containing javascript fields files.
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -90,6 +90,7 @@ en:
                 promptMessage: "Enter http timeout duration"
                 success: "The http timeout has been set to: {{ timeout }}"
       cms:
+        describe: "Commands for working with the CMS."
         subcommands:
           lighthouseScore:
             describe: "Score a theme using Google lighthouse."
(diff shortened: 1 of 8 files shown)


packages/cli/commands/process.js
```diff
@@ -15,7 +15,7 @@ const { trackProcessFieldsUsage } = require('../lib/usageTracking');
 const i18nKey = 'cli.commands.process';
 
 exports.command = 'process';
```
anthmatic, line 17: Thoughts on renaming this top level command to something more specific like `process-fields`?

@TanyaScales @brandenrodgers

bmatto: I agree it is more descriptive.

brandenrodgers: We have a `hs module` command bucket. This seems like it could go in there, rather than live at the top level. I also agree that `process` is too vague. `process-fields` is definitely better IMO, but it's still not clear that it is referring to module fields.

`hs module process-fields` seems reasonable. What do you think about that?

anthmatic: @brandenrodgers good call, I agree bucketing this would be better but this command also works for themes fields as well.

We could start a general `hs cms` bucket. Or actually I think you could make the case that whether you use this command for a theme or a module, it could be bucketed under `hs theme` since themes are the "parent feature" that you're using here.

TanyaScales: I'd almost ask use to reconsider the term `process` altogether -- It doesn't really explain what is going on, IMO. Using something like `convert` or `compile` is what comes to mind instead?

As to where it would sit -- we have what appears to be a top-level `cms` bucket already containing the `lighthouseScore` command. But, what comes to mind for me is a top level `fields` bucket instead, as there may be more commands in the future related to manipulating fields.json or fields.js files within themes/modules, and elsewhere.

So the end result would actually be `hs fields compile` or `hs fields compileJs` or something like that.

packages/cli-lib/lang/en.lyaml
```diff
@@ -786,6 +786,7 @@ en:
           initialUpload: "To upload the directory run \"hs upload\" beforehand or add the \"--initial-upload\" option when running \"hs watch\"."
           notUploaded: "The \"hs watch\" command no longer uploads the watched directory when started. The directory \"{{ path }}\" was not uploaded."
       process:
+        describe: "Processes a specific javascript fields file"
```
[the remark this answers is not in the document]
TanyaScales: Based on my comment from above: "Compiles a specific JavaScript module or theme fields file to JSON"

packages/cli/commands/process.js
anthmatic: I'd be +1 for 'convert` since it feels like that's closest to what we're doing.

I believe the ultimate goal is to consolidate commands into buckets so I'd be hesitant to introduce another top level `fields` bucket, especially since it would only have one command in it. I missed the `cms` bucket, though... `hs cms convertFields` seem good?

TanyaScales: We didn't get the change to talk about this specific case in the CLI sync today -- but I like the suggestion you have above @anthmatic and it aligns with what @brandenrodgers was talking about in the meeting (bucketing under `cms`). So, I'm in favor of the `hs cms convert-fields` command and flag as you have it laid out here.

packages/cli-lib/lang/en.lyaml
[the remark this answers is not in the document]
TanyaScales:
```suggestion
        describe: "Converts a specific JavaScript fields file of a module or theme to JSON"
```

Based on new comments in the above thread

TanyaScales (approved):

miketalley (approved):

PR #769, opened 2022-11-17
Handle subcomponents in hs project upload

camden11:
## Description and Context
This updates the `hs project upload` command  to work with the new subcomponent setup. It now uses the `/structure` endpoints to determine which components are top level, and shows only those when building and deploying.
https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/360

## Screenshots
Before (
Screen Shot 2022-11-17 at 4 51 25 PM
ungated)

After (ungated + matches gated)
Screen Shot 2022-11-17 at 4 58 28 PM

## TODO
There are a few minor issues here but they should theoretically be resolved on the backend.

## Who to Notify
@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -22,7 +22,9 @@ const {
 const {
   createProject,
   getBuildStatus,
+  getBuildStructure,
   getDeployStatus,
+  getDeployStructure,
   fetchProject,
   uploadProject,
 } = require('@hubspot/cli-lib/api/dfs');
@@ -340,6 +342,7 @@ const handleProjectUpload = async (
 
 const makePollTaskStatusFunc = ({
   statusFn,
+  structureFn,
   statusText,
   statusStrings,
   linkToHubSpot,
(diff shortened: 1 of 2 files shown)


packages/cli/lib/projects.js
```diff
@@ -372,9 +375,19 @@ const makePollTaskStatusFunc = ({
 
     spinnies.add('overallTaskStatus', { text: 'Beginning' });
 
-    const initialTaskStatus = await statusFn(accountId, taskName, taskId);
+    const [
+      initialTaskStatus,
+      { topLevelComponentsWithChildren: taskStructure },
+    ] = await Promise.all([
+      statusFn(accountId, taskName, taskId),
+      structureFn(accountId, taskName, taskId),
```
brandenrodgers, line 383: Just making sure. Will this be backwards compatible for accounts that are not using the dependent subcomponents BE yet?

camden11: Yep, it's backwards compatible in the same way that the build/deploy details pages are (the structure endpoint will just return all components as top level on older projects). Just need to make sure that gets deployed on prod before we release this, so might not make sense to merge this for a bit.

brandenrodgers (approved):

PR #770, opened 2022-11-18
Support annotation lookup by source

anthmatic:
## Description and Context
I didn't notice it before but the ability to get annotations straight from the source was removed in #556 which removes functionality that the VS Code extension relies on.

### DIFF
--- a/packages/cli-lib/templates.js
+++ b/packages/cli-lib/templates.js
@@ -32,6 +32,10 @@ const buildAnnotationValueGetter = file => {
     logger.error(`Error reading file annotations ${file}`);
     return;
   }
+  return getAnnotationsFromSource(source);
+};
+
+const getAnnotationsFromSource = source => {
   const match = source.match(ANNOTATIONS_REGEX);
   const annotation = match && match[1] ? match[1] : '';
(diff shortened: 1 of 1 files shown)


packages/cli-lib/templates.js
```diff
@@ -46,6 +50,7 @@ const isCodedFile = filePath => TEMPLATE_EXTENSION_REGEX.test(filePath);
 module.exports = {
   ANNOTATION_KEYS,
   getAnnotationValue,
+  getAnnotationsFromSource,
   buildAnnotationValueGetter,
```
anthmatic, line 54: It looks like this actually might not be used anymore, if it makes more sense I can just change `buildAnnotationValueGetter` to use source instead of file

brandenrodgers: > It looks like this actually might not be used anymore

It may not be used internally by us, but there's always the chance that it is being used externally (very unlikely, I know). I think what you currently have makes sense 👍

brandenrodgers (approved):

anthmatic:
Thanks @brandenrodgers

PR #772, opened 2022-11-22
Create module updates

miketalley:
## Description and Context
https://github.com/HubSpot/hubspot-cms-vscode/pull/135

- Moves `createModule` from `cli` to `cli-lib` to make it accessible to library consumers
- Moves `Sample.module` contents from `cli` to `cli-lib`
- Adds `options` argument to `createModule` and allows passing `name` of `''` along with `options.allowExistingDir: true` to place contents of `Sample.module` into existing directory


brandenrodgers:
This seems reasonable to me 👍. Seems like we should move towards a world where the cli-lib is responsible for all code generation logic? So external systems can more easily make use of it.

One idea - Maybe as a part of this update we can also get rid of all the default code examples that live in this repo? The more standard method of generating code in the CLI is to download it from a public GH repo. A benefit of doing it that way is that it decouples the code that gets generated from the CLI from the CLI itself. They can be updated independently without having to require a new CLI release. It will also move us slightly closer to singular pattern for code generation, which is nice.

miketalley:
> Maybe as a part of this update we can also get rid of all the default code examples that live in this repo?

I updated the PR to include moving the `createTemplate` functionality over to `cli-lib` as well.

brandenrodgers:
>  I think a follow-up PR for those changes would be prudent, so we do not block these updates.

I agree 👍

brandenrodgers (approved):
👌

PR #785, opened 2022-12-08
[WIP] Root CLI config

brandenrodgers:
## Description and Context

This PR does a few things, but the main thing is that it adds support for a new hidden, root-level configuration file. This is an attempt to handle CLI configuration in a more user-friendly manner. A few pros of storing the config file at the root are:
- Less of a chance of users accidentally committing their access tokens to public source control
- No clunky `hubspot.config.yml` file that gets generated in every directory that the user runs `hs init` in.
- Internally, it is simpler for us to locate + update the configuration file because we always know where it is.

### Updates in this PR
The work I have here adds support for the new root-level configuration file, but it shouldn't break any existing behavior. The code will initially check to see if there is a root level config. If there is one, then we will use it. If not, it will fall back to the current (I'm now calling them "legacy") configuration files. If those don't exist either, then it will default to using the new, root-level configuration flow.

All of this ^ logic for maintaining backwards compatibility happens in [cli-lib/lib/config.js](https://github.com/HubSpot/hubspot-cli/compare/root-config?expand=1#diff-316f96aa9438aa3faac1f763c2c44b70f5f7c801bd5dd6aa485f91c65bba9fd1R1). It is now acting as a proxy file to allow code to make use of either the new flow or the old flow.

The old config logic has been moved into [cli-lib/config/legacyConfig](https://github.com/HubSpot/hubspot-cli/compare/root-config?expand=1#diff-b9205147643832c6ac6dbb00e170cd760f1efacf123e795ebc2b0384e00124feR1).

The new config logic lives in a new class that I created, [CLIConfiguration](https://github.com/HubSpot/hubspot-cli/compare/root-config?expand=1#diff-f994b2649bb45dbab39e9fc39808eb7d538306a9033400015f416ade73023206R16). This is doing a lot of what the old configuration file was doing, except it is a class so some things are simpler...
- We only have to load options one time, and then they don't need to be passed around after that
- It stores the config as a field on the class, which IMO is an easier pattern to follow than using a global variable for it.

### New Command
As a part of this work, I created a new command that uses the new configuration logic. It is `hs accounts auth` and it does what `hs init` AND `hs auth` currently do. If there is no configuration file, it will create it. If there is one, it will update it. With this we can essentially get rid of `hs init` and also `hs auth`.

### Omitted work
One behavior of the current, legacy config files is that they support nesting. Devs can use the nested behavior to override the `defaultPortal` field on a per-directory basis. This is useful for devs who work on multiple projects.

We lose this behavior with the root-level config. However, one solution is to add a new `defaultAccountOverrides` section to the config file. This would look something along the lines of:
```yml
defaultAccountOverrides:
  src/brodgers/project-1: "account1"
  src/brodgers/project-2": "account2"
```

The key is the path to the desired directory, and the value is the account that should be used as the default for that directory. This should enable devs to utilize the same workflow that they were able to achieve with the nesting behavior.


brandenrodgers:
closing because we are going to roll this out as part of https://github.com/HubSpot/hubspot-local-dev-lib

PR #786, opened 2022-12-12
Fixing project download bug

brandenrodgers:
## Description and Context

This bug was causing `hs project download` to throw an error.


camden11 (approved):
:ship:

PR #787, opened 2022-12-15
Bugfix for convert-fields on Windows

anthmatic:
## Description and Context
There turned out to be a few different things that weren't working on Windows, mostly around file paths but a couple other nuances too. I've left comments on the diffs for specifics

### DIFF
--- a/packages/cli-lib/lib/dynamicImport.js
+++ b/packages/cli-lib/lib/dynamicImport.js
@@ -1,4 +1,5 @@
 const semver = require('semver');
+const { pathToFileURL } = require('url');
 const { getExt } = require('../path');
 const { logger } = require('../logger');
 // This is pulled into it's own file because it must be added to the esignore. Dynamics imports cause eslint _parser_ errors, which cannot be ignored.
@@ -10,7 +11,9 @@ const { logger } = require('../logger');
  */
 async function dynamicImport(filePath) {
   if (semver.gte(process.version, '13.2.0')) {
-    const exported = await import(filePath).then(content => content.default);
+    const exported = await import(pathToFileURL(filePath)).then(
+      content => content.default
+    );
     return exported;
   } else {
     if (getExt(filePath) == 'mjs') {
--- a/packages/cli-lib/lib/spawnFunction.js
+++ b/packages/cli-lib/lib/spawnFunction.js
@@ -3,7 +3,6 @@ const cp = require('child_process');
 const spawn = codeToExec => {
   // TODO: Sanitize input here, probably.
   const child = cp.spawnSync('node', ['-e', codeToExec], {
-    detached: true, //child process does not end when parent ends
     stdio: 'inherit',
   });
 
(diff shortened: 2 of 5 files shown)


packages/cli-lib/lib/watch.js
```diff
@@ -65,7 +65,8 @@ async function uploadFile(accountId, file, dest, options) {
     // Because we cannot clear ES Modules from the cache, we need to load them in a separate process.
     // So this spawns a new node process that calls `hs upload` with the same options as watch was called with.
     // More context: https://github.com/HubSpot/hubspot-cli/pull/712#discussion_r945056954
-    const fieldsJsProcessCode = `require('@hubspot/cli/commands/upload').handler(${parsedOptions})`;
+    const uploadCommandPath = require.resolve('@hubspot/cli/commands/upload');
+    const fieldsJsProcessCode = `require('${uploadCommandPath}').handler(${parsedOptions})`;
```
anthmatic, line 69: We need to resolve the path from the cli-lib module because if we require it directly in the child process, it expects the CWD to have `node_modules/@hubspot/cli`

anthmatic: NOTE: untested on Windows

packages/cli/commands/cms/convertFields.js
```diff
@@ -27,7 +27,10 @@ const invalidPath = src => {
 
 exports.handler = async options => {
   const src = path.resolve(getCwd(), options.src);
-  const projectRoot = path.dirname(getThemeJSONPath(src));
+  const themeJSONPath = getThemeJSONPath(src);
+  const projectRoot = themeJSONPath
+    ? path.dirname(themeJSONPath)
+    : path.dirname(getCwd());
```
anthmatic, line 33: `projectRoot` shouldn't be `undefined` because we'll fail to get the correct relative path in Windows

packages/cli-lib/lib/dynamicImport.js
```diff
@@ -10,7 +11,9 @@ const { logger } = require('../logger');
  */
 async function dynamicImport(filePath) {
   if (semver.gte(process.version, '13.2.0')) {
-    const exported = await import(filePath).then(content => content.default);
+    const exported = await import(pathToFileURL(filePath)).then(
+      content => content.default
+    );
```
anthmatic, line 16: this ensures that `filePath` is a valid URL (`file://c:\...` on Windows)

packages/cli-lib/lib/spawnFunction.js
```diff
@@ -3,7 +3,6 @@ const cp = require('child_process');
 const spawn = codeToExec => {
   // TODO: Sanitize input here, probably.
   const child = cp.spawnSync('node', ['-e', codeToExec], {
-    detached: true, //child process does not end when parent ends
```
anthmatic, line 6: Behavior with this option differs on Windows. Instead of a different process it creates a new shell.

According to https://nodejs.org/api/child_process.html#optionsdetached: 
>  If the parent's stdio is inherited, the child will remain attached to the controlling terminal.
So I believe it was being ignored anyways

TanyaScales (approved):

PR #788, opened 2022-12-15
[WIP] Hackathon boilerplate generation

Triippz:
## Description and Context

This is part of our internal team hackathon.

Enhance the CLI by adding commands that allow developers to create customized boilerplate components for the their project. Along with a command that allows a developer to use a boilerplate component template that can be created by non-hubspot developers, by pointing to a git repository that contains the boilerplate component. The concept behind this idea is similar to other generators such as Yeoman, Cookiecutter, JHipster, etc. This can help drive an open source community centered around HubSpot projects as well as help developers get started faster with their projects that may be more complex and contain multiple components.

While there are some pretty opinionated stances that are WIDE open to changes, the basic concepts here are:
- Using [ejs](https://ejs.co/) as a templating language
- Using [OctoKit](https://github.com/octokit/octokit.js) for GitHub interaction without the need for git directly.
- HubSpot will, internally, maintain a repository with a set of customized components, that various CMS, CRM, etc. teams can leverage to help bootstrap a developers development journey much faster.
- Potentially allow for any developer to create their own component customizations that can be "downloaded" or "generated" via this command, following a prescribed framework we, HubSpot, define.

## Who to Notify

@lilyhehehe


brandenrodgers:
Closing this b/c the hackathon is over, and we have the demo recorded already.

PR #791, opened 2022-12-19
Update polling timeout to 2.5s

m-roll:
## Description and Context
We get some performance gains, at least from the user's perspective, by reducing this polling amount. I still think we should consider opening up some websockets on the backend so we can give live status updates. 

I chose 2.5 since its just half of what existed.

## TODO
- [ ] Should this constant be configurable by the config yml or environment variables, instead of hard-coded?

## Who to Notify
cc @brandenrodgers @kemmerle @camden11

### DIFF
--- a/packages/cli-lib/lib/constants.js
+++ b/packages/cli-lib/lib/constants.js
@@ -99,7 +99,7 @@ const MARKETPLACE_FOLDER = '@marketplace';
 
 const FOLDER_DOT_EXTENSIONS = ['functions', 'module'];
 
-const POLLING_DELAY = 5000;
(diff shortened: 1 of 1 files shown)


brandenrodgers:
> Should this constant be configurable by the config yml or environment variables, instead of hard-coded?

At least for now, I'd prefer to have control over the timeout. I think we're okay keeping it hard coded.

brandenrodgers (approved):
👌

camden11:
Polling has since been reduced to 2000 ms in a different PR, don't think it makes sense to change to 2500 at this point.

PR #792, opened 2022-12-20
Move assets to cms-sample-assets repository and add methods for fetching contents from github

miketalley:
## Description and Context
Fetching assets, such as those residing in `@hubspot/cli-lib/defaults/` has some drawbacks because they are not `require`d in anywhere. We instead assume that these assets will be available at a specific path and load them at runtime.

For any consumers of the `cli-lib` package that are bundling, this can be a real problem because these assets do not get included in the bundle. This PR looks to move the assets into a new repository, [cms-sample-assets](https://github.com/HubSpot/cms-sample-assets), and update the reliance on them to using helper methods that fetch the contents from github.

#### Testing
Run `hs create module <name> <dest>` or `hs create template <name> <dest>`

#### TODO
- [x] Fix tests (API limit exceeded for GitHub)

## Who to Notify
@brandenrodgers @kemmerle @camden11 @anthmatic @jsines


packages/cli-lib/github.js
```diff
@@ -117,7 +124,83 @@ async function cloneGitHubRepo(dest, type, repoName, sourceDir, options = {}) {
...
+}
+
+/**
+ * Writes files from a HubSpot public repository to the destination folder
+ * @param {String} dest - Dir to write contents to
+ * @param {String} repoName - Name of GitHub repository to fetch contents from
+ * @param {String} path - Path to obtain contents from within repository
+ * @returns {Boolean} `true` if successful, `false` otherwise.
+ */
+async function downloadGitHubRepoContents(
```
jsines, line 151: Wondering for right-click create commands & any executing these programmatically; are these being redownloaded every call? Could we cache them client side to improve performance, since they'll likely not be changed often?

anthmatic:
I like that this allows us to get files by path from a repo. I wonder if we should split off `packaged/cli-assets` into its own GH Repo maybe with a more generic name like `sample-assets`? All of the other sample files exist in their own repos.

Also FWIW, with the current lerna config I believe this would get published on NPM

miketalley:
> I like that this allows us to get files by path from a repo. I wonder if we should split off `packaged/cli-assets` into its own GH Repo maybe with a more generic name like `sample-assets`? All of the other sample files exist in their own repos.
> 
> Also FWIW, with the current lerna config I believe this would get published on NPM

I like this idea. I thought it would make sense to have this as a package since it is related to the CLI, but I think it does make sense to keep it elsewhere because now we are going to be using it with the VSCode extension as well.

I'll set up another public repo with the assets and update this PR.

packages/cli-lib/github.js
miketalley: I think we could certainly do that, but to me, I think that the download is small enough that it makes sense to leave it like this, so any updates to the remote contents will come through immediately, without any need to invalidate the cache.

miketalley:
the sample assets now live in the [cms-sample-assets repo](https://github.com/HubSpot/cms-sample-assets) and the code has been updated to download from there.

packages/cli-lib/github.js
```diff
@@ -7,6 +9,11 @@ const { extractZipArchive } = require('./archive');
 const { GITHUB_RELEASE_TYPES } = require('./lib/constants');
 const { DEFAULT_USER_AGENT_HEADERS } = require('./http/requestOptions');
 
+const GITHUB_AUTH_HEADERS = {
+  authorization:
+    global && global.githubToken ? `Bearer ${global.githubToken}` : '',
+};
```
anthmatic, line 15: Does it make any difference to send and empty `authorization` header? I'm guessing it doesn't or even possibly gets stripped out but figured I'd ask

anthmatic (approved):
1 non blocking question, otherwise LGTM!

packages/cli-lib/github.js
kemmerle: ^^ I have the same question @miketalley

packages/cli-lib/templates.js
```diff
@@ -50,22 +51,16 @@ const getAnnotationsFromSource = source => {
...
-    __dirname,
-    './defaults/blog-listing-template.html'
-  ),
-  'blog-post-template': path.resolve(
-    __dirname,
-    './defaults/blog-post-template.html'
-  ),
-  'search-template': path.resolve(__dirname, './defaults/search-template.html'),
+  'page-template': 'templates/page-template.html',
+  partial: 'templates/partial.html',
```
kemmerle, line 55: Does partial need to be surrounded by '', as well, @miketalley?

packages/cli-lib/github.js
miketalley: Ah good call. This should be `null` -- will fix.

packages/cli-lib/templates.js
miketalley: It doesn't. Prettier puts quotes around properties with the `-` only.

PR #793, opened 2022-12-20
Rework fieldsjs child process

anthmatic:

### DIFF
--- a/packages/cli-lib/lib/processFieldsJs.js
+++ b/packages/cli-lib/lib/processFieldsJs.js
@@ -1,0 +1,133 @@
+const process = require('process');
+const path = require('path');
+const fs = require('fs');
+const semver = require('semver');
+const { pathToFileURL } = require('url');
+
+const dirName = process.env.dirName;
+const cwd = process.env.cwd;
+const fieldOptions = process.env.fieldOptions;
+const filePath = process.env.filePath;
+const writeDir = process.env.writeDir;
+
+// TODO: I think we can just give the process this dirName
+// Switch CWD to the dir of the fieldsjs. This is so that any calls to the file system that are written relatively will resolve properly.
+process.chdir(dirName);
+
+/*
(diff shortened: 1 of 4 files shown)


anthmatic:
This issue came in and it looks like this PR should fix the issue: https://github.com/HubSpot/hubspot-cli/issues/797

packages/cli-lib/lib/processFieldsJs.js
```diff
@@ -0,0 +1,123 @@
...
+  });
+  return JSON.stringify(fields, null, 2);
+}
+
+/**
+ * Takes in a path to a javascript file and either dynamically imports it or requires it, and returns, depending on node version.
+ * @param {string} filePath - Path to javascript file
+ * @returns {Promise | undefined} - Returns _default_ exported content if ESM, or exported module content if CJS, or undefined if node version < 13.2 and file is .mjs.
+ */
+async function dynamicImport(filePath) {
```
miketalley, line 110: 👍

packages/cli-lib/lib/processFieldsJs.js
```diff
@@ -0,0 +1,123 @@
...
+  }, []);
+}
+
+/**
+ * Converts an array of field objects to JSON
+ * @param {array} fields - Array of field objects
+ * @returns {string} - Returns converted JSON
+ */
+async function fieldsArrayToJson(fields) {
+  fields = await Promise.all(flattenArray(fields));
```
miketalley, line 98: Might be missing something, but why do we need `await Promise.all` here? It doesn't look like `flattenArray` returns a promise to me.

anthmatic: Good Q. I Believe `fields` items can be promises so this just makes sure they all get resolved

anthmatic: Ah, found this explanation: https://github.com/HubSpot/hubspot-cli/pull/738/files#r960887568

PR #794, opened 2022-12-21
Remove cli lib defaults and update paths

miketalley:
## Description and Context
Follow-up PR to https://github.com/HubSpot/hubspot-cli/pull/792 which will remove old assets from `cli-lib` and update paths. This is done in a two step process so that the previous PR's tests can pass with the new github content download functionality, but with the existing paths.

## Who to Notify
@brandenrodgers @kemmerle @camden11 @anthmatic @jsines


miketalley:
No longer need this PR as we can do everything in one PR with new changes to https://github.com/HubSpot/hubspot-cli/pull/792

PR #799, opened 2023-01-03
Nested subcomponents view on hs project upload

camden11:
## Description and Context
This updates the build/deploy status view shown when running `hs project upload` to include nested subcomponents. Also makes a few other UI improvements.

## Screenshots
https://user-images.githubusercontent.com/10413161/210836230-3ec4b766-1938-498d-a58d-2738962be35a.mov

Updated names
Screen Shot 2023-01-19 at 1 36 41 PM

## Testing
Due to a bug that's causing failing builds to hang, I haven't been able to test this on a failing build yet (though it works for failed deploys).

## Who to Notify
@kemmerle @markhazlewood


packages/cli/lib/projects.js
```diff
@@ -410,29 +414,61 @@ const makePollTaskStatusFunc = ({
       structureFn(accountId, taskName, taskId),
     ]);
 
-    const topLevelSubtasks = initialTaskStatus[statusText.SUBTASK_KEY].filter(
-      ({ id }) => !!taskStructure[id]
+    const flatTaskList = initialTaskStatus[statusText.SUBTASK_KEY].filter(
```
kemmerle, line 417: I am concerned about the complicated ways in which we're mapping over and changing this data. 

`flatTaskList` (Defined and then used once by `tasksById`)
`tasksById` (Defined and then used in `structuredTasks` in three places). 
`structuredTasks` (Defined and then used three separate times: once in defining `numComponents`, then iterated over in the `addTaskSpinner` function, and then used in defining `topLevelTask`). 
`addTaskSpinner` also defines four separate `const`'s within it.

It seems like we're introducing a lot of breaking points within the code. I'm worried that if we experience a bug, it's going to be very difficult to track down, but (and I apologize for this), I'm not quite sure if I know how to change the code to address this concern. Can we schedule a 1:1, so you can take me through the data structures you're getting handed down? Or is that too much? I'd love to have a longer conversation about this.

kemmerle: Talked this through via Zoom, and I'm all good!

PR #802, opened 2023-01-11
Sandbox delete - portal list updates

adamawang:
## Description and Context

When listing portals for the `hs sandbox delete` command, users were getting confused when presented options that were not sandbox portals. While the API would not allow any other portal type to be deleted (like prod for instance), the messaging was not clear and made it seem like customer portals could be deleted. 

The list of portals shown in the picker is now limited to sandboxes only. The default account picker remains the same. 

## Screenshots

Before:
Screen Shot 2023-01-11 at 16 21 16

After:
Screen Shot 2023-01-11 at 16 10 37

## Who to Notify

@kemmerle @brandenrodgers

### DIFF
--- a/packages/cli/lib/prompts/accountsPrompt.js
+++ b/packages/cli/lib/prompts/accountsPrompt.js
@@ -1,7 +1,19 @@
 const { updateDefaultAccount } = require('@hubspot/cli-lib/lib/config');
 const { promptUser } = require('./promptUtils');
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
-const { mapAccountChoices } = require('./sandboxesPrompt');
+
+const getSandboxType = type =>
+  type === 'DEVELOPER' ? 'development' : 'standard';
+
+const mapAccountChoices = portals =>
+  portals.map(p => {
+    const isSandbox = p.sandboxAccountType !== null;
+    const sandboxName = `[${getSandboxType(p.sandboxAccountType)} sandbox] `;
+    return {
+      name: `${p.name} ${isSandbox ? sandboxName : ''}(${p.portalId})`,
+      value: p.name || p.portalId,
+    };
+  });
 
 const i18nKey = 'cli.commands.accounts.subcommands.use';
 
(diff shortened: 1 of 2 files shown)


adamawang:
Last commit included a fix for older config entries that were missing a `sandboxAccountType` property. This was causing `hs accounts use` results to include an erroneous `[sandbox type]` label for accounts that may not have been sandboxes.

packages/cli/lib/prompts/sandboxesPrompt.js
```diff
@@ -6,15 +6,16 @@ const i18nKey = 'cli.lib.prompts.sandboxesPrompt';
...
-    const isSandbox = p.sandboxAccountType !== null;
-    const sandboxName = `[${getSandboxType(p.sandboxAccountType)} sandbox] `;
-    return {
-      name: `${p.name} ${isSandbox ? sandboxName : ''}(${p.portalId})`,
-      value: p.name || p.portalId,
-    };
-  });
+const mapSandboxAccountChoices = portals =>
+  portals
+    .filter(p => p.sandboxAccountType && p.sandboxAccountType !== null)
```
adamawang, line 11: Filtering out non-sandboxes from the delete sandbox list

packages/cli/lib/prompts/accountsPrompt.js
```diff
@@ -1,7 +1,19 @@
...
+
+const mapAccountChoices = portals =>
+  portals.map(p => {
+    const isSandbox = p.sandboxAccountType && p.sandboxAccountType !== null;
+    const sandboxName = `[${getSandboxType(p.sandboxAccountType)} sandbox] `;
+    return {
+      name: `${p.name} ${isSandbox ? sandboxName : ''}(${p.portalId})`,
+      value: p.name || p.portalId,
+    };
+  });
```
adamawang, line 16: Separated out the account mapping for the accounts prompt, and added in an undefined check for sandboxAccountType to support older configs

kemmerle (approved):
Reviewed the code and tested locally, everything is LGTM! 🚢

PR #803, opened 2023-01-12
Fix issue where files do not upload running `hs watch --initial-upload`

anthmatic:
## Description and Context
Since 4.1.0, passing the `--initial-upload` flag to the `hs watch` command does not actually upload any files because there was a step added to `uploadFolder` that was not accounted for in the watch command.

It also looks like the logging for `disable-initial` has been broken possibly since we moved from commander > yargs due to a collision where `debug` and `disable-initial` both share the same flag alias, `d`.

## Screenshots
image

### DIFF
--- a/packages/cli/lib/upload.js
+++ b/packages/cli/lib/upload.js
@@ -1,0 +1,50 @@
+const path = require('path');
+const { walk } = require('@hubspot/cli-lib/lib/walk');
+const { createIgnoreFilter } = require('@hubspot/cli-lib/ignoreRules');
+const { fieldsJsPrompt } = require('../lib/prompts/cmsFieldPrompt');
+const { isAllowedExtension } = require('@hubspot/cli-lib/path');
+const { isConvertableFieldJs } = require('@hubspot/cli-lib/lib/handleFieldsJs');
+
+/*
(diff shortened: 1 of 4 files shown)


packages/cli/lib/upload.js
```diff
@@ -0,0 +1,50 @@
...
+      if (choice !== filePath) continue;
+    }
+    uploadableFiles.push(filePath);
+  }
+  return uploadableFiles;
+};
+
+module.exports = {
+  getUploadableFileList,
+};
```
anthmatic, line 50: I think ideally we'd pull this out into `cli-lib` but it's relied on a `prompt` from `/cli`

miketalley (approved):
Looks good

miketalley (changes requested):
It looks like when I use `--initial-upload`, my watched files do not get uploaded. There is a message saying 

```
[SUCCESS] Completed uploading files in /Users/mtalley/HubSpot/prod/hubspot-cli/mytestfolder to OnlyATest in 101867970
```

though. `false`

miketalley (approved):
Sorry that was my bad. It looks to be uploading the files with `--initial-upload` specified.

PR #805, opened 2023-01-30
Error saying that environment variable config could not be loaded only shows when useEnv is specified

miketalley:
## Description and Context
When running commands in the CLI and specifying an account using `--account=<accountName>`, the config validation code path checks for an environment variable based config in addition to specifying the account, which would be invalid. This check always displays an error stating `Unable to load config from environment variables.`. This message can be confusing because it seems like the command did not work, even though it did.

This PR fixes the error from showing up when this validation check is made by passing in options when calling `loadConfigFromEnvironment` and checking for `useEnv` to be `true` in order to show the error message.

## Screenshots
Before: When running `hs upload ./tmp/MyFunctions.functions MyFunctions.functions --account=<accountName>`
image

After: When running `hs upload ./tmp/MyFunctions.functions MyFunctions.functions --account=<accountName>`
image

### DIFF
--- a/packages/cli-lib/lib/config.js
+++ b/packages/cli-lib/lib/config.js
@@ -259,7 +259,7 @@ const loadConfig = (
     useEnv: false,
   }
 ) => {
-  if (options.useEnv && loadEnvironmentVariableConfig()) {
+  if (options.useEnv && loadEnvironmentVariableConfig(options)) {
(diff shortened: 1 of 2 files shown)


brandenrodgers (approved):

brandenrodgers:
Tested this locally and it lgtm. I'm going to merge it.

PR #808, opened 2023-02-02
Update ignoreRules.js

m-roll:
## Description and Context
With some of the experimental features to the back-end, I'd like to open the possibility of reading `package-lock.json` from customer uploads. I can provide more context in our backroom channel, but I think the reason this was previously included in the ignore list is now moot since we moved over to our sub-component based build strategy for private Apps.

Previous PR that introduced the ignore: https://github.com/HubSpot/hubspot-cli/pull/477


m-roll:
Just realized this won't work since the CMS uses this ignore list too. Does it seem reasonable to have a separate ignore list for CMS and for projects?. cc @brandenrodgers @camden11 @kemmerle

PR #810, opened 2023-02-10
Sandbox sync command

adamawang:
## Description and Context

Adding a new command `hs sandbox sync` to allow users to sync their standard or development sandboxes. More discussion can be found here: https://git.hubteam.com/HubSpot/sandboxes-ui/issues/1590

Running the new command will take the default account, lookup the parent portal, and initiate a sync through the parent portal's token. This PR sets up the sync flow for both development and standard sandboxes. We deduce the portal type from the config, which gets saved after `hs auth`. 

This PR requires an update with personal access key UI, which allows users to add read scopes to the key. Users with existing keys will have to regenerate new keys with the correct scopes to run this. PR here: https://git.hubteam.com/HubSpot/personal-access-key-ui/pull/300

Error handling should cover no parent portal, running this with a non-sandbox default account, running this with a standard sandbox (it will exit, next PR will support this pathway), and if a PAK has the wrong scopes. 

Scripting functionality was added with an optional `-f` or `--force` to bypass the confirmation step. 

This PR also adds `cli-progress` as a new package to show sandbox sync progress. 

## Screenshots

New sync flow, and progress bars for standard sandbox sync and development sandbox sync:
Screenshot 2023-03-09 at 13 34 21
Screenshot 2023-03-09 at 13 35 35

Errors: 
Screenshot 2023-02-16 at 16 43 26
Screenshot 2023-02-16 at 16 44 31
Screenshot 2023-02-24 at 11 47 46
Sync in progress:
Screenshot 2023-03-09 at 13 37 12
Cancelled early:
Screenshot 2023-03-09 at 13 37 01

## TODO

n/a


packages/cli/commands/sandbox/create.js
```diff
@@ -10,105 +10,26 @@ const { logger } = require('@hubspot/cli-lib/logger');
...
-      account: validName,
-    })
-  );
-  uiFeatureHighlight([
-    'accountsUseCommand',
-    'accountOption',
-    'accountsListCommand',
-  ]);
-};
-
```
adamawang, line 111: Moved this to a lib file along with a few other helpers

packages/cli/commands/sandbox/sync.js
```diff
@@ -0,0 +1,244 @@
...
+
+const i18nKey = 'cli.commands.sandbox.subcommands.sync';
+
+exports.command = 'sync';
+exports.describe = i18n(`${i18nKey}.describe`);
+
+exports.handler = async options => {
+  await loadAndValidateOptions(options);
+
+  // TODO: add scripting options
```
adamawang, line 38: Will add scripting options in next commit

packages/cli-lib/api/sandboxes-sync.js
```diff
@@ -0,0 +1,37 @@
+const http = require('../http');
+const SANDBOXES_SYNC_API_PATH = 'sandboxes-sync/v1';
+
+// fromHubId is the parent portal here, which is required as the first param. Parent account key can only run the sync
```
kemmerle, line 4: What is `toHubId` here? I would think it's sandbox Id, but that seems to be represented here as `sandboxHubId`.

packages/cli-lib/api/sandboxes-sync.js
```diff
@@ -0,0 +1,37 @@
...
+
+async function fetchTaskStatus(accountId, taskId) {
+  return http.get(accountId, {
+    uri: `${SANDBOXES_SYNC_API_PATH}/tasks/${taskId}`,
+  });
+}
+
+async function fetchTypes(accountId, toHubId) {
+  return http.get(accountId, {
+    uri: `${SANDBOXES_SYNC_API_PATH}/types${
```
kemmerle, line 27: A little confused about what `types` represents here. Is this how we determine whether a portal is a sandbox account or a PROD account?

packages/cli-lib/api/sandboxes-sync.js
kemmerle: Ok, reading below in the `cli-lib/sandboxes.js` file, you write that `toHubId` is the `target account for the sync` and that `sandboxHubId` is `sandbox portal ID for the sync`. In what cases will these two IDs differ? I would think they'd always be the same.

packages/cli-lib/sandboxes.js
```diff
@@ -40,7 +51,65 @@ async function deleteSandbox(parentAccountId, sandboxAccountId) {
...
+    resp = await _fetchTaskStatus(accountId, taskId);
+  } catch (err) {
+    throw err;
+  }
+
+  return resp;
+}
+
+/**
+ * Fetches available sync types for a specified portal.
```
kemmerle, line 93: Would the types here be a regular sandbox vs a development sandbox?

packages/cli/commands/sandbox/sync.js
```diff
@@ -0,0 +1,251 @@
...
+  const config = getConfig();
+  const accountId = getAccountId(options);
+  const accountConfig = getAccountConfig(accountId);
+  const spinnies = new Spinnies({
+    succeedColor: 'white',
+  });
+
+  trackCommandUsage('sandbox-sync', null, accountId);
+
+  if (
```
kemmerle, line 48: I think the code here is very understandable, and I could read through it with relative ease. For convenience's sake, though, could you add a few comments to clarify some of these conditionals?

packages/cli/lib/sandboxes.js
```diff
@@ -0,0 +1,179 @@
...
+    cliProgress.Presets.rect
+  );
+  const mergeTasks = {
+    'lead-flows': 'forms', // Since UI only shows forms, merge lead-flows into forms progress
+  };
+  const barInstances = {};
+  return new Promise((resolve, reject) => {
+    const pollInterval = setInterval(async () => {
+      const taskResult = await fetchTaskStatus(accountId, taskId).catch(reject);
+      if (taskResult.tasks) {
```
kemmerle, line 127: This conditional is quite confusing. Is there any way to simplify this code?

kemmerle:
@adamawang, I reviewed your code and tested everything out locally.

In terms of testing, everything worked as expected. I particularly liked the extensive prompting you provided--I messed up the command several times, and I didn't need to refer to the PR at all. I just followed your prompts and was able to complete the sync.

In terms of the code, I did have a few clarifying questions for some variable names, and I was concerned about the `pollSyncStatus` function in the `lib/sandboxes` file. It seems pretty confusing to me, and I'm wondering whether there's a way to refactor to streamline it. Happy to chat about it, if that helps.

packages/cli-lib/api/sandboxes-sync.js
adamawang: Yep good catch, its a left over parameter that we never ended up stripping out on our API side that is a requirement still. There may be differences in the future with push to parent functionality, as well as sandbox to sandbox syncs so I think it'd be good to leave like this for now

packages/cli-lib/api/sandboxes-sync.js
adamawang: This one is to fetch sync types, eg: object schemas, workflows, etc. We have types that are only able to be run once per portal, so this fetch is necessary to give us what we can/can't sync.

packages/cli/commands/sandbox/sync.js
adamawang: For sure!

packages/cli/lib/sandboxes.js
adamawang: Let me add some more comments here to clarify things a bit, if its still too confusing I'd be happy to zoom whenever!

kemmerle: Sounds good to me!

packages/cli-lib/sandboxes.js
adamawang: Commented about this here: https://github.com/HubSpot/hubspot-cli/pull/810#discussion_r1119101532

packages/cli/commands/sandbox/sync.js
```diff
@@ -0,0 +1,292 @@
...
+      logErrorInstance(err);
+    }
+    logger.log('');
+
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  try {
+    // Handle manual exit for return key and ctrl+c
+    process.stdin.resume();
```
brandenrodgers, line 232: We have some utils here that may help with some of the exit handling code here: https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lib/process.js

packages/cli/commands/sandbox/sync.js
```diff
@@ -0,0 +1,292 @@
...
+    trackCommandUsage('sandbox-sync', { successful: false }, accountId);
+
+    logger.error(i18n(`${i18nKey}.failure.notSandbox`));
+
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  // Verify parent account exists in the config
+  let parentAccountId;
+  for (const portal of config.portals) {
```
brandenrodgers, line 63: I think you can use this [getAccountConfig](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lib/config.js#L318) util here.

adamawang: Good call, had lifted this from elsewhere and forgot to update. Pushing a new version soon

packages/cli-lib/errorHandlers/apiErrors.js
```diff
@@ -28,6 +28,23 @@ const isMissingScopeError = err =>
...
+  const categoryErr = category && err.error.category === category;
+  const subCategoryErr = subCategory && err.error.subCategory === subCategory;
+  return (
+    err.name === 'StatusCodeError' &&
+    statusCodeErr &&
+    categoryErr &&
+    subCategoryErr
+  );
+};
+
```
adamawang, line 47: Created a util to specify errors

packages/cli/commands/sandbox/sync.js
```diff
@@ -0,0 +1,298 @@
...
+          accountName: getAccountName(parentAccountConfig),
+        })
+      );
+    } else if (
+      isSpecifiedError(
+        err,
+        429,
+        'RATE_LIMITS',
+        'sandboxes-sync-api.SYNC_IN_PROGRESS'
+      )
```
adamawang, line 204: New usage. Must be as specific as possible in case BE changes categories, or else we'll fall back to the basic logErrorInstance at the end

packages/cli/commands/sandbox/sync.js
```diff
@@ -0,0 +1,298 @@
...
+    handleExit(onTerminate);
+    handleKeypress(key => {
+      if (
+        (key && key.ctrl && key.name == 'c') ||
+        key.name === 'enter' ||
+        key.name === 'return'
+      ) {
+        onTerminate();
+      }
+    });
```
adamawang, line 250: Replace old keypress handler with util

packages/cli/commands/sandbox/sync.js
```diff
@@ -0,0 +1,298 @@
...
+    // Poll sync task status to show progress bars
+    await pollSyncTaskStatus(parentAccountId, initiateSyncResponse.id);
+
+    logger.log('');
+    spinnies.add('syncComplete', {
+      text: i18n(`${i18nKey}.polling.syncing`),
+    });
+    spinnies.succeed('syncComplete', {
+      text: i18n(`${i18nKey}.polling.succeed`),
+    });
```
adamawang, line 263: I had to do this add + succeed here to get the "success" spinnies message to show after the poll func. The bar chart rendering was causing issues if I put the spinnies.add above the poll

packages/cli-lib/errorHandlers/apiErrors.js
brandenrodgers: Nice 👍

brandenrodgers (approved):
lgtm 👍

PR #815, opened 2023-03-02
Move module validation to BE

jsines:
## Description and Context
- Moves module validation to backend API calls
- Pulls some theme validation logic into functions to share between the two
- 🪓  all the old local validation logic that's no longer needed


packages/cli/lib/marketplace-validate.js
```diff
@@ -0,0 +1,138 @@
...
+        });
+      }
+    }
+    return null;
+  };
+
+  logger.log(chalk.bold(i18n(`${i18nKey}.results.required`)));
+  displayResults(validationResults.results['REQUIRED']);
+  logger.log();
+  logger.log(chalk.bold(i18n(`${i18nKey}.results.recommended`)));
```
jsines, line 127: @ssethia2 your API definition says this is not included for MVP, should we remove this? It was in for the theme validator already

packages/cli/lib/marketplace-validate.js
```diff
@@ -0,0 +1,138 @@
...
+
+const pollForValidationFinish = async (accountId, validationId) => {
+  try {
+    const checkValidationStatus = async () => {
+      const validationStatus = await getValidationStatus(accountId, {
+        validationId,
+      });
+
+      if (validationStatus === 'REQUESTED') {
+        await new Promise(resolve => setTimeout(resolve, SLEEP_TIME));
```
jsines, line 37: @ssethia2 is the Kafka producer ready for this yet? We can switch to that when it is

kemmerle:
Hi Jessica! I just reviewed your code, and everything is looking good to me. I especially appreciate that you've consolidated logic between theme and module validation. 

I had two minor comments from a design/UX perspective:

1) You must use the module's full path, eg. `test-module.module`, in the `<src>` parameter. This isn't an issue, but I think it might be helpful to include this in the command's description now, especially since we're validating modules uploaded to the customer's hubspot account, where we don't display the `.module` file extension.

2) We don't seem to be returning anything under `Recommended validation results` if there are no errors; could we show the same `[SUCCESS] No errors` message we display under `Required validation results`? Pic below:

Screenshot 2023-03-02 at 4 17 40 PM

packages/cli/lib/marketplace-validate.js
ssethia2: Can we just iterate through the top level response keys and show whatever exists instead of reading explicitly? That way we only have to worry about controlling the response on the BE.

packages/cli/lib/marketplace-validate.js
ssethia2: We have that planned as a low priority issue, but thanks for calling that out, I'll update this when we eventually do move to using that

jsines:
>You must use the module's full path, eg. test-module.module, in the <src> parameter. This isn't an issue, but I think it might be helpful to include this in the command's description now, especially since we're validating modules uploaded to the customer's hubspot account, where we don't display the .module file extension.

Good callout! Added a blurb about that.

kemmerle (approved):
Tested everything out again, and it's working great. Let's :shipit:

PR #816, opened 2023-03-03
Add scripting options for hs sandbox delete

adamawang:
## Description and Context

Support scripting by adding a --force flag to skip all confirmations. Added additional error logging to handle missing data/flags when skipping confirmations. 

## Screenshots

Error when the account config does not have `parentAccountId` set:
Screenshot 2023-03-03 at 17 03 24
Error when no account is specified:
Screenshot 2023-03-03 at 17 02 51

Success:
Screenshot 2023-03-03 at 17 20 02

## TODO

I'll do a separate PR for sandbox create

### DIFF
--- a/packages/cli/commands/sandbox/delete.js
+++ b/packages/cli/commands/sandbox/delete.js
@@ -48,10 +48,17 @@ exports.handler = async options => {
       accountPrompt = await deleteSandboxPrompt(config);
     } else {
       // Account is required, throw error if force flag is present and no account is specified
+      logger.log('');
       logger.error(i18n(`${i18nKey}.failure.noAccount`));
       process.exit(EXIT_CODES.ERROR);
     }
+    if (!accountPrompt) {
+      logger.log('');
+      logger.error(i18n(`${i18nKey}.failure.noSandboxAccounts`));
+      process.exit(EXIT_CODES.ERROR);
+    }
   }
+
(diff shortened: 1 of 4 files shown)


packages/cli/lib/prompts/sandboxesPrompt.js
```diff
@@ -17,6 +17,18 @@ const mapSandboxAccountChoices = portals =>
...
+    .filter(
+      p => p.sandboxAccountType === null || p.sandboxAccountType === undefined
+    )
+    .map(p => {
+      return {
+        name: `${p.name} (${p.portalId})`,
+        value: p.name || p.portalId,
+      };
+    });
+
```
adamawang, line 31: Edge case bug fix for when a config entry does not have a parentAccountId. It prompts you for a parent account, but previously it would show only sandbox entries which is incorrect.

packages/cli/commands/sandbox/delete.js
```diff
@@ -178,6 +198,9 @@ exports.builder = yargs => {
     describe: i18n(`${i18nKey}.options.account.describe`),
     type: 'string',
   });
+  yargs.option('force', {
+    type: 'boolean',
+  });
```
adamawang, line 203: For these scripting options, do we need a descriptor since its more of a hidden thing?

brandenrodgers (approved):
👍

PR #817, opened 2023-03-15
Fix package versions

jsines:
Made a goof when running prerelease and accidentally ran the command on a feature branch, this should correct the version numbers to the current published version


brandenrodgers (approved):
👌

PR #818, opened 2023-03-15
Guard release commands to only run on master branch

jsines:
## Description and Context

Adds a check before publish commands run to ensure that a non-master branch isn't checked out.


brandenrodgers (approved):
Great idea 👍

PR #819, opened 2023-03-16
Deliver modules separately

ssethia2:
## Description and Context
We have made some changes internally that require us to upload modules after all other assets have been uploaded.
This is done by delivering all other assets synchronously, before queueing up module uploads. 

## Who to Notify
cc @mattcoley @jmagnarelli-hs @jsines

### DIFF
--- a/packages/cli/commands/upload.js
+++ b/packages/cli/commands/upload.js
@@ -58,6 +58,56 @@ const logThemePreview = (filePath, accountId) => {
 };
 
 exports.handler = async options => {
+  async function uploadFiles(
+    accountId,
+    absoluteSrcPath,
+    dest,
+    {
+      mode,
+    },
+    options,
+    filePaths) {
+    return uploadFolder(
+      accountId,
+      absoluteSrcPath,
+      dest,
+      {
+        mode,
+      },
+      options,
+      filePaths
+    )
+      .then(results => {
+        if (!hasUploadErrors(results)) {
+          logger.success(
+            i18n(`${i18nKey}.success.uploadComplete`, {
+              dest,
+            })
+          );
+          logThemePreview(src, accountId);
+        } else {
+          logger.error(
+            i18n(`${i18nKey}.errors.someFilesFailed`, {
+              dest,
+            })
+          );
+          process.exit(EXIT_CODES.WARNING);
+        }
+      })
+      .catch(error => {
+        logger.error(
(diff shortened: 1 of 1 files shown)


packages/cli/commands/upload.js
```diff
@@ -58,6 +58,55 @@ const logThemePreview = (filePath, accountId) => {
 };
 
 exports.handler = async options => {
+  async function uploadFiles(
```
jsines, line 61: nit: since this inner function is being passed options and not using the variable from the outer function's scope it might be cleaner to separate it out and make `exports.handler` a bit shorter.

jsines (commented):
Code all looks good, one minor nit.

Also minor UX bit, we upload all the non-module files and then the success message is printed ("[SUCCESS] Uploading files to "-" in the Design Manager is complete
To preview this theme, visit..."), and then modules upload and another success message is printed. Might make sense to add some extra logging to make this behavior a little clearer. We could also pass a flag into options for whether to display the design previewer link or not.

What timeline were you looking at for rolling this change out?

packages/cli/commands/upload.js
```diff
@@ -204,45 +253,30 @@ exports.handler = async options => {
       absoluteSrcPath,
       options.convertFields
     );
-    uploadFolder(
+
+    const noModules = filePaths.filter(path => !path.includes('/modules/'));
```
brandenrodgers, line 257: We have to be careful when checking paths. Using slashes like this may result in bugs for certain filesystems that use backslash (like windows). 

You may be able to use some of the utils that we have like this one: https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/modules.js#L47

packages/cli/commands/upload.js
```diff
@@ -204,45 +253,30 @@ exports.handler = async options => {
...
+      absoluteSrcPath,
+      dest,
+      {
+        mode,
+      },
+      options,
+      noModules
+    );
+
+    const onlyModules = filePaths.filter(path => path.includes('/modules/'));
```
brandenrodgers, line 269: Not a blocker, but is there any easy way to build `onlyModules` and `noModules` without having to iterate `filePaths` twice?

jsines: Lodash has a `partition` that I think would work here

PR #822, opened 2023-03-28
Account for 429 error codes

kemmerle:
## Description and Context
@mattcoley recently refactored certain filemanager endpoints, so that they return a `429` status code when a customer hits a rate limit. This PR ensures that the CLI handles those errors gracefully. 

## Screenshots

Current error message:

Screenshot 2023-03-28 at 5 06 03 PM

New error message:

Screenshot 2023-03-28 at 5 07 13 PM

## Who to Notify

@camden11 @brandenrodgers @mattcoley

### DIFF
--- a/packages/cli-lib/errorHandlers/apiErrors.js
+++ b/packages/cli-lib/errorHandlers/apiErrors.js
@@ -167,6 +167,11 @@ function logApiStatusCodeError(error, context) {
         errorMessage.push(`The ${messageDetail} was not found.`);
       }
       break;
+    case 429:
+      errorMessage.push(
+        `The ${messageDetail} surpassed the rate limit. Retry in one minute.`
+      );
+      break;
     case 503:
       errorMessage.push(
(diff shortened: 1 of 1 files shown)


kemmerle:
@mattcoley, I'll update the summary and re-test, but this PR should cover 429 errors coming from any API endpoint 👍

kemmerle:
@mattcoley, I was testing out these changes with more commands that use `filemapper` API endpoints, and I discovered that we were calling the `logApiErrorInstance` function incorrectly in multiple places. Now we should have more detailed error messages for the affected commands, thanks!

camden11 (approved):
LGTM. Were all those `logApiErrorInstance` with `accountId` causing bugs before? Seems like they wouldhave

kemmerle:
@camden11, the `logApiErrorInstance` functions with `accountId` were returning, "An unknown error occurred." So not a bug per se, but certainly annoying for the customer.

PR #823, opened 2023-03-31
Adjust sandbox error messaging

adamawang:
## Description and Context

Sandboxes identified a few areas where error messaging could improve. 

- When a user hits their development sandbox limit (currently 1), we show an error based on whether or not they already have a development sandbox tied to the parent portal in the config. 
- If a user tries to interact with sandboxes and the parent portal PAK is invalid, a very generic message was thrown. Since the underlying logic of sandboxes is to use the parent portal PAK and not the sandbox PAK to interact with sandboxes, users may try to fix the sandbox PAK incorrectly. We intercept this message and display a more detailed version containing the portal that needs to be fixed. 

## Screenshots

Sandbox limit when dev sandbox is NOT in config:
Screenshot 2023-03-30 at 11 53 36

Sandbox limit when dev sandbox IS in the config (gives option to switch accounts with hs use):
Screenshot 2023-03-30 at 11 54 44

If the parent portal PAK is invalid, we intercept the error and show one specifically showing the portal that the PAK needs to be updated for:
Screenshot 2023-03-30 at 14 37 34

Note for this last one - the `HubSpotAuthError` class is set with a link that points to [designers.hubspot.com](https://github.com/HubSpot/hubspot-cli/blob/b441e5eb014e27633c811a54cc7f3bdcea87b6a5/packages/cli-lib/personalAccessKey.js#L38). Is this something that is still valid? 

## TODO

n/a

## Who to Notify

@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli/commands/sandbox/delete.js
+++ b/packages/cli/commands/sandbox/delete.js
@@ -148,7 +148,7 @@ exports.handler = async options => {
     logger.log('');
     logger.success(
       i18n(deleteKey, {
-        account: getAccountConfig(accountConfig),
+        account: account || accountPrompt.account,
         sandboxHubId: sandboxAccountId,
       })
     );
--- a/packages/cli/commands/sandbox/delete.js
+++ b/packages/cli/commands/sandbox/delete.js
@@ -14,7 +14,7 @@ const {
 const { logErrorInstance } = require('@hubspot/cli-lib/errorHandlers');
 const { deleteSandbox } = require('@hubspot/cli-lib/sandboxes');
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
-const { getConfig, getEnv } = require('@hubspot/cli-lib');
+const { getConfig, getEnv, getAccountConfig } = require('@hubspot/cli-lib');
 const { deleteSandboxPrompt } = require('../../lib/prompts/sandboxesPrompt');
 const {
   removeSandboxAccountFromConfig,
@@ -30,6 +30,8 @@ const { ENVIRONMENTS } = require('@hubspot/cli-lib/lib/constants');
 const {
   isSpecifiedError,
 } = require('@hubspot/cli-lib/errorHandlers/apiErrors');
+const { HubSpotAuthError } = require('@hubspot/cli-lib/lib/models/Errors');
+const { getAccountName } = require('../../lib/sandboxes');
 
 const i18nKey = 'cli.commands.sandbox.subcommands.delete';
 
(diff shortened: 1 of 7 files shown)


packages/cli/commands/sandbox/delete.js
```diff
@@ -170,7 +173,20 @@ exports.handler = async options => {
       sandboxAccountId
     );
 
-    if (
+    if (err instanceof HubSpotAuthError) {
+      // Intercept invalid key error
+      // This command uses the parent portal PAK to delete a sandbox, so we must specify which account needs a new key
+      const regex = /\bYour personal access key is invalid\b/;
```
kemmerle, line 179: This seems like an OK fix for now, but would it be possible to talk to the BE and see whether they can return the correct errors?

packages/cli/lib/sandboxes.js
```diff
@@ -35,6 +35,30 @@ function getAccountName(config) {
...
+      portal.sandboxAccountType &&
+      portal.sandboxAccountType === 'DEVELOPER'
+    ) {
+      return true;
+    }
+  }
+  return false;
+}
+
+function getDevSandboxLimit(message) {
```
kemmerle, line 55: Again, I'm ok with this solution for now, but is there any way to return the sandbox limit from the BE? This function seems pretty brittle, especially if the BE decides to change the error message.

adamawang: Yep let me check with BE to see if we can add an extra prop or something to pull from

kemmerle (approved):
I reviewed the code and tested out the PR, and everything is WAD. 

My one concern is that we're relying a lot on the content of the error messages, which I assume can change with little notice. Would it be possible to work with the BE team in charge of these endpoints to come up with less brittle functions in the future?

packages/cli/commands/sandbox/delete.js
kemmerle: Ok, interesting. Could you leave a comment here alerting anyone who changes this text that it will impact your code? I know we've got a couple of designers doing a content overview of the CLI, so we might have some changes coming up.

kemmerle: ^^ This is assuming there's no way to intercept the error at the point where the error message originates.

adamawang: Good call out - I would've ideally wanted to intercept by error instance AND the status code, but we're not currently passing that through... potentially another area of improvement here. I'll leave a note 👍

packages/cli/lib/sandboxes.js
adamawang: Talked with BE and they've added a new context property we can pull from to lessen the dependency on a string!

packages/cli/commands/sandbox/delete.js
adamawang: Left a note here for anyone making changes coming up as well

PR #824, opened 2023-04-03
Create `hs project add` command 

kemmerle:
## Description and Context
See [internal HubSpot issue](https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/371) for a full list of documents and related issues and PRs.

This PR adds an `hs project add` command that creates a sample component within a HubSpot project folder. It contains two options:
 
`--type`: The component type (determines what component will be downloaded from the `hs-project-component` repo). 
`--name`: The component's name

I have also generated the command prompts from the `config.json` file located in the `hubspot-project-components` repo. 

## Screenshots

The new command with full prompts: 

Screenshot 2023-04-07 at 12 58 10 PM

Component successfully added:

Screenshot 2023-04-07 at 1 09 34 PM

## TODO

- [x] Address Mark's feedback and refactor the command. 
- [x] Re-add the check to ensure that the location is within a project folder. 

## Who to Notify

@brandenrodgers @camden11 
@markhazlewood, if you have any suggestions on copy, I'd love to hear them. Will also link to this on Slack to see if Marisa has any thoughts (couldn't find her GitHub username).

### DIFF
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -164,7 +164,7 @@ async function downloadGitHubRepoContents(
       contentPath
     );
 
-    const downloadContent = async contentPiece => {
+    const downloadContentRecursively = async contentPiece => {
       const { path: contentPiecePath, download_url } = contentPiece;
       const downloadPath = path.join(
         dest,
@@ -182,30 +182,30 @@ async function downloadGitHubRepoContents(
         `Downloading content piece: ${contentPiecePath} from ${download_url} to ${downloadPath}`
       );
 
-      return fetchGitHubRepoContentFromDownloadUrl(downloadPath, download_url, {
-        headers: { ...DEFAULT_USER_AGENT_HEADERS, ...GITHUB_AUTH_HEADERS },
-      });
+      if (contentPiece.type === 'dir') {
+        const contentPath = contentPiece.path;
+        const innerDirContent = await getGitHubRepoContentsAtPath(
+          repoName,
+          contentPath
+        );
+        return innerDirContent.map(downloadContentRecursively);
+      } else {
+        return fetchGitHubRepoContentFromDownloadUrl(
+          downloadPath,
+          download_url,
+          {
+            headers: {
+              ...DEFAULT_USER_AGENT_HEADERS,
+              ...GITHUB_AUTH_HEADERS,
+            },
+          }
+        );
+      }
     };
 
-    const downloadDirectoriesContent = async directoryContent => {
-      const contentPath = directoryContent.path;
-      const innerDirContent = await getGitHubRepoContentsAtPath(
-        repoName,
-        contentPath
-      );
-      return innerDirContent.map(downloadContent);
-    };
+    const contentPromises = contentsResp.map(downloadContentRecursively);
 
-    if (contentsResp.download_url) {
-      return downloadContent(contentsResp);
-    }
-
-    const directories = contentsResp.filter(content => content.type === 'dir');
-    const files = contentsResp.filter(content => content.type !== 'dir');
-    const directoryPromises = directories.map(downloadDirectoriesContent);
-    const contentPromises = files.map(downloadContent);
-
-    return Promise.all([directoryPromises, contentPromises]);
+    return Promise.all([contentPromises]);
   } catch (e) {
     if (e.statusCode === 404) {
       if (e.error.message) {
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -205,7 +205,7 @@ async function downloadGitHubRepoContents(
 
     const contentPromises = contentsResp.map(downloadContentRecursively);
 
-    return Promise.all([contentPromises]);
+    return Promise.all(contentPromises);
   } catch (e) {
     if (e.statusCode === 404) {
       if (e.error.message) {
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -203,9 +203,14 @@ async function downloadGitHubRepoContents(
       }
     };
 
-    const contentPromises = contentsResp.map(downloadContentRecursively);
-
-    return Promise.all(contentPromises);
+    let contentPromises;
+    if (Array.isArray(contentsResp)) {
+      contentPromises = contentsResp.map(downloadContentRecursively);
+      return Promise.all(contentPromises);
+    } else {
+      contentPromises = downloadContentRecursively(contentsResp);
+      return Promise.resolve(contentPromises);
+    }
   } catch (e) {
     if (e.statusCode === 404) {
       if (e.error.message) {
@@ -213,7 +218,7 @@ async function downloadGitHubRepoContents(
       }
     }
 
-    throw new Error(`Failed to fetch contents: ${e.error.message}`);
+    throw new Error(e);
   }
 }
 
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -165,7 +165,11 @@ async function downloadGitHubRepoContents(
     );
 
     const downloadContentRecursively = async contentPiece => {
-      const { path: contentPiecePath, download_url } = contentPiece;
+      const {
+        path: contentPiecePath,
+        download_url,
+        type: contentPieceType,
+      } = contentPiece;
       const downloadPath = path.join(
         dest,
         contentPiecePath.replace(contentPath, '')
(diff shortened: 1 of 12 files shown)


kemmerle:
Hi team!

I've refactored this PR based on Mark's suggestions above, and I'd be grateful for some feedback from the team. I have three main questions:

1) Currently, I'm determining whether to add a file or folder extension (`.js` to crm cards and `.functions` to serverless functions) based on the component type listed in the config. This isn't ideal, however, because we should be able to fully automate this process. Teams adding a component to the `hubspot-project-components` repo should be able to merge their changes there and have their component immediately available for download with this command. Any suggestions for how to accomplish this? I was wondering whether I should replace the `type` property in the `config.json` with an `extension` property. But I wanted to check in first. 

2) I have pretty minimal error handling right now, and I would love some feedback about how to handle errors more thoroughly/gracefully. If you spot any portions of code where I could handle an error, let me know. 

3) I need to add back in a check to ensure that the component is being added to a project folder. Where would be the best place to do this? I was thinking within the `createProjectComponent` function, but that seems pretty late to be performing a check. 

@markhazlewood @brandenrodgers @camden11

packages/cli/lib/projects.js
```diff
@@ -618,6 +620,30 @@ const logFeedbackMessage = buildId => {
   }
 };
 
+const createProjectComponent = async (component, name) => {
+  let componentName = name;
+
+  if (component.type === PROJECT_COMPONENT_TYPES.APP_FUNCTION) {
```
brandenrodgers, line 626: I would vote to force the repo config to specify the extension. That way the CLI never needs to know anything about the type of asset that's being created. Ideally we can keep the CLI as "dumb" as possible here so teams can self-serve by adding new components and other assets to the `hubspot-project-components` repo without having to make any changes here in the CLI.

packages/cli/lib/prompts/projectAddPrompt.js
```diff
@@ -0,0 +1,55 @@
...
+const { i18n } = require('@hubspot/cli-lib/lib/lang');
+
+const i18nKey = 'cli.lib.prompts.projectAddPrompt';
+
+const createTypeOptions = async () => {
+  const config = await fetchJsonFromRepository(
+    'hubspot-project-components',
+    'main/config.json'
+  );
+  return config.components;
```
brandenrodgers, line 12: This is a similar suggestion to the one I had above. It's related to keeping the CLI blissfully unaware of any of the specifics of the config file in the `hubspot-project-components` repo. 

What if we could achieve the same functionality here, but in a more generic way without requiring a `components` key in the repo that the CLI needs to be aware of.

**Consider a flow along the lines of this:**
(This is just an idea to consider)

the repo is structured something like this:
```
config.json
components/
  config.json
  crm-card.js
  serverless-function.js
  app/
```
And the two configs look something along the lines of:
top level config
```json
{ "path": "components", "label": "Component" }
```
nested config
```json
{ "path": "app", "label": "App", "insertPath": "./", "isFile": false, "type": "PRIVATE_APP" }
```
This would enable us to run a command along the lines of `hs project add` without specifying any component or asset type. Then it would read from the top level config and use that to generate the prompt options for the user to select from. Once they select an option, you dive into the folder that they selected and you read the next config file and use that to dynamically generate the next prompt (and so on until the user selects an option that contains `insertPath`).

We would still support things like `hs project add component` because the command could be smart enough to know that it needs to immediately look inside of the "components" folder.

camden11 (commented):
Looked through and tested it out. I agree with both Branden's comments about higher level stuff, but code here all looks good.

Ran into some issues testing though. This might be a problem with the configs in `hs-project-component` rather than this PR, but when I tried to create an app or serverless function, I got an error. It seems like maybe the path to download the component to isn't accounting for the `src` directory.
Screenshot 2023-04-07 at 4 05 52 PM

kemmerle:
@camden11, just to check, did you pull down the latest version of the branch? I'm testing against my current branch, and everything is working.

packages/cli/commands/project/add.js
```diff
@@ -0,0 +1,60 @@
...
+    logErrorInstance(error);
+  }
+};
+
+exports.builder = yargs => {
+  yargs.options({
+    type: {
+      describe: i18n(`${i18nKey}.options.type.describe`),
+      type: 'string',
+      choices: PROJECT_COMPONENT_TEMPLATES.map(type => type.label),
```
brandenrodgers, line 49: I think for now we can probably exclude the `choices` b/c it somewhat breaks the pattern of letting the repo dictate the available options. Hardcoding these in the repo would mean that we would have to keep this list in sync with whatever templates or boilerplate code gets added to the repo.

kemmerle:
This PR is ready for a re-review.

After a lot of discussion last week, we determined that we don't yet have the ability to add items like a CRM card to a project and "hook them up" (ie. change all the necessary config files), so that they worked automatically. 

For the first iteration, we've decided to focus on apps that can be automatically added to the `src` directory of a project. Currently we have two types of apps: 1) An app with a CRM card in the middle panel, and 2) An app with a CRM card in the side panel. I adapted both apps from the [ui-extensions-example](https://github.com/HubSpot/ui-extensions-examples) repo. 

I've also refactored the `hs project create` command to download the getting started template from the `hs-project-component` repo, so we can consolidate repos and make it easier for other teams to add app and project templates. 

One note: There is no easy way to download a directory with subdirectories from the GitHub API (@brandenrodgers and I spent some time exploring the docs together), so I've created a function to download directories recursively.

packages/cli/lib/projects.js
```diff
@@ -121,19 +120,18 @@ const createProjectConfig = async (projectPath, projectName, template) => {
     }`
   );
 
-  if (template === 'no-template') {
+  if (template.name === 'no-template') {
```
camden11, line 123: I think you'll need to update this here for this to work https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/project/download.js#L61

camden11 (commented):
This looks great! About to test it out, but in the meantime I noticed one potential bug

packages/cli/lib/projects.js
kemmerle: Ooh, thanks so much for catching this!

kemmerle: Ooh, thanks so much for catching this!

packages/cli/lib/projects.js
```diff
@@ -618,6 +616,32 @@ const logFeedbackMessage = buildId => {
   }
 };
 
+const createProjectComponent = async (component, name) => {
+  const i18nKey = 'cli.commands.project.subcommands.add';
+  let componentName = name;
+
+  const directoryPath = path.resolve(getCwd());
+  const configInfo = await getProjectConfig(directoryPath);
```
brandenrodgers, line 624: You don't need to pass `directoryPath` in here. This util will use `CWD` by default if no path is passed in.

packages/cli/lib/projects.js
```diff
@@ -618,6 +616,32 @@ const logFeedbackMessage = buildId => {
...
+
+  if (!configInfo.projectDir && !configInfo.projectConfig) {
+    logger.error(i18n(`${i18nKey}.error.locationInProject`));
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  const src = `${configInfo.projectConfig.srcDir}/`;
+
+  const componentPath = path.resolve(
```
brandenrodgers, line 633: I think this should be:
```js
const componentPath = path.join(
  configInfo.projectDir,
  configInfo.projectConfig.srcDir,
  component.insertPath,
  componentName
);
```
My reasoning is
`configInfo.projectDir` is the absolute path to the project directory
`configInfo.projectConfig.srcDir` is the path to the src directory, relative to the project root dir
`component.insertPath` is the  path where the component should be inserted, relative to the src dir
`componentName` is the name of the component


We can use `path.join` here instead of `path.resolve` because of the fact that `configInfo.projectDir` is already an absolute path ([more context](https://stackoverflow.com/questions/35048686/whats-the-difference-between-path-resolve-and-path-join))

packages/cli-lib/github.js
```diff
@@ -182,26 +182,43 @@ async function downloadGitHubRepoContents(
         `Downloading content piece: ${contentPiecePath} from ${download_url} to ${downloadPath}`
       );
 
-      return fetchGitHubRepoContentFromDownloadUrl(downloadPath, download_url, {
-        headers: { ...DEFAULT_USER_AGENT_HEADERS, ...GITHUB_AUTH_HEADERS },
-      });
+      if (contentPiece.type === 'dir') {
+        const contentPath = contentPiece.path;
```
brandenrodgers, line 186: Nit, it looks like you've already destructured this `path` field [here](https://github.com/HubSpot/hubspot-cli/pull/824/files#diff-e33aaeaafb6096b335e754b56e4dbfab3afa2f136b6ad3f4e448ced2493e8fc9R168).

You could also destrcuture `type` for consistency

packages/cli/lib/projects.js
```diff
@@ -618,6 +616,32 @@ const logFeedbackMessage = buildId => {
...
+
+  const directoryPath = path.resolve(getCwd());
+  const configInfo = await getProjectConfig(directoryPath);
+
+  if (!configInfo.projectDir && !configInfo.projectConfig) {
+    logger.error(i18n(`${i18nKey}.error.locationInProject`));
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  const src = `${configInfo.projectConfig.srcDir}/`;
```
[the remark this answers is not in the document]
kemmerle: Note: Won't need this if we use `path.join` below.

camden11 (approved):
LGTM!

PR #825, opened 2023-04-04
Add hook for callback after initial upload in watch

jsines:
## Description and Context
Adds optional argument to watch function in `cli-lib` for a callback after the initial upload completes. This allows the VSCode extension to add a hook for informing the user that the initial upload has completed and for the remote file system display to refresh the view with the uploaded folder. See that PR here: https://github.com/HubSpot/hubspot-cms-vscode/pull/216


kemmerle (approved):
LGTM 🚢

PR #827, opened 2023-04-06
Adding `hs project dev` local dev command

brandenrodgers:
## Description and Context

This is WIP, but I wanted to post what I have so far. I'm trying to configure things so that we can share the upload polling work between this command, the upload command, and the watch command. I had to move a few things around so I figured it would be better to stop here to get some early feedback on it.

The Spinnies package was giving me a lot of trouble. You run into all sorts of issues when you try to do things like:
- Initialize and use two separate Spinnies instances
- Attempt to console log while a Spinnies instance is currently in use and tracking something

To get around this I added a wrapper around the Spinnies lib that helps manage the Spinnies instance across files without requiring us to pass it around as an argument everywhere.

The two main changes in this PR are:
- The addition of the SpinniesManager class
- The addition of the `pollProjectBuildAndDeploy` function which allows us to share polling logic between the different project commands.


packages/cli-lib/lang/en.lyaml
```diff
@@ -907,8 +898,33 @@ en:
         options:
           options:
             describe: "Options to pass to javascript fields files"
-
     lib:
+      projects:
```
brandenrodgers, line 902: I cleaned up the lang file a little bit. There were a handful of keys in here that weren't in the right section. We have a dedicated `lib` category for every util in the lib folder.

packages/cli/commands/project/upload.js
```diff
@@ -43,81 +42,34 @@ exports.handler = async options => {
 
   await ensureProjectExists(accountId, projectConfig.name, { forceCreate });
 
-  const startPolling = async (tempFile, buildId) => {
```
brandenrodgers, line 46: I ported this polling logic into the projects lib file so the dev command can use it too.

packages/cli/lib/SpinniesManager.js
```diff
@@ -0,0 +1,42 @@
+const Spinnies = require('spinnies');
+
+// Allows us to maintain a single instance of spinnies across multiple files
+class SpinniesManager {
```
brandenrodgers, line 4: The comment explains it, but this allows us to maintain a singe instance of Spinnies without having to pass the instance around as function arguments. This is useful because we have util files that instantiate and use Spinnies that can conflict with the command-level spinnies instances. 

I also overrode the `add` function to allow for a special option `isParent`. This lets us automatically nest future Spinnies lines under a parent. The spinnies api is pretty simple (only like 8 available functions to call), so it feels like it's ok to make this abstraction. It will also allow us to override other functions down the road if we want more control.

The requirement to use this new wrapper is to instantiate Spinnies with `SpinniesManager.init(options)` rather than `new Spinnies(options)`. I didn't update all the usages of spinnies with this new init pattern, but I think we should be able to safely do that later if we want to.

packages/cli/lib/projects.js
```diff
@@ -304,61 +308,142 @@ const uploadProjectFiles = async (
...
   const output = fs.createWriteStream(tempFile.name);
   const archive = archiver('zip');
 
-  output.on('close', async function() {
-    logger.debug(
-      i18n(`${i18nKey}.debug.compressed`, {
-        byteCount: archive.pointer(),
-      })
-    );
+  const result = new Promise(resolve =>
```
brandenrodgers, line 419: This `handleProjectUpload` function now returns a promise. I wanted to be able to wait until the callbackFunc had returned the necessary data before continuing. This enables us to make decisions based on the successes/failures that can happen in callbackFunc. There's an example of that [here](https://github.com/HubSpot/hubspot-cli/pull/827/files#diff-1989a521c805539a7ae5e60fa3f7b0f79d83106bbb225045cf33f831abf97757R45).

packages/cli/lib/SpinniesManager.js
```diff
@@ -0,0 +1,42 @@
...
+    const originalIndent = rest.indent || 0;
+
+    this.spinnies.add(key, {
+      ...rest,
+      indent: this.indentNext ? originalIndent + 1 : originalIndent,
+    });
+
+    if (isParent) {
+      this.indentNext = true;
+    }
```
camden11, line 38: This doesn't need to change now, but would be nice to be able to use the `isParent` functionality with something like the nested subcomponent spinners, which have multiple parents/children and some semi-complex indentation logic. I think with the way this is set up, you'd have to call `init` multiple times to reset the indent

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -0,0 +1,33 @@
...
+const { uiAccountDescription } = require('../ui');
+
+const i18nKey = 'cli.lib.prompts.folderOverwritePrompt';
+
+const selectTargetAccountPrompt = accountId => {
+  return promptUser([
+    {
+      name: 'targetAccountId',
+      type: 'list',
+      message: i18n(`${i18nKey}.prompt.account`),
```
camden11, line 12: Got a missing lang property error here when running

camden11 (commented):
Screenshot 2023-04-10 at 12 56 35 PM
A few things I noticed when running this:
* It starts uploading immediately without waiting for the prompt
* It seems like there's a bug with logging errors, probably related to the new polling helper. Both `dev` and `upload` commands don't log any errors

kemmerle:
In general, I think the code is looking good. I particularly appreciate your breakdown of the new `dev` command, since it made it much easier to understand and review the code.

I think my biggest overall suggestion would be concerning the `spinnies` library. Since it's causing so many issues, and now we're being forced to silence logs because of the library's behavior, would it make sense to replace the `spinnies` library in a separate PR? I remember that Anthony chose the library to replace `ora`, but it looks like the package isn't being maintained (last release was in 2019), and there's not many people using it. 

`oclif` recommends the [listr package](https://www.npmjs.com/package/listr) for spinners. That does have 2 million+ weekly downloads but also isn't regularly maintained (last release was also in 2019). I'm puzzled why we moved away from `ora` for spinners, since it seems the most popular (15 million+ weekly downloads) and the best maintained (last release 3 weeks ago).

brandenrodgers:
> A few things I noticed when running this: * It starts uploading immediately without waiting for the prompt * It seems like there's a bug with logging errors, probably related to the new polling helper. Both `dev` and `upload` commands don't log any errors

@camden11 thanks, I fixed both of these 👍

kemmerle (commented):
Hi @brandenrodgers! 

I looked at the changes you made to the code, and they all make sense to me! I also tested out the command locally once again, and I noticed some awkwardness:

Screenshot 2023-04-14 at 5 11 24 PM

With the prompt, `[--silent] Create new project in accountName?`, am I supposed to be able to choose yes or no? Is that what `silent` means? I'm never given the option, and instead, the dev server starts running automatically. I'm wondering whether I'm missing something here, since I know the command isn't complete yet. 

Otherwise, everything seems to be working correctly for me.

PR #829, opened 2023-04-14
V2 of `hs sandbox create`

adamawang:
## Description and Context

To bring the CLI closer to feature parity with the UI, sandboxes will now support the creation of standard sandboxes with the `hs sandbox create` command. A new flow has been designed where users will be prompted for a sandbox type before being prompted for a name. 

- Support creation of standard sandboxes with a picker, one of standard or development
  - Can use new `--type` flag to specify which type
  - Update to use sandbox hubs v2 API
  - Using v2 API, PAKs are generated automatically on BE and CLI can bypass the prompt to generate and copy a PAK in the browser. 
  - After creating a standard sandbox, we now prompt the user if they want to sync assets
- Duplicate name checking for items existing in config
  - Removal of the second name prompt, where sandbox name and config name were different. We now automatically transform the sandbox name to a config friendly version and save (with duplicate name check). 
- Updated lang file to support both standard and development sandboxes, and rearranged various messaging after design review with PD
- Extraction of sandbox create logic into util. Can be called in other commands such as `hs project dev`. 
- Extraction of sandbox sync logic into util. Can be called in other commands such as `hs sandbox create`.
- Continue to support `-f` or `--force` flags to skip all confirmation prompts for scripting.
- Add fix for sync polling where items would hit 100% before the sync task completed

## TODO

Auto PAK generation is not yet complete on BE, still have yet to remove PAK prompt from sandbox create flow.

### DIFF
--- a/packages/cli/commands/sandbox/sync.js
+++ b/packages/cli/commands/sandbox/sync.js
@@ -9,7 +9,6 @@ const { trackCommandUsage } = require('../../lib/usageTracking');
 const { logger } = require('@hubspot/cli-lib/logger');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
-const { ENVIRONMENTS } = require('@hubspot/cli-lib/lib/constants');
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 const { getAccountConfig, getEnv } = require('@hubspot/cli-lib');
 const { getHubSpotWebsiteOrigin } = require('@hubspot/cli-lib/lib/urls');
@@ -22,6 +21,7 @@ const {
   STANDARD_SANDBOX,
 } = require('../../lib/sandboxes');
 const { syncSandbox } = require('../../lib/sandbox-sync');
+const { getValidEnv } = require('@hubspot/cli-lib/lib/environment');
 
 const i18nKey = 'cli.commands.sandbox.subcommands.sync';
 
@@ -34,7 +34,7 @@ exports.handler = async options => {
   const { force } = options; // For scripting purposes
   const accountId = getAccountId(options);
   const accountConfig = getAccountConfig(accountId);
-  const env = getEnv(accountId) === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
+  const env = getValidEnv(getEnv(accountId));
 
   trackCommandUsage('sandbox-sync', null, accountId);
 
--- a/packages/cli/commands/sandbox/sync.js
+++ b/packages/cli/commands/sandbox/sync.js
@@ -7,26 +7,16 @@ const {
 } = require('../../lib/commonOpts');
 const { trackCommandUsage } = require('../../lib/usageTracking');
 const { logger } = require('@hubspot/cli-lib/logger');
-const Spinnies = require('spinnies');
-const { initiateSync } = require('@hubspot/cli-lib/sandboxes');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
-const { logErrorInstance } = require('@hubspot/cli-lib/errorHandlers');
 const { ENVIRONMENTS } = require('@hubspot/cli-lib/lib/constants');
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 const { getAccountConfig, getEnv } = require('@hubspot/cli-lib');
 const { getHubSpotWebsiteOrigin } = require('@hubspot/cli-lib/lib/urls');
 const { promptUser } = require('../../lib/prompts/promptUtils');
 const { uiLine } = require('../../lib/ui');
-const {
-  getAccountName,
-  getAvailableSyncTypes,
-  pollSyncTaskStatus,
-} = require('../../lib/sandboxes');
-const {
-  isMissingScopeError,
-  isSpecifiedError,
-} = require('@hubspot/cli-lib/errorHandlers/apiErrors');
+const { getAccountName } = require('../../lib/sandboxes');
+const { syncSandbox } = require('../../lib/sandbox-sync');
 
 const i18nKey = 'cli.commands.sandbox.subcommands.sync';
 
@@ -39,9 +29,7 @@ exports.handler = async options => {
   const { force } = options; // For scripting purposes
   const accountId = getAccountId(options);
   const accountConfig = getAccountConfig(accountId);
-  const spinnies = new Spinnies({
-    succeedColor: 'white',
-  });
+  const env = getEnv(accountId) === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
 
   trackCommandUsage('sandbox-sync', null, accountId);
 
@@ -51,10 +39,8 @@ exports.handler = async options => {
     accountConfig.sandboxAccountType === undefined ||
     accountConfig.sandboxAccountType === null
   ) {
-    trackCommandUsage('sandbox-sync', { successful: false }, accountId);
-
     logger.error(i18n(`${i18nKey}.failure.notSandbox`));
-
+    trackCommandUsage('sandbox-sync', { successful: false }, accountId);
     process.exit(EXIT_CODES.ERROR);
   }
 
@@ -67,6 +53,7 @@ exports.handler = async options => {
         sandboxName: getAccountName(accountConfig),
       })
     );
+    trackCommandUsage('sandbox-sync', { successful: false }, accountId);
     process.exit(EXIT_CODES.ERROR);
   }
 
@@ -100,12 +87,13 @@ exports.handler = async options => {
         },
       ]);
       if (!confirmed) {
+        trackCommandUsage('sandbox-sync', { successful: false }, accountId);
         process.exit(EXIT_CODES.SUCCESS);
       }
     }
   } else if (isStandardSandbox) {
     const standardSyncUrl = `${getHubSpotWebsiteOrigin(
-      getEnv(accountId) === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD
+      env
     )}/sandboxes-developer/${parentAccountId}/sync?step=select_sync_path&id=${parentAccountId}_${accountId}`;
 
     logger.log(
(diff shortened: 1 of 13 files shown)


packages/cli/lib/sandbox-create.js
```diff
@@ -0,0 +1,333 @@
...
+ */
+const buildSandbox = async ({
+  name,
+  type,
+  accountConfig,
+  env,
+  allowEarlyTermination = true,
+  skipDefaultAccountPrompt = false,
+  skipSyncStandardSandbox = false,
+  force = false,
```
adamawang, line 58: Moved all sandbox creation logic into this util with the following params to tap into

packages/cli/lib/sandbox-create.js
```diff
@@ -0,0 +1,333 @@
...
+      logErrorInstance(err);
+    }
+    throw err;
+  }
+
+  let sandboxConfigName;
+
+  try {
+    // Response contains PAK, save to config here
+    sandboxConfigName = await saveSandboxToConfig(env, result, force);
```
adamawang, line 254: After creation, the response should include a PAK where we'll auto save it to the config, bypassing the PAK prompt.

packages/cli/lib/sandbox-create.js
```diff
@@ -0,0 +1,333 @@
...
+    const syncI18nKey = 'cli.commands.sandbox.subcommands.sync';
+    const sandboxAccountConfig = getAccountConfig(result.sandbox.sandboxHubId);
+    const handleSyncSandbox = async () => {
+      await syncSandbox({
+        accountConfig: sandboxAccountConfig,
+        parentAccountConfig: accountConfig,
+        env,
+        allowEarlyTermination,
+      });
+    };
```
adamawang, line 294: If creating a standard sandbox, prompt user if they want to sync assets in the creation flow. Development sandboxes do not get this prompt.

packages/cli/lib/sandbox-sync.js
```diff
@@ -0,0 +1,173 @@
...
+ * @param {Object} parentAccountConfig - Account config of parent portal
+ * @param {String} env - Environment (QA/Prod)
+ * @param {Boolean} allowEarlyTermination - Option to allow a keypress to terminate early
+ * @returns
+ */
+const syncSandbox = async ({
+  accountConfig,
+  parentAccountConfig,
+  env,
+  allowEarlyTermination = true,
```
adamawang, line 31: Separated sandbox sync logic into this util. We call this during the creation flow for standard sandboxes

packages/cli/lib/sandboxes.js
```diff
@@ -66,25 +80,51 @@ async function getAvailableSyncTypes(parentAccountConfig, config) {
...
+ * @returns {String} validName saved into config
+ */
+const saveSandboxToConfig = async (env, result, force = false) => {
+  // const configData = { env, personalAccessKey: result.personalAccessKey };
+  // TODO: Temporary, remove
+  const configData = await personalAccessKeyPrompt({
+    env,
+    account: result.sandbox.sandboxHubId,
+  });
+  // End temporary section
```
adamawang, line 96: NOTE: BE does not currently return PAK just yet, so we are keeping the old PAK prompt for now as a temporary measure.

packages/cli/lib/sandboxes.js
```diff
@@ -66,25 +80,51 @@ async function getAvailableSyncTypes(parentAccountConfig, config) {
...
+        logger.log(
+          i18n(
+            `cli.lib.prompts.enterAccountNamePrompt.errors.accountNameExists`,
+            { name: nameForConfig }
+          )
+        );
+        const { name: promptName } = await enterAccountNamePrompt(
+          nameForConfig
+        );
+        validName = promptName;
```
adamawang, line 121: Only prompt for another name if we detect a duplicate name in the config

packages/cli/lib/sandboxes.js
```diff
@@ -163,16 +193,18 @@ function pollSyncTaskStatus(accountId, taskId, syncStatusUrl) {
...
+    handleKeypress(key => {
+      if (
+        (key && key.ctrl && key.name == 'c') ||
+        key.name === 'enter' ||
+        key.name === 'return'
+      ) {
+        onTerminate();
+      }
+    });
+  }
```
adamawang, line 207: Optional keypress handlers in case we want to run a sync in a different flow where there are steps required after the sync.

packages/cli/lib/sandboxes.js
```diff
@@ -148,6 +177,7 @@ function pollSyncTaskStatus(accountId, taskId, syncStatusUrl) {
     'lead-flows': 'forms', // lead-flows are a subset of forms. We combine these in the UI as a single item, so we want to merge here for consistency.
   };
   const barInstances = {};
+  let progressCounter = {};
```
adamawang, line 180: Keep track of each types progress so we never hit 100% before sync type status is complete

packages/cli/commands/sandbox/sync.js
```diff
@@ -39,9 +34,7 @@ exports.handler = async options => {
   const { force } = options; // For scripting purposes
   const accountId = getAccountId(options);
   const accountConfig = getAccountConfig(accountId);
-  const spinnies = new Spinnies({
-    succeedColor: 'white',
-  });
+  const env = getEnv(accountId) === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
```
brandenrodgers, line 37: I realize there are other commands doing this same thing, but I think we can make use of some existing utils here. Specifically [getValidEnv](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lib/environment.js#L11).

You should be able to do `getValidEnv(getEnv(accountId))`

packages/cli/lib/sandbox-create.js
```diff
@@ -0,0 +1,332 @@
...
+  env,
+  allowEarlyTermination = true,
+  allowSyncAssets = true,
+  skipDefaultAccountPrompt = false,
+  force = false,
+}) => {
+  const spinnies = new Spinnies({
+    succeedColor: 'white',
+  });
+  const accountId = accountConfig.portalId;
```
brandenrodgers, line 64: I'd use the [getAccountId](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lib/config.js#L326) util. That will protect this against any future name changes for the ID field (i.e. `accountId`)

packages/cli/lib/sandbox-sync.js
```diff
@@ -0,0 +1,173 @@
...
+}) => {
+  const accountId = accountConfig.portalId;
```
brandenrodgers, line 33: Same suggestion here for the accountId util

packages/cli/lib/sandboxes.js
```diff
@@ -66,25 +80,51 @@ async function getAvailableSyncTypes(parentAccountConfig, config) {
...
+            { name: nameForConfig }
+          )
+        );
+        const { name: promptName } = await enterAccountNamePrompt(
+          nameForConfig
+        );
+        validName = promptName;
+      } else {
+        // Basic invalid name handling when force flag is passed
+        validName = nameForConfig + Math.floor(Math.random() * 10);
```
brandenrodgers, line 124: This could theoretically still result in a conflict though right? If the user runs this multiple times. Would something like a timestamp work better?

adamawang: ah yeah good call, this should be super edge case-y but timestamp would be way better

packages/cli/lib/sandbox-create.js
```diff
@@ -0,0 +1,337 @@
...
+const { createSandbox } = require('@hubspot/cli-lib/sandboxes');
+const { promptUser } = require('./prompts/promptUtils');
+const { syncSandbox } = require('./sandbox-sync');
+const {
+  setAsDefaultAccountPrompt,
+} = require('./prompts/setAsDefaultAccountPrompt');
+const { updateDefaultAccount } = require('@hubspot/cli-lib/lib/config');
+const { getValidEnv } = require('@hubspot/cli-lib/lib/environment');
+
+const i18nKey = 'cli.commands.sandbox.subcommands.create';
```
brandenrodgers, line 44: Another small organization nit, but can this lang key be updated to point to the dedicated `cli.lib` section of the lang file?

packages/cli/lib/sandbox-create.js
```diff
@@ -0,0 +1,337 @@
...
+          }
+        )
+      );
+    }
+  }
+
+  // If creating standard sandbox, prompt user to sync assets
+  if (allowSyncAssets) {
+    if (sandboxType === STANDARD_SANDBOX) {
+      const syncI18nKey = 'cli.commands.sandbox.subcommands.sync';
```
brandenrodgers, line 288: Can this also move to the `cli.lib` section of the lang file now that it's not in the command file anymore?

packages/cli/lib/sandboxes.js
```diff
@@ -133,26 +147,35 @@ const isTaskComplete = task => {
...
   const i18nKey = 'cli.commands.sandbox.subcommands.sync.types';
-  const multibar = new cliProgress.MultiBar(
-    {
-      hideCursor: true,
-      format: '[{bar}] {percentage}% | {taskType}',
-      gracefulExit: true,
-    },
-    cliProgress.Presets.rect
-  );
+  // TODO: Extract cli progress into util
```
brandenrodgers, line 168: Is this todo still relevant?

packages/cli/lib/sandbox-create.js
adamawang: Good call! Will get these updated now

packages/cli/lib/sandbox-create.js
```diff
@@ -0,0 +1,337 @@
...
+        );
+      } else {
+        const baseUrl = getHubSpotWebsiteOrigin(getValidEnv(getEnv(accountId)));
+        logger.error(
+          i18n(
+            `${i18nKey}.failure.limit.developer.${plural ? 'other' : 'one'}`,
+            {
+              accountName: accountConfig.name || accountId,
+              limit: devSandboxLimit,
+              link: `${baseUrl}/sandboxes-developer/${accountId}/development`,
```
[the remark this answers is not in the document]
adamawang: I think our copy was based around showing the full url, but lemme see if I can trim it and use this instead

brandenrodgers (approved):
Looks good! 🎉 (approved pending the few small comments I left earlier this morning)

packages/cli/lib/sandbox-create.js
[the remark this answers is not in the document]
adamawang: Hmm so I remember attempting this earlier and uiLink doesn't seem to work when injecting it into a string as a param. I've also tried adding it as a separate string for some of the messages that have the URL at the very end, but the prompts are not allowing links to open for some reason... I'll continue to play around with these, but will exclude the links from this PR for now

PR #830, opened 2023-04-17
Adding file watch + upload behavior to the Dev Mode command

brandenrodgers:
## Description and Context

This is a continuation of https://github.com/HubSpot/hubspot-cli/pull/827

This adds the file watching + auto-uploading behavior to the dev mode command.

The LocalDevManager class is where most of the logic lives. It will be in charge of handling everything related to the dev mode process. It will listen for file changes and decide how to move forward depending on which type of change was made. Currently it will behave exactly the same way that `hs project watch` works. Eventually we can tie in our local dev server logic into it to avoid running uploads for every file change.

## TODO

- [x] Better error handling
- [ ] Fine tune the output of the command


packages/cli/lib/LocalDevManager.js
```diff
@@ -0,0 +1,382 @@
...
+      filePath,
+      remotePath: path.relative(this.projectSourceDir, filePath),
+    };
+
+    const isSupportedChange = await this.notifyServers(changeInfo);
+
+    if (isSupportedChange) {
+      this.addChangeToStandbyQueue({ ...changeInfo, supported: true });
+      return;
+    } else if (this.preventUploads) {
```
camden11, line 223: Should this check for `preventUploads` first?

camden11 (commented):
Looks good so far - just a few questions! Also, this PR is meant to be merged back into your other local dev branch right? Looks like its based off master but there are a few changes here that I think were also made in your other branch

packages/cli/commands/project/dev.js
```diff
@@ -0,0 +1,213 @@
...
+  const accounts = getConfigAccounts();
+  let targetAccountId = options.accountId;
+
+  if (!targetAccountId) {
+    const {
+      targetAccountId: promptTargetAccountId,
+    } = await selectTargetAccountPrompt(accounts);
+
+    targetAccountId = promptTargetAccountId;
+  }
```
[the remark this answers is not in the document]
brandenrodgers: Yeah I felt a little gross doing it this way. I wasn't sure if the inquirer library supported multiple different return types from a single prompt question. I can poke around in the lib to see if there's a better way to be handling this.

brandenrodgers:
> Also, this PR is meant to be merged back into your other local dev branch right?

@camden11 ah yeah sorry about that. I fixed the base branch so the diff should be better now

packages/cli/lib/LocalDevManager.js
brandenrodgers: Good call. I'll fix this

brandenrodgers: done in [a4151f7](https://github.com/HubSpot/hubspot-cli/pull/830/commits/a4151f731e6d699d8edd834e2142bdc77d7f0add)

PR #831, opened 2023-04-21
Sandboxes: prompt user for contact records sync 

adamawang:
## Description and Context

Adds option to sync contact records when creating or syncing a sandbox. 

- Add user prompt to include or exclude object records from sync for both development sandboxes and standard sandboxes
- Refactored util function and moved most of the logic back to the command file. We will be handling automatic syncing in a separate file when I get to work on linking up `hs project dev` with these items. 
- We make an additional sandbox usage call to verify whether or not a user is at their portal limit. If they are, we error out immediately. 

## Screenshots

Screenshot 2023-04-26 at 13 30 05

Screenshot 2023-04-26 at 13 30 34

## TODO

PAK generation is still not ready on BE, waiting to implement that

### DIFF
--- a/packages/cli-lib/sandboxes.js
+++ b/packages/cli-lib/sandboxes.js
@@ -2,6 +2,7 @@
 const {
   createSandbox: _createSandbox,
   deleteSandbox: _deleteSandbox,
+  getSandboxUsageLimits: _getSandboxUsageLimits,
 } = require('./api/sandbox-hubs');
 const {
   initiateSync: _initiateSync,
@@ -38,6 +39,18 @@ async function deleteSandbox(parentAccountId, sandboxAccountId) {
     sandboxAccountId,
     ...resp,
   };
+}
+
+async function getSandboxUsageLimits(parentAccountId, sandboxAccountId) {
+  let resp;
+
+  try {
+    resp = await _getSandboxUsageLimits(parentAccountId, sandboxAccountId);
+  } catch (err) {
+    throw err;
+  }
+
+  return resp.usage;
 }
 
 async function initiateSync(fromHubId, toHubId, tasks, sandboxHubId) {
@@ -79,6 +92,7 @@ async function fetchTypes(accountId, toHubId) {
 module.exports = {
   createSandbox,
   deleteSandbox,
+  getSandboxUsageLimits,
   initiateSync,
   fetchTaskStatus,
   fetchTypes,
--- a/packages/cli/commands/sandbox/create.js
+++ b/packages/cli/commands/sandbox/create.js
@@ -18,6 +18,8 @@ const {
   getAccountName,
   getAvailableSyncTypes,
   syncTypes,
+  STANDARD_SANDBOX,
+  getHasSandboxesByType,
 } = require('../../lib/sandboxes');
 const { getValidEnv } = require('@hubspot/cli-lib/lib/environment');
 const { logger } = require('@hubspot/cli-lib/logger');
@@ -29,6 +31,8 @@ const {
 const { promptUser } = require('../../lib/prompts/promptUtils');
 const { syncSandbox } = require('../../lib/sandbox-sync');
 const { logErrorInstance } = require('@hubspot/cli-lib/errorHandlers');
+const { getSandboxUsageLimits } = require('@hubspot/cli-lib/sandboxes');
+const { getHubSpotWebsiteOrigin } = require('@hubspot/cli-lib/lib/urls');
 
 const i18nKey = 'cli.commands.sandbox.subcommands.create';
 
@@ -57,12 +61,7 @@ exports.handler = async options => {
         sandboxName: accountConfig.name,
       })
     );
-    throw new Error(
-      i18n(`${i18nKey}.failure.creatingWithinSandbox`, {
-        sandboxType: getSandboxTypeAsString(accountConfig.sandboxAccountType),
-        sandboxName: accountConfig.name,
-      })
-    );
+    process.exit(EXIT_CODES.ERROR);
   }
 
   let typePrompt;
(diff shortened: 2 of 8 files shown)


packages/cli/lib/sandbox-sync.js
```diff
@@ -128,44 +139,46 @@ const syncSandbox = async ({
...
-    logger.log('');
-    logger.log('Sync progress:');
-    // Poll sync task status to show progress bars
-    await pollSyncTaskStatus(
-      parentAccountId,
-      initiateSyncResponse.id,
-      syncStatusUrl,
-      allowEarlyTermination
-    );
+  if (!skipPolling) {
```
adamawang, line 142: Option to skip polling, will be used when used with `hs project dev`

packages/cli/lib/sandbox-create.js
adamawang, line 1: Took another pass and ended up reworking the flow. I mistakenly included far too many complicated flows in the build sandbox util, and wanted to move it out back into the `hs sandbox create` command file to simplify things.

packages/cli/commands/sandbox/sync.js
```diff
@@ -143,12 +145,24 @@ exports.handler = async options => {
...
+    const availableSyncTasks = await getAvailableSyncTypes(
+      parentAccountConfig,
+      accountConfig
+    );
+    const syncTasks = await getSyncTypesWithContactRecordsPrompt(
+      accountConfig,
+      availableSyncTasks,
+      force
+    );
+
```
adamawang, line 157: Fetch and adjust sync tasks before passing into util

packages/cli/commands/sandbox/create.js
adamawang, line 1: As mentioned in the other comment, all the changes here are not new but just moved from the sandbox-create.js util file.

packages/cli/commands/sandbox/create.js
```diff
@@ -27,26 +47,221 @@ exports.handler = async options => {
...
+        }
+        trackCommandUsage('sandbox-create', { successful: false }, accountId);
+        process.exit(EXIT_CODES.ERROR);
+      }
+    }
+  } catch (err) {
+    logErrorInstance(err);
+    trackCommandUsage('sandbox-create', { successful: false }, accountId);
+    process.exit(EXIT_CODES.ERROR);
+  }
```
adamawang, line 170: We fetch usage limits for sandboxes, and if a portal has no available sandboxes (as per selected type), we exit early and display the following error:
Screenshot 2023-04-26 at 13 28 22
Screenshot 2023-04-26 at 13 28 35

packages/cli-lib/sandboxes.js
```diff
@@ -40,6 +41,18 @@ async function deleteSandbox(parentAccountId, sandboxAccountId) {
...
+async function getSandboxUsageLimits(parentAccountId, sandboxAccountId) {
+  let resp;
+
+  try {
+    resp = await _getSandboxUsageLimits(parentAccountId, sandboxAccountId);
+  } catch (err) {
+    throw err;
+  }
+
+  return resp.usage;
```
kemmerle, line 53: Seems a little strange to return `resp.usage` even if `resp` is undefined. Would it be possible to return something else like `null` in that case?

packages/cli/commands/sandbox/create.js
```diff
@@ -27,26 +47,221 @@ exports.handler = async options => {
...
+      })
+    );
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  let typePrompt;
+  let namePrompt;
+
+  if ((type && !sandboxTypeMap[type]) || !type) {
+    if (!force) {
```
kemmerle, line 71: Having a little trouble grasping this: What happens if `force` is true?

packages/cli/commands/sandbox/sync.js
kemmerle: OOC, what is the difference between `availableSyncTasks` and `syncTasks`? We only sometimes allow customers to sync contacts, is that correct? Why?

packages/cli/lib/sandboxes.js
```diff
@@ -81,6 +86,43 @@ async function getAvailableSyncTypes(parentAccountConfig, config) {
...
+ * @param {Boolean} skipPrompt - Option to skip contact records prompt and return all available sync tasks
+ * @returns {Array} Adjusted available sync task items
+ */
+const getSyncTypesWithContactRecordsPrompt = async (
+  accountConfig,
+  availableSyncTasks,
+  skipPrompt = false
+) => {
+  // Fetches sync types based on default account. Parent account required for fetch
+  let syncTasks = availableSyncTasks;
```
kemmerle, line 101: Why the need for this?

kemmerle:
Hi @adamawang, took a look over this while I was catching up on PR reviews and left you some comments.

packages/cli/commands/sandbox/sync.js
adamawang: Yes exactly. Long story short, we don't have a good way of syncing contacts multiple times due to ID changes, so we limit contact syncs to one for any sandbox. Our /types fetch will return a valid list of sync types, so if contacts have been synced already it won't be returned. Future plans are to change the sync behavior to support multiple syncs, but it may not be for a while

packages/cli-lib/sandboxes.js
adamawang: Good call! I'll update the fetch types response as well

packages/cli/commands/sandbox/create.js
adamawang: If a user were to run this command with the force option, we require both the type and a name for creation. If one or none of those are provided, we throw an error saying they must provide those options to proceed

packages/cli/lib/sandboxes.js
adamawang: For line 101, we filter the types based on the prompt on line 120. If no changes, we just return that set of tasks. Do you mean we can just ditch the var and filter on the param passed in?

packages/cli/commands/sandbox/create.js
adamawang: Would it be better to have defaults here? like if no type is provided, have a default selected and prefill a name?

packages/cli-lib/sandboxes.js
adamawang: Pushed changes for this to the new PR: https://github.com/HubSpot/hubspot-cli/pull/833

packages/cli/lib/sandboxes.js
kemmerle: > Do you mean we can just ditch the var and filter on the param passed in? 

Yes, that's what I was thinking, although this is really a nit. `syncTasks` has better readability, so I'm happy to keep it if you are!

adamawang: I've removed and pushed it into the latest branch #833 !

brandenrodgers (approved):
Looks good 👌

PR #833, opened 2023-04-26
`hs project dev` creates a new dev sandbox

adamawang:
## Description and Context

Plug in all the sandboxes utils and create a new development sandbox through the account picker. 

- User is prompted to enter a name, which we use to create a new dev sandbox. 
- Currently, the user is prompted to jump to the browser to grab their PAK. Sandboxes BE is working on getting that wrapped up this week. I'll remove this once we get that working
- The user will see a sandbox sync initiate, and a success message that the "Sandbox sync initiated.". 
  - The sync will run in the background and should complete rather quickly depending on how many object-schemas the user has. 
  - We also automatically sync up to 100 contacts and associations which should only take a minute or less. 

## Screenshots

Screenshot 2023-04-26 at 17 29 59


packages/cli/commands/project/dev.js
```diff
@@ -94,10 +107,42 @@ exports.handler = async options => {
...
-    logger.log(
-      'Creating new sandboxes is not supported yet. Use "hs sandbox create" and then run this command again.'
-    );
-    process.exit(EXIT_CODES.SUCCESS);
+    try {
+      await validateSandboxUsageLimits(accountConfig, DEVELOPER_SANDBOX, env);
+    } catch (err) {
+      logErrorInstance(err);
+      process.exit(EXIT_CODES.ERROR);
+    }
```
adamawang, line 115: We do a limit validation for the default account and throw if there are no development sandboxes available to create

packages/cli/lib/sandboxes.js
```diff
@@ -84,6 +94,128 @@ async function getAvailableSyncTypes(parentAccountConfig, config) {
...
+  return syncTasks;
+};
+
+/**
+ * @param {Object} accountConfig - Account config of sandbox portal
+ * @param {String} sandboxType - Sandbox type for limit validation
+ * @param {String} env - Environment
+ * @returns {null}
+ */
+const validateSandboxUsageLimits = async (accountConfig, sandboxType, env) => {
```
adamawang, line 140: Moved this into a util to use in multiple places

packages/cli/commands/project/dev.js
```diff
@@ -94,10 +107,42 @@ exports.handler = async options => {
...
+    try {
+      const { name } = await sandboxNamePrompt();
+      const { result } = await buildSandbox({
+        name,
+        type: DEVELOPER_SANDBOX,
+        accountConfig,
+        env,
+      });
```
adamawang, line 123: As mentioned in the description, this part still prompts the user to generate a PAK in a browser. Should be replaced later this week

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -25,18 +29,10 @@ const selectTargetAccountPrompt = async (accounts, nonSandbox = false) => {
...
-          targetAccountId: null,
-          chooseNonSandbox: false,
-          createNewSandbox: true,
-        },
-      },
       ...accounts.filter(isSandbox).map(accountConfig => {
         const accountId = getAccountId(accountConfig.name);
         return {
-          name: uiAccountDescription(accountId),
+          name: getAccountName(accountConfig),
```
adamawang, line 35: Updated to this util to show sandbox types

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -53,6 +49,16 @@ const selectTargetAccountPrompt = async (accounts, nonSandbox = false) => {
...
+    if (!isSandbox(defaultAccountConfig)) {
+      choices.unshift({
+        name: i18n(`${i18nKey}.createNewSandboxOption`),
+        value: {
+          targetAccountId: null,
+          chooseNonSandbox: false,
+          createNewSandbox: true,
+        },
+      });
+    }
```
adamawang, line 61: Still up in the air about this one, we may end up disabling this option instead with custom messaging

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -1,12 +1,16 @@
...
-const { isSandbox } = require('../sandboxes');
+const { isSandbox, getAccountName } = require('../sandboxes');
 const { getAccountId } = require('@hubspot/cli-lib');
 
 const i18nKey = 'cli.lib.prompts.projectDevTargetAccountPrompt';
 
-const selectTargetAccountPrompt = async (accounts, nonSandbox = false) => {
+const selectTargetAccountPrompt = async (
+  accounts,
+  defaultAccountConfig,
```
adamawang, line 11: Added an extra param for the sandbox check further down in this file

packages/cli/commands/project/dev.js
```diff
@@ -86,18 +101,50 @@ exports.handler = async options => {
...
+
+      targetAccountId = result.sandbox.sandboxHubId;
+
+      const sandboxAccountConfig = getAccountConfig(
+        result.sandbox.sandboxHubId
+      );
+      let syncTasks = await getAvailableSyncTypes(
```
camden11, line 132: Should this be a `const`?

packages/cli/lib/prompts/sandboxesPrompt.js
```diff
@@ -50,6 +50,26 @@ const sandboxNamePrompt = () => {
...
+        }
+        return accountNameExistsInConfig(val)
+          ? i18n(`${i18nKey}.name.errors.accountNameExists`, { name: val })
+          : true;
+      },
+      default: 'New development sandbox',
+    },
+  ]);
+};
+
```
camden11, line 72: This looks similar enough to the original `sandboxNamePrompt` that you could just pass a `isDevelopment` arg to it or something rather than making it its own function

packages/cli/lib/sandboxes.js
```diff
@@ -120,26 +130,114 @@ const getSyncTypesWithContactRecordsPrompt = async (
...
 const saveSandboxToConfig = async (env, result, force = false) => {
-  // const configData = { env, personalAccessKey: result.personalAccessKey };
-  // TODO: Temporary, remove once PAK is generated on BE
-  const configData = await personalAccessKeyPrompt({
-    env,
-    account: result.sandbox.sandboxHubId,
-  });
-  // End temporary section
+  let configData = { env, personalAccessKey: result.personalAccessKey };
+  if (!result.personalAccessKey || result.personalAccessKey === 'test') {
```
camden11, line 235: did you mean to leave this 'test' check in?

camden11 (commented):
This looks great, just had a few minor comments

packages/cli/lib/sandboxes.js
adamawang: Good catch - It was kinda necessary based on what we were returning from BE, but I think we're gonna get that keygen PR out today. Commits incoming

packages/cli/commands/project/dev.js
adamawang: Indeed it should! leftover from moving logic into that util

camden11 (approved):
Not sure if you also want to get a review from @brandenrodgers, but looks good to me

PR #834, opened 2023-04-28
Add projects list prompt to download command

kemmerle:
## Description and Context
I've added a prompt with a list of projects from which the customer can choose when using the `project download` command. 

An important note: I changed the functionality of this command a bit by making two positionals optional now (`name` and `dest`). Pros: This would bring the `download` command more in line with our current style guidelines. Cons: This would also constitute a breaking change and makes for a more complicated release. I'm wondering whether this is the best path forward. Would love some feedback, @brandenrodgers, @camden11. 

## Screenshots

Before:

Screenshot 2023-04-28 at 4 23 44 PM

After:

Screenshot 2023-04-28 at 4 24 13 PM
Screenshot 2023-04-28 at 4 25 14 PM

## TODO

 
- [x] I will possibly need to revise to bring back positional arguments. 

## Who to Notify

@camden11 @brandenrodgers

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -608,14 +608,13 @@ en:
             errors:
               downloadFailed: "Something went wrong downloading the project"
               projectNotFound: "Your project {{ projectName }} could not be found in {{ accountId }}"
-            positionals:
+            options:
+              buildNumber:
+                describe: "The build number to download"
               name:
                 describe: "The name of the project to download"
               dest:
                 describe: "Destination folder for the project"
-            options:
-              buildNumber:
-                describe: "The build number to download"
           open:
             describe: "Open available projects for the specified account"
             options:
@@ -1002,6 +1001,10 @@ en:
             nameRequired: "A project name is required"
             locationRequired: "A project location is required"
             invalidTemplate: "[--template] Could not find template {{ template }}. Please choose an available template."
+        downloadProjectPrompt:
+          selectProject: "Select a project to download:"
+          errors:
+            projectNotFound: "Your project {{ projectName }} could not be found in {{ accountId }}"
         projectAddPrompt:
           selectType: "[--type] Select your component type:"
           enterName: "[--name] Give your component a name: "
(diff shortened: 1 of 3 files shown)


packages/cli/commands/project/download.js
```diff
@@ -22,18 +22,24 @@ const {
...
+const {
+  downloadProjectPrompt,
+} = require('../../lib/prompts/downloadProjectPrompt');
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
 
 const i18nKey = 'cli.commands.project.subcommands.download';
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 
-exports.command = 'download <name> [dest]';
+exports.command = 'download';
```
brandenrodgers, line 33: This can be `download [--project]`. That's the format that we've been using to highlight command options that are semi-required. Semi-required options are options that can be omitted, but omitting them will result in a prompt.

Here's an example in [hs project deploy](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/project/deploy.js#L23). Also, as part of this pattern we're including the option in the prompt so that the user can quickly learn how to skip the prompt in the next run ([example](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli-lib/lang/en.lyaml#L1036)). I did a quick look and it seems like we're using `project` as the option in our other project commands.

packages/cli-lib/lang/en.lyaml
```diff
@@ -1002,6 +1001,10 @@ en:
             nameRequired: "A project name is required"
             locationRequired: "A project location is required"
             invalidTemplate: "[--template] Could not find template {{ template }}. Please choose an available template."
+        downloadProjectPrompt:
+          selectProject: "Select a project to download:"
+          errors:
+            projectNotFound: "Your project {{ projectName }} could not be found in {{ accountId }}"
```
brandenrodgers, line 1007: I like this pattern of validating the users input. Maybe we should add something to the end of this along the lines of "...not found in {{ accountId }}. Please select a valid project:"?

kemmerle: Nice, I like that pattern!

camden11:
Screenshot 2023-05-04 at 11 42 27 AM
Everything works as expected but I get a bunch of lang errors when running locally. Otherwise LGTM

camden11 (approved):

PR #835, opened 2023-05-03
Latest iteration on project dev mode command

brandenrodgers:
## Description and Context

This adds some new functionality to the command:
- The "prevent uploads" mode flow has been changed (see screenshot for new flow)
- I added a new concept to the spinnies manager class called `category`. This enables us to group spinnies instances into categories, which helps with bulk-removing entire categories from the console.
- I added a temporary option called `mockServers`. It'll force the dev mode command to pretend like there is a running server. It helps with testing out different flows.

## Screenshots

image (9)

## TODO

- [x] There's an annoying issue with the "view project in hubspot" link where long url's cause wrapping, which breaks the entire output of everything else in the console. I'm not sure what the fix is yet.


brandenrodgers:
Just to document it, I found the line that's causing issues for us with long urls being used as text in spinnies.

The Spinnies lib is [automatically breaking up the text](https://github.com/jbcarpanelli/spinnies/blob/d672dedcab8c8ce0f6de0bb26ca5582bf602afd7/utils.js#L68-L74) when it is longer than the width of the terminal window. AFAIK there's no way to get around this behavior. This is an issue because of the somewhat-hacky way that we're getting [text to render as a hyperlink in the console](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/ui.js#L42-L48). It's a little hacky, but it looks like it's actually the only way to get this behavior (looks like we grabbed it from [ansi-escapes](https://github.com/sindresorhus/ansi-escapes/blob/main/index.js#L105))

packages/cli/lib/LocalDevManager.js
```diff
@@ -265,23 +358,32 @@ class LocalDevManager {
     this.standbyChanges.push(changeInfo);
   }
 
-  async sendChanges({ event, filePath, remotePath }) {
+  async sendChanges(changeInfo) {
```
kemmerle, line 361: Nit: You could destructure changeInfo and make this slightly easier to read.

brandenrodgers: Good idea, done in [bbd694d](https://github.com/HubSpot/hubspot-cli/pull/835/commits/bbd694dec58a8f0cecad270b5f5135a0753efcb8) 👍

packages/cli/commands/project/dev.js
```diff
@@ -145,76 +146,69 @@ exports.handler = async options => {
     }
   }
 
-  // TODO programatically determine these values
-  const isNonSandboxAccount = false;
+  const isNonSandboxAccount = shouldTargetNonSandboxAccount;
+  // TODO programatically determine this
```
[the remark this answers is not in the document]
brandenrodgers: I updated the comment in [a50a9c7](https://github.com/HubSpot/hubspot-cli/pull/835/commits/a50a9c78a4a0a025ec9e04cd3c6d2601103536ba), but this is referring to the `isProjectUsingGitIntegration` variable. Right now it is hard-coded to `false`. I'm going to need to make a request to the BE to figure out if the project is actually using a github integration or not. If it is, we'll need to put Dev Mode into a state where it never uploads anything (b/c the git integration prevents any direct CLI uploads).

I'm going to do this in a follow up PR so it doesn't clutter up this one.

PR #836, opened 2023-05-03
Lazy load lang data, allow overriding of lang files

jsines:
## Description and Context
Updates `cli-lib/lib/lang.js` to wait until `i18n` is first called to load language data if it isn't present, instead of at import.

This allows for the VSCode extension to manually set the language data ahead of use. `loadLanguageFromYaml()` will error out when used in VSCode because `cli-lib` gets bundled with extension code using webpack, and the `en.lyaml` gets fetched directly from the filesystem. I've set up VSCode to keep and maintain a copy of `en.lyaml` so it can be bundled properly here: https://github.com/HubSpot/hubspot-cms-vscode/pull/226

This should hopefully just be a bandaid solution until migration is complete and we can add a build step on `cli-lib`'s end, but I've set it up to automatically download the `en.lyaml` of the resolved `cli-lib` version on precommit so it should be OK to stay as long as needed.

### DIFF
--- a/packages/cli-lib/lib/lang.js
+++ b/packages/cli-lib/lib/lang.js
@@ -68,6 +68,10 @@ const getTextValue = lookupDotNotation => {
 };
 
 const i18n = (lookupDotNotation, options = {}) => {
+  if (!languageObj) {
+    loadLanguageFromYaml();
+  }
+
   if (typeof lookupDotNotation !== 'string') {
     throw new Error(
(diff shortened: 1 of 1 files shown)


camden11 (approved):
LGTM. Also made an issue to make these changes in `hubspot-local-dev-lib` too so we don't forget whenever we end up doing that migration. Feel free to PR this there too, or I will when I have a chance

https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/391

PR #837, opened 2023-05-08
Add flag for custom repo 

kemmerle:
## Description and Context
In this PR, I've added a custom `--repoPath` flag to the `hs project create` command, so that customers can generate project templates from a GitHub repository of their choice.

Some notes: 

1) The `repoPath` flag must follow this pattern: `'<ownerName>/<repoName>'`. For example, the path to the repo `hubspot-project-components` owned by `HubSpot` is written as: `HubSpot/hubspot-project-components`. 

2) The repository must have a `config.json` file at its root that follows the pattern of the [config.json](https://github.com/HubSpot/hubspot-project-components/blob/main/config.json) file in the `hubspot-project-components` repo. All projects must be listed under the `projects` property in the config file.

## Screenshots

Screenshot 2023-05-09 at 12 38 23 PM

Erroring out with incorrect repo path:

Screenshot 2023-05-09 at 12 38 40 PM

## TODO

- [x] Reasonable error handling if customer inputs an incorrect `repoPath` flag with information about how to fix the config.json file. 

## Who to Notify

@brandenrodgers @camden11

### DIFF
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -135,13 +135,11 @@ async function cloneGitHubRepo(dest, type, repoName, sourceDir, options = {}) {
   return success;
 }
 
-async function getGitHubRepoContentsAtPath(repoName, path, repoPath = '') {
-  let contentsRequestUrl;
-  if (repoPath) {
-    contentsRequestUrl = `https://api.github.com/repos/${repoPath}/contents/${path}`;
-  } else {
-    contentsRequestUrl = `https://api.github.com/repos/HubSpot/${repoName}/contents/${path}`;
-  }
+async function getGitHubRepoContentsAtPath(
+  repoPath = 'HubSpot/hubspot-project-components',
+  path
+) {
+  const contentsRequestUrl = `https://api.github.com/repos/${repoPath}/contents/${path}`;
 
   return request.get(contentsRequestUrl, {
     json: true,
@@ -170,7 +168,6 @@ async function downloadGitHubRepoContents(
   dest,
   options = {
     filter: false,
-    repoPath: '',
   }
 ) {
   fs.ensureDirSync(path.dirname(dest));
@@ -178,8 +175,7 @@ async function downloadGitHubRepoContents(
   try {
     const contentsResp = await getGitHubRepoContentsAtPath(
       repoName,
-      contentPath,
-      options.repoPath
+      contentPath
     );
 
     const downloadContentRecursively = async contentPiece => {
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -34,11 +34,13 @@ async function fetchJsonFromRepository(
       headers: { ...DEFAULT_USER_AGENT_HEADERS, ...GITHUB_AUTH_HEADERS },
     });
   } catch (err) {
-    logger.error('An error occured fetching JSON file.');
     if (repoPath && err.statusCode === 404) {
-      return logger.error(
+      logger.error(
         i18n(`cli.lib.prompts.createProjectPrompt.errors.failedToFetchJson`)
       );
+      process.exit(1);
+    } else {
+      logger.error('An error occured fetching JSON file.');
     }
     logErrorInstance(err);
   }
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -19,9 +19,18 @@ const GITHUB_AUTH_HEADERS = {
  * @param {String} repoName - name of the github repository
  * @returns {Buffer|Null} Zip data buffer
  */
-async function fetchJsonFromRepository(repoName, filePath) {
+async function fetchJsonFromRepository(
+  repoName,
+  filePath,
+  customRepoPath = false
+) {
   try {
-    const URI = `https://raw.githubusercontent.com/HubSpot/${repoName}/${filePath}`;
+    let URI;
+    if (customRepoPath) {
+      URI = `https://raw.githubusercontent.com/${repoName}/${filePath}`;
+    } else {
+      URI = `https://raw.githubusercontent.com/HubSpot/${repoName}/${filePath}`;
+    }
     logger.debug(`Fetching ${URI}...`);
 
     return request.get(URI, {
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -20,14 +20,13 @@ const GITHUB_AUTH_HEADERS = {
  * @param {String} repoName - name of the github repository
  * @returns {Buffer|Null} Zip data buffer
  */
-async function fetchJsonFromRepository(repoName, filePath, repoPath = false) {
+async function fetchJsonFromRepository(
+  repoName = 'HubSpot/hubspot-project-components',
+  filePath,
+  repoPath = false
+) {
   try {
-    let URI;
-    if (repoPath) {
-      URI = `https://raw.githubusercontent.com/${repoName}/${filePath}`;
-    } else {
-      URI = `https://raw.githubusercontent.com/HubSpot/${repoName}/${filePath}`;
-    }
+    const URI = `https://raw.githubusercontent.com/${repoName}/${filePath}`;
     logger.debug(`Fetching ${URI}...`);
 
     return await request.get(URI, {
--- a/packages/cli-lib/github.js
+++ b/packages/cli-lib/github.js
@@ -1,13 +1,13 @@
 const request = require('request-promise-native');
 const path = require('path');
 const fs = require('fs-extra');
-const { i18n } = require('@hubspot/cli-lib/lib/lang');
 
 const { logger } = require('./logger');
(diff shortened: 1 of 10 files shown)


packages/cli/commands/project/create.js
```diff
@@ -61,6 +62,10 @@ exports.builder = yargs => {
       describe: i18n(`${i18nKey}.options.template.describe`),
       type: 'string',
     },
+    customRepoPath: {
```
brandenrodgers, line 65: Small nit, but IMO "custom" is a little redundant here. Would just `repoPath` work?

kemmerle: I was debating that, I can definitely make this change.

packages/cli-lib/github.js
```diff
@@ -124,8 +133,17 @@ async function cloneGitHubRepo(dest, type, repoName, sourceDir, options = {}) {
...
+  repoName,
+  path,
+  customRepoPath = ''
+) {
+  let contentsRequestUrl;
+  if (customRepoPath) {
+    contentsRequestUrl = `https://api.github.com/repos/${customRepoPath}/contents/${path}`;
+  } else {
+    contentsRequestUrl = `https://api.github.com/repos/HubSpot/${repoName}/contents/${path}`;
+  }
```
banderson, line 138: nit: Would we lose anything if we killed `isCustomRepoPath` entirely and used `HubSpot/hubspot-project-components` as the default repo value? That would let you work with the repo name/path as a single value and avoid adding conditional logic when you access the value, since both will follow the same structure of `org/repo`.

banderson (commented):
Definitely excited if the custom repo option works out, it's been on my wishlist! :)

packages/cli/commands/project/create.js
sejal27: I came here to just ask the same question as @brandenrodgers . :-) 
I think `repoPath` is short, and it's understood that it's going to be custom if you are using this flag.

sejal27: I also wonder if just `--path` is okay (to allow for non-GitHub sources). But that may be me overthinking it.

packages/cli-lib/github.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/837/commits/df3bddabd1b75142b473f1a965ca68a8dbebae97 and https://github.com/HubSpot/hubspot-cli/pull/837/commits/120f24a8aab3aaf08633bda45f1b66f584b8cde5

packages/cli-lib/lang/en.lyaml
```diff
@@ -1008,6 +1010,8 @@ en:
             nameRequired: "A project name is required"
             locationRequired: "A project location is required"
             invalidTemplate: "[--template] Could not find template {{ template }}. Please choose an available template."
+            failedtofetchJson: "Please ensure that the repo path you've entered is correct.
```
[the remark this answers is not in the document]
brandenrodgers: Seems pretty decent to me! 👌

packages/cli/commands/project/create.js
kemmerle: Right now, we're only set up to allow downloads from GitHub repositories (we're using the GH API), so I think it makes sense to keep as `repoPath`. Thanks for the feedback!

packages/cli-lib/lang/en.lyaml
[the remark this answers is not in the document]
kemmerle: Fixed in https://github.com/HubSpot/hubspot-cli/pull/837/commits/43f7a8ae0b6eddbe154c5638c0750f3fa68a128a

packages/cli-lib/github.js
```diff
@@ -19,17 +20,26 @@ const GITHUB_AUTH_HEADERS = {
  * @param {String} repoName - name of the github repository
  * @returns {Buffer|Null} Zip data buffer
  */
-async function fetchJsonFromRepository(repoName, filePath) {
+async function fetchJsonFromRepository(
```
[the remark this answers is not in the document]
kemmerle: This does need to be `async`, since we need to `await` the get request in the function ([link](https://github.com/HubSpot/hubspot-cli/blob/120f24a8aab3aaf08633bda45f1b66f584b8cde5/packages/cli-lib/github.js#L32)). 

I actually discovered a bug with this. We were originally simply returning the request without an `await`, and so we were never hitting the `catch` block, so originally, I was unable to make a custom error message.

packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -7,17 +7,19 @@ const { i18n } = require('@hubspot/cli-lib/lib/lang');
...
+  const isRepoPath = !!repoPath;
   const config = await fetchJsonFromRepository(
-    'hubspot-project-components',
-    'main/config.json'
+    repoPath,
+    'main/config.json',
+    isRepoPath
   );
 
   return config[PROJECT_COMPONENT_TYPES.PROJECTS];
```
brandenrodgers, line 18: not a blocker, but we could potentially do an additional error check here to make sure that the `config.json` in the repo has the right structure. I'm thinking about the scenario where the user puts a config.json in their repo, but it doesn't contain a `projects` field in it. Or if we want to go deeper, the values in the projects array don't contain the right fields. examples:

```js
const badConfig1 = {
  hubspot-projects: []
}

const badConfig2 = {
  projects: [
    { blah: ''}
  ]
}

```

kemmerle: I don't think it should be too hard to add some basic checks.

kemmerle: Added both checks in https://github.com/HubSpot/hubspot-cli/pull/837/commits/7d3c09c254666f4477d60bd043e23abf18360df0

packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -1,23 +1,45 @@
...
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
+const { logger } = require('@hubspot/cli-lib/logger');
 
 const i18nKey = 'cli.lib.prompts.createProjectPrompt';
 
-const createTemplateOptions = async () => {
+const hasAllProperties = projectList => {
+  return projectList.every(config =>
+    PROJECT_PROPERTIES.every(p =>
+      Object.prototype.hasOwnProperty.call(config, p)
```
kemmerle, line 17: I received an `eslint no-prototype-builtins` error when attempting to use `config.hasOwnProperty`. This is the generally accepted workaround.

kemmerle:
I've added two new errors:

1) When the `config.json` file is missing the `projects` field:

Screenshot 2023-05-09 at 3 21 37 PM

2) When any of the projects in the config are missing the required properties:

Screenshot 2023-05-09 at 3 21 12 PM

@brandenrodgers @banderson @sejal27

packages/cli-lib/github.js
```diff
@@ -19,17 +20,26 @@ const GITHUB_AUTH_HEADERS = {
...
     logger.debug(`Fetching ${URI}...`);
 
-    return request.get(URI, {
+    return await request.get(URI, {
       json: true,
       headers: { ...DEFAULT_USER_AGENT_HEADERS, ...GITHUB_AUTH_HEADERS },
     });
   } catch (err) {
     logger.error('An error occured fetching JSON file.');
+    if (repoPath && err.statusCode === 404) {
```
brandenrodgers, line 38: One last small nit. The console is noisy when we hit this error. Maybe we shouldn't log `logger.error('An error occurred fetching JSON file.');` when this check hits? So we only log one or the other. Here's an example of the console:

image

packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -1,23 +1,45 @@
...
+  const isRepoPath = !!repoPath;
   const config = await fetchJsonFromRepository(
-    'hubspot-project-components',
-    'main/config.json'
+    repoPath,
+    'main/config.json',
+    isRepoPath
   );
 
+  if (!config[PROJECT_COMPONENT_TYPES.PROJECTS]) {
```
brandenrodgers, line 30: We should probably add `process.exit` calls to the checks here. I also ran into an error in my console (screenshot above) where `config` is undefined. It may be worth adding a "not null" check at the start of this if.

kemmerle: Fixed in https://github.com/HubSpot/hubspot-cli/pull/837/commits/3d549ea4cb140ba9d558e364c834c2dd7aa11a9c

packages/cli-lib/github.js
kemmerle: Fixed in https://github.com/HubSpot/hubspot-cli/pull/837/commits/3d549ea4cb140ba9d558e364c834c2dd7aa11a9c

brandenrodgers (approved):
🚢

PR #838, opened 2023-05-09
Programmatically check for github integration

brandenrodgers:
## Description and Context

Our system blocks any direct uploads from the CLI whenever a project is using a git integration. Dev Mode needs to be smart enough to never attempt to upload to a project that it knows is using a git integration.

The Local Dev Manager now has three upload modes:
- always: always allow uploads without prompting the user. This is the mode that it's in when targeting a sandbox account
- manual: always ask the user if the CLI should upload changes. This is the mode that it's in when targeting a non-sandbox account
- never: never allow uploads, because the request would fail. This is the mode that it's in when targeting a project that's using a git integration.


camden11 (approved):
LGTM!

PR #839, opened 2023-05-09
Adjust `hs project dev` account prompt options

adamawang:
## Description and Context

Adds logic to disable the `Create new development sandbox` option based on the following criteria, as well as a custom message for the disabled option:
- Default account is NOT a production account.
  - `Option unavailable because your default account must be a production account.`
- Default account is a prod acct, but no more available sandboxes
  - `Option unavailable because you’ve reached the limit of x development sandboxes.`

Prompt now shows development sandboxes first to highlight dev sandbox usage, then standard sandboxes.

## Screenshots

Screenshots at a smaller screen width to show line break 

Screenshot 2023-05-09 at 10 50 18

Screenshot 2023-05-09 at 10 50 50

## TODO

Approved by content and design


packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -28,18 +39,48 @@ const selectTargetAccountPrompt = async (
...
       });
   } else {
+    let sandboxUsage = {};
+    try {
+      const accountId = getAccountId(defaultAccountConfig.portalId);
+      sandboxUsage = await getSandboxUsageLimits(accountId);
+    } catch (err) {
+      logger.debug('Unable to fetch sandbox usage limits: ', err);
+    }
+    const sandboxAccounts = accounts.reverse().filter(isSandbox);
```
adamawang, line 49: I've also reversed the accounts order out of the config to try and put the most recent accounts first. I figured since we don't have an accurate date to pull from, we can assume that new accounts are always added to the bottom of the config

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -28,18 +39,48 @@ const selectTargetAccountPrompt = async (
         };
       });
   } else {
+    let sandboxUsage = {};
+    try {
+      const accountId = getAccountId(defaultAccountConfig.portalId);
+      sandboxUsage = await getSandboxUsageLimits(accountId);
+    } catch (err) {
+      logger.debug('Unable to fetch sandbox usage limits: ', err);
```
adamawang, line 47: Only added a debug statement here, we have another sandbox limit validation after this prompt runs with more descriptive messaging

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
brandenrodgers: Seems reasonable to me 👌

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -28,18 +39,48 @@ const selectTargetAccountPrompt = async (
...
+    }
+    const sandboxAccounts = accounts.reverse().filter(isSandbox);
+    const createSandboxDisabled =
```
brandenrodgers, line 50: I don't think we need `createSandboxDisabled` if we're also making the same checks on lines 54 and 59-60 right?

brandenrodgers (approved):
🚢

PR #840, opened 2023-05-09
hs project deploy UX Improvements

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/389
This updates the `hs project deploy` command to prompt for a `buildId` when one is not specified, preventing users from accidentally deploying their latest build as was previously possible. The prompt defaults to the latest build unless its currently deployed, and includes validation that prevents users from entering the currently deployed build and builds that don't exist.

## Screenshots
Before
Screenshot 2023-05-09 at 3 14 03 PM

After
Screenshot 2023-05-09 at 3 14 31 PM

Error state
Screenshot 2023-05-09 at 3 14 58 PM
Screenshot 2023-05-09 at 3 14 51 PM
ates

## Who to Notify

### DIFF
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -1056,7 +1056,7 @@ en:
           bugPrompt: "Create an issue on Github in your browser?"
           generalPrompt: "Open the projects feedback form in your browser?"
         buildIdPrompt:
-          enterBuildId: "[--build] Deploy which build? (26)"
+          enterBuildId: "[--build] Deploy which build?"
           errors:
             buildIdDoesNotExist: "Build {{ buildId }} does not exist for project {{ projectName }}."
             buildAlreadyDeployed: "Build {{ buildId }} is already deployed."
--- a/packages/cli-lib/lang/en.lyaml
+++ b/packages/cli-lib/lang/en.lyaml
@@ -482,7 +482,8 @@ en:
               deploying: "Deploying project at path: {{ path }}"
             errors:
               deploy: "Deploy error: {{ details }}"
-              noBuildId: "No latest build ID was found."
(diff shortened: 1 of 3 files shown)


packages/cli/lib/prompts/buildIdPrompt.js
```diff
@@ -0,0 +1,35 @@
...
+        return;
+      }
+      return latestBuildId;
+    },
+    validate: val => {
+      if (Number(val) > latestBuildId) {
+        return i18n(`${i18nKey}.errors.buildIdDoesNotExist`, {
+          buildId: val,
+          projectName,
+        });
```
camden11, line 21: Per @mendel-at-hubspot, we could theoretically get into certain weird states where `latestBuild` is not the highest build ID, but practically it always will be. Figured it was worth it for the UX improvement, especially since using the flag is always an option.

packages/cli-lib/lang/en.lyaml
```diff
@@ -1054,6 +1055,11 @@ en:
             general: "[--general] Tell us about your experience with HubSpot's developer tools"
           bugPrompt: "Create an issue on Github in your browser?"
           generalPrompt: "Open the projects feedback form in your browser?"
+        buildIdPrompt:
+          enterBuildId: "[--build] Deploy which build? (26)"
```
brandenrodgers, line 1059: What's the purpose of the (26) here? Is it just to show that we're asking for a number? One suggestion is to instead to something like:

`[--build] Enter the number of the build you would like to deploy:`

I'm also wondering if we should support some way for them to easily choose the latest build (without knowing the exact number of it). Maybe that's not worthwhile though.

brandenrodgers (approved):
👌

PR #841, opened 2023-05-09
`hs project dev` add polling to project fetch in case of 403

adamawang:
## Description and Context

This PR specifically deals with fetchProject behavior when creating a new sandbox with `hs project dev`. Due to gating slowness (sometimes) when creating a new sandbox in prod, we are receiving 403's on initial fetch with a forbidden request error. 

We are adding a pollingFetchProject util that keeps retrying (up to max 30s) until a 404 or a success is received. This also ensures that the portal is ungated if the user needs to create a new project in the next step. 

We can remove this polling piece once CRM dev tools are ungated to all. 

## Screenshots

Adding a spinner to show progress
Screenshot 2023-05-09 at 15 20 23

Once we get a 404 or success, we remove the spinner
Screenshot 2023-05-09 at 15 20 39

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -168,6 +168,35 @@ const validateProjectConfig = (projectConfig, projectDir) => {
   }
 };
 
+const pollFetchProject = (accountId, projectName) => {
+  // Temporary solution for gating slowness. Retry on 403 statusCode
+  return new Promise((resolve, reject) => {
+    let pollCount = 0;
+    const spinnies = SpinniesManager.init();
+    spinnies.add('pollFetchProject', {
+      text: 'Fetching project status',
+    });
+    const pollInterval = setInterval(async () => {
+      try {
+        const project = await fetchProject(accountId, projectName);
+        if (project) {
+          spinnies.remove('pollFetchProject');
+          clearInterval(pollInterval);
+          resolve(project);
+        }
+      } catch (err) {
+        // Poll up to max 30s, throw immediately if 404
+        if (err.statusCode === 404 || pollCount >= 15) {
+          spinnies.remove('pollFetchProject');
+          clearInterval(pollInterval);
+          reject(err);
+        }
+        pollCount += 1;
+      }
+    }, 2000);
+  });
+};
+
 const ensureProjectExists = async (
   accountId,
   projectName,
@@ -175,7 +204,8 @@ const ensureProjectExists = async (
 ) => {
   const accountIdentifier = uiAccountDescription(accountId);
   try {
-    const project = await fetchProject(accountId, projectName);
+    // const project = await fetchProject(accountId, projectName);
+    const project = await pollFetchProject(accountId, projectName);
     return !!project;
   } catch (err) {
     if (err.statusCode === 404) {
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -39,6 +39,9 @@ const { EXIT_CODES } = require('./enums/exitCodes');
 const { uiLine, uiLink, uiAccountDescription } = require('../lib/ui');
 const { i18n } = require('@hubspot/cli-lib/lib/lang');
 const SpinniesManager = require('./SpinniesManager');
+const {
+  isSpecifiedError,
+} = require('@hubspot/cli-lib/errorHandlers/apiErrors');
 
 const i18nKey = 'cli.lib.projects';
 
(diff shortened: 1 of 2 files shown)


packages/cli/lib/projects.js
```diff
@@ -168,14 +168,51 @@ const validateProjectConfig = (projectConfig, projectDir) => {
...
+const pollFetchProject = async (accountId, projectName) => {
+  // Temporary solution for gating slowness. Retry on 403 statusCode
+  try {
+    const project = await fetchProject(accountId, projectName);
+    return project;
+  } catch (error) {
+    if (error.statusCode === 404) {
+      throw error;
+    }
+  }
```
adamawang, line 180: Try first for all other usages of `ensureProjectExists`, only poll if response is not successful or a 404

packages/cli/lib/projects.js
```diff
@@ -168,14 +168,51 @@ const validateProjectConfig = (projectConfig, projectDir) => {
...
+    let pollCount = 0;
+    const spinnies = SpinniesManager.init();
+    spinnies.add('pollFetchProject', {
+      text: 'Fetching project status',
+    });
+    const pollInterval = setInterval(async () => {
+      try {
+        const project = await fetchProject(accountId, projectName);
+        if (project) {
+          spinnies.remove('pollFetchProject');
```
adamawang, line 191: Rather than succeed, I thought removing would be cleaner since we already have a warning message: 
Screenshot 2023-05-09 at 15 20 39

Let me know what you think

packages/cli/lib/projects.js
```diff
@@ -168,14 +168,51 @@ const validateProjectConfig = (projectConfig, projectDir) => {
   }
 };
 
+const pollFetchProject = async (accountId, projectName) => {
```
brandenrodgers, line 171: Going off of the screenshot here, it was my understanding that the issue was that we are not able to successfully create the project in the newly-created sandboxes. This seems to be adding polling to the fetching of the project, not the creation of it in the target account. Let me know if I'm just misunderstanding what the original issue was!

Screenshot 2023-05-08 at 15 53 30

adamawang: Yep its both! The very last prompt was asking me to create a project, but then it fails and exits. This is before you merged in your other branch with an error message though. 

By adding the polling to the fetch, we can ensure that upon a 404 or success, the portal has been successfully ungated and project create in the next step will work

packages/cli/lib/projects.js
```diff
@@ -168,14 +168,51 @@ const validateProjectConfig = (projectConfig, projectDir) => {
...
+      try {
+        const project = await fetchProject(accountId, projectName);
+        if (project) {
+          spinnies.remove('pollFetchProject');
+          clearInterval(pollInterval);
+          resolve(project);
+        }
+      } catch (err) {
+        // Poll up to max 30s
+        if (err.statusCode === 404 || pollCount >= 15) {
```
brandenrodgers, line 197: Should we also abort on 500 status codes? Just to cover our bases

packages/cli/lib/projects.js
```diff
@@ -168,14 +168,51 @@ const validateProjectConfig = (projectConfig, projectDir) => {
...
+
 const ensureProjectExists = async (
   accountId,
   projectName,
   { forceCreate = false, allowCreate = true, noLogs = false } = {}
 ) => {
   const accountIdentifier = uiAccountDescription(accountId);
   try {
-    const project = await fetchProject(accountId, projectName);
+    const project = await pollFetchProject(accountId, projectName);
```
brandenrodgers, line 215: I'm thinking that we should only use this new polling logic for the Dev Mode work. Maybe we can add a new `withPolling` option to this arguments in `ensureProjectExists`. I don't think this issue will have much of an effect on the existing project-related commands.

adamawang: Yep exactly - I made it so behavior for all other usages stayed the same, but I like the prop idea better

brandenrodgers:
I tested this locally and it's all working correctly 👌 I'll take a quick look, but this should be good to go after you get the updates in.

packages/cli/lib/projects.js
adamawang: Yeah if its not ungated, we'll get a 403 with the following message: 
Screenshot 2023-05-08 at 15 28 17

adamawang:
Updated in my last commit - tested with `hs project dev`, `hs project watch` and `hs project upload` to make sure `ensureProjectExists` works as intended, looking good on my end

brandenrodgers (approved):
🚢

PR #847, opened 2023-05-10
Move lang utils to hubspot-cli package

camden11:
## Description and Context
Soon, we'll be pulling `cli-lib` out of this repo and into its own repo. Currently, all lang data for both `cli-lib` and `hubspot-cli` is defined within `cli-lib`, and separating the two packages means that any PR to `hubspot-cli` with lang changes will necessitate a separate PR to `cli-lib`. This PR copies the lang utils from `cli-lib` to `hubspot-cli` and adds a separate `en.lyaml` file to `hubspot-cli` so we can avoid this.

## Who to Notify
@brandenrodgers @kemmerle


brandenrodgers (approved):
lgtm 👍 I ran through a bunch of commands and it's looking good

PR #848, opened 2023-05-10
Move generateThemeFieldSelectors into CLI command

jsines:
## Description and Context
The themes and modules team has added the ability to link fields to CSS selectors so that when a content editor hovers over the field, the associated area of the page gets highlighted. This feature requires devs to add a `selectors.json` file to their theme's root directory to then be referenced by the page editor. To aid developers in creating this file, jmin has created a script that generates a "best guess" at what the linking should be. This PR incorporates that script into a cli command

`hs theme generate-selectors <themePath>`

Also tested on Windows and seems to work fine 👍


packages/cli/commands/generateThemeFieldSelectors.js
```diff
@@ -0,0 +1,210 @@
...
+    fieldsJson = generateInheritedSelectors(fieldsJson, fieldsJson);
+  }
+
+  const selectorsMap = generateSelectorsMap(fieldsJson);
+  const selectorsPath = `${themePath}/selectors.json`;
+
+  fs.writeFileSync(selectorsPath, `${JSON.stringify(selectorsMap, null, 2)}\n`);
+
+  logger.success(
+    `Selectors map generated at ${selectorsPath}. These are best guesses for the selectors for each field and is not exhaustive, please double check before uploading these changes.`
```
Jonah-Min, line 199: We probably want to reword this as well haha, same with the errors on line 35 and 175, these were just placeholder text as I was testing the script. Maybe something like:

Line 35: `Unable to find theme's fields.json.`
Line 175: `No selectors found.`
Line 199: `Success! Selectors generated for ${themePath}, please double check the selectors generated at ${selectorsPath} before uploading the theme.`

Also worth localizing these if possible as well.

jsines: Good call 👀 I'll move these to `en.lyaml`

We can't actually localize these yet since it's a public repo and our localization pipeline only works internally, but we'll eventually ™️  have a way to sync it with an internal repo so it'd be good to keep them there

Jonah-Min: Makes sense, was definitely curious how this would be hooked up to our localization tools haha

brandenrodgers:
@jsines just a heads up, [we recently split our lyaml files into separate files](https://github.com/HubSpot/hubspot-cli/pull/847) (one for cli-lib and one for cli). We tried to do it when there were no in-flux PR's to avoid conflicts, but it looks like this branch was created before we made the split. Resolving shouldn't be too hard 🤞

jsines:
@brandenrodgers Thanks for the heads up! 🙏 got that merged in

packages/cli/commands/theme.js
```diff
@@ -8,10 +9,10 @@ exports.command = 'theme';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.builder = yargs => {
-  addConfigOptions(yargs, true);
-  addAccountOptions(yargs, true);
```
jsines, line 12: Should be safe to remove these here I think - we don't need them for `generate-selectors` and they're present a second time in the exports.builder of  `theme/marketplace-validate.js`

Jonah-Min (approved):
Thanks for porting the script over and cleaning up some of my mess! Excited to see this released soon 🎉

kemmerle:
Hi @jsines, I was testing out this PR, and I'm running into a bit of trouble. I'm going to list out the steps I took to see where I might have gone wrong.

1) I created a website theme using the `hs create website-theme` command.
2) I navigated into the theme folder in the terminal (Located in the `tmp/website-theme` folder in my `hubspot-cli` repo). 
3) I ran the command and received a success message, but I can't locate the `selectors.json` file in my theme folder. 

Screenshot 2023-05-11 at 4 54 45 PM

Am I missing something?

Jonah-Min:
@kemmerle Interesting if it reaches that success message then the `selectors.json` should be generated 🤔  I just ran my local non-cli version of the script taking the same steps and I see the `selectors.json` show up. If you could show me how to set up the CLI to run it locally I can help debug.

jsines:
@kemmerle this is a weird side effect of running it as `yarn hs` instead of using a symlink - you can see if you add a `console.log(process.cwd())`.

With `yarn hs` I get: `/Users/jsines/src/hubspot-cli`
With `hs` symlinked I get: `/Users/jsines/src/hubspot-cli/website-theme`, my actual cwd.

When you do `yarn hs` it'll go up the directory tree until it finds a package.json and then fetch  the `hs` binary from that project's `node_modules/.bin` and run it from that project's directory. You can also see the side effect of this by adding a new pckage.json with `yarn init`, it'll try to fetch it from `/Users/jsines/src/hubspot-cli/website-theme/node_modules/.bin` instead, and fail to find the command since it's empty!

kemmerle:
@jsines, that's exactly what happened! Thanks for the explanation!

PR #849, opened 2023-05-10
`hs project dev` add missing scope error handling

adamawang:
## Description and Context

When attempting to create a sandbox with `hs project dev`, we throw an error if the PAK does not have sandboxes scopes. 

This PR repurposes a sandboxes scoping error message we had for `hs sandbox create` to help guide the user to deactivate and regenerate a new PAK with new scopes. The two places that this error is thrown are both sandboxes specific so the error message applies perfectly. 

## Screenshots

Screenshot 2023-05-10 at 15 37 31


brandenrodgers (approved):
Nice 👌

PR #850, opened 2023-05-10
Shortened dev mode urls

brandenrodgers:
## Description and Context

More context on the issue here: https://github.com/HubSpot/hubspot-cli/pull/835#issuecomment-1533240834

Here's one example of a way that we could handle long URLs that don't play nicely with the spinnies lib. This uses our running server to redirect urls.

### Before
image

### After
image


packages/cli/lib/LocalDevManager.js
```diff
@@ -159,14 +160,10 @@ class LocalDevManager {
...
     });
-    const projectDetailUrl = getProjectDetailUrl(
-      this.projectConfig.name,
-      this.targetAccountId
-    );
     this.spinnies.add('viewInHubSpotLink', {
       text: uiLink(
         i18n(`${i18nKey}.header.viewInHubSpotLink`),
-        projectDetailUrl,
+        `${this.devServerPath}/hs/project`,
```
kemmerle, line 166: What happens if the `this.devServerPath` remains `null`? Is that possible?

packages/cli/lib/LocalDevManager.js
```diff
@@ -510,23 +509,24 @@ class LocalDevManager {
   }
 
   async startServers() {
-    // TODO spin up local dev servers
-    return true;
+    this.devServerPath = await DevServerManager.start({
```
kemmerle, line 512: Should we be declaring this above line 166, where we reference `this.devServerPath`?

kemmerle (approved):
I had two small questions which I believe should require only minor work, if any. Otherwise, LGTM :shipit:

packages/cli/lib/LocalDevManager.js
brandenrodgers: This `startServers` function actually [does get called before the code on line 166 executes](https://github.com/HubSpot/hubspot-cli/pull/850/files#diff-0391b654b5bab2f8d2b05906b406db63c1c591cc6b937dfba92d6a7d475cf878R93). So we should be all set there

packages/cli/lib/LocalDevManager.js
brandenrodgers: done in [fe1c962](https://github.com/HubSpot/hubspot-cli/pull/850/commits/fe1c9622912af7e681d4f05cb2e3484b5849afd9)

PR #852, opened 2023-05-16
Support uie local dev server

brandenrodgers:
## Description and Context

Initial pass to support local dev servers inside of the Dev Mode command. Currently there are no servers configured, but this should enable teams to start testing out the connection.

I have this is a semi-decent state. To run this locally you need to do a few things:
- make sure that the hubspot-cli and ui-extensibility repos live in the same directory (because I am importing the ui-extensibility files via relative path for now)
- Run `npm install` in the ui-extensibility directory
- Run the command with this port: `yarn hs project dev --port=5173`


packages/cli/lib/DevServerManager.js
```diff
@@ -26,7 +32,33 @@ class DevServerManager {
     }
   }
 
-  async start({ projectConfig, accountId, port }) {
+  generateURL(path) {
+    return this.path ? `${this.path}/${path}` : null;
+  }
+
+  async getProjectFiles(projectSourceDir) {
```
camden11, line 39: Do we need this util? Looks like it doesnt do anything aside from call walk and is only used once

camden11 (approved):
Might ask you for a walk through tomorrow to make sure i'm 100% up to speed with what's going on here but I think this looks good

brandenrodgers:
This is replaced by: https://github.com/HubSpot/hubspot-cli/pull/879

PR #855, opened 2023-05-18
Dev mode account select ux updates

brandenrodgers:
## Description and Context

Making some changes to the init flow for the Dev Mode command
- New and more concise intro message
- Only show account selection prompt when the default account is a non-sandbox account
- Updates to the header for the running process view
- The console no longer clears while the process is running, so status messages persist
## Screenshots

image

## TODO

- [x] ~I need a URL to the docs for this command~ ignoring learn more links for now
## Who to Notify

cc/ @markhazlewood

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -181,21 +181,26 @@ exports.handler = async options => {
   const spinnies = SpinniesManager.init();
 
   if (!projectExists) {
-    uiLine();
-    logger.warn(
-      i18n(`${i18nKey}.logs.projectMustExistExplanation`, {
-        accountIdentifier: uiAccountDescription(targetAccountId),
-        projectName: projectConfig.name,
-      })
-    );
-    uiLine();
+    // Create the project without prompting if this is a newly created sandbox
+    let shouldCreateProject = createNewSandbox;
 
-    const shouldCreateProject = await confirmPrompt(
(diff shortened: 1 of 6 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -76,45 +71,32 @@ exports.handler = async options => {
...
-  let chooseNonSandbox = false;
+  const defaultAccountIsSandbox = isSandbox(accountConfig);
+
+  if (!targetAccountId && defaultAccountIsSandbox) {
+    targetAccountId = accountId;
+  }
 
   if (!targetAccountId) {
+    logger.log(
+      uiLink(i18n(`${i18nKey}.logs.learnMoreLink`), 'https://www.google.com')
```
kemmerle, line 82: I would suggest flagging this with a `TODO` to avoid any heartbreak.

kemmerle:
Left one very minor comment. Unfortunately, I'm unable to test the command out locally, because I'm having trouble linking the `cli-lib` repo. Hopefully, that gets resolved tomorrow/Monday, and I can leave feedback if necessary.

packages/cli/commands/project/dev.js
brandenrodgers: Good call. I may actually delete this for now because I don't think we have a doc setup for this yet.

PR #857, opened 2023-05-19
Better error handling for locked projects in Dev Mode

brandenrodgers:
## Description and Context

This adds some better error handling for the situation where a user runs the `hs project dev` command against a project that is currently locked.

The current copy that I'm using is this:
> Your project is locked. This may mean that another user is running the {{#bold}}`hs project dev`{{/bold}} command for this project. If this is you, unlock the project in Projects UI.

cc/ @markhazlewood for thoughts

### BEFORE
image

### AFTER
image

### DIFF
--- a/packages/cli/commands/project/upload.js
+++ b/packages/cli/commands/project/upload.js
@@ -59,13 +59,6 @@ exports.handler = async options => {
   );
 
   if (result.error) {
-    logApiErrorInstance(
-      result.error,
-      new ApiErrorContext({
-        accountId,
-        projectName: projectConfig.name,
-      })
-    );
     if (
       isSpecifiedError(result.error, {
         subCategory: ERROR_TYPES.PROJECT_LOCKED,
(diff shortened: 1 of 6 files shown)


packages/cli/commands/project/upload.js
```diff
@@ -50,6 +58,25 @@ exports.handler = async options => {
     message
   );
 
+  if (result.error) {
+    logApiErrorInstance(
+      result.error,
+      new ApiErrorContext({
+        accountId,
+        projectName: projectConfig.name,
```
camden11, line 66: Is there a reason we're logging the error here but not in `dev`?

camden11 (approved):
LGTM - just one question

packages/cli/commands/project/upload.js
brandenrodgers: I think it was for backwards compatibility with how it worked before. But I think I can actually remove this 👍

PR #861, opened 2023-05-20
build(runtime): use lts/hydrogen

AugustinMauroy:
## Description and Context

Update to nodejs lts


kemmerle:
Hi @AugustinMauroy, 

Thanks for your interest in the HubSpot CLI! We've decided not to move forward with your PR but wanted to leave you some feedback:

1) Setting our version of node to `lts/hydrogen` would require our customers to be on node 18 when using the CLI. It would be great if everyone were on the latest version of node, but we don't require that of our customers. 

2) Prefixing require statements with `node:` bypasses the require cache, but that's not a feature we typically need in this particular repo. If there's no change required to improve performance or functionality, we generally like to leave the code as is. 

We hope this proves helpful to you and you continue on your coding journey!

PR #872, opened 2023-06-02
Update `hs ls` to include @hubspot folder, improve logging a little

jsines:
## Description and Context
`getDirectoryContentsByPath` omits the `@hubspot` default asset folder. Recently it became possible to query the folder directly, leading to some strange behavior where `hs ls` didn't list `@hubspot`, but `hs ls @hubspot` would print the directory contents just fine. This update aims to bring this behavior more in line with the current DMUI and VSCE behavior and display this content for users.

The i18n key for the error was also wrong so fixed that. `In` -> `At`

Also went ahead and adjusted the logging a little bit, the `@hubspot` and `@marketplace` now display in bold, and files now display in cyan instead of black. We can change this color scheme a bit if anyone has Opinions 

## Screenshots

Screenshot 2023-06-02 at 2 22 44 PM

## Who to Notify

@TheWebTech

### DIFF
--- a/packages/cli/commands/list.js
+++ b/packages/cli/commands/list.js
@@ -52,52 +52,35 @@ exports.handler = async options => {
     process.exit(EXIT_CODES.SUCCESS);
   }
 
-  if (contentsResp.children) {
-    const contents =
-      directoryPath === '/'
-        ? ['@hubspot', ...contentsResp.children]
-        : contentsResp.children;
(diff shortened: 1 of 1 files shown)


kemmerle (approved):
Hi @jsines, thanks for handling this! The code looks great to me; I also tested it out, and the `@hubspot` folder displays for me. I think the colors for files/folders are fine for now. If we get any negative feedback from customers, we can always change the colors 👍

PR #874, opened 2023-06-08
Add tracking for template type on project create

brandenrodgers:
## Description and Context

Adding usage tracking for the template that the user selects during the `hs project create` command flow.


kemmerle (approved):
LGTM, looking forward to seeing these results! :shipit:

PR #876, opened 2023-06-12
`hs sandbox create` duplicate name error handling

adamawang:
## Description and Context

Issue to track: https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/405

Updates "error" message when a duplicate name is detected to be more clear about where the duplicate name was found. I've also used a logger.warn to make it clearer, and updated the default account name to append the newly created sandbox hub ID to avoid suggesting yet another duplicate name. 

## Screenshots

Before:
image

After: 
Screenshot 2023-06-12 at 15 27 13

### DIFF
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -998,7 +998,7 @@ en:
             invalidName: "You entered an invalid name. Please try again."
             nameRequired: "The name may not be blank. Please try again."
             spacesInName: "The name may not contain spaces. Please try again."
-            accountNameExists: "Account name \"{{ name }}\" already exists, please enter a different name."
+            accountNameExists: "Account with name \"{{ name }}\" already exists in the CLI config, please enter a different name."
         personalAccessKeyPrompt:
           enterAccountId: "Enter the account ID for your account (the number under the DOMAIN column at https://app.hubspot.com/myaccounts-beta ): "
           enterClientId: "Enter your OAuth2 client ID: "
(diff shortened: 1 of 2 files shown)


packages/cli/lib/sandboxes.js
```diff
@@ -256,19 +256,20 @@ const saveSandboxToConfig = async (env, result, force = false) => {
...
         );
         const { name: promptName } = await enterAccountNamePrompt(
-          nameForConfig
+          nameForConfig + `_${result.sandbox.sandboxHubId}`
         );
         validName = promptName;
       } else {
         // Basic invalid name handling when force flag is passed
-        validName = nameForConfig + `_${Date.now()}`;
+        validName = nameForConfig + `_${result.sandbox.sandboxHubId}`;
```
adamawang, line 272: Use the sandboxHubId instead of a date string for readability

packages/cli/lang/en.lyaml
```diff
@@ -1077,7 +1077,7 @@ en:
             errors:
               invalidName: "You entered an invalid name. Please try again."
               nameRequired: "The name may not be blank. Please try again."
-              accountNameExists: "Account name \"{{ name }}\" already exists, please enter a different name."
+              accountNameExists: "Account with name \"{{ name }}\" already exists in the CLI config, please enter a different name."
```
adamawang, line 1080: Adjusted in two places to make it clearer what the "name" error is referencing

kemmerle: Nice, this does make it much more clear 👍

packages/cli/lib/sandboxes.js
kemmerle: Agreed, this makes more sense to me, as well.

kemmerle (approved):
Reviewed the code and tested out locally, and these changes LGTM! 🚢

PR #877, opened 2023-06-15
Add error messaging on failed builds in project dev mode command

brandenrodgers:
## Description and Context

The Dev Mode command was previously failing silently whenever the initial build failed. It This updates it to show the user that something went wrong in their build. It also adds logging for the specific build errors (rather than just saying "something went wrong").

## Screenshots

Screenshot shows the initial state of the dev mode command when the initial upload + build fails. In the before screenshot, we're telling the user that everything went fine. In the after screenshot, we're doing a better job of letting the user know what the status of their project is.
### Before
image

### After
image

## TODO

- [x] Verify the UX of the output for the status when builds fail
- [x] Verify the UX of the output of the build errors
- [x] Verify that the error handling for failed deploys also works

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -240,11 +240,11 @@ exports.handler = async options => {
     isParent: true,
   });
 
-  let result;
+  let initialUploadResult;
   // Create an initial build if the project was newly created in the account or if
   // our upload permission is set to "always"
   if (!projectExists || uploadPermission === UPLOAD_PERMISSIONS.always) {
-    result = await handleProjectUpload(
+    initialUploadResult = await handleProjectUpload(
       targetAccountId,
       projectConfig,
       projectDir,
@@ -253,9 +253,9 @@ exports.handler = async options => {
     );
   }
 
-  if (result && result.error) {
+  if (!initialUploadResult || initialUploadResult.uploadError) {
     if (
-      isSpecifiedError(result.error, {
+      isSpecifiedError(initialUploadResult.uploadError, {
         subCategory: ERROR_TYPES.PROJECT_LOCKED,
       })
     ) {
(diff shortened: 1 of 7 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -253,9 +253,9 @@ exports.handler = async options => {
     );
   }
 
-  if (result && result.error) {
+  if (!initialUploadResult || initialUploadResult.uploadError) {
```
kemmerle, line 256: If `initialUploadResult` is undefined, do we need to account for that in this condition? Will this error out if `initialUploadResult.uploadError` is undefined?

packages/cli/commands/project/upload.js
```diff
@@ -59,9 +59,9 @@ exports.handler = async options => {
       message
     );
 
-    if (result.error) {
+    if (result.uploadError) {
```
kemmerle, line 62: To check, does this reflect changes on the BE?

kemmerle: Nevermind, just saw the changes you made in [lib/projects.js](https://github.com/HubSpot/hubspot-cli/pull/877/files#diff-fa4fa15b34bd2259fc2c24ec990fdbaa6492e6030d940700895318b399058174)

kemmerle (approved):
I left one question about a possible edge case with error messaging, but otherwise, LGTM. 

I tested it out locally, changed the name of the CRM card file in the serverless.json file to nonsense, and the new error messages were returned 👍

markhazlewood42:
Discussed when we met, but quick CLI layout notes:
* Add line breaks before each new build
* Indent output (success or errors) for a particular build *under* that main build headline.

@brandenrodgers

packages/cli/commands/project/dev.js
brandenrodgers: Good call, I added some edge case handling in [5e8e5be](https://github.com/HubSpot/hubspot-cli/pull/877/commits/5e8e5be8f83f0dcab2cd305559c756fd4c367540)

PR #878, opened 2023-06-20
Improve tracking for sandbox commands

adamawang:
## Description and Context

Sandbox commands were missing sandbox types in the tracking code. We need this level of granularity now that we have standard sandbox actions such as creating and syncing. 

## Screenshots

n/a, all related to tracking data

## TODO

n/a

### DIFF
--- a/packages/cli/commands/sandbox/create.js
+++ b/packages/cli/commands/sandbox/create.js
@@ -79,6 +79,10 @@ exports.handler = async options => {
   }
   const sandboxType = sandboxTypeMap[type] || sandboxTypeMap[typePrompt.type];
 
+  const trackingSandboxType =
+    sandboxType === DEVELOPER_SANDBOX ? 'DEVELOPER' : 'STANDARD';
+  trackCommandUsage('sandbox-create', { type: trackingSandboxType }, accountId);
+
   // Check usage limits and exit if parent portal has no available sandboxes for the selected type
   try {
     await validateSandboxUsageLimits(accountConfig, sandboxType, env);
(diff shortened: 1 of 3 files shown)


packages/cli/commands/sandbox/delete.js
```diff
@@ -44,6 +44,8 @@ exports.handler = async options => {
   const { account, force } = options;
   const config = getConfig();
 
+  trackCommandUsage('sandbox-delete', null);
```
adamawang, line 47: Previously we were not tracking this command if it exited early. Moved this to the top of the command

packages/cli/commands/sandbox/create.js
```diff
@@ -79,6 +79,10 @@ exports.handler = async options => {
   }
   const sandboxType = sandboxTypeMap[type] || sandboxTypeMap[typePrompt.type];
 
+  const trackingSandboxType =
```
brandenrodgers, line 82: I think we should try to find a pattern where we can log metadata for a command without having it look like commands are getting more usage than they actually are. Typically we try to log one user event per command and that lets us get an accurate count for total number of command runs.

Maybe we need some new [action](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/usageTracking.js#L37) type other than `cli-command`. Maybe something like `cli-command-metadata`? Then we can fire that one as often as we like, and it won't interfere with any of the total counts from the `cli-command` action.

adamawang: We actually just had another meeting on this yesterday with the team! We're going to go ahead and remove all the duplicate event tracking with extra metadata and were planning on creating another `cli-command` event with a separate name to track sandbox events. 

The `cli-command-metadata` one sounds a bit cleaner though, would we have the same metadata limitations as we do for the current implementation?

brandenrodgers: Great! That sounds like a good plan 👌 

> would we have the same metadata limitations as we do for the current implementation?

Yeah I think so. AFAIK those limitations exist everywhere in usage tracking, regardless of the action type we're using. I think we hit a company-wide limit on new field types so we have to re-use existing ones.

PR #880, opened 2023-06-27
More control over spinnies lib

brandenrodgers:
## Description and Context

I'm creating a slightly better pattern for the spinnies lib. I ported in most of the spinnies lib into our project so we can have more control over how it works. One real example of where this extra control will be useful is outlined [here](https://github.com/HubSpot/hubspot-cli/pull/879/files#diff-d4e938987cb7e9bb301c67ee17c9885e9d2d2b74be40068d31c5a66936668cacR74-R77).

 I'm also switching to a better pattern for how files should use the SpinniesManager.

```js
// Old pattern
const spinnies = SpinniesManager.init();

spinnies.add();

// New pattern
SpinniesManager.init();

SpinniesManager.add();
```

All I've done in this initial PR is port the core code from the spinnies package into this repo and merge it with the custom logic that we already had in SpinniesManager. I took a look at the license and it seems like this is fine to do.


packages/cli/lib/SpinniesManager.js
```diff
@@ -1,104 +1,346 @@
-const Spinnies = require('spinnies');
+'use strict';
```
brandenrodgers, line 1: This is mostly just a 1:1 port from the spinnies library [here](https://github.com/jbcarpanelli/spinnies/blob/master/index.js)

camden11 (approved):
Code looks good (assuming all those utils are ported directly from spinnies). Tried it out with a few commands and all works as expected

kemmerle (approved):
Interesting to read through the SpinniesManager and spinniesUtils files to see how they work.  I'm excited that we'll be able to fully customize this to our needs in the future!

I also tried out a bunch of commands, and everything is WAD!

PR #881, opened 2023-06-28
Disable printing logs in project logs command

kemmerle:
## Description and Context
Because of a change in the logging endpoint, we are currently always returning a 404 status code error when attempting to fetch logs in the `hs project logs` command. 

It will take some time for the logging team to address this issue, and so in the meantime, I've disabled the `Logs not found` comment and simply provide the customer to a link to the Logs UI, which is correct and up-to-date. 

One question for @camden11 or @brandenrodgers: I've simply commented out the code that generates the logs for now, assuming that we will re-enable this feature once the API endpoint is fixed. Would it be better simply to delete this code for now? 

## Screenshots

Currently, we always tell the customer that logs cannot be found:

Screenshot 2023-06-28 at 12 33 36 PM

With these changes, we simply link to the logs:

Screenshot 2023-06-28 at 12 34 03 PM

## Who to Notify

@brandenrodgers @camden11

### DIFF
--- a/packages/cli/commands/project/logs.js
+++ b/packages/cli/commands/project/logs.js
@@ -9,7 +9,7 @@ const {
 } = require('../../lib/commonOpts');
 const { trackCommandUsage } = require('../../lib/usageTracking');
 const { logger } = require('@hubspot/cli-lib/logger');
-const { outputLogs } = require('@hubspot/cli-lib/lib/logs');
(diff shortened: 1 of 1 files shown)


brandenrodgers (approved):
👍 Seems good to me. We can uncomment and update when there's a new endpoint available for us to use.

PR #882, opened 2023-06-28
Fix bugs in bulk adds/deletes in dev mode

brandenrodgers:
## Description and Context

This fixes an issue that pops up when you try to add or remove many files simultaneously from a project. You can test it out by creating a folder of several files and then alternate dragging it into and out of the project while the Dev Mode command is running.

The main issue was happening because we were using `await` on the call to [this.uploadQueue.addAll](https://github.com/HubSpot/hubspot-cli/compare/br-fix-dev-mode-bulk-actions?expand=1#diff-0391b654b5bab2f8d2b05906b406db63c1c591cc6b937dfba92d6a7d475cf878L638). This was causing the process to wait until that finished before setting `this.standbyChanges = [];` (happens right afterwards). This opened us up to the race condition where:
1. User drags a folder that contains two files (`file1.js` and `file2.js`) into the project
2. The watcher picks up the addition of file1. It adds it to the queue and flushes the changes, which triggers an upload
3. The watcher picks up the addition of file2 (before the flushing of the file1 changes as finished) and it flushes the changes
4. The standbyChanges list hasn't been cleared out yet, so file1 changes and file2 changes get uploaded

Also, I need to test this out, but I don't think we need `unlinkDir` because our BE doesn't care about empty folders anyways.

### DIFF
--- a/packages/cli/lib/LocalDevManager.js
+++ b/packages/cli/lib/LocalDevManager.js
@@ -40,6 +40,7 @@ const WATCH_EVENTS = {
   add: 'add',
   change: 'change',
   unlink: 'unlink',
+  unlinkDir: 'unlinkDir',
 };
 
 const UPLOAD_PERMISSIONS = {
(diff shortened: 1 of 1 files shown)


camden11 (approved):
Code LGTM! Trying to test out now but if you know it works feel free to merge if you're in a hurry

PR #884, opened 2023-06-30
Add option to turn on automatic uploads on prod accounts

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/419

This adds a third option to manual upload prompt that allows users to turn on automatic uploads for all future uploads. Wasn't exactly sure what we wanted the design for this to look like since it's not _really_ a prompt but took some guesses!

## Screenshots
Screenshot 2023-06-30 at 12 57 50 PM

Screenshot 2023-06-30 at 12 58 00 PM

## TODO
Maybe finalize design + copy

## Who to Notify


kemmerle (commented):
The code makes sense to me, but I think we need input from @markhazlewood before we can merge this. I'll ping him on Slack to get some eyes on this before @camden11 returns from vacation. 

@brandenrodgers

PR #885, opened 2023-07-10
Add project platform version to certain project commands (Other part in cli-lib)

kemmerle:
## Description and Context
This PR adds a project platform version to certain project commands, so we can pass that version off to BE endpoints (See https://github.com/HubSpot/cli-lib/pull/12):

`hs project dev`
`hs project watch`
`hs project upload`

We are using a new API endpoint `developer/projects/v1/platformVersion` to fetch the default project platform version. 

Furthermore, Caitlyn suggested that we add a message alerting the user if they haven't set a project version in their `hsproject.json` command. I've incorporated Mark's feedback about where to place this message and its contents.  

The [other half](https://github.com/HubSpot/cli-lib/pull/12) of this PR is located in the `cli-lib` repo.

To test out this PR, you will need to check out the [above branch](https://github.com/HubSpot/cli-lib/pull/12) in the `cli-lib` repo and then use the `yarn link @hubspot/cli-lib` command to use the symlinked local package. [Docs](https://github.com/HubSpot/hubspot-cli/blob/master/CONTRIBUTING.md#local-development-with-cli-lib) for more details. 

## Screenshots

With message alerting the user about setting a project platform version:

`hs project upload`:

Screenshot 2023-07-19 at 4 09 49 PM

`hs project watch`:

Screenshot 2023-07-19 at 4 10 11 PM

## TODO

- [x] Address any feedback about where to place the message about adding a project platform version. @markhazlewood 

## Who to Notify

@markhazlewood @brandenrodgers

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -18,6 +18,9 @@ const {
   SPINNER_STATUS,
 } = require('@hubspot/cli-lib/lib/constants');
 const {
+  fetchDefaultVersion,
+} = require('@hubspot/cli-lib/lib/projectPlatformVersion');
+const {
   createProject,
   getBuildStatus,
   getBuildStructure,
@@ -810,10 +813,15 @@ const createProjectComponent = async (component, name) => {
   );
 };
 
-const showPlatformVersionWarning = platformVersion => {
+const showPlatformVersionWarning = async (accountId, platformVersion) => {
   if (!platformVersion) {
+    const defaultVersion = await fetchDefaultVersion(accountId);
     logger.log('');
-    logger.log(i18n(`${i18nKey}.showPlatformVersionWarning.noPlatformVersion`));
+    logger.log(
+      i18n(`${i18nKey}.showPlatformVersionWarning.noPlatformVersion`, {
+        defaultVersion,
+      })
+    );
     logger.log('');
   }
 };
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -813,11 +813,13 @@ const createProjectComponent = async (component, name) => {
   );
 };
 
-const showPlatformVersionWarning = async (accountId, platformVersion) => {
+const showPlatformVersionWarning = async (accountId, projectConfig) => {
+  const platformVersion = projectConfig.platformVersion;
+
   if (!platformVersion) {
     const defaultVersion = await fetchDefaultVersion(accountId);
     logger.log('');
-    logger.log(
+    logger.warn(
       i18n(`${i18nKey}.showPlatformVersionWarning.noPlatformVersion`, {
         defaultVersion,
       })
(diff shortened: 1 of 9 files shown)


brandenrodgers:
Just leaving a comment so we don't forget to do it. We'll need to make these same changes in local dev lib as well so it doesn't get out of sync.

packages/cli/lib/projects.js
```diff
@@ -807,6 +810,14 @@ const createProjectComponent = async (component, name) => {
   );
 };
 
+const showPlatformVersionWarning = platformVersion => {
```
brandenrodgers, line 813: This util would probably be a little more useful if you set it up to just take in `projectConfig`. Then none of the consuming commands will be required to know the name of the `platformVersion` field. That would make it easier to change the name of the platform version field down the road if we ever decide to do that. There will be less placed in the code that directly reference it.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/885/commits/36ba46aa3a3f52236a7629e31d9c7167b626a156

packages/cli/lib/projects.js
```diff
@@ -807,6 +810,14 @@ const createProjectComponent = async (component, name) => {
...
+  if (!platformVersion) {
+    logger.log('');
+    logger.log(i18n(`${i18nKey}.showPlatformVersionWarning.noPlatformVersion`));
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/885/commits/36ba46aa3a3f52236a7629e31d9c7167b626a156

packages/cli/lib/projects.js
```diff
@@ -807,6 +813,21 @@ const createProjectComponent = async (component, name) => {
...
+    logger.log('');
+    logger.warn(
+      i18n(`${i18nKey}.showPlatformVersionWarning.noPlatformVersion`, {
+        defaultVersion,
+      })
+    );
+    logger.log('');
+  }
+};
+
```
camden11, line 830: Wondering if there's some way this could all be refactored to return the default platform version so we can pass it in to the functions that need it when `projectConfig.platformVersion` doesn't exist. That way, we won't need to make a second request to get the default version in cli-lib

kemmerle: I think we should keep the function that shows the warning and the function that returns the default version separate, just so we don't risk displaying the same message 15 times in the future. This is a straightforward API call that we're only using with three commands, so I don't think it's worth complicating the code. Going to resolve this, but happy to discuss over Slack.

camden11: That's fair, I agree that that wouldn't actually be the best way to do it. It just doesn't feel great making two extra requests every time you run these commands. Might even be worth exploring a way to just edit the config if you don't have a version specified to the request only needs to be made once. @brandenrodgers curious what your thoughts are here.

Anyway, don't need to over complicate things for now - i'll approve

camden11 (approved):

PR #887, opened 2023-07-11
Support uie dev server package (v2 behavior for dev mode)

brandenrodgers:
## Description and Context

This adds a new option to the `hs project dev` command that will make it easier for internal teams to iterate on the local dev package. Now we will be able to yarn link the WIP dev server package.

Steps to link:
1. Run `yarn link` in the dev server folder
2. Run `yarn link "@hubspot/ui-extensions-dev-server"` in the hubspot cli folder
3. Run `yarn install --force` in the hubspot cli folder
4. From the cli root folder, create a project inside of `tmp/` (create the tmp folder if necessary)
5. CD into the project, and run `yarn hs project dev --useDevServer --extension=<js-extension-filename>`

Steps to unlink:
1. Run `yarn unlink` in the dev server folder
2. Run `yarn unlink "@hubspot/ui-extensions-dev-server"` in the hubspot cli folder
3. Run `yarn install --force` in the hubspot cli folder

This also makes some feature changes to the dev mode command (only when you include the `--extension=` option. This only works via yarn linking for now because the latest version of the dev server package has not been published yet):
- No more usage of Spinnies to control the console output. We ran into too many issues with it. The main issue is that it's incredibly cpu intensive because it's constantly clearing and writing to the output stream every 50 ms. It also breaks catastrophically whenever the height of the output exceeds the height of the console
- We skip all uploads except for the situation where the project does not exist in the target account

### DIFF
--- a/packages/cli/lib/LocalDevManager.js
+++ b/packages/cli/lib/LocalDevManager.js
@@ -55,7 +55,7 @@ class LocalDevManager {
     this.projectConfig = options.projectConfig;
     this.projectDir = options.projectDir;
     this.extension = options.extension;
-    this.devServerPath = options.devServerPath;
+    this.useDevServer = options.useDevServer;
     this.uploadPermission =
       options.uploadPermission || UPLOAD_PERMISSIONS.always;
     this.debug = options.debug || false;
@@ -102,7 +102,7 @@ class LocalDevManager {
 
     await this.devServerStart();
 
-    if (!this.devServerPath) {
+    if (!this.useDevServer) {
       this.uploadQueue.start();
       await this.startWatching();
     } else {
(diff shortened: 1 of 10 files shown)


packages/cli/lib/LocalDevManager.js
```diff
@@ -1,5 +1,6 @@
 const chokidar = require('chokidar');
```
brandenrodgers, line 1: Keeping this file around makes it easy to maintain the existing functionality while we work to build out the v2 behavior that will eventually replace this. We fallback to using this version of the dev manager when the user does not pass in the `--extension` option.

camden11 (commented):
Code here all looks great to me! Having a little trouble actually testing - will talk to you tomorrow about that

camden11 (approved):

PR #889, opened 2023-07-17
[WIP] dev mode setup step for dev servers

brandenrodgers:
## Description and Context

This rips out the old behavior of `hs project dev`. We should only merge this when we are ready to fully make the jump to the new behavior.

This adds a new `setup` lifecycle method that the dev servers can use as a opportunity to prompt for any information that they might need to run. It passes down an object that contains useful project structure information. It looks like this:
```js
{
  <app_name>: {
    config: <parsed app config>,
    path: <path to app dir>
  }
}
```

It also passes down the [promptUser](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/prompts/promptUtils.js#L3) util function that we use in the CLI to prompt the user for input.

As part of this, I also simplified the amount of information that gets passed down to the servers on `start` because they should already have all of that information via the `setup` step.


brandenrodgers:
Closing this in favor of the backwards compatible replacement branch: https://github.com/HubSpot/hubspot-cli/pull/892

PR #891, opened 2023-07-20
Sandboxes: tracking intermediate actions with new action

adamawang:
## Description and Context

For actions that happen during a command, eg: syncing a sandbox in the sandbox create flow, we now track them using a new action name `cli-command-metadata`. 

Usages in this PR:
- User responds `yes` to a prompt for syncing a newly created sandbox within the `hs sandbox create` flow
- User creates a new sandbox within the `hs project dev` flow

I re-used the `step` metadata prop, but open to changing it to use another. 

Related sandboxes issue: https://git.hubteam.com/HubSpot/sandboxes-ui/issues/1930

## Screenshots

Sandbox sync in sandbox create:
Screenshot 2023-07-20 at 15 07 33

Sandbox create in project dev:
Screenshot 2023-07-20 at 15 08 23

### DIFF
--- a/packages/cli/lib/usageTracking.js
+++ b/packages/cli/lib/usageTracking.js
@@ -7,6 +7,30 @@ const { version } = require('../package.json');
 const { getPlatform } = require('./environment');
 const { setLogLevel } = require('./commonOpts');
 
+/* **
+Available tracking meta properties
+
+  ImmutableMap
+    .<EventProperty, String>builder()
+    .put(EventProperty.ACTION, "The specific action taken in the CLI")
+    .put(EventProperty.OS, "The user's OS")
+    .put(EventProperty.NODE_VERSION, "The user's version of node.js")
+    .put(EventProperty.NODE_MAJOR_VERSION, "The user's major version of node.js")
+    .put(EventProperty.VERSION, "The user's version of the CLI")
+    .put(
+      EventProperty.COMMAND,
+      "The specific command that the user ran in this interaction"
+    )
+    .put(EventProperty.AUTH_TYPE, "The configured auth type the user has for the CLI")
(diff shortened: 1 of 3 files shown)


packages/cli/lib/usageTracking.js
```diff
@@ -7,6 +7,25 @@ const { version } = require('../package.json');
...
+  - authType: "The configured auth type the user has for the CLI"
+  - step: "The specific step in the auth process"
+  - assetType: "The CMS asset type"
+  - mode: "The CLI mode (draft or publish"
+  - type: "The upload type (file or folder)"
+  - file: "Whether or not the 'file' flag was used"
+  - successful: "Whether or not the CLI interaction was successful"
+
+*/
+
```
adamawang, line 28: I figure this could be useful to know what properties are available to pass through to the tracker, let me know what you think!

packages/cli/lib/usageTracking.js
```diff
@@ -132,10 +151,48 @@ const trackAuthAction = async (command, authType, step, accountId) => {
   }
 };
 
+function trackCommandMetadataUsage(command, meta = {}, accountId) {
```
adamawang, line 154: Elected to create a new function rather than adjusting `trackingCommandUsage`, but if you feel its easier to re-use that one let me know

packages/cli/lib/usageTracking.js
brandenrodgers: This is great, thanks!

packages/cli/lib/usageTracking.js
brandenrodgers: I like this 👌

brandenrodgers (approved):

PR #892, opened 2023-07-21
Dev Mode - support setup step

brandenrodgers:
## Description and Context

This PR is replacing #889 . This one is backwards compatible with the existing dev mode command behavior.

There are two new hidden options:
- `--local` - runs the local dev server for ui extensions
- `--local-all` - runs the local dev server for ui extensions and app functions

When `--local-all` is specified, we pass `alpha: true` to the `setup` and `start` lifecycle methods for the DevServerManager.

This also adds a new `setup` lifecycle method that the dev servers can use as a opportunity to prompt for any information that they might need to run. It passes down an object that contains useful project structure information. It looks like this:
```js
{
  <app_name>: {
    config: <parsed app config>,
    path: <path to app dir>
  }
}
```

It passes down the [promptUser](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/prompts/promptUtils.js#L3) util function that we use in the CLI to prompt the user for input.

As part of this, I also simplified the amount of information that gets passed down to the servers on `start` because they should already have all of that information via the `setup` step.

### DIFF
--- a/packages/cli/lib/DevServerManager.js
+++ b/packages/cli/lib/DevServerManager.js
@@ -1,28 +1,35 @@
 const express = require('express');
 const bodyParser = require('body-parser');
 const cors = require('cors');
-const { walk } = require('@hubspot/cli-lib/lib/walk');
+const httpClient = require('@hubspot/cli-lib/http');
+const { logger } = require('@hubspot/cli-lib/logger');
 const { getProjectDetailUrl } = require('./projects');
+const { COMPONENT_TYPES } = require('./projectStructure');
 const { i18n } = require('./lang');
 const { EXIT_CODES } = require('./enums/exitCodes');
-const { logger } = require('@hubspot/cli-lib/logger');
+const { promptUser } = require('./prompts/promptUtils');
 
 const i18nKey = 'cli.lib.DevServerManager';
 
 const DEFAULT_PORT = 8080;
-
 class DevServerManager {
   constructor() {
     this.initialized = false;
+    this.started = false;
+    this.componentsByType = {};
     this.server = null;
     this.path = null;
     this.devServers = {};
+    this.debug = false;
   }
 
   safeLoadServer() {
     try {
       const { DevModeInterface } = require('@hubspot/ui-extensions-dev-server');
-      this.devServers['uie'] = DevModeInterface;
+      this.devServers['uie'] = {
+        componentType: COMPONENT_TYPES.app,
+        serverInterface: DevModeInterface,
+      };
     } catch (e) {
       logger.debug('Failed to load dev server interface: ', e);
     }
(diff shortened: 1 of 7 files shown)


packages/cli/lib/DevServerManager.js
```diff
@@ -33,75 +40,99 @@ class DevServerManager {
...
     for (let i = 0; i < serverKeys.length; i++) {
       const serverKey = serverKeys[i];
-      const serverInterface = this.devServers[serverKey];
-      await callback(serverInterface, serverKey);
+      const devServer = this.devServers[serverKey];
+
+      const compatibleComponents =
+        this.componentsByType[devServer.componentType] || {};
+
+      if (Object.keys(compatibleComponents).length) {
```
brandenrodgers, line 48: I am now only calling the lifecycle methods for the dev server interfaces if there are available components for the server to interact with.

packages/cli/lib/LocalDevManagerV2.js
```diff
@@ -31,14 +38,51 @@ class LocalDevManagerV2 {
...
+
+    const componentTypes = Object.keys(componentsByType);
+
+    if (!componentTypes.length) {
+      logger.log();
+      logger.error(i18n(`${i18nKey}.noComponents`));
+      process.exit(EXIT_CODES.SUCCESS);
+    }
+
+    const runnableComponentsByType = componentTypes.reduce((acc, type) => {
```
brandenrodgers, line 54: `runnableComponentsByType` is a subset of the `componentsByType` object. It only includes components that have `runnable` set to true. For apps, this means it only includes apps that contain react extensions.

packages/cli/lib/projectStructure.js
```diff
@@ -0,0 +1,92 @@
...
+        );
+        const isLegacy = getIsLegacyApp(parsedAppConfig, appPath);
+
+        if (!componentsByType[COMPONENT_TYPES.app]) {
+          componentsByType[COMPONENT_TYPES.app] = {};
+        }
+
+        componentsByType[COMPONENT_TYPES.app][parsedAppConfig.name] = {
+          config: parsedAppConfig,
+          runnable: !isLegacy,
```
brandenrodgers, line 79: Using this `runnable` field lets us log more useful information to the user on start up.

[no components in the project] - "This project has no components. Exiting"
[no runnable components in the project] - "This project has no components that support local development. Exiting"

packages/cli/lib/DevServerManager.js
```diff
@@ -1,28 +1,35 @@
...
     this.path = null;
     this.devServers = {};
+    this.debug = false;
   }
 
   safeLoadServer() {
     try {
       const { DevModeInterface } = require('@hubspot/ui-extensions-dev-server');
-      this.devServers['uie'] = DevModeInterface;
+      this.devServers['uie'] = {
```
brandenrodgers, line 29: I'm grouping servers with component types. This makes it easier to know whether or not we should even bother calling the lifecycle methods for certain servers.

camden11: I know this isn't new code but seems like it could be a good candidate for a constant / enum. `SERVER_KEYS.UIE` or something

camden11 (approved):
This all looks great - just had one minor suggestion (which is not really related so feel free to ignore)

packages/cli/lib/DevServerManager.js
brandenrodgers: 👍 done in [0df9f69](https://github.com/HubSpot/hubspot-cli/pull/892/commits/0df9f69e75b68e378f92bdb5e4bc291434a94743)

PR #893, opened 2023-07-21
Unhide project commands and account for different errors (other half in cli-lib)

kemmerle:
## Description and Context
Since Developer Projects is now in public beta, we can unhide project commands, give them descriptions, and link to our documentation. 

In this PR, I've unhidden and added descriptions for certain project commands. I've also added some better handling for gating errors, and I'm seeking feedback about downgrading errors. 

In order to fully test out the error handling for gating, you will need to check out [the appropriate branch](https://github.com/HubSpot/cli-lib/pull/17) in the cli-lib repo and then use the yarn link @hubspot/cli-lib command to use the symlinked local package. [Docs](https://github.com/HubSpot/hubspot-cli/blob/master/CONTRIBUTING.md#local-development-with-cli-lib) for more details.

## Screenshots

New project commands help menu (revised and with beta tags):

Screenshot 2023-07-25 at 2 01 47 PM

New error message when there's a gating error (revised):

Screenshot 2023-07-24 at 6 07 26 PM

## TODO

- [x] Address any feedback from Mark. 
- [x] Add handling for downgrading errors if necessary (not necessary at the moment).

@markhazlewood, thanks for your feedback! 

## Who to Notify

@markhazlewood @brandenrodgers @camden11

### DIFF
--- a/packages/cli/bin/cli.js
+++ b/packages/cli/bin/cli.js
@@ -115,7 +115,7 @@ const argv = yargs
   .option('debug', {
     alias: 'd',
     default: false,
-    describe: 'set log level to debug',
+    describe: 'Set log level to debug',
     type: 'boolean',
   })
   .option('noHyperlinks', {
--- a/packages/cli/commands/project.js
+++ b/packages/cli/commands/project.js
@@ -21,16 +21,16 @@ exports.builder = yargs => {
   addAccountOptions(yargs, true);
 
   // TODO: deploy must be updated
-  yargs.command(deploy).demandCommand(1, '');
   yargs.command(create).demandCommand(0, '');
-  yargs.command(upload).demandCommand(0, '');
+  yargs.command(add).demandCommand(0, '');
   yargs.command(watch).demandCommand(0, '');
-  yargs.command(listBuilds).demandCommand(0, '');
+  yargs.command(dev).demandCommand(0, '');
+  yargs.command(upload).demandCommand(0, '');
+  yargs.command(deploy).demandCommand(1, '');
   yargs.command(logs).demandCommand(1, '');
+  yargs.command(listBuilds).demandCommand(0, '');
   yargs.command(download).demandCommand(0, '');
   yargs.command(open).demandCommand(0, '');
-  yargs.command(dev).demandCommand(0, '');
-  yargs.command(add).demandCommand(0, '');
 
   return yargs;
 };
--- a/packages/cli/commands/project.js
+++ b/packages/cli/commands/project.js
@@ -1,4 +1,5 @@
 const { addConfigOptions, addAccountOptions } = require('../lib/commonOpts');
+const { i18n } = require('../lib/lang');
 const deploy = require('./project/deploy');
 const create = require('./project/create');
 const upload = require('./project/upload');
@@ -10,8 +11,10 @@ const open = require('./project/open');
 const dev = require('./project/dev');
 const add = require('./project/add');
 
+const i18nKey = 'cli.commands.project';
+
 exports.command = 'project';
-exports.describe = false; //'Commands for working with projects';
+exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.builder = yargs => {
   addConfigOptions(yargs, true);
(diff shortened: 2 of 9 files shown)


brandenrodgers:
Code lgtm! 👌

markhazlewood42:
Thanks for the updates @kemmerle! I have a bunch of comments + questions. In no particular order:

* What determines the order of subcommands in the help text? It seems somewhat arbitrary right now, or at least I can't tell if it's following some pattern. Would be great if we can list them in some meaningful way, like roughly the order you'd normally use/encounter them in (create, add, watch, dev, upload, deploy, etc ...) But if that's not possible, maybe at least alphabetical.

* Does `hs project add` actually work as expected right now? I know we were working on those commands but I don't think I've ever used them.

* `open` should probably say something about opening the projects details page in your browser.

* Could we not advertise the `--portal` or `-p` arguments? Those go against approved vocab guidelines.

* In the error message, the link needs to be an anonymized account link without an account ID. Looks like maybe you pasted in your own. Needs to be exactly this: `https://app.hubspot.com/l/whats-new/betas?productUpdateId=13860216`, which will then present the person with an account picker if they have more than one account.

* Also in the error message, why does it lead with "Your production account ..."? Wouldn't it be for any account type? If so, change that to, "The current target account ([number]) doesn't have access ... ". The rest of that message looks good!

* `set log level to debug` should use sentence casing like the other descriptions.
----

On the "downgrading" scenario ... I have thoughts too. 

In theory the projects framework is open for any kind of development (or should be). It's at the *feature-level* where we need to be imposing restrictions -- so if someone tries to build/deploy a project with features that aren't supported by the target account, it will fail. This is all in theory and IMO the way things should be working, but we haven't really built support for this yet since the main available features we're focused on right now (CRM cards) require those enterprise tiers. So we just block the whole thing if you don't have that (with some exceptions).

So all that said, I think in the scenario you illustrated that error is showing up at the right place (trying to build a project with features that aren't supported by the target account). But it's probably happening by accident. And the messaging isn't great. Really that second error is the important one. So we should discuss with the rest of the team + component team(s) the right way to handle this.

kemmerle:
Hi, @markhazlewood42! Thanks for the comments, they were super helpful! I'm going to answer your questions by bullet point. 

1) _What determines the order of subcommands in the help text?_ The order of subcommands is determined by their order in the project command file: https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/commands/project.js. I've re-ordered the subcommands in that file in the order the customer would usually encounter them, as you suggested.

2) _Does hs project add actually work as expected right now?_ Yes, `hs project add` does work. It can add apps (either in the middle or side panel of the CRM record) to a project folder. 

3) _open should probably say something about opening the projects details page in your browser._ I've made this change 👍 

4) _Could we not advertise the --portal or -p arguments? Those go against approved vocab guidelines._ I need to check in with the FE team about whether we can hide the `portal` options here. Currently, `portal` is defined as the flag, while `p`, `account`, and `a` are its aliases. I could hide `portal` and `p` by breaking them into their own separate flag, but I'm not sure whether that would constitute a breaking change. @brandenrodgers @camden11, would appreciate your input. 

5) Corrected the error message according to your suggestions 👍 

6) _set log level to debug should use sentence casing like the other descriptions._ Fixed sentence casing 👍 

--

Regarding the downgrading error message, that sounds like it falls outside the scope of this PR at the moment. I'll mark it down for the UX sync and get into contact with the components team.

brandenrodgers:
> Could we not advertise the --portal or -p arguments? Those go against approved vocab guidelines. I need to check in with the FE team about whether we can hide the portal options here. Currently, portal is defined as the flag, while p, account, and a are its aliases. I could hide portal and p by breaking them into their own separate flag, but I'm not sure whether that would constitute a breaking change. @brandenrodgers @camden11, would appreciate your input.

Yeah I'd actually vote to keep the portal option around. That would be another breaking change, and I'd like to minimize the number of big changes that we're making in this. Part of the work that @camden11 is doing with the new library will make changes to the config file. One of those changes is renaming `portalId` to `accountId` in the config. Whenever we roll that out would probably also be a good time to remove `portalId` as a valid option.

brandenrodgers (approved):

PR #894, opened 2023-07-21
Dev Mode - support upload warning messages

brandenrodgers:
## Description and Context

This adds a util to log consistently-formatted upload warnings to the user whenever a manual upload is recommended. I'm doing one initial check to make sure that the local version of the project matches the deployed version of it. Without that check you can get into some confusing situations where it looks like everything is running fine, but there is nothing showing up in HubSpot.

How to get the initial startup warning to show:
1. Upload a project to HubSpot
2. Locally, change the title of an extension, but do not upload the change
3. Run the dev mode command

This also passes down a `onUploadRequired` callback to the dev servers in the `setup` method. They can use this whenever they want to tell the user to manually upload. This helps us keep the message formatting consistent.
## Screenshots

**Initial startup message:**
image

**Example message coming from a dev server:**
image

### DIFF
--- a/packages/cli/lib/LocalDevManagerV2.js
+++ b/packages/cli/lib/LocalDevManagerV2.js
@@ -50,39 +50,26 @@ class LocalDevManagerV2 {
     SpinniesManager.removeAll();
     SpinniesManager.init();
 
-    const componentsByType = await findProjectComponents(this.projectSourceDir);
+    const components = await findProjectComponents(this.projectSourceDir);
 
-    const componentTypes = Object.keys(componentsByType);
-
-    if (!componentTypes.length) {
+    if (!components.length) {
       logger.log();
       logger.error(i18n(`${i18nKey}.noComponents`));
       process.exit(EXIT_CODES.SUCCESS);
     }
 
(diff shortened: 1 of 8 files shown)


packages/cli/lib/LocalDevManagerV2.js
```diff
@@ -133,12 +131,64 @@ class LocalDevManagerV2 {
...
+          accountId: this.targetAccountId,
+        })
+      );
+    } else {
+      logger.log(i18n(`${i18nKey}.uploadWarning.runUpload`));
+    }
+    logger.log(i18n(`${i18nKey}.uploadWarning.restartDev`));
+  }
+
+  compareLocalProjectToDeployed(runnableComponents) {
```
brandenrodgers, line 153: This compares the contents of `deployedBuild.subbuildStatuses` to see if we have any components or subcomponents locally that don't exist in the deployed build. Currently this is only checking for CRM cards, but we can eventually expand it to check for more than that if we want.

camden11 (approved):
Looks good to me, tested locally successsfully!

kemmerle (approved):
Reviewed the code and was able to test locally (thanks to your help!), and LGTM 🚢

Issue #744, opened 2022-09-15
Add --remove arg to hs upload


johnsfuller:
## Description and Context

Similar to the `hs watch --remove` [command](https://developers.hubspot.com/docs/cms/developer-reference/local-development-cli#watch), a `--remove` flag would be a nice addition to the `hs upload` command. 

It would ideally delete missing files from the design manager. Maybe even an additional argument could give it some safe guards (i.e. `--hard` or `--soft`).
```
hs upload --remove --hard  # removes / deletes even if dependencies exist for a file
hs upload --remove --soft  # removes / deletes if no dependencies or warnings exist for a file
```

It could even be used in [HubSpot's CMS Deploy GitHub Action](https://github.com/HubSpot/hubspot-cms-deploy-action/) and resolve [one of it's issues](https://github.com/HubSpot/hubspot-cms-deploy-action/issues/10).

Issue #361, opened 2020-10-14
dependency command


TheWebTech:
This idea is really @ajlaporte's. 

A command that can list the dependencies for a module/template/file. Just like what shows within the design manager.
Just brings us closer to parity between the design manager and CLI functionality. Because the functionality is in the design manager this is not an urgent need.

Proposed command:
`hs dependents <file/path>`
file/path being within the account's developer file system.

It would need to work for multiple asset types and that makes things a little tricky:
- modules
- templates
- themes
- files

TanyaScales:
Adding onto this -- it would be great to bubble up dependency information within the watch/upload commands if files with dependents are going to be removed locally and that removal is going to be pushed remotely to the Design Manager.

chiragchadha1:
Thanks for the suggestion @TheWebTech. Going to close this one out. It's been open for 5 years without further demand and dependency info remains available in the Design Manager. Feel free to re-open if a CLI version of this becomes important to your workflow. Thanks!

PR #895, opened 2023-07-24
Add overwrite option to hs upload

j-malt:
## Description and Context

Addresses #744 by adding `--overwrite`. Deletes the remote copy of dest directory, and then uploads the source, effectively acting as an "overwrite".

## Who to Notify
@HubSpot/cms-developer-tooling


packages/cli/commands/upload.js
```diff
@@ -204,6 +209,33 @@ exports.handler = async options => {
       absoluteSrcPath,
       options.convertFields
     );
+
+    if (options.remove) {
+      //  If remove is true, will first delete the dest folder and then upload src. Cleans up files that only exist on HS.
+      const remoteFileList = await getDirectoryContentsByPath(accountId, dest);
```
jsines, line 215: Hmm I'm a little iffy on this prompting strategy, leads to kinda weird scenario like here - I'm prompted for if I want to delete `package.json` specifically, and if I say no, it's still overwritten, so my answer kinda didn't matter...

Screenshot 2023-07-28 at 2 26 07 PM

I think we might be OK, instead of getting directory contents from the dest and prompting about deleting each, to just prompt about deleting the contents at `dest` once for the whole directory and blasting the whole thing in one go. 

FWIW, the `--remove` on `hs watch` doesn't prompt at all. So I don't think we need to be _too_ in depth with this

packages/cli/lang/en.lyaml
```diff
@@ -1115,6 +1121,8 @@ en:
           errors:
             buildIdDoesNotExist: "Build {{ buildId }} does not exist for project {{ projectName }}."
             buildAlreadyDeployed: "Build {{ buildId }} is already deployed."
+        deletePrompt:
+          message: Are you sure you want to remove {{ filePath }} from HubSpot?
```
jsines, line 1125: Shouldn't this need double quotes? It's working somehow without them on my machine!? Not sure how the parsing logic for this is done but maybe they're not necessary? Haha

packages/cli/commands/upload.js
j-malt: Yeah, I was thinking of using "overwrite" for naming as opposed to remove, but it's also kind of awkward.

I'm fine with just prompting once before the command is run. Probably can just remove `--force` in that case.

packages/cli/lang/en.lyaml
j-malt: I have no clue how I missed the double quotes but apparently [YAML doesn't need quotes at all](https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml) lol. Will fix for consistency.

jsines: TIL!

packages/cli/lang/en.lyaml
```diff
@@ -1115,6 +1119,8 @@ en:
           errors:
             buildIdDoesNotExist: "Build {{ buildId }} does not exist for project {{ projectName }}."
             buildAlreadyDeployed: "Build {{ buildId }} is already deployed."
+        overwritePrompt:
+          message: "Are you sure you want to overwrite {{ filePath }} on HubSpot?"
```
kemmerle, line 1123: Would it clearer to say "in your HubSpot account #{{accountId}}"?

kemmerle (approved):
Reviewed the code and tested out these changes locally, and everything LGTM, besides the one copy suggestion I made. @jsines, do you also want to take a final look at this?

packages/cli/commands/upload.js
TanyaScales: cc @markhazlewood  -- we were discussing this internally but wanted to get your opinion on the language used here for the option `--overwrite` as opposed to other suggestions that you might have for the functionality. We don't have a designer on our side and since you were heading up the CLI guidelines originally, I thought you'd be a good confirmation here.

markhazlewood42: @TanyaScales Thanks for looping me in! Not to throw a wrench in, but when I see "removes the original and completely overwrites it with this one" my mind immediately goes to `--clean`. A "clean" operation (clean build, or clean install, etc.) usually means to completely remove the current one and replace it, rather than trying to merge them together.

jsines: Hmm yeah overwrite is a little awkward too. When I hear it my assumption is that it overwrites the remote files with my versions, not deletes ones that aren't there. I think we should keep it as `--remove` to remain in line with `hs watch`'s option since it's effectively the same behavior. Some of the copy for that can maybe be borrowed.

packages/cli/lang/en.lyaml
```diff
@@ -792,13 +792,16 @@ en:
...
+          deleteFailed: "Deleting \"{{ path }}\" from account {{ accountId }} failed"
         options:
           options:
             describe: "Options to pass to javascript fields files"
           saveOutput:
             describe: "If true, saves all output from javascript fields files as 'fields.output.json'."
           convertFields:
             describe: "If true, converts any javascript fields files contained in module folder or project root."
+          overwrite:
+            describe: "Overwrites the destination folder."
```
jsines, line 804:
```suggestion
            describe: "Will cause upload to delete files in your HubSpot account that are not found locally."
```
Stole wording from the `hs watch` arg

packages/cli/commands/upload.js
```diff
@@ -204,6 +205,24 @@ exports.handler = async options => {
       absoluteSrcPath,
       options.convertFields
     );
+
+    if (options.overwrite) {
+      //  If overwrite is true, will first delete the dest folder and then upload src. Cleans up files that only exist on HS.
+      const overwriteFile = await overwriteFilePrompt(dest);
```
jsines, line 211: If we do go with prompting here I would want a `--force` option for CI/CD usages. But since `hs watch` doesn't prompt like this maybe we just shouldn't? @TanyaScales ?

packages/cli/lang/en.lyaml
```diff
@@ -810,6 +813,7 @@ en:
           uploadComplete: "Uploading files to \"{{ dest }}\" in the Design Manager is complete"
         uploading: "Uploading files from \"{{ src }}\" to \"{{ dest }}\" in the Design Manager of account {{ accountId }}"
         notUploaded: "There was an error processing \"{{ src }}\". The file has not been uploaded."
+        overwriting: "Overwriting \"{{ path }}\" on account {{ accountId }}..."
```
jsines, line 816: Some of the verbiage from `cli-lib`: https://github.com/HubSpot/cli-lib/blob/f2283d961c2a12e3e78793dc6991d0f47e34caba/lib/watch.js#L202-L221

packages/cli/commands/upload.js
jsines: Actually, the `--remove` on `hs watch` functions slightly differently doesn't it! My assumption without checking code was that it removes on the initial upload _AND_ any files you delete while its running but it actually only does the second. Should we change that?

packages/cli/lang/en.lyaml
TanyaScales: I do think we need to specify in the language that this is the remote file -- So, this suggestion makes sense to me. 👍

packages/cli/commands/upload.js
TanyaScales: (If I'm reading this right) For the `--remove` flag on watch -- if you also flag `--initial-upload` are you saying that files that don't exist locally aren't removed on that initial upload to the DM?

Also, the suggestion here is to have `hs watch --remove` and `hs upload --remove` to behave the same -- With the obvious difference being the underlying function of the main command? And that flag behavior being -- removal of files remotely that don't exist locally.

Additionally, using the `--initial-upload` flag paired with the `--remove` flag on the `watch` command would then do, basically, an "overwrite/clean" of the dest (mirroring `hs upload --remove`).

I think I am for making the removal of files behave the same across the board in the two commands. Purely because I think that it keep consistency across the cli in what that particular word means for file management.

packages/cli/commands/upload.js
TanyaScales: I'm slightly terrified of developers deleting their entire dest source because they mess something up locally -- I hear that we aren't prompting on watch. So, I am conflicted. If we were to remove prompting on each file -- which I do agree could be verbose in some cases -- I would almost want a confirmation prompt initially. 

`You are about to remove any remote files on your HubSpot account ${accountName} at ${filepath} that don't exist locally. Are you sure you want to do this? y/N`

^ Does that seem like a fair compromise?
---
Could also use a `--force` flag in this case for CI/CD usage

packages/cli/commands/upload.js
jsines: >For the --remove flag on watch -- if you also flag --initial-upload are you saying that files that don't exist locally aren't removed on that initial upload to the DM?

Yep, never noticed it before while working on CAT but I figured it from reading the code - just ran it on growth theme to check: 
Screenshot 2023-08-14 at 11 50 04 AM
Screenshot 2023-08-14 at 11 58 08 AM

>Also, the suggestion here is to have hs watch --remove and hs upload --remove to behave the same -- With the obvious difference being the underlying function of the main command? And that flag behavior being -- removal of files remotely that don't exist locally.
>Additionally, using the --initial-upload flag paired with the --remove flag on the watch command would then do, basically, an "overwrite/clean" of the dest (mirroring hs upload --remove).
>I think I am for making the removal of files behave the same across the board in the two commands. Purely because I think that it keep consistency across the cli in what that particular word means for file management.

Yeah this is my feeling. If the two are going to have the same argument name they should either both do a "full clean" of the remote directory, or if they're going to be different then have different names. 

Is it possible we'd be breaking something by changing this? I wonder if there are devs that rely on it behaving like this & keep certain files in their DFS but not locally and just expect them to stay there. Maybe pretty niche use case.

Also, should it only do the initial removal if you use both `--initial-upload` AND `--remove`? Or is just the `--remove` enough? Is there a better way to distinguish these behavior options than combinations of these two args?

TanyaScales: > Is it possible we'd be breaking something by changing this? I wonder if there are devs that rely on it behaving like this & keep certain files in their DFS but not locally and just expect them to stay there. Maybe pretty niche use case.

This would be my concern about changing it -- however, we can be pretty brazen about communicating the change in a version update, I think. Those developers should probably not be using the `--remove` flag in that case altogether. (IMO)...

I agree with pairing down to one argument that does _one_ thing -- not two different scenarios.

> Also, should it only do the initial removal if you use both --initial-upload AND --remove? Or is just the --remove enough? Is there a better way to distinguish these behavior options than combinations of these two args?

Is there a better way to distinguish these? That is a good question. I do feel like there isn't a great combinator of two, maybe `hs watch --clean` like Mark was suggesting. That would mean it could upload all files and remove the ones that aren't present locally. But, that leaves the door open to having `hs watch --initial-upload` still exist for those folks that don't want to use the removal feature. And, to me, the two args `--initial-upload` and `--remove` are a little more self-explanatory about what is going to happen?

TanyaScales: Also, on another note, a question: Should we be moving forward with this as it is and determine some changes that make sense to the CLI in subsequent issues? I feel like there are quite a few things that we could determine need refactoring or more thought put into if we really wanted to. The current solution does solve the issue at hand -- even if it is a bit disjointed. Without having to go into multiple refactors during this PR -- maybe we should consider a different issue to bring the two removal options more in line with each other?

packages/cli/lang/en.lyaml
TanyaScales: If we refactor the option to prompt once at the start of the command -- then we can probably change this to "Are you sure you want to overwrite all files at `{{ filePath }}` in your HubSpot account `#{{ accountID }}`?

TanyaScales:
My one question about the overall deletion of files for this method is -- what if a file has dependencies? And if the current `watch` command doesn't protect against a developer deleting a file that has dependencies -- are we okay with that? This, again, could be address subsequently. However, its brought up in the original issue and I think it is something worth discussing and at least making a call to revisit.

j-malt:
> What if a file has dependencies?

For the case of `upload`, I think the fact that any remote deletion first requires the file to be deleted locally means we shouldn't really need to worry about it. If the dev deletes something that has dependencies, they'll see errors locally (and also have their local recycling bin to revert the delete if they do end up uploading with doing the remove upload). I'm personally fine with the current behaviour `watch`.  

Long-term, if we start doing FieldsJS dependency tracking like Jessica and I were discussing for FieldsJS watch, maybe we could abstract that tooling to provide some alerting here, but really I think it's quite low-priority.

The original idea of a "hard delete" that deletes all dependencies seems like too niche of a use case for something that could be potentially pretty catastrophic. Not sure if we should be giving that level of foot-gun.  

> Clean vs Remove

For the reason Mark gave of this being similar in functionality to a clean build/install, I really like `--clean`. I prefer it over overwrite, for sure. I _don't_ think we should be changing the current behaviour of `watch --remove`, and since this one behaves differently, I think it ought to have a different name. So, my vote is for "clean". I also like the idea of adding a `--clean` param to hs watch that does an initial removal. 

> Force command

I agree with prompting once with the string [Tanya gave](https://github.com/HubSpot/hubspot-cli/pull/895#discussion_r1293534977), and adding back the `--clean --force` command for CI/CD.

> Waiting for a refactor

IMO it's fine to merge this in now and move stuff around when the time comes for a refactor.

TanyaScales:
> For the case of `upload`, I think the fact that any remote deletion first requires the file to be deleted locally means we shouldn't really need to worry about it. If the dev deletes something that has dependencies, they'll see errors locally (and also have their local recycling bin to revert the delete if they do end up uploading with doing the remove upload). I'm personally fine with the current behaviour `watch`.

Ah, when I say dependencies -- What I was actually referring to was assets built with the templates/modules -- Ex. If I had an about.html template in the Design Manager that 3 live pages are built from and I go and delete the file locally, which deletes the remote/DM file -- the live pages are 💥 for editing.

I suppose since we have the removal feature already on the watch command, that this hasn't been an issue thus far with developers? It still might be good for future thought to think about how to bubble this information up locally as a preventative.

j-malt:
>  What I was actually referring to was assets built with the templates/modules

Oops, that makes way more sense. Yeah I totally agree that we should think about how to get that info to devs in the future.

TanyaScales:
Just added a comment to an existing issue [here](https://github.com/HubSpot/hubspot-cli/issues/361) about dependencies

jsines (approved):
Looks good! 🎉

packages/cli/lang/en.lyaml
```diff
@@ -1119,8 +1121,8 @@ en:
           errors:
             buildIdDoesNotExist: "Build {{ buildId }} does not exist for project {{ projectName }}."
             buildAlreadyDeployed: "Build {{ buildId }} is already deployed."
-        overwritePrompt:
-          message: "Are you sure you want to overwrite {{ filePath }} on HubSpot?"
+        cleanUploadPrompt:
+          message: "You are about to remove any remote files in \"{{ filePath }}\" on HubSpot account {{ accountId }} that don't exist locally. Are you sure you want to do this? y/N"
```
TanyaScales, line 1125: Nit: remove y/N text in this string

TanyaScales (approved):

packages/cli/lang/en.lyaml
jsines: I think it's helpful to give an indicator what the default is

j-malt: Looks like the prompting lib already supplies the y/N so no need to have it in the string. Removed.

image

jsines: Oh interesting! My brain must have trimmed that out while testing 😆

PR #896, opened 2023-07-26
`hs accounts clean` to remove invalid accounts from config

adamawang:
## Description and Context

Now that LocalDevAuth returns an invalid portal error with more detailed info, we can utilize this error for a new command `hs accounts clean` which will loop through all accounts and check for a refresh token. If an invalid portal error is caught, we add that portal to a list and prompt the user for removal from the CLI config. 

These errors should only pop up if a user deletes a portal from the UI but the config entry still exists for the CLI. Users may not know exactly which portals have been deleted, so this command should be able to help clean things up. This command does not clean up portals that you have been removed from (admin removed you or archived you).

NOTE: there are some issues for internal users on the VPN regarding EU1 hublet portals. When on the VPN, any portals on EU1 may be marked for deletion as it cannot find it on the correct hublet. Users can run the command with an api override to get around this like so: `HUBAPI_DOMAIN_OVERRIDE=api.hubspot hs accounts clean`. If you're not on the VPN, this is not an issue. 

I added a `--qa` flag for internal users where if ran with the flag, it will only check for QA portals. No flag will only check non-qa marked portals. However, when running the command with the `HUBAPI_DOMAIN_OVERRIDE` set to `api.hubspot`, it will think QA portals are all invalid since those require `api.hubspotqa` as the value. Its a bit confusing, but since these are all issues for internal users, @brandenrodgers and I thought it would still be useful to the vast majority of users. 

Let me know your thoughts! 

## Screenshots

If no accounts are found for deletion:
Screenshot 2023-07-26 at 12 11 15

If accounts are found, we add it to a list and prompt the user for confirmation before removing: 
Screenshot 2023-07-26 at 12 11 34

Help entry:
Screenshot 2023-07-26 at 12 38 10

## TODO

~~PENDING a QA and prod deploy from BE for these new error messages so I will leave this in draft for now.~~ 
Once it is deployed, I'll get some test scenarios up for all of us to confirm the usage.

In the future we may want to use some sort of hublet aware endpoints for internal VPN users to avoid the overrides.

## Who to Notify

@brandenrodgers @kemmerle @camden11


packages/cli/commands/accounts/clean.js
```diff
@@ -0,0 +1,120 @@
...
+  SpinniesManager.init({
+    succeedColor: 'white',
+  });
+  SpinniesManager.add('accountsClean', {
+    text: i18n(`${i18nKey}.loading.add`),
+  });
+
+  for (const account of filteredTestPortals) {
+    try {
+      await accessTokenForPersonalAccessKey(account.portalId);
```
adamawang, line 56: PAK validation will fail after authdata expires. I figure we use this instead of hitting the refresh token API every time, but I had another version that fetched tokens directly. Let me know if you have any thoughts on this

packages/cli/commands/accounts/clean.js
```diff
@@ -0,0 +1,120 @@
...
+    } catch (error) {
+      if (error && error instanceof HubSpotAuthError) {
+        if (
+          error.statusCode === 404 &&
+          error.subCategory === 'LocalDevAuthErrorType.INVALID_PORTAL_ID'
+        ) {
+          portalsToRemove.push(account);
+        }
```
adamawang, line 64: We only check for this specific error of INVALID_PORTAL_ID

packages/cli/commands/accounts/clean.js
```diff
@@ -0,0 +1,124 @@
...
+    } catch (error) {
+      if (error && error instanceof HubSpotAuthError) {
+        if (
+          (error.statusCode === 401 &&
+            error.category === 'INVALID_AUTHENTICATION' &&
+            error.subCategory === 'LocalDevAuthErrorType.PORTAL_NOT_ACTIVE') ||
+          (error.statusCode === 404 &&
+            error.category === 'INVALID_AUTHENTICATION' &&
+            error.subCategory === 'LocalDevAuthErrorType.INVALID_PORTAL_ID')
+        ) {
```
adamawang, line 66: We handle two separate errors for invalid portals (dont exist at all) and inactive portals (archived but not deleted yet)

Error details here: https://git.hubteam.com/HubSpotProtected/LocalDevAuth/pull/138

adamawang:
For testing:
- In the CLI, create a standard sandbox account using `hs sandbox create` in either QA or PROD. Doesnt matter if you sync to it or not.
- Once created, go into the UI and delete the sandbox from the list. 
- This part can take a while, but since we use `accessTokenForPersonalAccessKey` to check for the key, the expiresAt in authData will be valid for about an hour and we'll have to wait for that to expire before we can get a result when running `yarn hs accounts clean`. Once it has expired, it will fail the refresh token flow and the account will be marked for deletion when running the command. 
- Hit yes when prompted, and the account will be removed from the list.

packages/cli/commands/accounts/clean.js
```diff
@@ -0,0 +1,124 @@
...
+    text: i18n(`${i18nKey}.loading.add`),
+  });
+
+  for (const account of filteredTestPortals) {
+    try {
+      await accessTokenForPersonalAccessKey(account.portalId);
+    } catch (error) {
+      if (error && error instanceof HubSpotAuthError) {
+        if (
+          (error.statusCode === 401 &&
```
brandenrodgers, line 60: You should be able to use the `isSpecifiedError` util here

brandenrodgers:
Just a small nit: can we refer to the HubSpot accounts as "accounts" instead of portals? For user facing copy and for variable naming. 

Everything else in here lgtm!

packages/cli/commands/accounts/clean.js
adamawang: Will update to use `isSpecifiedHubSpotAuthError` !

adamawang:
Good call on account vs portal naming! Update screenshots:

Screenshot 2023-08-02 at 10 12 44
Screenshot 2023-08-02 at 10 13 03
Screenshot 2023-08-02 at 10 13 33

brandenrodgers (approved):
lgtm!

adamawang:
Pending some content adjustments, will merge after that 👍

adamawang:
Last round of content updates with feedback from PD/CD: 

Screenshot 2023-08-02 at 15 58 06
Screenshot 2023-08-02 at 17 09 31
Screenshot 2023-08-02 at 17 10 08

PR #898, opened 2023-07-26
Add preview command

jsines:
## Description and Context

!! Requires gate `Cms:LocalHublPreviews` !! 

CLI frontend for the `hs preview` experience. Moving forward this will serve as the base branch for any additions until we're ready to add it to a prerelease, similar to our VSCode release pattern.

See also: https://github.com/HubSpot/cli-lib/pull/21

To test:

1) pull this branch and the same named branch in `cli-lib`, install packages

1) in `cli-lib`
```
yarn link
```

1) in `hubspot-cli/packages/cli`

```
yarn link @hubspot/cli-lib
yarn link
```

1) in the directory you wish to test with 

```
yarn link @hubspot/cli
```

Then run the command on a folder similar to how you would with `hs watch`

```
hs preview ./cms-growth-theme test-growth-theme [--skipInitial]
```

The folder will be uploaded to the remote directory and an `hs watch` process will begin, and an http service will be launched on `localhost:3000`. `skip-initial`, or `skip` can be used to bypass the initial upload.

## LIMITATIONS

- Site proxying using buffered assets does not yet work - some BE changes are still needed for this functionality. Currently you can just view pages of your site essentially as they are live.

## Screenshots

Screenshot 2023-10-12 at 12 27 02 PM
Screenshot 2023-10-12 at 12 32 36 PM


jsines:
👋 This is waiting until other components are in place, it can't really be tested properly until a few other components are in place so sitting on the backburner until then.

packages/cli/commands/preview.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+  yargs.positional('dest', {
+    describe: i18n(`${i18nKey}.positionals.dest.describe`),
+    type: 'string',
+  });
+  yargs.option('skipInitial', {
+    alias: 'skip',
+    describe: i18n(`${i18nKey}.options.skipInitial.describe`),
+    type: 'boolean',
+  });
```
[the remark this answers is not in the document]
jsines: Ah yes, this was an option I added without much thought to speed things up during restarts while testing, but you're right it should be the other way around to match the `watch` pattern!

packages/cli/commands/preview.js
```diff
@@ -0,0 +1,72 @@
...
+  return yargs;
+};
+
+const validateSrcPath = src => {
+  const logInvalidPath = () => {
+    logger.log(
+      i18n(`${i18nKey}.errors.invalidPath`, {
+        path: src,
+      })
```
TanyaScales, line 58: Should we consider making this `logger.error` instead of the standard `log` since that will help stylize the output for this invalid path error?

packages/cli/commands/preview.js
```diff
@@ -0,0 +1,72 @@
...
+  });
+  yargs.option('initial-upload', {
+    alias: 'i',
+    describe: i18n(`${i18nKey}.options.initialUpload.describe`),
+    type: 'boolean',
+  });
+  return yargs;
+};
+
+const validateSrcPath = src => {
```
TanyaScales, line 53: Would love to see this function pushed above the builder and handler to follow the pattern in other commands -- or better, moved to the `cli-lib` in order to only have the handler and builder in this file (also similar to other commands)

packages/cli/lang/en.lyaml
```diff
@@ -837,6 +837,20 @@ en:
           disableInitial: "Passing the \"--disable-initial\" option is no longer necessary. Running \"hs watch\" no longer uploads the watched directory by default."
           initialUpload: "To upload the directory run \"hs upload\" beforehand or add the \"--initial-upload\" option when running \"hs watch\"."
           notUploaded: "The \"hs watch\" command no longer uploads the watched directory when started. The directory \"{{ path }}\" was not uploaded."
+      preview:
+        describe: "Description of preview command"
```
TanyaScales, line 841:
```suggestion
        describe: "Watch and upload a directory on your computer for changes and start a local development server to preview files"
```

packages/cli/lang/en.lyaml
```diff
@@ -837,6 +837,20 @@ en:
...
+            describe: "Path to the local directory your files are in, relative to your current working directory"
+          dest:
+            describe: "Path in HubSpot Design Tools. Can be a net new path"
+        errors:
+          invalidPath: "The path \"{{ path }}\" is not a path to a directory"
+        options:
+          notify:
+            describe: "Log to specified file when a watch task is triggered and after workers have gone idle. Ex. --notify path/to/file"
+          initialUpload:
+            describe: "Initial upload to preview buffer (how can we better explain this to end user?)"
```
TanyaScales, line 853:
```suggestion
            describe: "Upload directory before watching for updates"
```

I think we can keep close to the watch command here and keep it simple -- we don't have to expose information about the actual technical details to the developer on where their initial upload is going or what is happening behind the scenes

TanyaScales (approved):
Outside of the few comments I made -- nothing seems particularly non-blocking -- so approving this portion once you've cleared those

packages/cli/commands/preview.js
```diff
@@ -0,0 +1,72 @@
...
+const { getUploadableFileList } = require('../lib/upload');
+const { trackCommandUsage } = require('../lib/usageTracking');
+
+const i18nKey = 'cli.commands.preview';
+
+exports.command = 'preview <src> <dest>';
+exports.describe = i18n(`${i18nKey}.describe`);
+
+exports.handler = async options => {
+  const { src, dest, notify, initialUpload } = options;
```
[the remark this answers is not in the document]
jsines: Took the whole notify option out while debugging and forgot to put this bit back in 👍 😭

packages/cli/commands/preview.js
jsines: Looks like others do have their supporting functions toward the top - moved it there. 

I don't like moving it to `cli-lib` though, since it's only used before the `preview` process is called and not at all during it.

Maybe while adding support to VSCE we find we want that for both and it makes sense to just bake it into the command - but in the mean time a single helper function I don't think is a big deal. `hs watch` does some of the same logic.

packages/cli/commands/preview.js
```diff
@@ -0,0 +1,93 @@
...
+      return false;
+    }
+  } catch (e) {
+    logInvalidPath();
+    return false;
+  }
+  return true;
+};
+
+exports.command = 'preview <src> <dest>';
```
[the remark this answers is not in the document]
TanyaScales: yeah, and with cms turning into Content Hub -- I'm starting to wonder even more about bucketing things under the "cms" terminology and whether or not that is going to morph as well. 

Either way, I do agree with the sentiment here that this could probably move to `hs cms preview` to distinguish the command is for cms only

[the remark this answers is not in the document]
TanyaScales: Yeah, I would agree that this probably makes sense to bucket under `hs cms preview` to distinguish that it is for cms usage only

packages/cli/lang/en.lyaml
```diff
@@ -837,6 +837,22 @@ en:
           disableInitial: "Passing the \"--disable-initial\" option is no longer necessary. Running \"hs watch\" no longer uploads the watched directory by default."
           initialUpload: "To upload the directory run \"hs upload\" beforehand or add the \"--initial-upload\" option when running \"hs watch\"."
           notUploaded: "The \"hs watch\" command no longer uploads the watched directory when started. The directory \"{{ path }}\" was not uploaded."
+      preview:
+        describe: "Watch and upload a directory on your computer for changes and start a local development server to preview files"
```
brandenrodgers, line 841: Maybe 
`Watch and upload a directory on your computer for changes`
would be better as:
`Upload a directory on your computer and watch for changes`

TanyaScales: Updated

packages/cli/lang/en.lyaml
```diff
@@ -1158,6 +1177,12 @@ en:
           errors:
             buildIdDoesNotExist: "Build {{ buildId }} does not exist for project {{ projectName }}."
             buildAlreadyDeployed: "Build {{ buildId }} is already deployed."
+        previewPrompt:
+          enterSrc: "[--src] Enter a local theme directory to preview."
+          enterDest: "[--dest] Enter the destination path for the src theme in the design manager."
```
kemmerle, line 1182: Change design manager to HubSpot Design Tools (as above)?

packages/cli/commands/theme/preview.js
```diff
@@ -0,0 +1,175 @@
...
+const { loadAndValidateOptions } = require('../../lib/validation');
+const { previewPrompt } = require('../../lib/prompts/previewPrompt');
+const { EXIT_CODES } = require('../../lib/enums/exitCodes');
+const { FileUploadResultType } = require('@hubspot/cli-lib/lib/uploadFolder');
+const i18nKey = 'cli.commands.preview';
+const cliProgress = require('cli-progress');
+const {
+  ApiErrorContext,
+  logApiUploadErrorInstance,
+} = require('@hubspot/cli-lib/errorHandlers');
```
camden11, line 23: If possible would like to add as few new imports to `cli-lib` as possible. For the new stuff you've added there its alright, but for anything else (`path` and `logger` for example) we should use the local-dev-lib version.\

For `errorHandlers`, these actually exist within the CLI now so can import them from `lib/errorHandlers`

packages/cli/lib/prompts/previewPrompt.js
```diff
@@ -0,0 +1,39 @@
+const path = require('path');
+const { getCwd } = require('@hubspot/cli-lib/path');
```
camden11, line 2: same here re: path import

PR #899, opened 2023-07-27
Invalid portal error handling for `hs project dev`

adamawang:
## Description and Context

With new errors being thrown for invalid portals and unauthorized users (https://git.hubteam.com/HubSpotProtected/LocalDevAuth/pull/138), this PR serves to update error handling for the `ensureProjectExists` function and display more detailed messaging. 

Note: if your default account is set to a sandbox, the error will get caught and thrown in the `loadAndValidateOptions` check at the top of the command. If the default account is not a sandbox, it will show a list of sandboxes but it will not perform the same validation on selection which could result in selecting a deleted portal or one that the user is unauthorized for. 

## Screenshots

Before: 
Screenshot 2023-07-27 at 14 39 49

After:

Inactive account:
Screenshot 2023-07-27 at 14 33 01

Unauthorized user/invalid PAK:
Screenshot 2023-07-27 at 14 33 06

## TODO

n/a

### DIFF
--- a/packages/cli/commands/sandbox/delete.js
+++ b/packages/cli/commands/sandbox/delete.js
@@ -28,8 +28,8 @@ const { promptUser } = require('../../lib/prompts/promptUtils');
 const { getHubSpotWebsiteOrigin } = require('@hubspot/cli-lib/lib/urls');
 const {
   isSpecifiedError,
+  isSpecifiedHubSpotAuthError,
 } = require('@hubspot/cli-lib/errorHandlers/apiErrors');
-const { HubSpotAuthError } = require('@hubspot/cli-lib/lib/models/Errors');
 const { getAccountName } = require('../../lib/sandboxes');
 const { getValidEnv } = require('@hubspot/cli-lib/lib/environment');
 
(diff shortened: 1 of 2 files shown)


packages/cli/lib/projects.js
```diff
@@ -280,6 +281,10 @@ const ensureProjectExists = async (
         return false;
       }
     }
+    if (err.statusCode === 401 && err instanceof HubSpotAuthError) {
```
brandenrodgers, line 284: I think you can use `isSpecifiedError` here and pass in the statusCode option. It's a little verbose but is more consistent

brandenrodgers (approved):
Approved pending use of the error util

packages/cli/lib/projects.js
adamawang: Ah so that `isSpecifiedError` specifically checks for an error with `err.name === 'StatusCodeError'` but in this case, the error comes through as error name `HubSpotAuthError`

I'll create a new util for these specifically, probably `isHubSpotAuthError` and add that in all the appropriate spots

packages/cli/commands/sandbox/delete.js
```diff
@@ -168,7 +168,7 @@ exports.handler = async options => {
   } catch (err) {
     debugErrorAndContext(err);
 
-    if (err instanceof HubSpotAuthError && err.statusCode === 401) {
+    if (isSpecifiedHubSpotAuthError(err, { statusCode: 401 })) {
```
adamawang, line 171: Util PR here: https://github.com/HubSpot/cli-lib/pull/24

PR #901, opened 2023-07-31
Add build + deploy IDs while polling in dev mode

kemmerle:
## Description and Context
In Dev mode, we currently show `Building <project name>` and `Deploying <project name>` (see screenshot), but we do not include the build and deploy numbers. This PR adds the build numbers while polling. @brandenrodgers, our general pattern has been to write `Deploy build #` rather than giving a deploy number. I've followed that pattern, but let me know if that's not correct. 

## Screenshots

Before

8e02e94b-3f8a-4792-99ea-64a6ce4284fb

After

Screenshot 2023-07-31 at 10 43 07 AM

## TODO

- [x] Get feedback on copy from @markhazlewood 
- [x] Clarify with @brandenrodgers about deploy vs build IDs (see question above). 

## Who to Notify

@brandenrodgers @camden11 @markhazlewood

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -610,7 +610,7 @@ const makePollTaskStatusFunc = ({
     SpinniesManager.update(overallTaskSpinniesKey, {
       text: `${statusStrings.INITIALIZE(
         taskName,
-        taskId
+        displayId
       )}\n${componentCountText}`,
     });
 
@@ -691,11 +691,11 @@ const makePollTaskStatusFunc = ({
           if (isTaskComplete(taskStatus)) {
             if (status === statusText.STATES.SUCCESS) {
               SpinniesManager.succeed(overallTaskSpinniesKey, {
-                text: statusStrings.SUCCESS(taskName, taskId),
+                text: statusStrings.SUCCESS(taskName, displayId),
               });
             } else if (status === statusText.STATES.FAILURE) {
               SpinniesManager.fail(overallTaskSpinniesKey, {
-                text: statusStrings.FAIL(taskName, taskId),
+                text: statusStrings.FAIL(taskName, displayId),
               });
 
               if (!silenceLogs) {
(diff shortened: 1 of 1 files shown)


packages/cli/lib/projects.js
```diff
@@ -608,7 +608,10 @@ const makePollTaskStatusFunc = ({
         ) + '\n';
 
     SpinniesManager.update(overallTaskSpinniesKey, {
-      text: `${statusStrings.INITIALIZE(taskName)}\n${componentCountText}`,
+      text: `${statusStrings.INITIALIZE(
+        taskName,
+        taskId
```
brandenrodgers, line 613: Does` taskId` for deploys represent the deploy ID or the build ID?

kemmerle: Ooh interesting, it's `deployId` here (Just checked on my end). Should I fix it to the correct build ID, or rewrite the text to display deploy ID?

markhazlewood42:
Hm, I guess I'd expect this output to just reuse / look exactly like what we have for build + deploy when an `upload` happens. Is there a reason they wouldn't match up?

project_build_deploy

brandenrodgers:
@markhazlewood42 it would be possible to do the same polling that currently happens for `hs project upload`. We can do that if that sounds like a better UX. I think we had originally opted for the condensed version in this case because the output of the standard `hs project upload` build and deploy is pretty verbose and takes up a lot of space.

markhazlewood42:
Ah OK I didn't realize it was more complicated than reusing output blocks.

We can keep the condensed version then and revisit later, I'd just change the sentence structure a bit:

> `Building **[project name]** #[build number]`

For example:

> `Building **project0731** #4`

The "Building build" redundancy sounded a little weird  😄

packages/cli/lib/projects.js
kemmerle: Fixed to correct build ID in https://github.com/HubSpot/hubspot-cli/pull/901/commits/0ccf74f6f549e3e5121bca2aff262e5f522f2d7d

brandenrodgers (approved):
💯

PR #902, opened 2023-07-31
Replace ! with emoji in local dev prod warning message

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/432
This replaces the `!` in the local dev prod warning message with the❗ emoji. 

Did a little research into whether its safe to use emojis on the CLI without a fallback. It seems like actually checking emoji support is non-trivial but they should be supported in the vast majority of cases so this should be fine

## Screenshots
Screenshot 2023-07-31 at 11 58 41 AM

## Who to Notify

### DIFF
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -1011,7 +1011,7 @@ en:
       prompts:
         projectDevTargetAccountPrompt:
           createNewSandboxOption: "<Test on a new development sandbox>"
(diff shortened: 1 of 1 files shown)


kemmerle (approved):
I'm a little leery about not having a fallback for the emoji, but if you're confident :shipit:

brandenrodgers (approved):
Thanks!

PR #904, opened 2023-07-31
default sandbox confirmation prompt

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/431

This adds a prompt for users to confirm using their default account for local dev when their default account is a sandbox. Previously, having a sandbox as your default account would skip the account prompt entirely.

## Screenshots
Screenshot 2023-07-31 at 4 17 10 PM

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -27,6 +27,7 @@ const { uiAccountDescription, uiBetaMessage, uiLine } = require('../../lib/ui');
 const { confirmPrompt } = require('../../lib/prompts/promptUtils');
 const {
   selectTargetAccountPrompt,
+  confirmDefaultSandboxAccountPrompt,
 } = require('../../lib/prompts/projectDevTargetAccountPrompt');
 const SpinniesManager = require('../../lib/SpinniesManager');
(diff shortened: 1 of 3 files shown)


brandenrodgers:
This looks great. I think we can actually exit the process if they choose "no" here (maybe we could add some message to explain why we need to exit, similar to [this one](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lang/en.lyaml#L455)). The follow-up prompt is intended to be used when the default CLI account is a prod one.

camden11:
Ah gotcha ok. I can make that change, though IMO bringing up the prompt does seem helpful if they want to switch to a different sandbox or make a new one. Just seems like if someone ran the command they probably want to develop on one of their accounts. Is the preference always that they change the default account in these cases?

camden11:
@brandenrodgers Updated! Will now exit if you decline to use the sandbox

Screenshot 2023-08-02 at 5 11 59 PM

brandenrodgers (approved):
Looks like the tests are failing, but I'm not sure why. Maybe try just re-running them? Code looks good though 👌

PR #905, opened 2023-08-01
Dev Mode - Add file watch behavior to watch for config file changes

brandenrodgers:
## Description and Context

This sets up chokidar to watch for file changes while the dev mode command is running. The CLI will surface a warning to the user whenever they make an edit to a project configuration file.

There is logic in place to make sure that we don't spam the same upload warning to the user repeatedly.


packages/cli/lib/LocalDevManager.js
```diff
@@ -143,36 +160,53 @@ class LocalDevManager {
...
+      this.uploadWarnings[reason] = true;
     }
-    logger.log(
-      i18n(`${i18nKey}.uploadWarning.restartDev`, {
-        command: uiCommandReference('hs project dev'),
-      })
-    );
+  }
+
+  monitorConsoleOutput() {
```
brandenrodgers, line 200: This allows us to intercept any `console.log` calls to help prevent us from spamming upload warnings to the console. It pairs with the logic in `logUploadWarning`. The behavior is like this:

If the last thing logged to the console was an upload warning...
- Prevent any upload warning that has the exact same `reason` from logging to the console again.
- Allow any upload warning that has a different `reason` to log to the console
- Allow any other console logs that are not upload warnings.

If the last thing logged to the console was not an upload warning...
- Allow all console logs, including upload warnings

packages/cli/lib/LocalDevManager.js
```diff
@@ -217,10 +251,54 @@ class LocalDevManager {
...
+
+  handleWatchEvent(filePath, event, configPaths) {
+    if (configPaths.includes(filePath)) {
+      this.logUploadWarning(
+        i18n(`${i18nKey}.uploadWarning.configEdit`, {
+          path: path.relative(this.projectDir, filePath),
+        })
+      );
+    } else {
+      this.devServerFileChange(filePath, event);
```
brandenrodgers, line 295: This makes it so the dev servers can make their own checks for config changes within their respective components. If they detect a file change that requires an upload, they can call the `onUploadRequired` callback that gets passed down to them in `setup`.

kemmerle (approved):
I reviewed the code, and everything makes sense to me. @brandenrodgers, thanks so much for the explanatory notes. Should we perhaps add in some code comments for clarity? That's not a blocker. 

I was also able to test out the changes locally, and everything is working great! 🚢

camden11 (approved):

PR #907, opened 2023-08-01
Refactor `hs project download` (Other half in cli-lib)

kemmerle:
## Description and Context
1) Currently, there is a bug in the `hs project download` command, in which the command will not run outside a project directory. I was able to fix this issue by eliminating an unnecessary call to get a project config. 

2) I also refactored the download command to use a different download endpoint (`/archive-full`) that downloads the `src` of the project folder and the `hsproject.json`  (see the other part of this work in the [cli-lib PR](https://github.com/HubSpot/cli-lib/pull/23)).

To test out this PR, you will need to check out the [above branch](https://github.com/HubSpot/cli-lib/pull/23) in the cli-lib repo and then use the yarn link @hubspot/cli-lib command to use the symlinked local package. [Docs](https://github.com/HubSpot/hubspot-cli/blob/master/CONTRIBUTING.md#local-development-with-cli-lib) for more details.

## Who to Notify

@brandenrodgers @camden11

### DIFF
--- a/packages/cli/commands/project/download.js
+++ b/packages/cli/commands/project/download.js
@@ -17,10 +17,7 @@ const {
   downloadProject,
   fetchProjectBuilds,
 } = require('@hubspot/cli-lib/api/dfs');
-const {
-  createProjectConfig,
-  ensureProjectExists,
-} = require('../../lib/projects');
+const { ensureProjectExists } = require('../../lib/projects');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const {
   downloadProjectPrompt,
@@ -63,17 +60,6 @@ exports.handler = async options => {
 
     const absoluteDestPath = dest ? path.resolve(getCwd(), dest) : getCwd();
 
-    const projectConfigCreated = await createProjectConfig(
-      absoluteDestPath,
-      projectName,
-      { name: 'no-template' }
-    );
-
-    if (!projectConfigCreated) {
-      logger.log(i18n(`${i18nKey}.logs.downloadCancelled`));
-      process.exit(EXIT_CODES.SUCCESS);
-    }
-
     let buildNumberToDownload = buildNumber;
 
     if (!buildNumberToDownload) {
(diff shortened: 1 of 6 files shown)


packages/cli/commands/project/download.js
```diff
@@ -36,6 +33,12 @@ exports.describe = i18n(`${i18nKey}.describe`);
 exports.handler = async options => {
   await loadAndValidateOptions(options);
 
+  const { projectConfig } = await getProjectConfig();
```
brandenrodgers, line 36: If a project config already exists, should we abort the command and exit? It doesn't really make much sense to download a project within another project.

packages/cli/commands/project/download.js
```diff
@@ -92,16 +84,15 @@ exports.handler = async options => {
     const zippedProject = await downloadProject(
       accountId,
       projectName,
-      buildNumberToDownload
+      buildNumberToDownload,
+      platformVersion
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/907/commits/6907f42236eb9b7dc011ee7d35912316902208cf

packages/cli/commands/project/download.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/907/commits/6907f42236eb9b7dc011ee7d35912316902208cf

camden11 (approved):
Looks good! Tested and everything works as expected

packages/cli/commands/project/download.js
```diff
@@ -34,10 +34,11 @@ exports.handler = async options => {
...
 
   const { projectConfig } = await getProjectConfig();
-  const platformVersion =
-    projectConfig && projectConfig.platformVersion
-      ? projectConfig.platformVersion
-      : '';
+
+  if (projectConfig) {
+    logger.warn(i18n(`${i18nKey}.warnings.cannotDownloadWithinProject`));
+    process.exit(EXIT_CODES.WARNING);
```
brandenrodgers, line 40: The `WARNING` exit status is reserved for things like validation issues. This is less of a warning and more of an error. I think we should use error status here because this is a scenario where the command cannot possibly run.

brandenrodgers (approved):

PR #908, opened 2023-08-01
Fix polling hang when gated

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/434
When users who haven't been ungated for projects run `hs project dev`, they'll get stuck as `ensureProjectExists` polls indefinitely. This was due to a bug in `pollFetchProject`. For good measure, I also made it only poll when it needs to, i.e. when you've created a new sandbox.

## Who to Notify
@brandenrodgers @kemmerle


brandenrodgers (approved):
🔥

PR #913, opened 2023-08-08
`hs project dev` account picker only shows sandboxes for default account

adamawang:
## Description and Context

`hs project dev` prompt was updated to say "choose a sandbox **under** xyz account" but the prompt choices show all available sandboxes in your config. 

This PR filters down the list to sandboxes belonging to the default account only. This list will only show if the default account is a prod account, if a sandbox is the default it will automatically skip the prompt. 

## Screenshots

Before:
Screenshot 2023-08-08 at 14 52 50

After:
Screenshot 2023-08-08 at 14 52 31

## TODO

n/a


adamawang:
Slack thread for reference: https://hubspot.slack.com/archives/C01GV708YH3/p1691506693246309

brandenrodgers (approved):
🙏

PR #915, opened 2023-08-09
Remove console clears from dev mode init flow

brandenrodgers:
## Description and Context

We decided that we don't want to be so aggressive with the console clears in the init flow for the dev command

Edit: I updated this PR to do a few more things. It now also includes some changes to the errors and logging that we do during the init flow.
- We weren't failing the spinnies for project create failures
- Adding a suggestion to run `hs project upload` whenever we detect that the project has no deployed build.

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -234,13 +234,14 @@ exports.handler = async options => {
     if (shouldCreateProject) {
       await showPlatformVersionWarning(accountId, projectConfig);
 
+      SpinniesManager.add('createProject', {
+        text: i18n(`${i18nKey}.status.creatingProject`, {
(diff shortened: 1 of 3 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -317,7 +309,7 @@ exports.handler = async options => {
 
       logger.log();
       failedSubTasks.forEach(failedSubTask => {
-        console.log(failedSubTask.errorMessage);
+        console.error(failedSubTask.errorMessage);
```
brandenrodgers, line 312: I also switched this to log as errors because that's what they are

camden11 (approved):
🚢 Looks good + works well! Good call on adding the upload message too, I thought it was a bit confusing without it

PR #916, opened 2023-08-09
Fetch empty template from github

camden11:
## Description and Context
This modifies the create command to fetch the empty template from `hubspot-project-components` rather than creating it from scratch so that we can make sure new projects have the correct platform version.

See https://github.com/HubSpot/hubspot-project-components/pull/11

## Who to Notify
@brandenrodgers @kemmerle


kemmerle (approved):
:shipit:

PR #920, opened 2023-08-14
Improve error message in deploy command when project doesn't exist

kemmerle:
## Description and Context
Currently, when anyone runs the `hs project deploy` and the project doesn't exist in the target account, they receive a very vague error message:

Screenshot 2023-08-14 at 5 55 42 PM

With this PR, I've added a more detailed error message that gives the customer clear instructions to proceed:

Screenshot 2023-08-14 at 5 58 31 PM

NOTE: I originally took a different approach and wanted to give the customer the option to create and build/deploy the project directly within the `hs project deploy` command. However, the deploy and upload commands work differently, in that uploading requires a project path. That would require making significant changes to the `deploy` command, which I don't think we want to do right now. 

## Who to Notify

@brandenrodgers @camden11

### DIFF
--- a/packages/cli/commands/project/deploy.js
+++ b/packages/cli/commands/project/deploy.js
@@ -1,3 +1,4 @@
+const chalk = require('chalk');
 const {
   addAccountOptions,
   addConfigOptions,
@@ -12,7 +13,11 @@ const {
 const { logger } = require('@hubspot/cli-lib/logger');
 const { deployProject, fetchProject } = require('@hubspot/cli-lib/api/dfs');
 const { loadAndValidateOptions } = require('../../lib/validation');
-const { getProjectConfig, pollDeployStatus } = require('../../lib/projects');
+const {
+  ensureProjectExists,
+  getProjectConfig,
+  pollDeployStatus,
+} = require('../../lib/projects');
 const { projectNamePrompt } = require('../../lib/prompts/projectNamePrompt');
 const { buildIdPrompt } = require('../../lib/prompts/buildIdPrompt');
 const { i18n } = require('../../lib/lang');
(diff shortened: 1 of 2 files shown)


packages/cli/commands/project/deploy.js
```diff
@@ -42,6 +47,21 @@ exports.handler = async options => {
...
+  const projectExists = await ensureProjectExists(accountId, projectName, {
+    allowCreate: false,
+    noLogs: true,
+  });
+
+  if (!projectExists) {
+    logger.error(
+      i18n(`${i18nKey}.errors.projectNotFound`, {
+        projectName: chalk.bold(projectName),
+        accountId: chalk.bold(accountId),
```
brandenrodgers, line 59: We have a UI util for referencing an account in the logs: [uiAccountDescription](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/ui.js#L81)

packages/cli/lang/en.lyaml
```diff
@@ -517,6 +517,7 @@ en:
               deploy: "Deploy error: {{ details }}"
               noBuilds: "Deploy error: no builds for this project were found."
               noBuildId: "You must specify a build to deploy"
+              projectNotFound: "{{ projectName }} does not exist in account {{ accountId }}. Run `{{#yellow}}hs project upload{{/yellow}}` to upload your project files to HubSpot."
```
brandenrodgers, line 520: There's also a UI util for logging command references: [uiCommandReference](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/ui.js#L97)

brandenrodgers (approved):

PR #921, opened 2023-08-15
update project create next steps + remove deprecated ui beta warning util

brandenrodgers:
## Description and Context

Two small changes:
- Updating the next steps list in the project create flow to include a reference to the dev command
- Remove the uiBetaWarning util in favor of the new uiBetaMessage one
- Also, I removed code in uiLink that is no longer needed

### DIFF
--- a/packages/cli/commands/project/create.js
+++ b/packages/cli/commands/project/create.js
@@ -45,8 +45,7 @@ exports.handler = async options => {
   logger.log('');
   logger.log(chalk.bold(i18n(`${i18nKey}.logs.welcomeMessage`)));
   uiFeatureHighlight([
-    'projectUploadCommand',
-    'projectDeployCommand',
+    'projectDevCommand',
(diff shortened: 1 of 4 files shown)


packages/cli/commands/project/create.js
```diff
@@ -45,8 +45,7 @@ exports.handler = async options => {
   logger.log('');
   logger.log(chalk.bold(i18n(`${i18nKey}.logs.welcomeMessage`)));
   uiFeatureHighlight([
-    'projectUploadCommand',
-    'projectDeployCommand',
```
brandenrodgers, line 49: I removed these two because the dev command should probably be the first thing that a user does. But I can see the argument for keeping these in

camden11: I think it makes sense to remove and put more focus on dev

camden11 (approved):
🚢

PR #924, opened 2023-08-16
Clearer disabled messages at sandbox limit

adamawang:
## Description and Context

To reduce confusion around available development sandbox accounts while running `hs project dev`, we've updated the disabled messaging to suggest running `hs auth` if the target portal is at the limit of sandboxes, but the CLI config only has one or none of the accounts listed. 

We've also updated the messaging to state the **portal** limit has been reached. 

## Screenshots

Screenshot 2023-08-16 at 12 19 57
Screenshot 2023-08-16 at 12 21 49

## TODO

n/a

### DIFF
--- a/packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
+++ b/packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
@@ -1,6 +1,6 @@
 const { promptUser } = require('./promptUtils');
 const { i18n } = require('../lang');
-const { uiAccountDescription } = require('../ui');
+const { uiAccountDescription, uiCommandReference } = require('../ui');
 const { isSandbox, getAccountName } = require('../sandboxes');
 const { getAccountId } = require('@hubspot/cli-lib');
 const { getSandboxUsageLimits } = require('@hubspot/cli-lib/sandboxes');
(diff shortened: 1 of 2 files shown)


packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -34,9 +34,15 @@ const selectTargetAccountPrompt = async (accounts, defaultAccountConfig) => {
   let disabledMessage = false;
 
   if (sandboxUsage['DEVELOPER'] && sandboxUsage['DEVELOPER'].available === 0) {
-    disabledMessage = i18n(`${i18nKey}.sandboxLimit`, {
-      limit: sandboxUsage['DEVELOPER'].limit,
-    });
+    if (sandboxAccounts.length < sandboxUsage['DEVELOPER'].limit) {
+      disabledMessage = i18n(`${i18nKey}.sandboxLimitWithSuggestion`, {
+        limit: sandboxUsage['DEVELOPER'].limit,
```
brandenrodgers, line 39: There's a new util that we're using whenever we want to log out command references. It's [uiCommandReference](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/ui.js#L89). Can you use that here?

Just a note, it's possible that it won't play nicely with the prompt lib and how that logs. Feel free to keep this as is if uiCommandReference isn't working for it.

adamawang: Nice, looks great with the util usage:
Screenshot 2023-08-16 at 13 40 26

brandenrodgers (approved):

PR #927, opened 2023-08-17
Hubspot project components tag

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/449 and https://github.com/HubSpot/cli-lib/pull/26
This updates the `hs project create` and `hs project add` commands to fetch from a specific version tag of `hubspot-project-components.

Keeping this as a draft for now since I still need to test everything

## Who to Notify
@brandenrodgers


packages/cli/commands/create/webpack-serverless.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     cloneGitHubRepo(
       dest,
       assetType,
-      'cms-webpack-serverless-boilerplate',
+      'HubSpot/cms-webpack-serverless-boilerplate',
```
camden11, line 9: Ended up having to change a few extra files to include `HubSpot` in the github repo URLs. Thought it was confusing that some of the github functions required it, while some didn't

kemmerle (approved):
Reviewed the code, and changes look good to me. Standardizing how we pass in repo paths is chef's kiss.

PR #931, opened 2023-09-21
Setting up foundation for lint command

TanyaScales:
## Description and Context
Finalizing the `hs lint` command in order to return HubL validation errors and warnings. This is currently a standalone command that would need to be run manually on a file in order to get its validation results returned. The expectation in the future would be that we could integrate this into the watch/upload command so that developers can utilize it in order to catch these results before the file is uploaded to the DM (currently the only place that you can see them).

- Related lib PR here: https://github.com/HubSpot/cli-lib/pull/27

#### Linting command run and return:
Screenshot 2023-09-21 at 3 53 11 PM

## TODO
- [ ] Testing: Run lint on a variety of file types to ensure that proper errors are thrown if linting command fails
- [ ] Testing: Run lint on templates with a variety of `template_errors` to make sure `SYNTAX_ERROR` and `DEPRECATED_MODULE` errors/warnings only come back
- [ ] Run command on folder - refactor to surface warnings and highlight while file those came from
- [ ] Determine if there are other errors/warnings that come from the `cos-rendering/v1/internal/validate` endpoint that should be surfaced in this command
- [ ] Add linting command as an optional flag onto the `upload` and `watch` commands
    - As the lint command is run on each file individually -- it might be nice to think about how this would work when uploading a large amount of files via watch/upload -- the validation warnings could get overwhelming if it is reporting back each of them
- [ ] Should determine if there is a flag to only show errors vs warnings -- in case developers don't want to be warned about the deprecated modules and whatnot
- [ ] Move linting logic from `cli-lib` to `hubspot-local-dev-lib`
- [ ] Determine why the line count for warnings is off -- in the example, these warnings are for different lines of the code and definitely no line `19` -- but this is what is being returned from the endpoint -- investigate.

### DIFF
--- a/packages/cli/commands/lint.js
+++ b/packages/cli/commands/lint.js
@@ -9,6 +9,7 @@ const {
   addConfigOptions,
   addAccountOptions,
   getAccountId,
+  addUseEnvironmentOptions,
 } = require('../lib/commonOpts');
 const { resolveLocalPath } = require('../lib/filesystem');
 const { trackCommandUsage } = require('../lib/usageTracking');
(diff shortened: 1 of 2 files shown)


packages/cli/commands/lint.js
```diff
@@ -19,44 +20,47 @@ const i18nKey = 'cli.commands.lint';
...
-  trackCommandUsage('lint', null, accountId);
+  let errorCount = 0;
 
   logger.group(groupName);
-  let count = 0;
+
   try {
     await lint(accountId, localPath, result => {
-      count += printHublValidationResult(result);
+      errorCount = printHublValidationResult(result);
```
TanyaScales, line 43: `+=` removed because the `printHubLValidationResult` function returns a count of the errors, we don't need to do it here.

packages/cli/commands/lint.js
```diff
@@ -19,44 +20,47 @@ const i18nKey = 'cli.commands.lint';
...
-  );
+
+  logger.groupEnd();
+  errorCount === 0
+    ? logger.log(i18n(`${i18nKey}.noIssuesFound`))
+    : logger.log(
+        i18n(`${i18nKey}.issuesFound`, {
+          errorCount,
+        })
+      );
```
TanyaScales, line 57: I found that when `0` was returned for the errorCount, the log would show ` issues found` instead of `0 issues found` -- even though `0` was being passed into the interpolator. I couldn't track down why that was happening, so I added in another key here for that scenario with `No issues found` string.

packages/cli/commands/lint.js
```diff
@@ -19,44 +20,47 @@ const i18nKey = 'cli.commands.lint';
...
 };
 
 exports.builder = yargs => {
-  addConfigOptions(yargs, true);
-  addAccountOptions(yargs, true);
+  addConfigOptions(yargs);
+  addAccountOptions(yargs);
```
TanyaScales, line 62: Secondary param here of `true/false` is no longer required after the migration from commander took place -- this is just tech debt from that, so I'm removing here

packages/cli/commands/lint.js
jsines: This was due to a bug in i18n string interpolation - fixed it [here](https://github.com/HubSpot/hubspot-cli/pull/933/files)

TanyaScales: Thanks! I tried tracing that into the interpolator but I couldn't pinpoint that line as the issue -- I can revert this back -- it would be better to keep this simple

packages/cli/commands/lint.js
jsines: Given the `exports.describe` - I'm assuming this was meant in the case of running the command on folders to add up the results count for each file.

Here I've added two test templates to `./src/templates` with the same deprecated module and run it on the folder before and after reverting the `+=`. Since whatever template it saw last had no issue it reports none found, even though it did find some.

Screenshot 2023-09-25 at 2 19 45 PM

Related to the case of running it on folders, it would be a nice add to report which specific file the errors were found in, doesn't seem we do that now which makes them a quit a bit harder to pull apart

kemmerle:
@TanyaScales, I wanted to give you a heads up that I have a [PR](https://github.com/HubSpot/hubspot-cli/pull/934) that will create a (hopefully minor) merge conflict for you. Also, would you like me to test out this PR, or are you in the middle of revisions? Happy to be a second pair of eyes 👀

TanyaScales:
Thanks for the heads up @kemmerle -- I do have some comments and revisions to make from @jsines before this PR is back up to review. Once it is in a better state, it would be great to get an extra set of eyes on it though. I can tag you in at that point for a review.

joe-yeager:
Closing this PR since it's been sitting for a year

PR #932, opened 2023-09-22
Port modules that log or display data (Part 1)

kemmerle:
## Description and Context

As part of our efforts to switch from the `cli-lib` library to the `hubspot-local-dev-lib` library, we are making `hubspot-local-dev-lib` agnostic about consoles, displays, etc. This means that we need to move modules that log or display data directly to `hubspot-cli`. 

NOTE: I have ported `table.js` to the `cli` package as part of this PR, but it is also referenced in the `serverless-dev-runtime` package. Should we consider keeping this as part of `hubspot-local-dev-lib`? EDIT: Addressed this by reverting the port of `table.js`. @brandenrodgers suggested that we have a dedicated folder in `hubspot-local-dev-lib` for utils that are CLI/console specific. 

This is the first of two PRs, so we can test more easily. 

## Testing

Need to test the following commands to ensure that everything is working as expected:

`process.js`:

- [x] hs auth (with oauth)
- [x] hs init (with oauth) 
- [x] hs project dev 
- [x] hs project watch
- [x] hs project logs
- [x] hs sandbox create
- [x] hs sandbox sync 
- [x] hs logs
- [x] hs functions deploy

`functions.js`:

- [x] hs functions list

`projectsWatch.js`:

- [x] hs project watch

## TODO

- [x] Complete testing
- [x] Address feedback

## Who to Notify

@brandenrodgers @camden11


kemmerle:
@brandenrodgers @camden11, I am not quite sure why the current build is failing. I've attempted searching for a solution in StackOverflow, etc., and there's very few that cover installing the package as part of a GitHub action (ie. they'll recommend deleting the `node-modules` file, etc).

packages/cli/lib/functions.js
```diff
@@ -0,0 +1,18 @@
...
+      secretNames.join(', '),
+      moment(created).format(),
+      moment(updated).format(),
+    ];
+  });
+};
+
+module.exports = {
+  getFunctionArrays,
+};
```
camden11, line 18: Might make sense to just rename this file `getFunctionArrays` since `functions` isn't very descriptive

camden11 (approved):
Looks good! Haven't tested yet but if you're confident everything works then I think this is probably good to merge

PR #933, opened 2023-09-25
Fix i18n branch interpolation

jsines:
Fixes logic in i18n interpolation to not break for falsey values - previously any time an argument of "0" was passed into the string it would be replaced with an empty string!


TanyaScales:
I would be curious if there are other cases where correcting this might break the current way a message is being displayed? I can't think of a situation off the top of my head.

TanyaScales (approved):

jsines:
>I would be curious if there are other cases where correcting this might break the current way a message is being displayed? I can't think of a situation off the top of my head.

Only situation I could see is if someone was passing something falsey as the i18n argument expecting it to revert to `''`. Which seems like a strange practice we should discourage anyway 😄

PR #934, opened 2023-09-26
Port modules that log or display data (Part 2)

kemmerle:
## Description and Context
As part of our efforts to switch from the `cli-lib` library to the `hubspot-local-dev-lib` library, we are making `hubspot-local-dev-lib` agnostic about consoles, displays, etc. This means that we need to move modules that log or display data directly to `hubspot-cli`.

This is the second of two PRs, so we can test more easily.

## TODO

- [x] Address any feedback 

## Who to Notify

@camden11 @brandenrodgers


packages/cli/lib/watch.js
```diff
@@ -0,0 +1,255 @@
...
+    });
+    triggerNotify(notify, 'Changed', filePath, uploadPromise);
+  });
+
+  return watcher;
+}
+
+module.exports = {
+  watch,
+};
```
camden11, line 255: @kemmerle We actually decided to keep a slightly refactored version of watch in `local-dev-lib` since it's used by the vscode extension (see https://github.com/HubSpot/hubspot-local-dev-lib/pull/28). My bad for not updating the issue!

camden11 (commented):
Need to remove watch - everything else looks good

packages/cli/lib/watch.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/934/commits/d5cd90f9d4931b2d378f9e1f3e6b6e312bcf5e25

camden11 (approved):

PR #935, opened 2023-09-28
Add debug logs for files filtered by ignore rules during project uploads

brandenrodgers:
## Description and Context

The CLI has a set of [default ignore rules](https://github.com/HubSpot/cli-lib/blob/main/ignoreRules.js#L6-L30) that it uses on upload which can be confusing to users when their content gets unexpectedly filtered. This PR adds the filtered content to some debug logs so it's easier for us to catch this when it happens. Ultimately it would be a little better if we could differentiate between a default HubSpot ignore rule and a user-defined one, but this is a good place to start.

I added some special casing to catch nodule_modules. It prevents us from logging tons of lines of debug logs for everything in the node modules folder.

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -522,9 +522,15 @@ const handleProjectUpload = async (
 
   archive.pipe(output);
 
-  archive.directory(srcDir, false, file =>
-    shouldIgnoreFile(file.name, true) ? false : file
-  );
+  archive.directory(srcDir, false, file => {
+    const ignored = shouldIgnoreFile(file.name, true);
+    if (ignored) {
+      logger.debug(`${i18nKey}.handleProjectUpload.fileFiltered`, {
+        filename: file.name,
+      });
+    }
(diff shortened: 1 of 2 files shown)


packages/cli/lib/projects.js
```diff
@@ -522,9 +522,27 @@ const handleProjectUpload = async (
...
+            filename: file.name,
+          })
+        );
+      }
+
+      if (isNodeModule && !loggedIgnoredNodeModule) {
+        loggedIgnoredNodeModule = true;
+      }
+    }
+    return ignored ? false : file;
```
kemmerle, line 544: This isn't a blocker, but I'm thinking about the TypeScript headache here when we eventually convert the CLI to TS.

brandenrodgers: This is a good call out, but I think I may skip changing it for now. I took a look at the [docs](https://www.archiverjs.com/docs/archiver/#directory) and couldn't find any obvious ways to make this better without potentially introducing some obscure bug.

PR #941, opened 2023-10-16
Port error handlers to CLI

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/481

The CLI uses these `errorHandler` utils to log errors in a user friendly fashion. Since they are interface related, we're moving them from `cli-lib` to the CLI. There is one instance in `webpack-cms-plugins` where I had to copy some stuff over, but don't think that one spot is worth keeping these in `cli-lib` for.

## Who to Notify
@brandenrodgers

### DIFF
--- a/packages/cli/lib/errorHandlers/apiErrors.js
+++ b/packages/cli/lib/errorHandlers/apiErrors.js
@@ -12,6 +12,9 @@ const {
   logErrorInstance,
   ErrorContext,
 } = require('./standardErrors');
+const { i18n } = require('../lang');
+
+const i18nKey = 'cli.lib.errorHandlers.apiErrors';
 
 const isApiStatusCodeError = err =>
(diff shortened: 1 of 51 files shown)


packages/cli/lib/errorHandlers/apiErrors.js
```diff
@@ -0,0 +1,343 @@
...
+  const subCategoryErr = !subCategory || err.subCategory === subCategory;
+  return (
+    err.name === 'HubSpotAuthError' &&
+    statusCodeErr &&
+    categoryErr &&
+    subCategoryErr
+  );
+};
+
+const contactSupportString =
```
brandenrodgers, line 67: This copy should probably live in the translations file.

brandenrodgers (approved):
Tested locally and lgtm 👌

PR #942, opened 2023-10-17
Swap cli-lib 'text' dep for local-dev-lib version

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/476

This adds `@hubspot/local-dev-lib` as a dependancy to the CLI and imports its version of `text` in places where the `cli-lib` version was being used. `text` is a really simple function, so figured this would be a good way to test the waters before swapping any other deps

## TODO

- [ ] Publish `v1.0.0` of `@hubspot/local-dev-lib`
- [ ] Update the dependency here

## Who to Notify
@brandenrodgers


packages/cli/commands/lint.js
```diff
@@ -1,7 +1,7 @@
 const { lint } = require('@hubspot/cli-lib/validate');
 const { printHublValidationResult } = require('../lib/hublValidate');
 const { logger } = require('@hubspot/cli-lib/logger');
-const { logErrorInstance } = require('../lib/errorHandlers/standardErrors');
+const { logErrorInstance } = require('@hubspot/cli-lib/errorHandlers');
```
brandenrodgers, line 4: Should this import be coming from within the cli now?

camden11: Ohh yep, looks like that accidentally got reset somehow

brandenrodgers (approved):
🎉

PR #947, opened 2023-10-26
Use local dev lib config

camden11:
## Description and Context
This switches the CLI from using `@hubspot/cli-lib` for `config` to using `@hubspot/local-dev-lib`. The config module is functionally identical between both, so everything should behave exactly the same. Once this PR is merged, we'll be able to start porting everything else over to `local-dev-lib

## TODO
* This PR will need to be merged and released in tandem with this `cli-lib` PR https://github.com/HubSpot/cli-lib/pull/22
As well as a PR for the VSCode extension
* It would probably be good to stress test this pretty thoroughly before doing a full release.

## Who to Notify
@brandenrodgers


packages/cli/commands/project/deploy.js
```diff
@@ -18,7 +18,7 @@ const { projectNamePrompt } = require('../../lib/prompts/projectNamePrompt');
 const { buildIdPrompt } = require('../../lib/prompts/buildIdPrompt');
 const { i18n } = require('../../lib/lang');
 const { uiBetaTag } = require('../../lib/ui');
-const { getAccountConfig } = require('@hubspot/cli-lib');
+const { getAccountConfig } = require('@hubspot/cli-lib/');
```
brandenrodgers, line 21: Why does this need the trailing slash?

brandenrodgers:
The code in here lgtm (looks like tests are failing though). I'll hammer on this locally for a bit before approving though

packages/cli/commands/project/deploy.js
camden11: I'm not sure how that got there. This should actually be a local-dev-lib import anyway

packages/cli/commands/project/deploy.js
```diff
@@ -18,7 +18,7 @@ const { projectNamePrompt } = require('../../lib/prompts/projectNamePrompt');
 const { buildIdPrompt } = require('../../lib/prompts/buildIdPrompt');
 const { i18n } = require('../../lib/lang');
 const { uiBetaTag } = require('../../lib/ui');
-const { getAccountConfig } = require('@hubspot/cli-lib');
+const { getAccountConfig } = require('@hubspot/local-dev-lib');
```
brandenrodgers, line 21: Missing `/config` here I think

brandenrodgers:
I cross checked all of the exports from `local-dev-lib/config`. There are a few more exports that you may want to update so they pull from local-dev-lib:
- getAccountId (exported from the old `commonOpts` file in cli-lib)
  - Edit: I realized this is pulling from `cli/lib/commonOpts` 😨. We should probably look into fixing up that file to remove confusion between those utils and the ones exported from local-dev-lib
- isConfigFlagEnabled (I think I saw one or two missing updates for this one)
- getEnv (One or two missing updates here too)
-

brandenrodgers (approved):
Tested locally and this lgtm! We should make sure there isn't any other WIP PR's that need to be merged soon because this should probably sit on the beta release for a bit so we can be sure that there aren't any bugs

PR #950, opened 2023-11-03
Swap cli lib deps 2

camden11:
## Description and Context
This switches to using hubspot-local-dev-lib for the following dependancies:
* `@hubspot/cli-lib/lib/validate`
* `@hubspot/cli-lib/lib/environment`
* `@hubspot/cli-lib/lib/ignoreRules`

Also bumps the deps on `local-dev-lib` and `cli-lib` to the latest versions

## Who to Notify
@brandenrodgers


brandenrodgers (approved):
lgtm! The syntax of `getValidEnv` is different in local-dev-lib than it is in cli-lib, but it looks like it doesn't matter because nothing in here uses the second arg anyways.

PR #951, opened 2023-11-03
Validate project config `srcDir` within project root

mendel-at-hubspot:
## Description and Context

Adds validation within `validateProjectConfig` to ensure the srcDir of a project points to either the project root, or a subdirectory. This will help ensure a user doesn't attempt to upload a parent or root directory by mistake.

Adds a suite of jest tests to confirm our current validations, as well as the new validation logic.

To support the testing with mocks around `process.exit`, added `return` statements to all the spots in `validateProjectConfig` where the whole node process would be existing anyway. Open to reverting those changes and using exceptions in the mocking of `process.exit` if that'd be cleaner / clearer.

## Screenshots
Screenshot 2023-11-06 at 12 24 53 PM

cc @camden11 @brandenrodgers

### DIFF
--- a/packages/cli/lib/__tests__/projects.test.js
+++ b/packages/cli/lib/__tests__/projects.test.js
@@ -71,9 +71,6 @@ describe('@hubspot/cli/lib/projects', () => {
         expect(errorSpy).toHaveBeenCalledWith(
           expect.stringContaining('srcDir: ".."')
         );
-        expect(errorSpy).toHaveBeenCalledWith(
-          expect.stringContaining(`within ${projectDir}`)
-        );
       });
 
       it('for root directory', () => {
@@ -86,9 +83,6 @@ describe('@hubspot/cli/lib/projects', () => {
         expect(errorSpy).toHaveBeenCalledWith(
           expect.stringContaining('srcDir: "/"')
         );
-        expect(errorSpy).toHaveBeenCalledWith(
-          expect.stringContaining(`within ${projectDir}`)
-        );
       });
 
       it('for complicated directory', () => {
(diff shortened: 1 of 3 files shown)


packages/cli/lib/__tests__/projects.test.js
```diff
@@ -0,0 +1,153 @@
...
+describe('@hubspot/cli/lib/projects', () => {
+  describe('validateProjectConfig()', () => {
+    let realProcess;
+    let projectDir;
+    let exitMock;
+    let errorSpy;
+
+    beforeAll(() => {
+      projectDir = fs.mkdtempSync(path.join(os.tmpdir(), 'projects-'));
+      fs.mkdirSync(path.join(projectDir, 'src'));
```
brandenrodgers, line 16: TIL about `mkdtempSync`. Do these temp folders remove themselves if the tests fail to complete for some reason? There are scenarios where the after lifecycle methods never run, which could cause the folders to be left around in the project. Can you do this without actually creating the temp folder? It looks like the only fs method that might require the folder to actually exist is  [existsSync](https://github.com/HubSpot/hubspot-cli/pull/951/files#diff-fa4fa15b34bd2259fc2c24ec990fdbaa6492e6030d940700895318b399058174R188), so you could mock that instead and avoid having to write anything to the filesystem.

Thanks for adding tests btw! 😄

brandenrodgers: Yeah avoiding mocking wherever possible is a good rule of thumb. For filesystem methods though we typically try to avoid ever actually writing because it can potentially have an impact on the repo structure depending on what the utils are doing. I did a little reading and it looks like [the better solution is to use memfs ](https://medium.com/nerd-for-tech/testing-in-node-js-easy-way-to-mock-filesystem-883b9f822ea4)(not currently a dep of the CLI though and probably out of the scope of this PR).

The main issue I'm trying to avoid is someone running these tests locally, unintentionally modifying the project structure, and then committing test files ([I accidentally did this last week](https://github.com/HubSpot/hubspot-local-dev-lib/pull/54#pullrequestreview-1708649549)). Now that I've read that article I'm realizing we should switch up the approach at some point, but regardless the current pattern we've been using is to mock fs to prevent any actual writes.

brandenrodgers (approved):

PR #955, opened 2023-11-13
Use local-dev-lib for urls module

camden11:

### DIFF
--- a/packages/cli/commands/customObject/schema/create.js
+++ b/packages/cli/commands/customObject/schema/create.js
@@ -18,7 +18,7 @@ const { createSchema } = require('@hubspot/cli-lib/api/schema');
 const {
   createSchema: createSchemaFromHubFile,
 } = require('@hubspot/cli-lib/api/fileTransport');
-const { getHubSpotWebsiteOrigin } = require('@hubspot/cli-lib/lib/urls');
+const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
 const { i18n } = require('../../../lib/lang');
 
 const i18nKey =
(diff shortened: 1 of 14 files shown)


packages/cli/commands/cms/convertFields.js
```diff
@@ -3,7 +3,7 @@ const fs = require('fs');
 const { createIgnoreFilter } = require('@hubspot/local-dev-lib/ignoreRules');
 const { isAllowedExtension, getCwd } = require('@hubspot/cli-lib/path');
 const { logger } = require('@hubspot/cli-lib/logger');
-const { walk } = require('@hubspot/cli-lib/lib/walk');
+const { walk } = require('@hubspot/local-dev-lib/fs');
```
brandenrodgers, line 6: Not sure if it matters, but there's a small change in the behavior of `walk`. It used to log a debug log and then return an empty array. Now it throws on errors in the catch block. We may need to update all of the usages of `walk` to catch the thrown error and default to an empty array so we have parity in the behavior.

packages/cli/lib/usageTracking.js
```diff
@@ -1,4 +1,4 @@
-const { trackUsage } = require('@hubspot/cli-lib/api/fileMapper');
+const { trackUsage } = require('@hubspot/local-dev-lib/trackUsage');
```
brandenrodgers, line 1: I realized that our [trackUsage in local-dev-lib](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/lib/trackUsage.ts#L61) never returns anything. There are areas in the CLI that rely on a response being returned (like [here](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/usageTracking.js#L141)).

packages/cli/commands/cms/convertFields.js
camden11: Ooh good catch!

camden11:
@brandenrodgers Fixed both of those things. I have a bit more work on this coming in my following PR, but it looks like the existing `logErrorInstance` should be able to handle most errors thrown by local-dev-lib

brandenrodgers (approved):

Issue #953, opened 2023-11-10
Can`t cancel a local dev server


NahuelBarbieri:
There is not possible to cancel local dev server if the last terminal with the server running was closed. Only allow to cancel qith 'q' if the terminal window is still open.

PR #956, opened 2023-11-14
Fix project local dev exit cleanup

brandenrodgers:
## Description and Context

Fixes https://github.com/HubSpot/hubspot-cli/issues/953

There were a handful of issues in our `handleExit` util. It's the util that attempts to catch the various different exit signals, which allows us to gracefully exit the process. The major things I changed:
- Only trigger the exit once (We were sometimes triggering multiple times, especially when uncaughtException is triggered)
- Prevent interacting with stdout when the parent terminal is closed
- Adds a debug log to the termination callback to make it easier to figure out what's happening

This article was a useful resource: https://thomashunter.name/posts/2021-03-08-the-death-of-a-nodejs-process

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -338,7 +338,7 @@ exports.handler = async options => {
 
   await LocalDev.start();
 
-  handleExit(() => LocalDev.stop());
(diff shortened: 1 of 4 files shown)


camden11 (approved):
LGTM

PR #960, opened 2023-11-30
Use local-dev-lib for archive and gitignore

camden11:
## Description and Context
This swaps `@hubspot/cli-lib/lib/git` and `@hubspot/cli-lib/archive` for `@hubspot/local-dev-lib/gitignore` and `@hubspot/local-dev-lib/archive` respectively. Also adds a util for using logcallbacks, and updates `debugErrorAndContext` to include information about the `cause` of errors

## Who to Notify
@brandenrodgers

### DIFF
--- a/packages/cli/commands/init.js
+++ b/packages/cli/commands/init.js
@@ -8,7 +8,9 @@ const {
 } = require('@hubspot/local-dev-lib/config');
 const { addConfigOptions } = require('../lib/commonOpts');
 const { handleExit } = require('../lib/process');
-const { checkAndUpdateGitignore } = require('@hubspot/cli-lib/lib/git');
+const {
+  checkAndAddConfigToGitignore,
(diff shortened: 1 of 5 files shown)


packages/cli/commands/init.js
```diff
@@ -117,7 +119,7 @@ exports.handler = async options => {
     );
     const configPath = getConfigPath();
 
-    checkAndUpdateGitignore(configPath);
+    checkAndAddConfigToGitignore(configPath);
```
brandenrodgers, line 122: It looks like in cli-lib the `checkAndUpdateGitignore ` function was set up to purposely fail silently in the catch block. Maybe we want to add a try/catch here so we can keep the silent fail behavior?

camden11:
Fixed both of these!

brandenrodgers (approved):

PR #963, opened 2023-12-06
Use local-dev-lib for path

camden11:
## Description and Context
This swaps the `path` dep to use `local-dev-lib`. `path`'s functionality is identical between the two packages, so this should be pretty safe despite things changing in a lot of files.

## Who to Notify
@brandenrodgers


brandenrodgers (approved):
lgtm!

PR #968, opened 2023-12-18
Use local dev lib v0.1.0

camden11:
## Description and Context
Bumps local dev lib dep to 0.1.0, which requires updating the import path of `lint`

## Who to Notify
@brandenrodgers


brandenrodgers (approved):
Thanks 🙏

PR #975, opened 2024-01-11
Use local-dev-lib for filemapper and filemanager

camden11:
## Description and Context
This switches `fileManager` and `fileMapper` to use `local-dev-lib` as deps + adds log callbacks for them. I've tested this pretty thoroughly but could the following commands could still use another run through:
* `hs fetch`
* `hs filemanager fetch`
* `hs upload`
* `hs filemanager upload`

Also will need to release local-dev-lib and update dep here before merging

## Who to Notify
@brandenrodgers


brandenrodgers (approved):
I left one comment about potentially updating the pattern for using the logger util, but it's not blocking.

PR #976, opened 2024-01-11
Adding to create command to support React modules

TanyaScales:
## Description and Context
 Pairs with https://github.com/HubSpot/hubspot-local-dev-lib/pull/78 -- Adding more to the create prompt and passing some additional options through the create module command

## Screenshots
See [this cli-lib PR](https://github.com/HubSpot/cli-lib/pull/44) for Loom of the command

## TODO
- [x] Update ATs


packages/cli/commands/create.js
```diff
@@ -111,6 +112,10 @@ exports.builder = yargs => {
     describe: i18n(`${i18nKey}.positionals.dest.describe`),
     type: 'string',
   });
+  yargs.option('internal', {
+    describe: 'Internal HubSpot version of creation command',
+    type: 'boolean',
+  });
```
j-malt, line 118: Perhaps we should make this a hidden option? I think you can just add `hidden: true` to the props here but I don't remember the yargs API that well.

j-malt (approved):
👍 similar question as in `cli-lib` about adding support for regular JS modules, but code LGTM.

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/cli",
-  "version": "5.1.1",
+  "version": "5.1.2",
```
brandenrodgers, line 3: You shouldn't need to manually bump this. lerna will handle it for you when you run the release command.

packages/cli/commands/create.js
TanyaScales: Yep - just checked that out in docs -- added hidden property to that option

packages/cli/package.json
TanyaScales: Took it back down

PR #977, opened 2024-01-12
Update some more of the cms imports to use local-dev-lib

brandenrodgers:
## Description and Context

Updating imports for `cli-lib/functions`, `cli-lib/modules`, and `cli-lib/templates` to pull from their new local-dev-lib counterparts.


brandenrodgers:
This is ready for review now, although it's dependent on the latest changes over in local-dev-lib. We will need to cut a new patch release of LDL before merging/releasing this.

camden11 (approved):
Tested and works well! Should be good to merge once LDL update is out

packages/cli/lib/commonOpts.js
```diff
@@ -66,8 +69,16 @@ const setLogLevel = (options = {}) => {
   const { debug } = options;
   if (debug) {
     Logger.setLogLevel(LOG_LEVEL.DEBUG);
+
+    // Update the log level in local-dev-lib's instance of the logger
+    // This will evenutally replace cli-lib's version of it
+    setLocalDevLibLogLevel(LOG_LEVEL.DEBUG);
```
brandenrodgers, line 75: I realized that certain log statements in local-dev-lib were not working correctly. It was because the default log level is `NONE`, and we were only setting the log level to `LOG` for the cli-lib instance of the logger.

camden11 (approved):

camden11:
Good catch with the log level stuff. Also this whole PR is a pretty solid case against log callbacks. Looks like there are some tests failing but otherwise looks good

brandenrodgers:
I updated everything that uses logCallbacks in response to: https://github.com/HubSpot/hubspot-local-dev-lib/pull/95

camden11 (approved):
Log callback stuff looks good!

PR #979, opened 2024-01-17
Default account names for hs init and hs auth

camden11:
## Description and Context
Now that the `localDevAuth` endpoints returns `hubName`, we can use that name as a default when setting up a new account on the CLI via `hs init` or `hs auth`. This slightly refactors both of those commands to add support for this.

cc @adamawang This also fixes the lang errors with sandbox gate syncing

## Screenshots
Screenshot 2024-01-17 at 2 05 45 PM

## TODO
* Need to merge https://github.com/HubSpot/hubspot-local-dev-lib/pull/83 and release local dev lib

## Who to Notify
@brandenrodgers

### DIFF
--- a/packages/cli/commands/auth.js
+++ b/packages/cli/commands/auth.js
@@ -44,6 +44,7 @@ const { trackAuthAction, trackCommandUsage } = require('../lib/usageTracking');
 const { authenticateWithOauth } = require('../lib/oauth');
 const { EXIT_CODES } = require('../lib/enums/exitCodes');
 const { uiFeatureHighlight } = require('../lib/ui');
+const { logErrorInstance } = require('@hubspot/cli-lib/errorHandlers');
 
 const i18nKey = 'cli.commands.auth';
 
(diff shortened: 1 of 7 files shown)


packages/cli/commands/auth.js
```diff
@@ -98,7 +104,11 @@ exports.handler = async options => {
     case PERSONAL_ACCESS_KEY_AUTH_METHOD.value:
       configData = await personalAccessKeyPrompt({ env, account });
 
-      updatedConfig = await updateConfigWithPersonalAccessKey(
+      token = await getAccessToken(configData.personalAccessKey, env);
```
brandenrodgers, line 107: Does this need a try/catch like it had in the lib?

brandenrodgers:
IIRC there might be some sandboxes commands that make use of `updateConfigWithPersonalAccessKey `. Probably worth double checking for that.

brandenrodgers (approved):
🚢

PR #984, opened 2024-01-24
Adding `get-react-module` command

TanyaScales:
## Description and Context

Closes: https://git.hubteam.com/HubSpot/cms-developer-tooling/issues/159
Closes: https://git.hubteam.com/HubSpot/cms-developer-tooling/issues/158

This PR adds a subcommand to the `hs cms` command in order to retrieve a sample asset from a designated repo. In this case, it is currently pointing to the `cms-sample-assets` repo as the intended repo [`cms-react`](https://github.com/HubSpot/cms-react) is not public yet. Once it is, I can go ahead with updating this command to look at that instead.

Breakdown of the command:

`hs cms get-react-module [--name] [--dest]`

The user should be able to use the default command in order to see a list of the available directories at the specified repo path. In this case, we are going to point it to a `modules/` directory in the [cms-react](https://github.com/HubSpot/cms-react) repo. This will return a list of the modules like so:

```
React modules available to download:
NameOne
NameTwo
NameThree
``` 

The user should be able to choose what they want to grab

`hs cms get-react-module NameOne src/modules/` in order to download that module's content's to their src/modules path on their local machine, relative to where they are currently in their filesystem.


packages/cli/commands/cms/reactModules.js
```diff
@@ -0,0 +1,80 @@
...
+  if (list) {
+    try {
+      const contents = await listGithubRepoContents(
+        'HubSpot/cms-sample-assets',
+        'modules/',
+        'dir'
+      );
+
+      logger.group('React modules available to download:');
+      contents.map(x => {
```
bmatto, line 34: not really a canonical use of `map`. Generally Map would be used to create a new array based on some transformation of they current array. Here you're just iterating on contents and logging - I think a `forEach` would make more sense.

TanyaScales: Yep - makes sense - Updated the code to reflect

brandenrodgers:
> hs cms get-react-module [name] [dest] [--list]

> The user should be able to use the --list option in order to see a list of the available directories at the specified repo path. In this case, we are going to point it to a modules/ directory in the [cms-react](https://github.com/HubSpot/cms-react) repo. This will return a list of the modules like so:

Is `--list` necessary here? A pattern that we've used in some other places is to fall back to the "list" behavior whenever input is required, but not included.

So for example:
Skip the list if user runs `hs cms get-react-module NameOne ./modules/`, but show the list if the user omits the "name" input. An example of this is `hs project download`. It will show you a list of available projects to download if you omit the `--project` option.

This pattern would rely on you updating the args to be options rather than positional args (i.e. `get-react-module [--name] [--dest]`)

I also wonder if there's a better name for the command itself. Are we using "get" as a verb anywhere else? We have a few existing commands that will download code from GH repos (like `hs create module`). Would it make sense to bundle this functionality into one of those? Maybe something like `hs create react-module` (seems related to [this PR](https://github.com/HubSpot/hubspot-cli/pull/976)

Tagging @markhazlewood in case he has some suggestions on format too

TanyaScales:
> > hs cms get-react-module [name] [dest] [--list]
> 
> > The user should be able to use the --list option in order to see a list of the available directories at the specified repo path. In this case, we are going to point it to a modules/ directory in the [cms-react](https://github.com/HubSpot/cms-react) repo. This will return a list of the modules like so:
> 
> Is `--list` necessary here? A pattern that we've used in some other places is to fall back to the "list" behavior whenever input is required, but not included.
> 
> So for example: Skip the list if user runs `hs cms get-react-module NameOne ./modules/`, but show the list if the user omits the "name" input. An example of this is `hs project download`. It will show you a list of available projects to download if you omit the `--project` option.
> 
> This pattern would rely on you updating the args to be options rather than positional args (i.e. `get-react-module [--name] [--dest]`)

Ah, thank you, I do like this pattern better -- I was mostly going off of the initial request and trying to combine the two (get and list) commands into one. I can follow the fall back pattern here to keep the command clean and less cumbersome.

> I also wonder if there's a better name for the command itself. Are we using "get" as a verb anywhere else? We have a few existing commands that will download code from GH repos (like hs create module). Would it make sense to bundle this functionality into one of those? Maybe something like hs create react-module (seems related to https://github.com/HubSpot/hubspot-cli/pull/976

Yeah, again, going off the initial request -- but I'm wondering about this too. The difference between the two is that hs create module get you to a boilerplate sample module -- and this is supposed to be a fetch of the default react modules that the CAT team is publicizing. So, less boilerplate and more of a copy of the default modules and all of their available versions. Its kind of a create, but not exactly the intended use case (once that repo is up and running).

brandenrodgers:
> The difference between the two is that hs create module get you to a boilerplate sample module -- and this is supposed to be a fetch of the default react modules that the CAT team is publicizing. So, less boilerplate and more of a copy of the default modules and all of their available versions. Its kind of a create, but not exactly the intended use case (once that repo is up and running).

Ah I see, thanks for clarifying. Is it more similar to our existing `fetch` commands? I'm thinking maybe not. It's possible that this doesn't fit cleanly into any of our existing patterns, and `get` is more accurate. fwiw here are the existing and somewhat relevant patterns that I can think of:
- `download`: downloads developer project source code from HubSpot to your machine
- `create`: grabs source code from GH and writes it to your machine (mainly for cms modules and templates)
- `fetch`: downloads the latest version of your cms assets from Hubspot to your machine

TanyaScales:
> > The difference between the two is that hs create module get you to a boilerplate sample module -- and this is supposed to be a fetch of the default react modules that the CAT team is publicizing. So, less boilerplate and more of a copy of the default modules and all of their available versions. Its kind of a create, but not exactly the intended use case (once that repo is up and running).
> 
> Ah I see, thanks for clarifying. Is it more similar to our existing `fetch` commands? I'm thinking maybe not. It's possible that this doesn't fit cleanly into any of our existing patterns, and `get` is more accurate. fwiw here are the existing and somewhat relevant patterns that I can think of:
> 
> * `download`: downloads developer project source code from HubSpot to your machine
> * `create`: grabs source code from GH and writes it to your machine (mainly for cms modules and templates)
> * `fetch`: downloads the latest version of your cms assets from Hubspot to your machine

download and fetch are close but they are grabbing from HubSpot/portal as the source -- this is looking at a GH repo with the default modules. We aren't able to fetch these like the standard modules in the Design Manager/DFS because they aren't represented there.

TanyaScales:
> > The difference between the two is that hs create module get you to a boilerplate sample module -- and this is supposed to be a fetch of the default react modules that the CAT team is publicizing. So, less boilerplate and more of a copy of the default modules and all of their available versions. Its kind of a create, but not exactly the intended use case (once that repo is up and running).
> 
> Ah I see, thanks for clarifying. Is it more similar to our existing `fetch` commands? I'm thinking maybe not. It's possible that this doesn't fit cleanly into any of our existing patterns, and `get` is more accurate. fwiw here are the existing and somewhat relevant patterns that I can think of:
> 
> * `download`: downloads developer project source code from HubSpot to your machine
> * `create`: grabs source code from GH and writes it to your machine (mainly for cms modules and templates)
> * `fetch`: downloads the latest version of your cms assets from Hubspot to your machine

download and fetch are close but they are grabbing from HubSpot/portal as the source -- this is looking at a GH repo with the default modules. We aren't able to fetch these like the standard modules in the Design Manager/DFS because they aren't represented there.

`create` _is_ the closest pattern, but there is quite a bit of logic built into the current library function there the various module and manipulating what is returned from the repo as well as the prompting for scaffolding out various module types and metadata. Which is why I chose to not lean on that and instead scale back to a simple listing and retrieving functionality.

Maybe there is something to breaking up the createModule method in the lib to smaller pieces would help with leveraging different actions/bits of the logic in both the create flow and this one. If that were to happen, its possible that we could be smarter about the `create module` prompts, and resulting GH retrieval, etc. but its a bit of a larger rework at this point.

PR #987, opened 2024-01-29
Remove cli-lib/http dep + use local-dev-lib for OAuth

camden11:
## Description and Context
This removes three more dependancies on cli-lib modules:
* `http` is no longer needed (see https://hubspot.slack.com/archives/C01GV708YH3/p1706644044503449)
* Switches to local-dev-lib for `oauth`
* Switches to local-dev-lib for `OAuth2Manager`

This also fixes a big that would cause `hs auth --type=oauth2` to display an error message even when working properly.

## Todo
* Need to merge + release this https://github.com/HubSpot/hubspot-local-dev-lib/pull/92

## Who to Notify
@brandenrodgers

### DIFF
--- a/packages/cli/lib/validation.js
+++ b/packages/cli/lib/validation.js
@@ -118,7 +118,7 @@ async function validateAccount(options) {
       logger.error(
         `The OAuth2 configuration for account ${accountId} is incorrect`
       );
-      logger.error('Run "hscms auth oauth2" to reauthenticate');
+      logger.error('Run "hs auth --type=oauth2" to reauthenticate');
       return false;
     }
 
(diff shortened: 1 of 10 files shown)


packages/cli/lib/validation.js
```diff
@@ -118,7 +118,7 @@ async function validateAccount(options) {
       logger.error(
         `The OAuth2 configuration for account ${accountId} is incorrect`
       );
-      logger.error('Run "hscms auth oauth2" to reauthenticate');
```
brandenrodgers, line 121: Wow `hscms`, this really hasn't been touched in a long time

brandenrodgers (approved):
lgtm!

PR #990, opened 2024-01-31
Sandbox error handling updates

adamawang:
## Description and Context

For context, a migration was made to use local-dev-lib from cli-lib and some of the error message formats have changed. The util functions we are using on hubspot-cli (eg `isSpecifiedError`, `isMissingScopeError`) were outdated and were expecting a different error format. 

This PR aims to fix error handling for `hs sandbox` commands by migrating to the local-dev-lib versions of the util functions. I've also added a missing error scenario for `hs sandbox create` where users may not be ungated for development sandboxes. 

Required changes in local-dev-lib: https://github.com/HubSpot/hubspot-local-dev-lib/pull/93

## Screenshots

Before 1:
before1

After 1:
after1

Before 2:
before2

After 2:
after2

## TODO

NOTE: I wasn't able to migrate `isSpecifiedError` to the local-dev-lib version for `cli/lib/projects.js` and `cli/commands/project/dev.js`. The APIs being called in those files still come from cli-lib and use request-promise-native so the new LDL version of `isSpecifiedError` wont work since they're not AxiosErrors. 

## Who to Notify

@shannon-lichtenwalter

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -61,9 +61,11 @@ const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
 const {
   logApiErrorInstance,
   ApiErrorContext,
-  isMissingScopeError,
   isSpecifiedError,
 } = require('../../lib/errorHandlers/apiErrors');
+const {
+  isMissingScopeError,
+} = require('@hubspot/local-dev-lib/errors/apiErrors');
 const { logErrorInstance } = require('../../lib/errorHandlers/standardErrors');
 
 const i18nKey = 'cli.commands.project.subcommands.dev';
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -61,7 +61,7 @@ const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
 const {
   logApiErrorInstance,
   ApiErrorContext,
-  isSpecifiedError,
+  isSpecifiedError, // Migrate isSpecifiedError to local-dev-lib version only after uploadProject is migrated to local-dev-lib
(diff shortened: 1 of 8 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -61,9 +61,11 @@ const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
 const {
   logApiErrorInstance,
   ApiErrorContext,
-  isMissingScopeError,
   isSpecifiedError,
 } = require('../../lib/errorHandlers/apiErrors');
+const {
+  isMissingScopeError,
+} = require('@hubspot/local-dev-lib/errors/apiErrors');
```
adamawang, line 68: I migrated over `isMissingScopeError` to the LDL version since its for a sandboxes endpoint, but left `isSpecifiedError` as the local version. The usage of `isSpecifiedError` in this file is for `handleProjectUpload` which is still pulling `uploadProject` from the old `@hubspot/cli-lib` dependency where the http base is still request-promise-native. 

Same goes for `cli -> lib -> projects.js` in this repo since `fetchProject` is also pulled in from `@hubspot/cli-lib`. Usage is limited to just those two in projects files - I'll leave a comment in the code to reduce any confusion once that migration happens

packages/cli/lib/sandbox-create.js
```diff
@@ -101,6 +101,22 @@ const buildSandbox = async ({
         })
       );
       logger.log('');
+    } else if (
+      isSpecifiedError(err, {
+        statusCode: 403,
+        category: 'BANNED',
+        subCategory: 'SandboxErrors.DEVELOPMENT_SANDBOX_ACCESS_NOT_ALLOWED',
```
adamawang, line 108: Added a gating error scenario for when users are not ungated for crm-developement-beta

brandenrodgers (approved):
🚢

PR #993, opened 2024-02-06
Fix sandbox automatic PAK save

adamawang:
## Description and Context

This PR fixes an issue where PAKs were not being saved automatically when creating sandboxes. I tracked it down to an older PR here where a negate expression was missed: https://github.com/HubSpot/hubspot-cli/pull/964/files#diff-1571027f94f3fbe8d93b2e70d16ab063244781cd29fee699c8b8783e9eb163a4L237-L240

## Screenshots

n/a

## TODO

n/a


camden11 (approved):
Ah my bad! Ty for fixing

PR #995, opened 2024-02-07
Cleanup ignore tasks for sandbox sync

adamawang:
## Description and Context

Sandboxes introduced a bug where an extra sync type was being returned. This resulted in a very ugly sync progress display as shown below in the screenshots section.

We've hidden that sync type from our side so I'm removing the ignore code that was added as a hotfix for this issue here: https://github.com/HubSpot/hubspot-cli/pull/979/files#diff-1571027f94f3fbe8d93b2e70d16ab063244781cd29fee699c8b8783e9eb163a4R335

This PR also adds an additional check to make sure if this happens again, we don't show the user that missing translation error and just hide it. 

## Screenshots

Before:
Screenshot 2024-01-23 at 4 21 04 PM

Expected view:
Screenshot 2024-02-07 at 11 12 22

### DIFF
--- a/packages/cli/lib/lang.js
+++ b/packages/cli/lib/lang.js
@@ -52,7 +52,7 @@ const getTextValue = lookupDotNotation => {
       previouslyCheckedProp = prop;
     });
   } catch (e) {
-    logger.error(
+    logger.debug(
       `Unable to access language property: ${lookupProps.join(
         '.'
       )}. Failed to access prop "${previouslyCheckedProp}".`
@@ -92,4 +92,5 @@ const setLangData = (newLocale, newLangObj) => {
 module.exports = {
   i18n,
   setLangData,
+  MISSING_LANGUAGE_DATA_PREFIX,
 };
--- a/packages/cli/lib/lang.js
+++ b/packages/cli/lib/lang.js
@@ -57,6 +57,7 @@ const getTextValue = lookupDotNotation => {
         '.'
       )}. Failed to access prop "${previouslyCheckedProp}".`
     );
+    logger.error('Unable to access language property.');
     return missingTextData;
   }
 
--- a/packages/cli/lib/sandboxes.js
+++ b/packages/cli/lib/sandboxes.js
@@ -332,7 +332,6 @@ function pollSyncTaskStatus(
   const mergeTasks = {
     'lead-flows': 'forms', // lead-flows are a subset of forms. We combine these in the UI as a single item, so we want to merge here for consistency.
   };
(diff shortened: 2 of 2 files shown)


packages/cli/lib/lang.js
```diff
@@ -52,7 +52,7 @@ const getTextValue = lookupDotNotation => {
       previouslyCheckedProp = prop;
     });
   } catch (e) {
-    logger.error(
+    logger.debug(
```
adamawang, line 55: I changed this to log only in debug mode but let me know if you'd rather keep it as error. I figure that the end user may not want to see missing translation logs, but I can also see us missing some of these missing translation warnings while working in the CLI

camden11: Hmm that's a tricky one. My gut feeling is that this will _probably_ lead to us missing more errors - personally I don't always test things with the `--debug` flag.

camden11 (commented):
The fix for the `gates` task looks good to me, but think it might be good to get a second opinion on what to do about the lang error logging

cc @brandenrodgers @kemmerle

packages/cli/lib/lang.js
kemmerle: :two-cents: I'm ok with keeping this as `debug`, since this isn't an error that will break the actual command.

kemmerle (approved):
I'm good with the change to `logger.debug` 👍

packages/cli/lib/lang.js
brandenrodgers: Yeah I'm seeing both sides here. I agree that it's pretty gross for us to show the user the long `Unable to access language property some.long.key.string. Failed to access...` error message. But I think we'd end up missing a ton of errors if we don't log anything by default, just because it can be tough to notice the absence of a log.

I wonder if we could split the difference. What if we have two different versions of what gets logged here? By default we could log something that doesn't include the lang key and is much shorter.

**For example:**

With `--debug` (log what we log now)
`[Debug] Unable to access language property some.long.key.string. Failed to access...`

Without `--debug` (log a condensed version)
`[Error] Unable to access language property.`

brandenrodgers: Also something to remember is even in the HubSpot app we will show the "[missing translation en.something.blah]" strings to users (that's not great either tbh. Just making the point that it's not a new thing)

adamawang: Yeah good points there. I think the two log approach is a good happy medium, less gross to the end user but will still be shown as an error. Will push an update shortly

brandenrodgers (approved):

PR #996, opened 2024-02-07
Changes pre-ET

jsines:
Lots of changes - I'll comment throughout

See also https://github.com/HubSpot/cli-lib/pull/48


packages/cli/commands/theme/preview.js
```diff
@@ -32,21 +37,32 @@ const validateSrcPath = src => {
...
   const accountId = getAccountId(options);
-  const absoluteSrc = path.resolve(getCwd(), src);
 
+  const previewPromptAnswers = await previewPrompt(options);
+  const src = options.src || previewPromptAnswers.src;
+  let dest = options.dest || previewPromptAnswers.dest;
+  if (!dest) {
+    logger.error(i18n(`${i18nKey}.errors.destinationRequired`));
+    return;
+  }
```
jsines, line 56: Making src and dest optional and then prompting for them as Branden suggested

kemmerle:
Hi! I'm back from MAT leave, looking over this PR to catch up on what's happening with the Dev Tooling team 👋 

OOC, how does the theme preview command work? I see there's a local `src` theme directory and a remote `dest` directory in Design Manager. When using the command, does the `src` theme upload to the `dest` theme directory, and then the theme previewer automatically opens? I think you can add a description of the command that makes this more explicit.

TanyaScales (approved):

packages/cli/commands/theme/preview.js
```diff
@@ -32,21 +37,32 @@ const validateSrcPath = src => {
   return true;
 };
 
-exports.command = 'preview <src> <dest>';
-exports.describe = i18n(`${i18nKey}.describe`);
+exports.command = 'preview';
```
brandenrodgers, line 40: Not a blocker, but you could update this to `preview [--src] [--dest]` to help make the options more discoverable.

brandenrodgers (approved):
Thanks for making these changes! Looks great

PR #997, opened 2024-02-07
Port api modules 2

camden11:
## Description and Context
This switches deps for the following api modules to use `local-dev-lib`
* `api/customObject`
* `api/schema` (consolidated with `customObject`)
* `api/fileTransport`

## Testing
The following commands need to be tested
* `hs custom-object create`
* `hs custom-object scheme create`
* `hs custom-object schema delete`
* `hs custom-object schema fetch`
* `hs custom-object scheme update`

## TODO
Need to merge this PR and release LDL https://github.com/HubSpot/hubspot-local-dev-lib/pull/99

## Who to Notify
@kemmerle @brandenrodgers

### DIFF
--- a/packages/cli/commands/customObject/schema/delete.js
+++ b/packages/cli/commands/customObject/schema/delete.js
@@ -6,7 +6,9 @@ const {
 const { loadAndValidateOptions } = require('../../../lib/validation');
 const { trackCommandUsage } = require('../../../lib/usageTracking');
 const { getAccountId } = require('../../../lib/commonOpts');
-const { deleteSchema } = require('@hubspot/cli-lib/api/schema');
+const {
+  deleteObjectSchema,
+} = require('@hubspot/local-dev-lib/api/customObjects');
 const { i18n } = require('../../../lib/lang');
 
 const i18nKey =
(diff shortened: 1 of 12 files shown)


packages/cli/commands/customObject/schema/delete.js
```diff
@@ -25,7 +27,7 @@ exports.handler = async options => {
   trackCommandUsage('custom-object-schema-delete', null, accountId);
 
   try {
-    await deleteSchema(accountId, name);
+    await deleteObjectSchema(accountId, name);
```
brandenrodgers, line 30: Should we swap out `logErrorInstance` for `logApiErrorInstance` here too?

packages/cli/commands/customObject/schema/update.js
```diff
@@ -61,7 +62,8 @@ exports.handler = async options => {
       );
     }
   } catch (e) {
-    logErrorInstance(e, { accountId });
+    console.log(e);
```
brandenrodgers, line 65: Looks like a leftover console.log

camden11:
Fixed those things + added the new LDL dep so this should be good to go

brandenrodgers (approved):
looks great!

PR #999, opened 2024-02-13
Include flags at the beginning of the upload prompt

kemmerle:
## Description and Context
In a Slack thread, @brandenrodgers pointed out that it's difficult to discover the `--forceCreate` and `--message` flags in the `hs project upload` command. 

@brandenrodgers, you noted in Slack that the two flags weren't showing up in the help menu for the `create` command, but they only exist for the `upload` command, and they do appear in that help menu. 

I took your suggestion to include an example of the command with the `--forceCreate` command at the beginning of the upload flow. This isn't the usual pattern for us (the closest we come to it is in the `hs project add` command), so wanted to get your feedback. I know the message flag's message is a bit too long right now. 

## Screenshots

Screenshot 2024-02-13 at 3 01 26 PM

## TODO

- [ ] Address feedback (What should the message example read? Should it be included at all?) 

## Who to Notify

@brandenrodgers @camden11

### DIFF
--- a/packages/cli/lib/projects.js
+++ b/packages/cli/lib/projects.js
@@ -238,6 +238,7 @@ const ensureProjectExists = async (
     allowCreate = true,
     noLogs = false,
     withPolling = false,
+    uploadCommand = false,
   } = {}
 ) => {
   const accountIdentifier = uiAccountDescription(accountId);
@@ -251,10 +252,11 @@ const ensureProjectExists = async (
       let shouldCreateProject = forceCreate;
 
       if (allowCreate && !shouldCreateProject) {
+        const promptKey = uploadCommand ? 'createPromptUpload' : 'createPrompt';
         const promptResult = await promptUser([
           {
             name: 'shouldCreateProject',
-            message: i18n(`${i18nKey}.ensureProjectExists.createPrompt`, {
+            message: i18n(`${i18nKey}.ensureProjectExists.${promptKey}`, {
               projectName,
               accountIdentifier,
             }),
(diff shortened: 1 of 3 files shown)


brandenrodgers:
> you noted in Slack that the two flags weren't showing up in the help menu for the create command, but they only exist for the upload command, and they do appear in that help menu.

Whoops, that was a mistake on my part. I must have used `hs project create --help` accidentally while initially looking into this.

> I took your suggestion to include an example of the command with the --forceCreate command at the beginning of the upload flow. This isn't the usual pattern for us (the closest we come to it is in the hs project add command), so wanted to get your feedback. I know the message flag's message is a bit too long right now.

Sorry, I should have been more clear. I don't think we should add all of those log statements during the execution of the command. Instead, we should follow the prompt pattern and do something like this:
`$ [--forceCreate] The project <projectName> does not exist in <targetAccount>. Would you like to create it? (Y/n)`
Where we add an example of the option that can be used to skip the prompt at the beginning of the prompt question.

kemmerle:
@brandenrodgers, thanks so much for clarifying! That makes a lot of sense. I've added the `forceCreate` flag directly to the prompt flow (image attached). Because only the upload command uses the `forceCreate` flag, I had to add some minor logic to make sure it only appears when running the upload command. Happy to change it up, if that doesn't work!

packages/cli/lib/projects.js
```diff
@@ -238,6 +238,7 @@ const ensureProjectExists = async (
     allowCreate = true,
     noLogs = false,
     withPolling = false,
+    uploadCommand = false,
```
brandenrodgers, line 241: hmmm the arguments here are getting a little out of hand, but I see why you added this. Seems fine to me for now 👍

brandenrodgers (approved):

PR #1000, opened 2024-02-13
`hs project dev` supports dev test accounts

adamawang:
## Description and Context

If a user's default account is a developer test account, `hs project dev` should show a confirmation like we do if your default account is a sandbox. No picker will be shown. 

I also updated `hs accounts list` to support dev test accounts (and app developer accounts but this is not really a thing yet until https://git.hubteam.com/HubSpot/AppLifecycle/issues/368 is done) with the hierarchy view like sandboxes. 

I'm using `[dev test account]` as the account "type" but will get feedback from Mark. `developer test account` seemed really long and `developer test` may be confusing. 

## Screenshots

Screenshot 2024-02-13 at 16 06 00

Question: for app developer accounts, should we show it like this or add something like `[app deveveloper]` ? 

Screenshot 2024-02-13 at 16 02 12

`hs project upload` and the rest of the projects commands work as expected
Screenshot 2024-02-15 at 15 19 51

and same with the sandboxes commands after all the name/helper changes

## TODO

This PR does NOT support app developer accounts yet, that is still to come. This is only for developer test accounts support. 

## Who to Notify

@shannon-lichtenwalter


packages/cli/commands/accounts/list.js
```diff
@@ -22,29 +29,32 @@ exports.describe = i18n(`${i18nKey}.describe`);
...
+    .filter(p =>
+      p.accountType
+        ? isSandbox(p) || isDeveloperTestAccount(p)
+        : p.sandboxAccountType !== null
+    )
+    .forEach(p => {
+      if (p.parentAccountId) {
+        mappedPortalData[p.parentAccountId] = [
+          ...(mappedPortalData[p.parentAccountId] || []),
+          p,
```
adamawang, line 54: All backwards compatible with older configs without `accountType`, however the validation PR (https://github.com/HubSpot/hubspot-local-dev-lib/pull/105) just got merged so these should be populated as soon as any command with a validation step is run

packages/cli/commands/project/dev.js
```diff
@@ -96,19 +100,27 @@ exports.handler = async options => {
...
+  );
 
-  if (!targetAccountId && defaultAccountIsSandbox) {
+  if (
+    !targetAccountId &&
+    (defaultAccountIsSandbox || defaultAccountIsDeveloperTestAccount)
+  ) {
     logger.log();
-    const useDefaultSandboxAccount = await confirmDefaultSandboxAccountPrompt(
+    const useDefaultAccount = await confirmDefaultAccountPrompt(
```
adamawang, line 112: I renamed this to be more generic since the copy in the confirmation could be used for both sandbox and dev test account

packages/cli/lib/developerTestAccounts.js
```diff
@@ -0,0 +1,14 @@
+const {
+  HUBSPOT_ACCOUNT_TYPES,
+} = require('@hubspot/local-dev-lib/constants/config');
+
+const DEV_TEST_ACCOUNT_STRING = 'dev test account';
```
adamawang, line 5: @markhazlewood let me know your thoughts on this naming - I posted some screenshots in the main PR body

packages/cli/lib/sandboxes.js
```diff
@@ -51,11 +54,24 @@ const sandboxApiTypeMap = {
...
+
+const isSandbox_OLD = config =>
+  config.sandboxAccountType && config.sandboxAccountType !== null;
 
 const isSandbox = config =>
-  config.sandboxAccountType && config.sandboxAccountType !== null;
+  config.accountType
+    ? config.accountType === HUBSPOT_ACCOUNT_TYPES.STANDARD_SANDBOX ||
+      config.accountType === HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX
+    : isSandbox_OLD(config);
```
adamawang, line 74: Updated these to use `accountType` but still backwards compatible

packages/cli/lib/sandboxes.js
```diff
@@ -51,20 +58,38 @@ const sandboxApiTypeMap = {
...
 
-  const message = `${config.name} ${isSandbox(config) ? sandboxName : ''}(${
-    config.portalId
-  })`;
+function getAccountName(config, bold = true) {
+  const message = `${config.name} ${
+    isSandbox(config)
+      ? getSandboxName(config)
+      : isDeveloperTestAccount(config)
```
adamawang, line 89: Added developer test account support here so we get the correct hub tag in other command usages

packages/cli/lib/developerTestAccounts.js
brandenrodgers: I'll defer to Mark here for final say, but should we be consistent with the use of `dev` and `developer` in the account names (i.e. `dev test account` vs. `developer sandbox`)? I'm partial to the shorter `dev` naming.

packages/cli/lib/sandboxes.js
```diff
@@ -51,20 +58,38 @@ const sandboxApiTypeMap = {
...
 };
 
-const getSandboxTypeAsString = type =>
-  type === 'DEVELOPER' ? 'development' : 'standard';
+const getSandboxTypeAsString = accountType => {
+  if (
+    accountType === HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX ||
+    accountType === 'DEVELOPER'
+  ) {
+    return 'development';
```
brandenrodgers, line 66: Could this return the constants that you have for these names?

packages/cli/lib/sandboxes.js
brandenrodgers: We have a [ui util for logging the user-friendly account name](https://github.com/HubSpot/hubspot-cli/blob/master/packages/cli/lib/ui.js#L73). It's pretty similar to what this is doing (but currently isn't very smart about handling different account types). Seems like maybe we could consolidate the functionality so we just have one util that we can use everywhere.

adamawang: ah nice, yeah I can move the logic over to that util so everyone gets it

packages/cli/lib/developerTestAccounts.js
markhazlewood42: Agree. For external consumption we should always use "developer test account" fully spelled out.

adamawang (commented):
Updated to use constants for sandbox type checks as well

packages/cli/lib/developerTestAccounts.js
adamawang: Cool - will update to the full name 👍  thanks!

packages/cli/lib/sandboxes.js
```diff
@@ -1,73 +1,66 @@
...
+  DEVELOPER_SANDBOX_TYPE,
+  DEVELOPER_SANDBOX,
+  STANDARD_SANDBOX,
+} = require('./constants');
 
 const syncTypes = {
   OBJECT_RECORDS: 'object-records',
 };
 
 const sandboxTypeMap = {
```
brandenrodgers, line 28: Do we need this just because of how `sandboxAccountType` works? Seems like we can probably also remove this once we're ready to fully remove `sandboxAccountType`, right?

packages/cli/lib/developerTestAccounts.js
adamawang: One more question for you @markhazlewood42 - do we want to differentiate between standard accounts (customer portals) and app developer accounts? Like do we want to show `[app developer]` next to those accounts

packages/cli/lib/sandboxes.js
```diff
@@ -1,73 +1,66 @@
...
-  personalAccessKeyPrompt,
-} = require('./prompts/personalAccessKeyPrompt');
-const { logErrorInstance } = require('./errorHandlers/standardErrors');
-
-const STANDARD_SANDBOX = 'standard';
-const DEVELOPER_SANDBOX = 'developer';
+  HUBSPOT_ACCOUNT_TYPES,
+} = require('@hubspot/local-dev-lib/constants/config');
+const {
+  STANDARD_SANDBOX_TYPE,
```
brandenrodgers, line 18: Are these different than what we have in `HUBSPOT_ACCOUNT_TYPES `? Is there any way we could get rid of the need for these constants in favor of the export from local-dev-lib?

packages/cli/lib/sandbox-create.js
```diff
@@ -2,10 +2,7 @@ const SpinniesManager = require('./SpinniesManager');
 const {
```
brandenrodgers, line 2: Nit, but most of the files in the CLI use camel case. Would you mind updating this for consistency?

packages/cli/lib/sandboxes.js
adamawang: yeah good point... its a bit confusing cause our own sandboxes endpoints use `DEVELOPER` and `STANDARD` for sandbox types which is why I had kept them, but I'll revisit all these once I get to the `sandboxAccountType` migration issue

packages/cli/lib/ui.js
```diff
@@ -65,16 +70,22 @@ const uiLink = (linkText, url) => {
...
+ * @param {boolean} bold
  * @returns {string}
  */
-const uiAccountDescription = accountId => {
+const uiAccountDescription = (accountId, bold = true) => {
   const account = getAccountConfig(accountId);
-  return chalk.bold(
-    account.name ? `${account.name} (${account.portalId})` : account.portalId
-  );
+  const message = `${account.name} ${
```
brandenrodgers, line 81: Not a blocker, but sometimes the nested ternary expressions can be tricky to follow. This would probably also be ok as a few if statements.

packages/cli/lib/sandboxes.js
adamawang: Yep - same as the other comment, I think I'll get around to swapping these once I get to the `sandboxAccountType` migration issue

packages/cli/lib/sandbox-create.js
adamawang: yes not a prob, I did also notice inconsistencies but I'll at least get these to camel case

brandenrodgers:
Not a blocker (feel free to look into this later if you want), but I noticed that the account type labels for `hs accounts list` only show up when the parent account is also configured in the CLI config file.

Example:
image

In here, `prodsandbox` is a dev sandbox account and `app-test-account-2` is a dev test account

brandenrodgers:
I'm not seeing the `[developer test account]` label being included for my dev test account here:
image


Also ignore the failed upload. It's unrelated

adamawang:
> Not a blocker (feel free to look into this later if you want), but I noticed that the account type labels for hs accounts list only show up when the parent account is also configured in the CLI config file.

Oh wow... this bug has actually been around since I wrote it way back, dont know how I missed testing that scenario but good catch! Will push up a fix shortly

For the second one - is the entry missing an `accountType` or is it set to `STANDARD` somehow? I just ran it locally and it seemed to work as expected: 
Screenshot 2024-02-15 at 16 56 13

brandenrodgers:
> For the second one - is the entry missing an accountType or is it set to STANDARD somehow? I just ran it locally and it seemed to work as expected for both QA and PROD:

Ah sorry, I realized that I had some issue with my local version of your branch. For some reason it got disconnected from the remote version so my pulls weren't doing anything, and I didn't notice. This is all WAD now! 🎉

brandenrodgers (approved):
🚢

PR #1002, opened 2024-02-14
chore: Update dev server contract

joe-yeager:
## Description and Context

- Update the dev server setup contract to take in a `logger` and a `urls` object that contains the `api` url and the `web`  
- Update the dev server start contract to take in the `httpClient` from the new `local-dev-lib`

### DIFF
--- a/packages/cli/lib/DevServerManager.js
+++ b/packages/cli/lib/DevServerManager.js
@@ -9,6 +9,7 @@ const {
   stopPortManagerServer,
   requestPorts,
 } = require('@hubspot/local-dev-lib/portManager');
+const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
 
 const i18nKey = 'cli.lib.DevServerManager';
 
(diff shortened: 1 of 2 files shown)


packages/cli/lib/DevServerManager.js
```diff
@@ -75,6 +81,11 @@ class DevServerManager {
             debug,
             onUploadRequired,
             promptUser,
+            logger,
+            urls: {
```
joe-yeager, line 85: I thought we initially needed the `env`, but we were exchanging that for the api and web urls using the legacy `httpClient.getRequestOptions` method.  To get rid of the need for that I found these two methods to get the urls.  It seems to be the same as what we want.

brandenrodgers: Just fyi, `getRequestOptions` was replaced with [getAxiosConfig](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/http/getAxiosConfig.ts). But that being said, I like what you have here better than using getAxiosConfig 👍

packages/cli/lib/DevServerManager.js
```diff
@@ -75,6 +80,11 @@ class DevServerManager {
             debug,
             onUploadRequired,
             promptUser,
+            logger,
```
brandenrodgers, line 83: Just a thought, do we still need to pass in `debug` if we're also passing in the logger? I'm thinking the `debug` might be outdated by at this point since `logger.debug()` should do the same thing.

brandenrodgers (approved):
lgtm! I had one suggestion about `debug`, but not a blocker. Feel free to merge whenever you're ready for it. It's backwards compatible so should be safe.

packages/cli/lib/DevServerManager.js
joe-yeager: Yeah, good call. I'll update this PR and the dev server to remove it

PR #1003, opened 2024-02-15
Swap logger imports to pull from local-dev-lib

brandenrodgers:
## Description and Context

Swapping over the logger imports. This should have no impact on any log statements, so everything should work the exact same way. We also have to swap over imports from the `local-dev-lib/logger` folder, but I think we may want to revisit the pattern there before committing to that. I found [the reasoning for why we originally added logging/ui utils to the local-dev-lib repo](https://github.com/HubSpot/hubspot-local-dev-lib/pull/43), but I think we may want to reconsider (even if it means potentially duplicating some code). The local-dev-lib library was intended to help abstract away the business logic from the UI (aka CLI), so adding ui utils into local-dev-lib blurs the line of separation.


camden11 (approved):
Looks like you have some (presumably very simple) merge conflicts but LGTM once those are fixed

PR #1004, opened 2024-02-16
[WIP] Port logging UI utils back from local-dev-lib

brandenrodgers:
## Description and Context

Relies on: https://github.com/HubSpot/hubspot-local-dev-lib/pull/108

Porting all of the deleted logging UI utils from local-dev-lib into the CLI. The idea is to keep the UI utils here in the CLI, which is effectively the UI of our local dev tooling. I reorganized things in the `lib/` folder because the file structure in that folder is so flat that it's tough to find things in there. Now we have `lib/ui/` which can be a place for all of our UI utils. We're planning to eventually expand on the ui utils at some point this year so this sets us up for that.

I mentioned it in the LDL PR, but there is a little bit of duplicated code in here because of the other two packages that live in this repo. My opinion is that we should work towards breaking those two packages out into their own repos. There isn't a ton of maintenance that goes into supporting them, so it wouldn't be a big lift.


packages/cli/lang/en.lyaml
```diff
@@ -1036,6 +1036,17 @@ en:
               linkText: "HubSpot's sample projects"
               url: "https://developers.hubspot.com/docs/platform/sample-projects?utm_source=cli&utm_content=project_create_whats_next"
               message: "See {{ link }}"
+        git:
```
brandenrodgers, line 1039: I ported some of the git-related logging utils back into here, so I had to also add the lang keys for them

packages/cli/lib/ui/serverlessFunctionLogs.js
```diff
@@ -0,0 +1,103 @@
+const moment = require('moment');
```
brandenrodgers, line 1: This serverless logging functionality is duplicated in here and in the serverless-dev-runtime package. It's a little gross, but ideally we can deprecate that package soon in favor of the developer projects version of local dev for serverless functions.

packages/webpack-cms-plugins/package.json
```diff
@@ -17,7 +17,6 @@
     "test": "echo \"Error: run tests from root\" && exit 1"
   },
   "dependencies": {
-    "@hubspot/cli-lib": "^9.0.0",
```
camden11, line 20: just beautiful

camden11 (approved):
One very minor comment, otherwise this looks good to me!

packages/cli/commands/project/logs.js
```diff
@@ -9,15 +9,12 @@ const {
 } = require('../../lib/commonOpts');
 const { trackCommandUsage } = require('../../lib/usageTracking');
 const { logger } = require('@hubspot/local-dev-lib/logger');
-// const { outputLogs } = require('@hubspot/cli-lib/lib/logs');
+// const { outputLogs } = require('../lib/ui/serverlessFunctionLogs');
```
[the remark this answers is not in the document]
brandenrodgers: I guess we could delete all of them. I think they were kept around because we plan to bring the functionality back sometime soon hopefully.

PR #1005, opened 2024-02-16
Replace `sandboxAccountType` usage with new `accountType`

adamawang:
## Description and Context

This PR removes the old `sandboxAccountType` property which is now deprecated, and replaces it with the new `accountType` property. Also included is a new `HUBSPOT_ACCOUNT_TYPES_STRING` const from local-dev-lib which is used for copy purposes when putting together messages + error message bodies. 

Related LDL PR: https://github.com/HubSpot/hubspot-local-dev-lib/pull/109

## Screenshots

Will post comment screenshots to illustrate some of the changes

## Who to Notify

@shannon-lichtenwalter


packages/cli/lib/sandboxes.js
```diff
@@ -14,53 +14,49 @@ const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
...
-  STANDARD: STANDARD_SANDBOX_TYPE,
-  standard: STANDARD_SANDBOX_TYPE,
+  DEV: HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX,
+  dev: HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX,
+  DEVELOPER: HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX,
+  developer: HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX,
+  DEVELOPMENT: HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX,
+  development: HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX,
+  STANDARD: HUBSPOT_ACCOUNT_TYPES.STANDARD_SANDBOX,
+  standard: HUBSPOT_ACCOUNT_TYPES.STANDARD_SANDBOX,
```
adamawang, line 30: Keeping this map here for users who use the --type flag where they can type whatever they want. This just handles random strings that they may type, but let me know if this is overkill and we can just remove it

packages/cli/lang/en.lyaml
```diff
@@ -724,7 +724,7 @@ en:
               developmentSandbox: "Sync CRM object definitions to {{#cyan}}{{#bold}}{{ sandboxName }}{{/bold}}{{/cyan}} from {{#bold}}{{ parentAccountName }}{{/bold}}?"
               standardSandbox: "Sync all supported assets to {{#cyan}}{{#bold}}{{ sandboxName }}{{/bold}}{{/cyan}} from {{#bold}}{{ parentAccountName }}{{/bold}}?"
             failure:
-              notSandbox: "Sync must be run in a sandbox account. Your default account is a production account. Run {{#bold}}hs auth{{/bold}} to connect your sandbox account to the CLI or {{#bold}}hs accounts use{{/bold}} to change your default account, then try again."
+              notSandbox: "Sync must be run in a sandbox account. Your default account is a {{ accountType }}. Run {{#bold}}hs auth{{/bold}} to connect your sandbox account to the CLI or {{#bold}}hs accounts use{{/bold}} to change your default account, then try again."
```
camden11, line 727: Super minor but should we also include the account name here like with `invalidAccountType`?

packages/cli/lib/sandboxes.js
camden11: Hmm,  think having the different variations of "developer" makes sense, but we could also `toLowerCase` or `toUpperCase` any strings that we're using as keys for this and not need the different capitalizations

camden11 (approved):
Haven't had a chance to test yet, but all the code here LGTM. Couple minor comments but nothing blocking

packages/cli/lib/sandboxes.js
adamawang: good call, pushed an update with this and the lang key change!

PR #1008, opened 2024-02-22
Port `handlefieldsJS` to local-dev-lib (Corresponding PR in local-dev-lib)

kemmerle:
## Description and Context

Corresponding local-dev-lib PR: https://github.com/HubSpot/hubspot-local-dev-lib/pull/110

Imports from `handleFieldsJS.js` appear in three files:
- `lib/upload`. `isConvertibleFieldJs` appears in the `getUploadableFileList` function. `getUploadableFileList` appears in two commands with the `--convertFields`: `hs upload` and `hs watch`.  
- `commands/upload`. The upload command directly imports `FieldsJs`, `isConvertibleJS`, and `cleanupTmpDirSync` functions. 
- `commands/cms/convert-fields`. The convert-fields command imports `FieldsJs` and `isConvertibleFieldJs` directly. 

I did need to fix one import in `local-dev-lib` (See linked PR). I encountered multiple bugs but was able to address them with the help of the whole team. Also, there was an issue where `src` wasn't declared as a positional in the command's export, so it could only be used as a flag. I've fixed that, as well. 

## Testing

Background info: The `convert-fields` flag and command allows CMS developers to generate a `fields.json` file from a `fields.js` file. 

To test out this PR, you need to:

1) Create a module in your `tmp` directory using the `hs create module` command. Delete the `fields.json` from the module directory, and replace it with a `fields.js` file. You can copy and paste the following code into `fields.js`:

```
module.exports = fieldOptions => {
  return [
    {
      required: true,
      locked: false,
      help_text: '',
      inline_help_text: '',
      name: 'button_text',
      label: 'Button text',
      type: 'text',
      default: 'Add a button link here',
    },
  ];
};
```

2) To test the upload command, run `yarn hs upload tmp/yourModule.module --convert-fields`. 

3) To test the watch command, run `yarn hs watch tmp/yourModule.module --initialUpload --convert-fields`

4) To test the convert-fields command, run `yarn hs cms convert-fields tmp/yourModule.module`. **You will need to use a symlinked version of the local-dev-lib library on the correct branch.**

## Who to Notify

@brandenrodgers @camden11


packages/cli/commands/cms/convertFields.js
```diff
@@ -15,7 +15,7 @@ const { trackConvertFieldsUsage } = require('../../lib/usageTracking');
 const { logErrorInstance } = require('../../lib/errorHandlers/standardErrors');
 const i18nKey = 'cli.commands.convertFields';
 
-exports.command = 'convert-fields';
+exports.command = 'convert-fields [src] [--fieldOptions]';
```
brandenrodgers, line 18: `src` is an optional argument so we shouldn't list it here. We should also change it to be an option instead of a positional since we are preferring the use of options over positionals (and that way this change won't be a breaking change, b/c currently the command does not support passing in src as a positional).

brandenrodgers (approved):
Just one comment about removing the optional arguments from the description, and switching `src` to be an option rather than a positional. But otherwise lgtm 👍

packages/cli/commands/cms/convertFields.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1008/commits/cd2468ad58eea69e033939ec6fe39a472fe25da8

camden11 (approved):
Works great locally! Should be good to go once local-dev-lib is out

PR #1011, opened 2024-03-04
Support public app config

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/510
This makes a few behind-the-scenes changes to `hs project dev` to make parsing public app configs a bit cleaner and less error prone. Previously, `findProjectComponents` would look for any file _ending_ in `app.json`, which included `public-app.json`. Now it looks for specific config files, and then adds their corresponding `type` to the component object that gets passed into the dev servers. This also changes the `type` for private apps from `app` to `private-app`. From what I could tell, `type` isn't currently being used by the dev server, but will likely be useful when there are multiple types of components supported.

At this point, running `hs project dev` on a public app will still fail, since they're not supported by the UIE dev server yet. This PR shouldn't change the behavior of `hs project dev` at all - to test, make sure local development on private apps still works.

## Who to Notify
@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli/lib/projectStructure.js
+++ b/packages/cli/lib/projectStructure.js
@@ -103,14 +103,13 @@ async function findProjectComponents(projectSourceDir) {
       const parsedAppConfig = loadConfigFile(projectFile);
 
       if (parsedAppConfig && parsedAppConfig.name) {
-        const appPath = `${dir}/`;
-        const isLegacy = getIsLegacyApp(parsedAppConfig, appPath);
+        const isLegacy = getIsLegacyApp(parsedAppConfig, dir);
 
         components.push({
           type: getTypeFromConfigFile(base),
           config: parsedAppConfig,
           runnable: !isLegacy,
-          path: appPath,
+          path: dir,
         });
       }
     }
--- a/packages/cli/lib/projectStructure.js
+++ b/packages/cli/lib/projectStructure.js
@@ -9,15 +9,10 @@ const COMPONENT_TYPES = Object.freeze({
   publicApp: 'public-app',
 });
 
-const PRIVATE_APP_COMPONENT_CONFIG = 'app.json';
-const PUBLIC_APP_COMPONENT_CONFIG = 'public-app.json';
-
 const CONFIG_TYPE_MAP = {
-  [PRIVATE_APP_COMPONENT_CONFIG]: COMPONENT_TYPES.privateApp,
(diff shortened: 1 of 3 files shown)


packages/cli/lib/projectStructure.js
```diff
@@ -5,10 +5,23 @@ const { logger } = require('@hubspot/local-dev-lib/logger');
 const { logErrorInstance } = require('./errorHandlers/standardErrors');
```
brandenrodgers, line 5: Not related to your PR, but I had forgotten about this file. I wonder if there's a case for moving this into local-dev-lib at some point?

camden11: I could be mixing thing up, but wasn't the idea here that we moved this _out_ of cli-lib/local-dev-lib because its specific to how the CLI displays the errors thrown by the libraries?

brandenrodgers: re: `./errorHandlers/standardErrors`, yeah you're totally right. I was actually talking about this entire projectStructure.js file though.

camden11: Ohhh gotcha. Yeah that's a good question. It does seem like it would make sense in local-dev-lib, though could also see it making sense to wait until its needed in another package (like a theoretical vscode extension)

brandenrodgers: Agreed 👍 maybe just something to keep an eye on for now

packages/cli/lib/projectStructure.js
```diff
@@ -84,24 +97,21 @@ async function findProjectComponents(projectSourceDir) {
...
-          const isLegacy = getIsLegacyApp(parsedAppConfig, appPath);
-
-          components.push({
-            type: COMPONENT_TYPES.app,
-            config: parsedAppConfig,
-            runnable: !isLegacy,
-            path: appPath,
-          });
-        }
+        const appPath = `${dir}/`;
```
brandenrodgers, line 106: You may need to use `path` here instead of manually inserting the slash to make sure it's compatible with other OS's

camden11: Oooh good call

camden11: Just tested and looks like it doesn't actually need the slash

brandenrodgers (approved):
lgtm!

PR #1014, opened 2024-03-06
Swap cli lib constants

camden11:
## Description and Context
This removes _almost_ all of the dependancies on `@hubspot/cli-lib/lib/constants` and replaces them with either imports from `local-dev-lib` or the constants file within the repo. Still a fair amount of testing to do here - these are minor changes but a lot of files have been touched.

## TODO
* `GITHUB_RELEASE_TYPES` is used by the `cli-lib` github functions, but not the `local-dev-lib` github functions, so leaving it as is for now with the hope that we can remove it completely once we switch to local-dev-lib for github stuff.

## Who to Notify
@brandenrodgers @kemmerle


packages/cli/lib/projects.js
```diff
@@ -44,6 +45,10 @@ const { HUBSPOT_PROJECT_COMPONENTS_GITHUB_PATH } = require('./constants');
 
 const i18nKey = 'cli.lib.projects';
 
+const SPINNER_STATUS = {
```
brandenrodgers, line 48: You could also export VALID_STATUSES from SpinniesUtils

brandenrodgers (approved):
👌 (I had one comment about potentially DRY-ing up a constant, but not a blocker)

PR #1016, opened 2024-03-12
Add platform version errors

kemmerle:
## Description and Context

This PR adds platform version errors to commands like `hs project upload`. To do so, I've created an `overrideErrors` function that can "intercept" our normal error handlers to display custom errors for specific use cases, ie. with platform version errors. 

## Screenshots

See [mocks](https://zenhub.hubteam.com/workspaces/dpg---developer-projects-62546eb6da7ec20010a53332/issues/hubspot/developer-projects-ui/692)

Screenshot 2024-03-12 at 10 22 18 AM

In the mocks, the ERROR header is red. Looking through our code base, however, I couldn't find any instances where we used chalk within i18n translations. Is there another way to add color to this text? @brandenrodgers @camden11 

## TODO

- [ ] Address feedback
- [ ] Test with Roger and Caitlyn. 

## Who to Notify

@brandenrodgers @camden11


packages/cli/lib/errorHandlers/overrrideErrors.js
```diff
@@ -0,0 +1,95 @@
...
+    i18n(`${i18nKey}.platformVersionErrors.betaLink`, {
+      docsLink,
+    })
+  );
+  logger.log(createDashes());
+}
+
+function overrideErrors(err) {
+  if (
+    isSpecifiedError(err, {
```
kemmerle, line 55: My instinct was to create a helper function to determine the three platform version specified errors, since `isSpecifiedError` is a little clunky to write out every time. But I realized that I would still need to include the helper function three times and include the subcategory each time. So I just left as is with `isSpecifiedError`.

packages/cli/lib/errorHandlers/overrrideErrors.js
```diff
@@ -0,0 +1,95 @@
...
+const { uiLink } = require('../ui');
+
+const i18nKey = 'cli.lib.errorHandlers.overrideErrors';
+
+function createDashes() {
+  return '-'.repeat(79);
+}
+
+function createPlatformVersionError(subCategory, errorData) {
+  const platformVersion = errorData.context.PLATFORM_VERSION[0];
```
kemmerle, line 15: @ypzheng, will there only ever be one item in the `PLATFORM_VERSION` array? Could this simply be a string? It feels brittle to me right now.

packages/cli/lib/errorHandlers/overrrideErrors.js
```diff
@@ -0,0 +1,95 @@
...
+    translationKey = 'platformVersionRetired';
+  } else if (
+    subCategory ===
+    PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST
+  ) {
+    translationKey = 'nonExistentPlatformVersion';
+  }
+
+  logger.log(createDashes());
+  logger.log(i18n(`${i18nKey}.platformVersionErrors.header`));
```
kemmerle, line 38: I broke up the text a bit to make the i18n translations more contained, especially since I needed to interpolate the platform version and docs link. Is there a better approach here?

packages/cli/lib/errorHandlers/overrrideErrors.js
```diff
@@ -0,0 +1,95 @@
+const { isSpecifiedError } = require('@hubspot/local-dev-lib/errors/apiErrors');
+const { logger } = require('@hubspot/local-dev-lib/logger');
+
+const { PLATFORM_VERSION_ERROR_TYPES } = require('../constants');
+const { i18n } = require('../lang');
+const { uiLink } = require('../ui');
+
+const i18nKey = 'cli.lib.errorHandlers.overrideErrors';
+
+function createDashes() {
```
brandenrodgers, line 10: I think we have a uiLine component that does this.

packages/cli/lib/errorHandlers/overrrideErrors.js
```diff
@@ -0,0 +1,95 @@
+const { isSpecifiedError } = require('@hubspot/local-dev-lib/errors/apiErrors');
```
brandenrodgers, line 1: Looks like you snuck an extra "r" into override in the file name

packages/cli/lib/errorHandlers/overrrideErrors.js
brandenrodgers: This looks good to me 👍. Should this top log line be using `logger.error()`? Then you wouldn't need to manually insert the "[ERROR]" [here](https://github.com/HubSpot/hubspot-cli/pull/1016/files#diff-f1b852efd5f578b3521c0020953a63d55a5670e3bfe42219ffb0f77bae7c4cdeR1346).

packages/cli/lib/errorHandlers/overrrideErrors.js
```diff
@@ -0,0 +1,95 @@
...
+    isSpecifiedError(err, {
+      subCategory:
+        PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST,
+    })
+  ) {
+    createPlatformVersionError(
+      PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST,
+      err.response.data
+    );
+    return true;
```
brandenrodgers, line 88: I like this pattern of returning a boolean because it gives us the flexibility to override an error but still let the BE-returned error get logged if we ever want to do that (by returning false)

packages/cli/lib/errorHandlers/overrrideErrors.js
kemmerle: Ooh nice, thanks Branden!

packages/cli/lib/errorHandlers/overrrideErrors.js
kemmerle: 🤦‍♀️ Woops, thanks for catching that!

packages/cli/lib/errorHandlers/apiErrors.js
```diff
@@ -257,6 +258,7 @@ function logApiErrorInstance(error, context) {
   // Use the new local-dev-lib error handler
   // NOTE: This will eventually replace the logic in logApiStatusCodeError
   if (error.isAxiosError) {
+    if (overrideErrors(error)) return;
```
camden11, line 261: This doesn't need to be blocking if you don't agree, but this pattern seems a little confusing to me + `overrideErrors` is kind of a vague name and doesn't really say what it's actually doing.

Seeing that we do this for `statusCodeErrors`
```
 if (isApiStatusCodeError(error)) {
    logApiStatusCodeError(error, context);
    return;
  }
```

Would it make sense to just make a `isPlatformVersionError` function and then do
```
 if (isPlatformVersionError(error)) {
    logPlatformVersionError(error, context);
    return;
  }
```
where `logPlatformVersionError` is basically just the current `overrideErrors`

packages/cli/lib/errorHandlers/overrrideErrors.js
```diff
@@ -0,0 +1,95 @@
...
+
+function createPlatformVersionError(subCategory, errorData) {
+  const platformVersion = errorData.context.PLATFORM_VERSION[0];
+  const docsLink = uiLink(
+    i18n(`${i18nKey}.platformVersionErrors.docsLink`),
+    'https://developers.hubspot.com/docs/platform/platform-versioning'
+  );
+  let translationKey;
+
+  if (
```
camden11: this could be a bit cleaner as an object that maps subcategory to i18nkey, i.e.
```
{
  [PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_RETIRED]: 'platformVersionRetired'
}

**Review by brandenrodgers — APPROVED** [review]
looks good to me!

File: packages/cli/lib/errorHandlers/overrrideErrors.js (hunk 2 above)
Comment (ypzheng) [reply packages/cli/lib/errorHandlers/overrrideErrors.js]: Did look into this a bit. It seems like there is no easy way to make it a string due to how errors are transformed and parsed along backend clients & services.

However, there will only be one string in `PLATFORM_VERSION` because the way we construct platform version related error on backend is always matching one single string to an error token (i.e. PLATFORM_VERSION_RETIRED). 

I'll keep this in mind though. If I find anything better in the future.

File: packages/cli/lang/en.lyaml [hunk 9]
```diff
@@ -1341,6 +1341,15 @@ en:
             portalMissingScope: "Your account does not have access to this action. Talk to an account admin to request it."
             userMissingScope: "You don't have access to this action. Ask an account admin to change your permissions in Users & Teams settings."
             genericMissingScope: "Your access key does not allow this action. Please generate a new access key by running `hs auth personalaccesskey`."
+        overrideErrors:
+          platformVersionErrors:
+            header: "{{#bold}}[ERROR]{{/bold}} Platform version update required"
```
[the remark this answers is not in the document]
kemmerle: Fixed in https://github.com/HubSpot/hubspot-cli/pull/1016/commits/10533707319cdd36640c67207b0af901b2a8c93d. Used the `logger.error()` function to display the `[ERROR]` tag in red 👍

packages/cli/lib/errorHandlers/apiErrors.js
kemmerle: Purposely chose this pattern, so we can reuse `overrideErrors` later, when we need to override normal error handlers to display custom errors for specific use cases. I'm planning to migrate missing scopes errors in projects to this pattern, and I think the sandboxes team has a few use cases, as well 👍

packages/cli/lib/errorHandlers/overrrideErrors.js
kemmerle: Great idea! Addressed in https://github.com/HubSpot/hubspot-cli/pull/1016/commits/93170ea1751b72db2582750d67c21b76b20d4766

PR #1017, opened 2024-03-13
Press q to cancel for hs preview

jsines:
## Description and Context
Adds better exit handling for theme preview dev server


camden11 (approved):
LGTM

PR #1018, opened 2024-03-13
Use local-dev-lib for Github utils

camden11:
## Description and Context
This switches to local-dev-lib for the github utils, which is the last major dependency we need to switch!

## Testing
I've tested this pretty thoroughly, but for good measure, these are the affected commands
* `hs create api-sample`
* `hs create app`
* `hs create react-app`
* `hs create vue-app`
* `hs create webpack-serverless`
* `hs create website-theme`
* `hs project create`
* `hs project add`

## TODO
I thought this was going to be the last PR, but there are still a few stray `cli-lib` deps out there so there will be at least one more

## Who to Notify
@brandenrodgers @kemmerle


kemmerle (approved):
I love that we've streamlined the helper functions, particularly `cloneGithubRepo`. I also tested out all commands, and everything is working as expected. Nice!

PR #1021, opened 2024-03-20
Prevent same dev server from being run multiple times during hs project dev

camden11:
## Description and Context
`hs project dev` is set up to iterate through multiple different servers that handle different component types, but wasn't handling the case of two different component types using the same server interface. This would cause it to call `setup` `start` etc on the UIE dev server multiple times in projects that contain both a private app and public app. This adds a check to make sure each server interface can only be iterated over once.

## Who to Notify


brandenrodgers (approved):
Good idea 👍

camden11:
Closing this because
* a) we don't need to account for the scenario of a single server supporting multiple component types at the same time just yet 
* b) this doesn't actually fix the problem - it just skips over all but the first component type

PR #1022, opened 2024-03-20
Log error instance correctly in project watch command (platformVersion warning fix)

kemmerle:
## Description and Context
While testing out the new platformVersion warnings/errors in the CLI, Roger discussed that we were not displaying the correct error in the `hs project watch` command in two instances:

1) Running the `watch` command in projects with an empty platform version.
2) Running the `watch` command in projects with an unsupported platform version (ie. '2025.7').

## Screenshots

Before:

Screenshot 2024-03-20 at 12 59 00 PM

After:

Screenshot 2024-03-20 at 12 59 38 PM

## TODO

- [x] Address any feedback

## Who to Notify

@brandenrodgers @camden11


camden11 (approved):
Could be worth trying to get the error message to display within spinnies like the previous set up, but maybe not worth the effort for an old command like watch

PR #1023, opened 2024-03-21
Clean up error handling

camden11:
## Description and Context
Now that `cli-lib` is (soon to be) out of this repo, we can get rid of all the CLI specific error handlers and try to reduce repetition of error handling code as much as possible. This PR takes a few steps towards that, though there may be more to good.
* Remove `isSpecifiedError` and use the local-dev-lib version
* Remove `isSpecifiedHubSpotAuthError` and use the local-dev-lib version
* Remove `isMissingScopeError` and use the local-dev-lib version
* Remove `isSystemError` and use the local-dev-lib version
* Remove `isFatalError`
* Remove `parseValidationErrors` and use the local-dev-lib version
* Remove `logApiStatusCodeError` since we can now safely use `getAxiosErrorWithContext` for all API errors
* Remove remaining reference to `StatusCodeError`s
* Remove `logFileSystemErrorInstance` logic in favor of `getFileSystemError` from local-dev-lib

## Testing
This is going to be a tricky one to test - it touches a lot of different commands and a lot of errors can be tricky to reproduce. Not sure what the best move is there, but whenever this goes out it we should give it plenty of time in beta

## TODO
* Merge [this PR ](https://github.com/HubSpot/hubspot-local-dev-lib/pull/120)and release LDL

## Who to Notify
@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -62,10 +62,10 @@ const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
 const {
   logApiErrorInstance,
   ApiErrorContext,
-  isSpecifiedError, // Migrate isSpecifiedError to local-dev-lib version only after uploadProject is migrated to local-dev-lib
 } = require('../../lib/errorHandlers/apiErrors');
 const {
   isMissingScopeError,
+  isSpecifiedError,
 } = require('@hubspot/local-dev-lib/errors/apiErrors');
 const { logErrorInstance } = require('../../lib/errorHandlers/standardErrors');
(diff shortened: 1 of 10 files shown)


kemmerle (approved):
@camden11, I tested out each of the error types that you replaced. With some, I was able to throw errors with my HubSpot setup (ie. the `missingScopes` error only required that I generate a PAK with few scopes). For others like system errors, I just threw errors in the code itself. I agree that we should let this sit on a beta for a while, since it's so hard to reproduce every scenario. Could we cut that beta on/after April 1, after we cut the minor release for the platform version work?

PR #1025, opened 2024-03-26
Developer test accounts support in `hs project dev`

adamawang:
## Description and Context

This PR adds developer test account support and creation within the `hs project dev` CLI command. 

If your default account is an app developer account ([supporting PR](https://github.com/HubSpot/hubspot-local-dev-lib/pull/118)), `hs project dev` will list all the developer test accounts associated with the app developer account. We perform some usage validation to enabled/disable the create option as shown in the screenshots below. Newly created developer test accounts will return a pre-generated PAK that we automatically save into the CLI config. 

The behavior of this flow is more or less matching the sandboxes flow with development sandboxes so it will be familiar to the end users who've used those before. 

@markhazlewood to confirm on some of the copy here, I pulled most of it out from the Figma you created a little while back. Also re: [standardized account names](https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/512), I added in `[app developer account]` as a label (screenshots below), but let me know if we'd want changes to these

I've also updated `hs accounts list` to show app developer accounts labeled as such, and have developer test accounts show as child accounts. 

Related PR in LDL: https://github.com/HubSpot/hubspot-local-dev-lib/pull/121

## Screenshots

Select developer test accounts belonging to the default account (app developer account)
Screenshot 2024-03-26 at 11 22 45

Disabled create option if user is at the limit of test portals (10)
Screenshot 2024-03-26 at 11 26 22

Ability to create new dev test accounts and go directly to deploying a project
Screenshot 2024-03-26 at 11 23 11

`hs accounts list` updates where app developer accounts are labeled, and dev test accounts are child portals
Screenshot 2024-03-26 at 12 12 01

If PAK does not have dev test account scopes, we'll see this error when attempting to create a new dev test account:
Screenshot 2024-03-27 at 15 45 25

If PAK has an error for both app developer or dev test account, auth errors are thrown: 
Screenshot 2024-03-27 at 15 47 49

## TODO

Next steps after this would be to link this up to the init flow where the CLI detects whether or not its a private vs public app. 

Once determined, if a user is in a production account (default account) and tries to run `hs project dev` in a public app project, we'd need to direct them to choose an app developer account as the default. 

Also as a follow up PR, I'll be making updates to `hs project clean` to handle cleaning up non-existing developer test accounts

## Who to Notify

@shannon-lichtenwalter


packages/cli/commands/project/dev.js
```diff
@@ -215,6 +232,62 @@ exports.handler = async options => {
     }
   }
 
+  if (createNewDeveloperTestAccount) {
+    let currentPortalCount = 0;
+    let maxTestPortals = 10;
```
adamawang, line 237: Default here should always be 10, but this will get overridden with whatever gets returned from BE which gets pulled from a config value

packages/cli/lang/en.lyaml
```diff
@@ -1065,10 +1066,12 @@ en:
       prompts:
         projectDevTargetAccountPrompt:
           createNewSandboxOption: "<Test on a new development sandbox>"
+          createNewDeveloperTestAccountOption: "<Test on a new developer test account>"
           chooseDefaultAccountOption: "<{{#bold}}\U00002757{{/bold}} Test on this production account {{#bold}}\U00002757{{/bold}}>"
-          promptMessage: "[--account] Choose a sandbox under {{ accountIdentifier }} to test with:"
+          promptMessage: "[--account] Choose a {{ accountType }} under {{ accountIdentifier }} to test with:"
           sandboxLimit: "Your account reached the limit of {{ limit }} development sandboxes"
           sandboxLimitWithSuggestion: "Your account reached the limit of {{ limit }} development sandboxes. Run {{ authCommand }} to add an existing one to the config."
+          developerTestAccountLimit: "Your account reached the limit of {{ limit }} developer test accounts."
```
adamawang, line 1074: @markhazlewood let me know if you'd prefer adding the extra blurb `Run hs auth to add an existing one to the config`

packages/cli/lib/developerTestAccountCreate.js
```diff
@@ -0,0 +1,194 @@
...
+      logger.info(
+        i18n(`${i18nKey}.create.failure.scopes.instructions`, {
+          accountName: uiAccountDescription(accountId),
+          url,
+        })
+      );
+    } else if (
+      isSpecifiedError(err, {
+        statusCode: 400,
+        errorType: 'TEST_PORTAL_LIMIT_REACHED',
```
adamawang, line 155: Dev test accounts BE returns a slightly different error format which doesnt have a category/subcategory, so I had to update the isSpecifiedError def in LDL to support this

packages/cli/lib/prompts/developerTestAccountNamePrompt.js
```diff
@@ -0,0 +1,29 @@
...
+        if (typeof val !== 'string') {
+          return i18n(`${i18nKey}.name.errors.invalidName`);
+        } else if (!val.length) {
+          return i18n(`${i18nKey}.name.errors.nameRequired`);
+        }
+        return accountNameExistsInConfig(val)
+          ? i18n(`${i18nKey}.name.errors.accountNameExists`, { name: val })
+          : true;
+      },
+      default: `Developer test account ${currentPortalCount + 1}`,
```
adamawang, line 22: This naming convention matches what we do on the dev test accounts UI

packages/cli/lib/developerTestAccountCreate.js
```diff
@@ -0,0 +1,194 @@
...
+    updatedConfig = await updateConfigWithAccessToken(
+      token,
+      personalAccessKey,
+      env
+    );
+  } catch (e) {
+    logErrorInstance(e);
+  }
+  if (!updatedConfig) {
+    throw new Error('Failed to update config with personal access key.');
```
camden11, line 56: Might be worth trying to incorporate this into the error logged on line 53. Looks like the only place `saveDevTestAccountToConfig` this error is being caught and logged

packages/cli/lib/developerTestAccountCreate.js
```diff
@@ -0,0 +1,194 @@
...
+  try {
+    // Response contains PAK, save to config here
+    devTestAcctConfigName = await saveDevTestAccountToConfig(
+      env,
+      result,
+      force
+    );
+  } catch (err) {
+    logErrorInstance(err);
+    throw err;
```
camden11, line 183: Will re-throwing these errors cause them to be logged twice?

camden11 (commented):
Need to spend some time testing this out but looking good! Left some pretty minor comments

packages/cli/lib/developerTestAccountCreate.js
adamawang: yeah probably very rare for that to fail, I'll just move the error up and get rid of this undefined block

packages/cli/lib/developerTestAccountCreate.js
adamawang: Yeah good catch, removing the logging in `saveDevTestAccountToConfig`

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -20,66 +28,117 @@ const mapSandboxAccount = accountConfig => ({
...
-    ...sandboxAccounts
-      .filter(a => a.accountType === HUBSPOT_ACCOUNT_TYPES.STANDARD_SANDBOX)
-      .map(mapSandboxAccount),
-    {
-      name: i18n(`${i18nKey}.createNewSandboxOption`),
-      value: {
-        targetAccountId: null,
-        createNewSandbox: true,
+    choices = [
+      ...devTestAccounts.map(mapSandboxAccount),
```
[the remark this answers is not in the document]
adamawang: yup I like that, updated!

packages/cli/lib/prompts/developerTestAccountNamePrompt.js
```diff
@@ -0,0 +1,29 @@
+const { promptUser } = require('./promptUtils');
+const { i18n } = require('../lang');
+const { accountNameExistsInConfig } = require('@hubspot/local-dev-lib/config');
+
+const i18nKey = 'cli.lib.prompts.developerTestAccountPrompt';
+
+const developerTestAccountNamePrompt = currentPortalCount => {
```
brandenrodgers, line 7: Not a blocker, but I wonder if we could DRY things up by making the `accountNamePrompt` abstract enough to support this and also the sandboxNamePrompt use cases.

packages/cli/lib/developerTestAccountCreate.js
```diff
@@ -0,0 +1,186 @@
...
+  getAccessToken,
+  updateConfigWithAccessToken,
+} = require('@hubspot/local-dev-lib/personalAccessKey');
+const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
+const { uiAccountDescription } = require('./ui');
+const {
+  personalAccessKeyPrompt,
+} = require('./prompts/personalAccessKeyPrompt');
+const { enterAccountNamePrompt } = require('./prompts/enterAccountNamePrompt');
+
```
brandenrodgers, line 31: Also not a blocker, but looks like there's a lot of very similar logic between this file and lib/sandboxCreate.js . Maybe there's a way to consolidate somehow?

adamawang: yeah I had considered this but felt it might have been too many changes at once, I'll prob elect to break it out into a separate PR and consolidate a bunch of things

brandenrodgers: Yeah that makes sense 👌

packages/cli/lib/prompts/developerTestAccountNamePrompt.js
adamawang: Will be taken care of here: https://git.hubteam.com/HubSpot/sandboxes-ui/issues/2309

packages/cli/commands/accounts/list.js
```diff
@@ -53,21 +56,28 @@ const sortAndMapPortals = portals => {
...
-        if (set.length > 1) {
+        if (hasParentPortal && set.length > 1) {
           name = `↳ ${name}`;
         }
       } else if (isDeveloperTestAccount(portal)) {
         name = `${portal.name} [${
           HUBSPOT_ACCOUNT_TYPE_STRINGS[HUBSPOT_ACCOUNT_TYPES.DEVELOPER_TEST]
         }]`;
-        if (set.length > 1) {
+        if (hasParentPortal && set.length > 1) {
```
adamawang, line 74: If multiple portals referenced the same parent portal ID but that parent portal didn't exist in the config, it would display incorrect name indentations

Before:
Screenshot 2024-04-03 at 14 36 07

After:
Screenshot 2024-04-03 at 14 35 59

camden11 (approved):
This is looking good to me! Wondering now about how to best integrate it with the rest of the local dev work.

I think an important step that needs to happen _before_ the account selection flow starts is a check to see whether your project contains a private or public app. Pretty sure that's what we want to determine whether or not the prompt to create a developer test account shows up (for example, right now, it shows up even when you run dev in a private app project). I'm going to work on setting up logic for that next.

There's also the matter of what happens when you run `dev` in a public app project but with a standard production account as your default account. Could make sense to handle that case separately once we have something that can actually determine the type of project.

Since this doesn't affect local dev for private apps, I think it'd be fine to merge now but could also wait until some of my work is done.

adamawang:
Just chatted with Camden and we're cool with merging it in as is and we can work on the app check separately 🚢

PR #1029, opened 2024-04-03
Set custom request header on all axios requests for easier debugging

brandenrodgers:
## Description and Context

Depends on: https://github.com/HubSpot/hubspot-local-dev-lib/pull/123

Sets a custom request header that will apply to all requests. This makes it easier for us to debug issues in our backend systems because it lets us link incoming requests to certain CLI versions.

I'm open to suggestions for better naming on the header.


packages/cli/bin/cli.js
```diff
@@ -110,9 +111,13 @@ const performChecks = argv => {
   }
 };
 
+const setRequestHeaders = () => {
+  addRequestHeader('X-Parent-User-Agent', `HubSpot CLI/${pkg.version}`);
```
m-roll, line 115: Is this a standard header?

Can this be `HubSpot CLI-lib/{version}`?

brandenrodgers:
Relevant supporting PR: https://github.com/HubSpot/hubspot-local-dev-lib/pull/123

packages/cli/bin/cli.js
brandenrodgers: Just to close the loop here, I updated this to use the existing `User-Agent` header (see my latest comment)

kemmerle (approved):
Tested and reviewed the code, everything is working great for me 👍

PR #1030, opened 2024-04-08
Public App Init Flow First Pass + Clean up dev command

camden11:
## Description and Context
This PR does the following

#### Sets up _most_ of the public app local dev init flow
* Determine what components are in the project before determining what kind of account the user should use
* Error if the project contains both public apps and private apps (not supported)
* Error if the user is not using a supported account type - for example trying to develop a private app with a developer test account
* Set the target ID for project upload to the parent account if developing a public app

#### Cleans up code for dev command
`dev.js` was over 400 lines long and just its imports previously took up my entire screen. While I was going through and trying to figure out exactly what was going on, I tried to make the code more readable by extracting each discrete step of the dev flow into its own function. These functions now live in `lib/localDev.js`. I also tried to reconcile some of the shared logic between sandboxes and test accounts and clean things up there.

## Testing
From what I can tell, this PR doesn't break the existing private app flow, and gets the public app flow to the point where things are passed off to the UIE dev server (which errors). A lot has changed here though, and there are a lot of different permutations, so should try to test all of them thoroughly.

Off the top of my head, here are some of them
* Developing a private app with a production account
* Developing a private app with an existing sandbox
* Developing a private app with a new sandbox
* Developing a private app with a developer account (should error in a clear way)
* Developing a private app with a developer test account (should error in a clear way)
* Developing a public app with a non-developer account (should error in a clear way)
* Developing a public app with a developer account and creating a new test account
* Developing a public app with developer account and using an existing test account
* Developing a public app with a developer test account

## TODO
This PR includes most of the updates we'll need for the init flow, but still need to add the check to see if the public app is installed in the test account. Going to do this in a later PR.

## Who to Notify
@brandenrodgers @kemmerle @adamawang

### DIFF
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -492,6 +492,8 @@ en:
               noProjectConfig: "No project detected. Please run this command again from a project directory."
               projectLockedError: "Your project is locked. This may mean that another user is running the {{#bold}}`hs project watch`{{/bold}} command for this project. If this is you, unlock the project in Projects UI."
               invalidProjectComponents: "Projects cannot contain both private and public apps. Move your apps to separate projects before attempting local development."
+              standardAccountNotSupported: "This project contains a public app. Local development of public apps is only supported on developer accounts and developer test accounts. Change your default account using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
+              appDeveloperAccountNotSupported: "Your current default account is a Developer Account. To develop this project locally, switch to a standard account using using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
             examples:
               default: "Start local dev for the current project"
           create:
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -946,10 +946,10 @@ en:
           declineDefaultAccountExplanation: "To develop on a different account, run {{ useCommand }} to change your default account, then re-run {{ devCommand }}."
         checkCorrectParentAccountType:
           standardAccountNotSupported: "This project contains a public app. Local development of public apps is only supported on developer accounts and developer test accounts. Change your default account using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
-          appDeveloperAccountNotSupported: "Your current default account is a Developer Account. To develop this project locally, switch to a standard account using using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
         suggestRecommendedNestedAccount:
           nonSandboxWarning: "Testing in a sandbox is strongly recommended. To switch the target account, select an option below or run {{#bold}}`hs accounts use`{{/bold}} before running the command again."
-          nonDeveloperTestAccountWarning: "Local development of public apps is only supported in {{#bold}}developer test accounts{{/bold}}."
+          publicAppNonDeveloperTestAccountWarning: "Local development of public apps is only supported in {{#bold}}developer test accounts{{/bold}}."
+          privateAppNonDeveloperTestAccountWarning: "Local development of private apps is only supported in {{#bold}}developer test accounts{{/bold}}"
         createNewProjectForLocalDev:
           projectMustExistExplanation: "The project {{ projectName }} does not exist in the target account {{ accountIdentifier}}. This command requires the project to exist in the target account."
           createProject: "Create new project {{ projectName}} in {{#bold}}[{{ accountIdentifier }}]{{/bold}}?"
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -491,6 +491,7 @@ en:
             errors:
               noProjectConfig: "No project detected. Please run this command again from a project directory."
               projectLockedError: "Your project is locked. This may mean that another user is running the {{#bold}}`hs project watch`{{/bold}} command for this project. If this is you, unlock the project in Projects UI."
+              invalidProjectComponents: "Projects cannot contain both private and public apps. Move your apps to separate projects before attempting local development."
             examples:
               default: "Start local dev for the current project"
           create:
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -475,25 +475,10 @@ en:
             describe: "Start local dev for the current project"
             logs:
               betaMessage: "HubSpot projects local development"
-              nonSandboxWarning: "Testing in a sandbox is strongly recommended. To switch the target account, select an option below or run {{#bold}}`hs accounts use`{{/bold}} before running the command again."
-              nonDeveloperTestAccountWarning: "Local development of public apps is only supported in {{#bold}}developer test accounts{{/bold}}."
               placeholderAccountSelection: "Using default account as target account (for now)"
-              projectMustExistExplanation: "The project {{ projectName }} does not exist in the target account {{ accountIdentifier}}. This command requires the project to exist in the target account."
-              choseNotToCreateProject: "Exiting because this command requires the project to exist in the target account."
-              initialUploadMessage: "HubSpot Local Dev Server Startup"
-              declineDefaultAccountExplanation: "To develop on a different account, run {{ useCommand }} to change your default account, then re-run {{ devCommand }}."
-            status:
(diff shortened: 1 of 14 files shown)


packages/cli/lang/en.lyaml
```diff
@@ -491,6 +491,9 @@ en:
             errors:
               noProjectConfig: "No project detected. Please run this command again from a project directory."
               projectLockedError: "Your project is locked. This may mean that another user is running the {{#bold}}`hs project watch`{{/bold}} command for this project. If this is you, unlock the project in Projects UI."
+              invalidProjectComponents: "Projects cannot contain both private and public apps. Move your apps to separate projects before attempting local development."
+              standardAccountNotSupported: "This project contains a public app. Local development of public apps is only supported on developer accounts and developer test accounts. Change your default account using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
+              appDeveloperAccountNotSupported: "Your current default account is a Developer Account. To develop this project locally, switch to a standard account using using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
```
camden11, line 496: Should probably run this copy by someone, this can all be considered placeholder

packages/cli/lib/LocalDevManager.js
```diff
@@ -47,6 +46,7 @@ class LocalDevManager {
     this.isGithubLinked = options.isGithubLinked;
     this.watcher = null;
     this.uploadWarnings = {};
+    this.runnableComponents = this.getRunnableComponents(options.components);
```
camden11, line 49: We now find project components outside the `LocalDevManager` so we can determine what apps a project contains and therefore what account to recommend the user use. Components are passed into the constructor, so I moved the logic to find runnable components from setup to the constructor

packages/cli/lib/accountTypes.js
```diff
@@ -0,0 +1,34 @@
...
+  isAccountType(config, HUBSPOT_ACCOUNT_TYPES.STANDARD_SANDBOX);
+
+const isDevelopmentSandbox = config =>
+  isAccountType(config, HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX);
+
+const isDeveloperTestAccount = config =>
+  isAccountType(config, HUBSPOT_ACCOUNT_TYPES.DEVELOPER_TEST);
+
+const isAppDeveloperAccount = config =>
+  isAccountType(config, HUBSPOT_ACCOUNT_TYPES.APP_DEVELOPER);
```
camden11, line 25: Moved all these to one place since `isStandardAccount` didn't really fit anywhere else

packages/cli/lib/localDev.js
```diff
@@ -0,0 +1,376 @@
...
+
+module.exports = {
+  confirmDefaultAccountIsTarget,
+  checkCorrectParentAccountType,
+  suggestRecommendedNestedAccount,
+  createSandboxForLocalDev,
+  createDeveloperTestAccountForLocalDev,
+  createNewProjectForLocalDev,
+  createInitialBuildForNewProject,
+};
```
camden11, line 376: The logic in this file is pretty close to 1:1 with the old version of `dev.js`. With the exception of new logic for public apps in the account selection functions, nothing was changed, just moved

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -12,133 +12,145 @@ const { logger } = require('@hubspot/local-dev-lib/logger');
...
-      },
-      {
-        name: i18n(`${i18nKey}.chooseDefaultAccountOption`),
-        value: {
-          targetAccountId: defaultAccountId,
-          createNewSandbox: false,
-        },
-      },
-    ];
+const selectDeveloperTestTargetAccountPrompt = async (
```
camden11, line 92: Found the old setup a little confusing so I separated the prompts for selecting sandboxes and selecting test accounts

packages/cli/lib/accountTypes.js
brandenrodgers: Seems like all of this could maybe even live in local-dev-lib, but I like this spot for now 👌

packages/cli/lib/localDev.js
```diff
@@ -0,0 +1,376 @@
...
+  PROJECT_ERROR_TYPES,
+  PROJECT_BUILD_TEXT,
+  PROJECT_DEPLOY_TEXT,
+} = require('./constants');
+const {
+  logApiErrorInstance,
+  ApiErrorContext,
+} = require('./errorHandlers/apiErrors');
+
+// TODO: Maybe give these to their own lang key
```
brandenrodgers, line 56: +1 to this todo

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -12,133 +12,145 @@ const { logger } = require('@hubspot/local-dev-lib/logger');
...
-    const sandboxAccounts = accounts
-      .reverse()
-      .filter(
-        config =>
-          isSandbox(config) && config.parentAccountId === defaultAccountId
-      );
-    let disabledMessage = false;
+  return selectTargetAccountPrompt(
+    defaultAccountId,
+    'sandbox account',
```
brandenrodgers, line 87: Does `sandbox account` not exist in `HUBSPOT_ACCOUNT_TYPE_STRINGS `?

brandenrodgers (approved):
I didn't do a ton of testing of this on public apps, but I did test on my private app project and this works great. I left a few comments, but none are blockers

packages/cli/lib/accountTypes.js
adamawang: Nice, this looks much nicer!

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -12,133 +12,145 @@ const { logger } = require('@hubspot/local-dev-lib/logger');
 const {
   fetchDeveloperTestAccounts,
 } = require('@hubspot/local-dev-lib/developerTestAccounts');
-const {
-  isAppDeveloperAccount,
-  isDeveloperTestAccount,
-} = require('../developerTestAccounts');
+const { isDeveloperTestAccount } = require('../accountTypes');
```
adamawang, line 15: Can combine with import on line 4

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
adamawang: Yeah it unfortunately doesn't, the usage here needs to be the generic sandbox type but `HUBSPOT_ACCOUNT_TYPE_STRINGS` only has standard/development sandbox. IIRC this is the only spot I used this at and didnt know if it'd be worth creating a const for

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
adamawang: yeah definitely good call 👍

adamawang:
```
Developing a private app with a developer account (should error in a clear way)
Developing a private app with a developer test account (should error in a clear way)
```
I think these are actually supposed to be supported at this time? IIRC Sejal has a small group of customers using this flow and we unlocked private apps on dev test accounts early on so they could start

camden11:
> ```
> Developing a private app with a developer account (should error in a clear way)
> Developing a private app with a developer test account (should error in a clear way)
> ```
> 
> I think these are actually supposed to be supported at this time? IIRC Sejal has a small group of customers using this flow and we unlocked private apps on dev test accounts early on so they could start

Oh interesting, didn't know that. Will look more into it then. I could see how developer test accounts could work, but don't developer accounts not even have contacts?

adamawang:
Yeah I believe it was for milestone 0 back in march where they're onboarding public app developers onto developer test accounts first and creating private apps to get used to the flow before public apps are available. 

> I could see how developer test accounts could work, but don't developer accounts not even have contacts?
Yep the developer accounts flow in `hs project dev` prompted you to either choose a dev test account or create a new one if a developer account was default, otherwise it goes straight into the local dev flow if its a dev test account default

camden11:
Gotcha alright. Well glad you caught that! I'll have to update that logic

camden11:
Updated to support private app development on developer test accounts!

adamawang (approved):
awesome, confirming the private app flow works as expected for app developer and dev test accounts 👍  I haven't done much testing on public apps but I trust you got it based on the listed scenarios in the original description. 

Might want to adjust the `Testing` section to have the two items you changed not be listed as `(should error in a clear way)` and combine that import in `projectDevTargetAccountPrompt`, rest LGTM!

PR #1032, opened 2024-04-08
Update product updates URL

adamawang:
## Description and Context

Issue: https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/532

The existing product updates URL was updated to a new format following a UI redesign that went live in December '23. This PR updates all usages of the link to the new format. 

## TODO

none


brandenrodgers (approved):
Thanks for fixing this

PR #1034, opened 2024-04-10
Fixes after switching default branch from master to main

brandenrodgers:
## Description and Context

Just making a few updates in spots where we were directly referencing the master branch name

For anyone who works on the CLI, you can follow these steps to help resolve any default branch issues:
image

## Who to Notify

cc/ @adamawang @jsines just to let you know that the default branch has been updated


camden11 (approved):
Looks good!

PR #1035, opened 2024-04-11
Theme project dev

jsines:
## Description and Context

Adds projects support for theme previewing

Some updates to ContentFileMapper and DesignManagerApi now allow `.hubl.html` and for us to upload to the @projects directory within the preview buffer to mock the theme that's normally built with projects.

So the theme component still gets built with CmsThemePipeline normally with `hs project upload`, but for the dev server we route it through ContentFileMapper instead to avoid the build pipeline

The difference of behavior is enabled by simply running it within a projects directory; outside a project directory you'll be prompted for the normal `src` and `dest` for a traditional theme. Inside a projects directory you'll be prompted to choose a theme component in the project, or if you run it within the theme component itself you're not prompted at all and it just runs that one. 

Down the line it would be nice to further integrate this with `hs project dev`, but untangling some of the server setup / port selection / allowing multiple to run at the same time is a slightly bigger chunk of work and supporting it through this avenue I think is currently sufficient


packages/cli/lib/projectStructure.js
```diff
@@ -102,13 +104,14 @@ async function findProjectComponents(projectSourceDir) {
...
+        const isHublTheme = base === CONFIG_FILES[COMPONENT_TYPES.hublTheme];
 
         components.push({
           type: getTypeFromConfigFile(base),
           config: parsedAppConfig,
-          runnable: !isLegacy,
+          runnable: !isLegacy && !isHublTheme,
           path: dir,
         });
       }
```
jsines, line 117: @brandenrodgers @camden11 to confirm I'm not making a mistake in adding like this. Running the server through `hs project dev` is going to be a bit bigger chunk of work I'd like to avoid for the moment but I'm using this util to fetch components and then filter to just the theme components to select from - marking as not runnable here to prevent it from getting picked up by `hs project dev`, and then we have a check for the name property that _seems_ unnecessary & is filtering `theme.json`s because they don't have a `name` field

packages/cli/lib/prompts/previewPrompt.js
```diff
@@ -34,6 +34,24 @@ const previewPrompt = (promptOptions = {}) => {
   ]);
 };
 
+const previewProjectPrompt = async themeComponents => {
```
brandenrodgers, line 37: Would you mind updating this file's name to be a little more specific? Maybe something along the lines of `themePreviewPrompt`?

packages/cli/lib/projectStructure.js
brandenrodgers: This seems like it should be safe! tbh I can't remember why we added in the `name` check, but it's probably not necessary. There's a bunch of validation that we could be doing here to make sure the JSON files are formatted correctly and contain the required fields, but that work can all be done separately from this. You can keep what you have here 👍 

Also, I'm noticing the `parsedAppConfig` doesn't really make sense as a variable name anymore since now it can include themes. Feel free to update it to be something more generic if you'd like to.

TanyaScales (approved):

j-malt (approved):

PR #1037, opened 2024-04-16
Pass correct account ids in project dev init flow

brandenrodgers:
## Description and Context

We had an issue where the CLI was passing the incorrect target account ID to the local dev servers when running public apps. The problem is that we need to keep the `targetAccountId` pointed to the parent account for all of the validation checks because the project lives in the parent (developer) account, but the testing that we do actually happens in the developer test accounts.

I split the account variable in two so we have `targetProjectAccountId` and `targetTestingAccountId`


kemmerle (approved):
Reviewed and tested the changes, and things are working great for me

PR #1038, opened 2024-04-17
Remove unnecessary lang file key

brandenrodgers:
## Description and Context

Removing the top level `cli` key in our lang file because we don't need it. This just cuts down on a bit of the nesting and makes the file slightly easier to parse. It should have no effect on the CLI's logged output.


camden11 (approved):
Oooh this is a lot, but I went all the way through and didn't notice any typos. Should be good to merge once the conflicts are fixed

PR #1041, opened 2024-04-23
`hs project dev` supports developer test accounts not in config 

adamawang:
## Description and Context

Issue: https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/546

After feedback from last weeks bug bash, the expected behavior for `hs project dev` with an app developer account as default is to list all available developer test accounts including those that do not exist in the CLI config. 

In this PR, we fetch all dev test accounts for an app developer account and mix them into the choices available when prompted to select an account. We track that behind the scenes and if a user is to select an account not yet in the config, we prompt them to continue and have them manually generate/save a PAK. 

Confirmation message approved by Mark H: https://hubspot.slack.com/archives/C066MUYG1UH/p1713888899890209?thread_ts=1713886322.515959&cid=C066MUYG1UH

## Screenshots

Declining the confirmation:
Screenshot 2024-04-23 at 13 28 28

Full flow:
Screenshot 2024-04-23 at 13 29 10

## TODO

It does look a bit long on screen but I'm not sure if theres any way to shorten it. 

## Who to Notify

@shannon-lichtenwalter

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -128,6 +128,8 @@ exports.handler = async options => {
 
     targetProjectAccountId = hasPublicApps ? parentAccountId : targetAccountId;
     targetTestingAccountId = targetAccountId;
+
+    // Only used for developer test accounts that are not yet in the config
     if (notInConfigAccount) {
       const useExistingDevTestAcct = await confirmUseExistingDeveloperTestAccountPrompt(
         notInConfigAccount
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -22,7 +22,7 @@ const {
   validateProjectConfig,
 } = require('../../lib/projects');
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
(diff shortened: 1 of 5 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -119,6 +129,37 @@ exports.handler = async options => {
...
+      const devTestAcctConfigName = await saveDevTestAccountToConfig(
+        env,
+        notInConfigAccount
+      );
+      logger.success(
+        i18n(`cli.lib.developerTestAccount.create.success.configFileUpdated`, {
+          accountName: devTestAcctConfigName,
+          authType: PERSONAL_ACCESS_KEY_AUTH_METHOD.name,
+        })
+      );
```
camden11, line 160: Trying to keep things as concise as possible within this file - could you move this logic to `lib/localDev`?

camden11 (approved):
Just tested and looks good otherwise

packages/cli/commands/project/dev.js
adamawang: yep was just thinking about this, will push that up in a bit

PR #1042, opened 2024-04-23
Fix issues with account flag + add clarification to new project warning

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/549
This does a few small things
* Fixes the `account` flag on both private and public app projects
* Adds an error if running a public app project with an account other than a developer test account
* Adds some clarification to the warning message when uploading a public app project - previously it said it didn't exist in the target account. For public apps, it needs to exist in the _parent_ of the target account.

## Screenshots
`--account` flag error
Screenshot 2024-04-23 at 4 48 38 PM

Upload warning
Screenshot 2024-04-23 at 4 48 22 PM

## Who to Notify
@brandenrodgers @kemmerle

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -42,6 +42,7 @@ const {
   confirmDefaultAccountIsTarget,
   suggestRecommendedNestedAccount,
   checkIfAppDeveloperAccount,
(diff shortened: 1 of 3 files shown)


brandenrodgers (approved):
lgtm!

PR #1043, opened 2024-04-24
Bumping the local-dev-lib dep version to 1.0.1

brandenrodgers:
## Description and Context

Bumping the CLI's dep on local-dev-lib to version 1.0.1. Switching the pattern to specify exact dep version so that we can have more control over it.


kemmerle (approved):
:shipit:

PR #1049, opened 2024-04-26
Fixing account id references for public app local dev

brandenrodgers:
## Description and Context

This fixes a few different small edge-case bugs, all related to local dev for public apps:
- We were showing the incorrect account id in the prompt when the user edits a configuration file while local dev is running
- We were showing the incorrect account to upload to when there is no deployed build for the project in the target account
  - This only shows when the project exists in the account, but there are no builds (repro by uploading a broken project)
- Fixing JS error when default account is a dev test account, and the parent dev account has not been configured in the CLI

## Screenshots

WIP copy for the message to tell the user that they need to configure their Developer Account:
image

## Who to Notify

cc/ @markhazlewood for input on copy

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -22,7 +22,11 @@ const {
   validateProjectConfig,
 } = require('../../lib/projects');
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
-const { uiBetaTag } = require('../../lib/ui');
+const {
+  uiAccountDescription,
+  uiBetaTag,
+  uiCommandReference,
+} = require('../../lib/ui');
 const SpinniesManager = require('../../lib/ui/SpinniesManager');
 const LocalDevManager = require('../../lib/LocalDevManager');
 const {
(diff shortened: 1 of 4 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -104,12 +108,30 @@ exports.handler = async options => {
...
+              `hs auth --account=${accountConfig.parentAccountId}`
+            ),
+          })
+        );
+        process.exit(EXIT_CODES.ERROR);
+      }
+      targetProjectAccountId = accountConfig.parentAccountId;
+    } else {
+      targetProjectAccountId = accountId;
+    }
```
camden11, line 134: I think this could be pulled into `lib/localDev`

camden11 (approved):
LGTM, one minor suggestion

packages/cli/commands/project/dev.js
brandenrodgers: @camden11 just to clarify, which part do you think should be pulled in? All this logic is really doing is checking to see if the parent account exists in the config or not.

brandenrodgers: I spent a little time trying to do this, but I couldn't get something I was happy with. There are so many args that need to be passed into the abstracted util that it ends up not feeling very valuable. I'm going to merge this as-is, but feel free to tweak this later on if you can think of a cleaner format for it!

packages/cli/lang/en.lyaml
```diff
@@ -479,6 +479,7 @@ en:
             errors:
               noProjectConfig: "No project detected. Please run this command again from a project directory."
               invalidProjectComponents: "Projects cannot contain both private and public apps. Move your apps to separate projects before attempting local development."
+              parentAccountNotConfigured: "To develop this project locally, run {{ authCommand }} to authenticate the App Developer Account {{ accountId}} associated with {{ accountIdentifier }}."
```
brandenrodgers, line 482: cc/ @adamawang for 👍 on the copy here

packages/cli/lang/en.lyaml
```diff
@@ -479,6 +479,7 @@ en:
             errors:
               noProjectConfig: "No project detected. Please run this command again from a project directory."
               invalidProjectComponents: "Projects cannot contain both private and public apps. Move your apps to separate projects before attempting local development."
+              parentAccountNotConfigured: "To develop this project locally, run {{ authCommand }} to authenticate the App Developer Account {{ accountId }} associated with {{ accountIdentifier }}."
```
adamawang, line 482: 💎 works for us!

PR #1050, opened 2024-04-26
Public app install prompt setup

camden11:
## Description and Context
This is kind of a weird WIP PR, but it won't affect the local dev flow at all and will allow @joe-yeager to start handling the dev server side of this.

What this does
* Passed a `setActiveApp` callback to the dev server. Once the `LocalDevManager` has the active app, it'll be able to issue the relevant prompts and warnings
* Slightly tweaks `ensureProjectExists` to return the project - noticed when working on this that we were making a redundant call to `fetchProject`

## TODO
Once the dev server is setting the active app, we can handle the rest of the flow in the CLI

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -166,13 +166,17 @@ exports.handler = async options => {
     }
   );
 
-  let deployedBuild = project.deployedBuild;
-  let isGithubLinked =
-    project.sourceIntegration && project.sourceIntegration.source === 'GITHUB';
+  let deployedBuild;
+  let isGithubLinked;
 
   SpinniesManager.init();
 
-  if (!projectExists) {
+  if (projectExists) {
+    deployedBuild = project.deployedBuild;
+    isGithubLinked =
+      project.sourceIntegration &&
+      project.sourceIntegration.source === 'GITHUB';
+  } else {
     await createNewProjectForLocalDev(
       projectConfig,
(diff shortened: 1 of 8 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -167,19 +166,9 @@ exports.handler = async options => {
...
-    const project = await fetchProject(
-      targetProjectAccountId,
-      projectConfig.name
-    );
-    deployedBuild = project.deployedBuild;
-    isGithubLinked =
-      project.sourceIntegration &&
-      project.sourceIntegration.source === 'GITHUB';
-  }
+  let deployedBuild = project.deployedBuild;
```
brandenrodgers, line 169: Is it possible for the project to be undefined here?

camden11: Ohhh good call, it is

packages/cli/lib/LocalDevManager.js
```diff
@@ -80,6 +81,12 @@ class LocalDevManager {
     return components.filter(component => component.runnable);
   }
 
+  async setActiveApp(app) {
+    this.activeApp = app;
+
+    // TODO: Show the install warnings the first time this gets set
```
joe-yeager, line 87: This this coming in a follow up PR?

joe-yeager (approved):
LGTM, just one question about the message being displayed

PR #1052, opened 2024-05-01
Feature: apps migration and cloning (Sister PR in local-dev-lib)

kemmerle:
## Description and Context
In this PR, I'm adding support for public apps migration and cloning. Because the necessary endpoints are still being worked on, I'm currently scaffolding out the commands using test data. I will update this PR, as endpoints become available. 

There is a sister [PR](https://github.com/HubSpot/hubspot-local-dev-lib/pull/130) in local-dev-lib. 

## Screenshots

`hs project create --migrateApp`:

Happy path: 
Screenshot 2024-05-09 at 4 03 36 PM

Sad path (No errors being returned yet):
Screenshot 2024-05-09 at 4 05 45 PM

## TESTING

The `hs project migrate-app` command is now ready for testing. You will need to go through the `yarn link` process to take advantage of the API endpoints in the local-dev-lib library (See PR linked above). Make sure that you switch to the correct branch before testing. 

Currently, the BE does not create and link an app, but we do create a new project in your developer account. I've also added downloading the project files to your local machine, so make sure to check that, as well.  

IMPORTANT CAVEAT: Currently, no errors are being returned, so we're just returning a generic error message. 

## TODO

- [x] Scaffold out the migration flow
- [x] Update endpoints as they become available
- [x] Update after feedback from @markhazlewood 
- [x] Address any feedback from FE/BE engineers
- [x] Test out different scenarios

## Who to Notify

@markhazlewood @brandenrodgers @camden11 @joe-yeager @mendel-at-hubspot


packages/cli/lang/en.lyaml
```diff
@@ -993,6 +997,13 @@ en:
         logFeedbackMessage:
           feedbackHeader: "We'd love to hear your feedback!"
           feedbackMessage: "How are you liking the new projects and developer tools? \n > Run `{{#yellow}}hs feedback{{/yellow}}` to let us know what you think!\n"
+      publicApps:
```
kemmerle, line 1000: @markhazlewood, if I get could your feedback on these error messages, I'd be very appreciative. Connected [Loom](https://www.loom.com/share/1476c63cc3814ce5bfe8330dda699e0d) for your convenience.

markhazlewood42:
@kemmerle Here's the error message I had proposed for both "no apps" and "no valid apps to migrate" cases. I don't think we have a doc/place to link to yet but we should in the future:

image

packages/cli/lang/en.lyaml
kemmerle: Canceling this request for comment. Going to rework the commands after the Projects UX sync today.

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
+const { downloadProject } = require('@hubspot/local-dev-lib/api/projects');
+const { extractZipArchive } = require('@hubspot/local-dev-lib/archive');
+const { getHubSpotWebsiteOrigin } = require('@hubspot/local-dev-lib/urls');
+const { fetchProject } = require('@hubspot/local-dev-lib/api/projects');
+
+const i18nKey = 'cli.commands.project.subcommands.migrateApp';
+
+exports.command = 'migrate-app';
+exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
```
brandenrodgers, line 41: Will we want to hide this initially?

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+    logger.error(
+      i18n(`${i18nKey}.errors.invalidAccountType`, {
+        accountName: accountConfig.name,
+        accountType: accountConfig.accountType,
+      })
+    );
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  let appId;
```
brandenrodgers, line 59: You could call `validateAppId` outside of the if blocks. Or you could shorten this to something along the lines of:
```js
  const appId =
    options.appId ||
    (await validateAppId(appId, accountId, accountConfig.name));
  await validateAppId(appId, accountId, accountConfig.name);
```

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  const { name, location } = await createProjectPrompt('', options, false);
+
+  const projectName = options.name || name;
+  const projectLocation = options.location || location;
+
+  let projectExists = false;
+  try {
+    await fetchProject(accountId, projectName);
```
brandenrodgers, line 75: Just to be safe, should we check to see that the response actually contains the project? I think this currently works because the request will 404 if the project doesn't exist, but doing something like `projectExists = !!project` is a bit more explicit.

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+    projectExists = true;
+  } catch (e) {
+    projectExists = false;
```
brandenrodgers, line 78: No need to set false here since it's already false

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  if (projectExists) {
```
brandenrodgers, line 81: If you put this in the try/catch block, then you wouldn't need to use the `projectExists` variable at all

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+    logger.error(
+      i18n(`${i18nKey}.errors.projectAlreadyExists`, {
+        projectName,
+      })
+    );
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  trackCommandUsage('migrate-app', {}, accountId);
```
brandenrodgers, line 90: I think this should go towards the top since the idea of this tracking is to track every time a user tries to migrate an app (successful or not)

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+  if (appId && projectName && projectLocation) {
```
brandenrodgers, line 91: What should happen if this IF block does not trigger?

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,190 @@
...
+  }
+};
+
+exports.builder = yargs => {
+  yargs.options({
+    name: {
+      describe: i18n(`${i18nKey}.options.name.describe`),
+      type: 'string',
+    },
+    location: {
```
brandenrodgers, line 171: nit (And something to run by Mark), but I'm partial to the `dest` option that the CMS commands are already using. It's shorted and slightly more informative IMO

packages/cli/lang/en.lyaml
```diff
@@ -496,6 +496,36 @@ en:
...
+              appConfig: "All supported app configuration will be moved to the {{#bold}}public-app.json{{/bold}} component definition file. Future updates to those features must be made through the project build and deploy pipeline, not the developer account UI."
+              buildAndDeploy: "This will create a new project with a single app component and immediately build and deploy it to your developer account (build #1)."
+              existingApps: "{{#bold}}This will not affect existing app users or installs.{{/bold}}"
+              copyApp: "We strongly recommend making a copy of your app to test this process in a development app before replacing production."
+            createAppPrompt: "Proceed with migrating this app to a project component?"
+            projectDetailsLink: "View project details in your developer account"
+            errors:
+              invalidAccountType: "Public apps must be migrated within an app developer account. Your current account {{#bold}}{{ accountName }}{{/bold}} is a {{ accountType }}.
+                \n- Run {{#bold}}hs accounts use{{/bold}} to switch your default account to an app developer account.
```
brandenrodgers, line 526: we have a `uiCommandReference` for lines like this. It'll help to keep every command reference standard

packages/cli/lang/en.lyaml
```diff
@@ -993,6 +1023,10 @@ en:
         logFeedbackMessage:
           feedbackHeader: "We'd love to hear your feedback!"
           feedbackMessage: "How are you liking the new projects and developer tools? \n > Run `{{#yellow}}hs feedback{{/yellow}}` to let us know what you think!\n"
+      publicApps:
+        errors:
+          noApps: "{{#bold}}No apps to convert{{/bold}}"
+          noAppsMessage: "The selected developer account {{#bold}}{{ accountName }}{{/bold}} doesn't have any apps that can be converted to the projects framework."
```
brandenrodgers, line 1029: Let's try to use the `uiAccountDescription` util as much as possible so that all of the account names in the logs are standardized

packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -48,11 +48,18 @@ const createTemplateOptions = async (templateSource, githubRef) => {
...
 
-const createProjectPrompt = async (githubRef, promptOptions = {}) => {
-  const projectTemplates = await createTemplateOptions(
-    promptOptions.templateSource,
-    githubRef
-  );
+const createProjectPrompt = async (
+  githubRef,
+  promptOptions = {},
+  templatesRequired = true
```
brandenrodgers, line 54: I'm a little confused by what this is doing. Can you explain it?

packages/cli/lib/publicApps.js
```diff
@@ -0,0 +1,60 @@
...
+  const { appId } = await selectPublicAppPrompt({
+    accountId,
+    accountName,
+    options,
+    migrateApp,
+  });
+  return appId;
+};
+
+const migratePublicApp = async (accountId, appId, name) => {
```
brandenrodgers, line 26: Seems like most of the utils in here are very thin wrappers around api calls from local dev lib. IMO most of the utils in here aren't necessary as they're adding a layer of abstraction without providing much value. The one that seems most useful is `validateAppId`, but since it's only used it one place I think it would be fine to live in migrateApp.js

packages/cli/lib/polling.js
```diff
@@ -0,0 +1,22 @@
+const { POLLING_DELAY } = require('./constants');
+
+const poll = (callback, accountId, taskId) => {
+  return new Promise((resolve, reject) => {
+    const pollInterval = setInterval(async () => {
+      const pollResp = await callback(accountId, taskId);
+      const { status } = pollResp;
+
+      if (status === 'SUCCESS') {
```
camden11, line 9: Might be worth making a constant for these statuses

packages/cli/commands/project/migrateApp.js
camden11: I think should be
```js
const appId =
    options.appId ||
    (await fetchPublicApp(accountId, accountConfig.name, options, true));
    
    await validateAppId(appId, accountId, accountConfig.name);
    
```

packages/cli/commands/project/migrateApp.js
camden11: You can actually just use `ensureProjectExists` from `lib/projects`. I updated it recently to return whether or not the project exists _and_ the project itself if you need that too

packages/cli/lib/publicApps.js
```diff
@@ -0,0 +1,60 @@
...
+
+const getMigrationStatus = async (accountId, id) => {
+  const response = await checkMigrationStatus(accountId, id);
+  return response;
+};
+
+const clonePublicApp = async (appId, name, location) => {
+  console.log('Cloning appId', appId);
+  console.log('Name:', name);
+  console.log('Location:', location);
```
camden11, line 39: Agree with what branden said about most of these being kind of superfluous. Remember to switch these to `logger`, wherever you end up putting them

packages/cli/commands/project/migrateApp.js
kemmerle: @markhazlewood

packages/cli/lib/prompts/createProjectPrompt.js
kemmerle: When we call the `createProjectPrompt` in the `migrate-app` command, we don't need the `template` option to appear in the prompts. This boolean allows us to hide that.

packages/cli/commands/project/migrateApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4

packages/cli/commands/project/migrateApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4

packages/cli/commands/project/migrateApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4

packages/cli/commands/project/migrateApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4

Realized this conditional is unnecessary--we already check whether each of these options exists in the prompts.

packages/cli/lang/en.lyaml
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4

packages/cli/lang/en.lyaml
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4

packages/cli/lib/prompts/createProjectPrompt.js
brandenrodgers: Oh I see, thanks. This might be a little clearer with a name like `skipTemplatePrompt`

packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -91,8 +98,9 @@ const createProjectPrompt = async (githubRef, promptOptions = {}) => {
           : i18n(`${i18nKey}.selectTemplate`);
       },
       when:
-        !promptOptions.template ||
-        !projectTemplates.find(t => t.name === promptOptions.template),
+        (templatesRequired && !promptOptions.template) ||
```
brandenrodgers, line 101: Not a blocker, but this could also be written as
```js
!skipTemplatePrompt && (!promptOptions.template || !projectTemplates.find(t => t.name === promptOptions.template))`
```
to avoid using templatesRequired twice (also this assumes you make the variable name change)

packages/cli/commands/project/migrateApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4

packages/cli/lib/publicApps.js
kemmerle: Deleted the `clonePublicApp` function in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/62039a6495b1b5eb20d8ddd746508b2daf69a8d4. Going to move these functions to the `migrateApp` folder in a later PR.

packages/cli/lib/prompts/createProjectPrompt.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/0930b7255c8ed88d0f7e21420974c9223016f4e4

packages/cli/lib/prompts/createProjectPrompt.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/0930b7255c8ed88d0f7e21420974c9223016f4e4

packages/cli/lib/publicApps.js
kemmerle: Removed publicApps util in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/5652758620136accfe8bedcb6f880ea864605757

packages/cli/lib/publicApps.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/5652758620136accfe8bedcb6f880ea864605757

packages/cli/lib/polling.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/d26a4b34b0899f3de6e354bfccf416f97b3278ec

packages/cli/commands/project/migrateApp.js
```diff
@@ -0,0 +1,210 @@
...
+    'appId' in options
+      ? options
+      : await selectPublicAppPrompt({
+          accountId,
+          accountName,
+          options,
+          migrateApp: true,
+        });
+
+  const validateAppId = async (appId, accountId, accountName) => {
```
camden11, line 82: Looks like you could probably just remove this function and run the next few lines inline if its not used anywhere else

packages/cli/lib/prompts/selectPublicAppPrompt.js
```diff
@@ -0,0 +1,73 @@
...
+    const filteredPublicApps = publicApps.filter(
+      app => !app.projectId && !app.sourceId
+    );
+
+    if (!filteredPublicApps.length) {
+      uiLine();
+      logger.error(i18n(`${i18nKey}.errors.noApps`));
+      logger.log(i18n(`${i18nKey}.errors.noAppsMessage`, { accountName }));
+      uiLine();
+      process.exit(EXIT_CODES.ERROR);
```
camden11, line 24: I am not 100% positive what the best practices here but I feel like not having any apps to migrate isn't _really_ an error. Might make more sense as a successful exit

packages/cli/lib/prompts/selectPublicAppPrompt.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return filteredPublicApps;
+  } catch (error) {
+    logger.debug(error);
```
camden11, line 28: Should use `logErrorInstance` (or maybe `logApiErrorInstance`) here

camden11 (approved):
Looking great! Just a few minor things. Will approve for now so you're not blocked, just make sure to get those things cleaned up + release LDL before you merge.

I also haven't had a chance to run this myself yet so let me know if you need help testing

packages/cli/commands/project/migrateApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/17da9c8463e4edcfc9166abc3ad864a30736e0dd

packages/cli/lib/prompts/selectPublicAppPrompt.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/17da9c8463e4edcfc9166abc3ad864a30736e0dd

packages/cli/lib/prompts/selectPublicAppPrompt.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1052/commits/17da9c8463e4edcfc9166abc3ad864a30736e0dd

PR #1054, opened 2024-05-02
Public app install prompt

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/523
This adds a few checks to `hs project dev` to ensure development of public apps goes smoothly. Now, the CLI will check to see if the public app is installed on the target developer test account, and prompt you to open your browser and install if it's not. This also fetches information about app installs, which sets us up to do the active install warning in a future PR (see https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/522)

## Screenshots
Screenshot 2024-05-02 at 4 36 16 PM

## TODO / Questions / Etc
* ~I'm having some trouble getting my backend to work properly on QA, so I haven't gotten all the way through the flow yet. I'm 90% sure it all works though~ Tested and works on QA!
* Right now, all this does is open the install page for the user. We don't have any way to confirm if they actually install the app, so we continue with local dev once they open the page. Is that OK for now or should there be another check of some kind?
* Need to release local-dev-lib and update dependancy

## Who to Notify
@brandenrodgers @kemmerle @joe-yeager @markhazlewood

### DIFF
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -1231,6 +1231,10 @@ en:
           destRequired: "You must specify a destination directory."
       cleanUploadPrompt:
         message: "You are about to remove any remote files in \"{{ filePath }}\" on HubSpot account {{ accountId }} that don't exist locally. Are you sure you want to do this?"
+      installPublicAppPrompt:
+        explanation: "Local development requires this app to be installed in the target test account"
+        prompt: "Open hubspot.com to install this app?"
+        decline: "To continue local development of this app, install it in your target test account and re-run {{#bold}}`hs project dev`{{/bold}}"
     convertFields:
       positionals:
         src:
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -927,6 +927,7 @@ en:
       exitingStart: "Stopping local dev server ..."
       exitingSucceed: "Successfully exited"
       exitingFail: "Failed to cleanup before exiting"
+      missingUid: "Could not find a uid for the selected app. Confirm that the app config file contains the uid field and re-run {{#bold}}`hs project dev`{{/bold}}."
       uploadWarning:
         appLabel: "[App]"
         uiExtensionLabel: "[UI Extension]"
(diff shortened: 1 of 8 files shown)


packages/cli/lib/LocalDevManager.js
```diff
@@ -179,6 +219,32 @@ class LocalDevManager {
     process.exit(EXIT_CODES.SUCCESS);
   }
 
+  getActiveAppInstallationData() {
+    return fetchAppInstallationData(
+      this.targetAccountId,
+      this.projectId,
+      this.activeApp.config.uid,
```
kemmerle, line 226: These variables are kind of a nightmare to read, but I don't think destructuring would help the code's readability overall. Not a blocker, but something to think about.

kemmerle: Honestly, maybe it's more a reflection on how our BE is structured...

packages/cli/lang/en.lyaml
```diff
@@ -927,6 +927,7 @@ en:
       exitingStart: "Stopping local dev server ..."
       exitingSucceed: "Successfully exited"
       exitingFail: "Failed to cleanup before exiting"
+      missingUid: "Could not find a uid for the selected app. Confirm that the app config file contains the uid field and re-run {{#bold}}`hs project dev`{{/bold}}."
```
brandenrodgers, line 930: There's a `uiCommandReference` util for logging out any commands for the user to run.

packages/cli/lib/LocalDevManager.js
camden11: Yeah unfortunately the endpoint just requires a LOT of info to work. Not sure if there's any winning for us here

brandenrodgers:
>  We don't have any way to confirm if they actually install the app, so we continue with local dev once they open the page. Is that OK for now or should there be another check of some kind?

Yeah I think that's okay for this first iteration. Maybe in a second pass at this we can make it re-fetch the installed apps to verify that the app actually was installed before continuing.

brandenrodgers (approved):
Left a comment about a ui util that we should use for command references, but otherwise this lgtm

PR #1055, opened 2024-05-02
Stop linking to old feedback form

brandenrodgers:
## Description and Context

`hs feedback` was linking to a feedback survey that we no longer use. We're working to get another survey up for the command to link to, but in the meantime running `hs feedback --general` will now link to this CLI repository and users can report feedback in here.


camden11 (approved):
🚢

PR #1056, opened 2024-05-07
Error on private app dev against Dev account

brandenrodgers:
## Description and Context

Making sure that we error and exit when the user attempts to do local dev of a private app on an App Developer Account. This isn't supported and there isn't anything useful that we can automate here. The user needs to select another account type to do local dev against in this scenario.


camden11 (approved):
Looks like theres a conflict (probably with the logging stuff I just did) but LGTM otherwise

PR #1065, opened 2024-05-10
Fix install url encoding

camden11:
## Description and Context
See https://hubspot.slack.com/archives/C066MUYG1UH/p1715359708440239
This fixes the encoding of scopes in install urls in `hs project dev`. Previously, install urls broke when apps required more than one score

## Questions
Is it safe to always just use the first `redirectUrl`? The `redirect_uri` param only supports one url

## Who to Notify
@brandenrodgers @kemmerle @robrown-hubspot


brandenrodgers:
> Is it safe to always just use the first redirectUrl? The redirect_uri param only supports one url

We can use the first item for now since it's the easiest option, but it's definitely possible for devs to include multiple urls in that field. My guess is that this default behavior will be reasonable for most use cases though.

brandenrodgers (approved):

PR #1076, opened 2024-05-17
WIP: Prompt upload on config change

camden11:

### DIFF
--- a/packages/cli/lib/LocalDevManager.js
+++ b/packages/cli/lib/LocalDevManager.js
@@ -307,7 +307,7 @@ class LocalDevManager {
 
     logger.log();
     this.isGithubLinked
-      ? logger.warn(i18n(`${i18nKey}.uploadWarning.header`, { warning }))
+      ? logger.warn(i18n(`${i18nKey}.uploadWarning.githubHeader`, { warning }))
       : logger.warn(warning);
 
     let uploadSuccess = false;
--- a/packages/cli/lib/LocalDevManager.js
+++ b/packages/cli/lib/LocalDevManager.js
@@ -18,7 +18,11 @@ const { PROJECT_CONFIG_FILE } = require('./constants');
 const SpinniesManager = require('./ui/SpinniesManager');
 const DevServerManager = require('./DevServerManager');
 const { EXIT_CODES } = require('./enums/exitCodes');
-const { getProjectDetailUrl } = require('./projects');
+const {
+  getProjectDetailUrl,
+  handleProjectUpload,
+  pollProjectBuildAndDeploy,
+} = require('./projects');
 const {
   CONFIG_FILES,
   COMPONENT_TYPES,
@@ -34,6 +38,10 @@ const {
 } = require('./ui');
 const { logErrorInstance } = require('./errorHandlers/standardErrors');
 const { installPublicAppPrompt } = require('./prompts/installPublicAppPrompt');
+const {
+  publicAppUploadPrompt,
+  privateAppUploadPrompt,
+} = require('./prompts/localDevUploadPrompt');
 
 const WATCH_EVENTS = {
   add: 'add',
@@ -282,7 +290,7 @@ class LocalDevManager {
       : uiCommandReference('hs project upload');
   }
 
-  logUploadWarning(reason) {
+  async warnAndPromptUpload(reason) {
     let warning = reason;
     if (!reason) {
       warning =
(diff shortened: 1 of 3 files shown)


packages/cli/lib/prompts/localDevUploadPrompt.js
```diff
@@ -0,0 +1,49 @@
...
+
+    const formattedMessage = `${chalk.green('?')} ${chalk.bold(
+      message
+    )} ${chalk.dim('(Y/n)')} `;
+
+    let promiseResolve;
+
+    const promptPromise = new Promise(resolve => {
+      promiseResolve = resolve;
+      rl.question(formattedMessage, answer => {
```
camden11, line 23: I tried a bunch of different things, but could not figure out a way to cancel prompts with our current version of `inquirer`. So for this one, just doing it manually with readline

packages/cli/lib/prompts/localDevUploadPrompt.js
```diff
@@ -0,0 +1,49 @@
+const readline = require('readline');
+const chalk = require('chalk');
+// const { promptUser } = require('./promptUtils');
+const { i18n } = require('../lang');
+
+const i18nKey = 'lib.prompts.localDevUploadPrompt';
+
+const makeUploadPrompt = message => {
+  return () => {
+    const rl = readline.createInterface({
```
brandenrodgers, line 10: I'm not exactly sure where it's coming from, but it seems like my inputs for each prompt are not clearing between prompts. You can reproduce this by getting the prompt to show up several times in a row. The first time when I type "n" I see "n" in the terminal. The second time I see "nn", then "nnn", and so on.

Somewhat related, I'm able to type `q` and the command immediately exits (I think that's WAD), but the cleanup logs show on the same line as the upload prompt. Maybe we want to exit the prompt before logging the cleanup commands? Here's an example:
image

camden11: hmmm that is...not ideal. looking into it

camden11: OK, I fixed the issue where keypresses were logging twice. Also opted to have `q` _not_ exit the command when the prompt is open. Felt a little weird since you can theoretically type anything into the prompt.

It now cleanly shuts down (and tells you to upload to see your config changes) when you press control C with the prompt open

packages/cli/lib/prompts/localDevUploadPrompt.js
```diff
@@ -0,0 +1,50 @@
+const readline = require('readline');
+const chalk = require('chalk');
+// const { promptUser } = require('./promptUtils');
```
brandenrodgers, line 3: This comment can probably go away since we can't use inquirer here

packages/cli/lib/prompts/localDevUploadPrompt.js
```diff
@@ -0,0 +1,50 @@
...
+const { i18n } = require('../lang');
+
+const i18nKey = 'lib.prompts.localDevUploadPrompt';
+
+const makeUploadPrompt = message => {
```
brandenrodgers, line 8: It may be worthwhile to add some comments in the code here, since this is breaking away from the standard prompt logic that we're using everywhere else

brandenrodgers (approved):
Code lgtm! Have we had a chance to run this by design yet? Should we wait until the sync later today before merging?

camden11:
Oh good call. I can demo today!

packages/cli/lib/DevServerManager.js
```diff
@@ -124,6 +125,16 @@ class DevServerManager {
     }
   }
 
+  updateConfigFile({ filepath, config }) {
+    if (this.started) {
+      this.iterateDevServers(async serverInterface => {
+        if (serverInterface.updateConfigFile) {
+          await serverInterface.updateConfigFile(filepath, config);
```
tamara-at-hubspot, line 132: Hi @camden11, I'm trying to familiarize myself with these lifecycle methods. What's the difference between `fileChange` and `updateConfigFile`?

camden11: Unfortunately to my knowledge there's no documentation (cc @brandenrodgers) but agreed that that's definitely something we should work on, especially if other teams are to develop their own dev servers.

As it stands now:
* `fileChange` is called by the CLI any time a file _other_ than an app config is changed. If an app config file is changed, the CLI assumes the dev server can't handle that change and prompts the user to upload their project.
* If that upload is successful, the CLI informs the dev server of the changed app config file via `updateConfigFile`

* If a card config file (or any other file the dev server doesn't want to handle or knows upload is required for) is changed, the CLI will still call `fileChange`. Within `fileChange`, the dev server will recognize the file as one that requires an upload, and call the `onUploadRequired` callback passed from the CLI. 
* This triggers the upload prompt. If it's successful, the CLI will then call `updateConfigFile` to let the dev server know the config file was updated

It is admittedly starting to seem like this prompt functionality doesn't work super nicely with the way the dev server interface was originally set up. Might be worth re-thinking entirely, but for the time being wanted something that works within the current constraints.

As for the inconsistency between `setup` and `updateConfigFile`, the current `updateConfigFile` API was recommended by @joe-yeager. Can definitely update to make it more similar to `setup` if you both think that would make more sense

joe-yeager: Another +1 for for formalizing the contract, that is definitely a gap.

> I'm also curious if there's any documentation that might explain the lifecycle of a dev server. For example, it would be helpful to know when is setup called, the expectation on the dev server, the contract between the dev server and the dev mode manager

`setup` is called [here](https://github.com/HubSpot/hubspot-cli/pull/1076/files/b4baee91d124c89dd75ef4ad36de5b5e65b6981c#diff-004104a9aade0cabe581c4be51b61d8b43745b9772cc1d3afedc9c5b64eb3d00R82) and should only be called when `hs project dev` is initializing.  It can technically be called multiple times because the `DevServerManager` iterates the dev servers defined in [this map](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/DevServerManager.js#L32-L42) and calls the `setup` method based on the components.  We have [this check](https://git.hubteam.com/HubSpot/ui-extensibility/blob/master/public-packages/ui-extensions-dev-server/src/lib/DevModeInterface.ts#L145) to prevent the dev server from running for multiple components types because right now the only components are public apps and private apps which are not supported in the same project.

> I feel like there's some inconsistency between the setup and updateConfigFile interfaces. Setup provides a data structure about the components ([here](https://git.hubteam.com/HubSpot/ui-extensibility/blob/master/public-packages/ui-extensions-dev-server/src/lib/DevModeInterface.ts#L42)), whereas updateConfigFile drops one level down to the file

The reason I recommended this API was that we can use the `filepath` to loop over the [extensionsConfig](https://git.hubteam.com/HubSpot/ui-extensibility/blob/master/public-packages/ui-extensions-dev-server/src/lib/DevModeInterface.ts#L67) which is an array of the configs for all of the currently running extensions and has the [`extensionsConfigPath`](https://git.hubteam.com/HubSpot/ui-extensibility/blob/master/public-packages/ui-extensions-dev-server/src/lib/types.ts#L13) prop, which should be the absolute file path of the extension config.  So the [projectComponentMap](https://git.hubteam.com/HubSpot/ui-extensibility/blob/master/public-packages/ui-extensions-dev-server/src/lib/types.ts#L100) seemed like extraneous information

>  What's the difference between fileChange and updateConfigFile?

> It is admittedly starting to seem like this prompt functionality doesn't work super nicely with the way the dev server interface was originally set up

This question and statement bring up a great point.  I'm just riffing, but maybe we could update the `fileChange` method to add an additional 'context' arg that could contain the config files contents to simplify the contract, we would also need to update the logic so that method is called even for app config files so the dev server can update it's local copy of the app config.  So it's contract could be something along the lines of:

```typescript
function fileChange(filePath: string, __event: unknown, context?: { config?: AppConfig | ExtensionConfig}){}
```

That feels like it gets us what we need while removing some of the complications.

joe-yeager: Here's a [placeholder issue](https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/581) to document the contract so that doesn't get left behind

camden11: After looking into this approach, @joe-yeager and I decided that it would be safer and more straightforward to instead add a `restart` method to the dev server that stops and restarts the dev server while preserving any relevant state from the previous session.

PR for that here: https://git.hubteam.com/HubSpot/ui-extensibility/pull/2306

packages/cli/lib/LocalDevManager.js
```diff
@@ -348,32 +367,130 @@ class LocalDevManager {
...
           })
         );
+    logger.log(
+      i18n(`${i18nKey}.uploadWarning.restartDev`, {
+        command: uiCommandReference('hs project dev'),
+      })
+    );
+  }
+
+  updateComponentAtPath(filepath) {
```
camden11, line 458: This re-parses the changed config file and updates `runnableComponents`. This is necessary because the initial parsing of config files happens outside `LocalDevManager`

camden11:
Update - realizing there are a LOT of edge cases this doesn't properly account for. Converting to draft for now

PR #1078, opened 2024-05-20
chore: Remove unused boolean args passed to options methods.

joe-yeager:
## Description and Context

When the migration to `yargs` occurred, some methods had a last argument left behind, and the methods no longer require them.

- Remove the last boolean argument from the invocations of the following methods:
   - [`addAccountOptions`](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/commonOpts.js#L18-L23)
   - [`addConfigOptions`](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/commonOpts.js#L25-L30)
   - [`addOverwriteOptions`](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/commonOpts.js#L32-L38)
   - [`addModeOptions`](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/commonOpts.js#L40-L53)
   - [`addTestingOptions`](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/commonOpts.js#L55-L61)
   - [`addUseEnvironmentOptions`](https://github.com/HubSpot/hubspot-cli/blob/main/packages/cli/lib/commonOpts.js#L63-L68)

### DIFF
--- a/packages/cli/commands/accounts.js
+++ b/packages/cli/commands/accounts.js
@@ -13,8 +13,8 @@ exports.command = 'accounts';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.builder = yargs => {
-  addConfigOptions(yargs, true);
-  addAccountOptions(yargs, true);
+  addConfigOptions(yargs);
+  addAccountOptions(yargs);
 
   yargs
     .command({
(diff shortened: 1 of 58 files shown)


brandenrodgers (approved):
Oh wow good find!

PR #1079, opened 2024-05-20
feat: Add link to local development docs

joe-yeager:
## Description and Context

Add a local dev docs link to the `hs project dev` command

### During setup phase
Screenshot 2024-05-20 at 11 55 44 AM

Screenshot 2024-05-21 at 10 41 53 AM

### After startup
Screenshot 2024-05-20 at 11 55 29 AM

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -25,6 +25,7 @@ const {
   uiAccountDescription,
   uiBetaTag,
   uiCommandReference,
+  uiLink,
 } = require('../../lib/ui');
 const SpinniesManager = require('../../lib/ui/SpinniesManager');
 const LocalDevManager = require('../../lib/LocalDevManager');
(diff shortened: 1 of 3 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -70,6 +71,13 @@ exports.handler = async options => {
 
   uiBetaTag(i18n(`${i18nKey}.logs.betaMessage`));
 
+  logger.log(
```
joe-yeager, line 74: Not sure if we want this here or not.  This is the initial log when `hs project dev` is ran, and this gets cleared when the account has been selected and the dev servers have started up.

camden11 (approved):
LGTM. I think it makes sense to have it in both places, may be worth running by @markhazlewood though

joe-yeager:
> LGTM. I think it makes sense to have it in both places, may be worth running by @markhazlewood though

I checked with Mark and he is onboard with having it in both places.

PR #1080, opened 2024-05-20
fix: Honor the `--account` flag for `hs project download`

joe-yeager:
## Description and Context

- The `hs project download` command was not honoring the account passed in with the `--account` flag and was using the default account for fetching the project list

### Before
I am prompted with a list of projects that are associated with my default account and not the account provided
Screenshot 2024-05-20 at 2 19 06 PM

### After
I am prompted with a list of projects for the account passed in with the `--account` flag
Screenshot 2024-05-20 at 2 19 47 PM


packages/cli/commands/project/download.js
```diff
@@ -67,7 +67,7 @@ exports.handler = async options => {
         })
       );
       let { name: promptedProjectName } = await downloadProjectPrompt(options);
-      let projectName = promptedProjectName || project;
+      projectName = promptedProjectName || project;
```
joe-yeager, line 70: This is unrelated to the account not being honored, but this was redeclaring a variable defined in the parent scope [here](https://github.com/HubSpot/hubspot-cli/pull/1080/files#diff-a746cff94477f3abe62a319178fa6394252aed5473144e39412aa31af3f87d7aR46)

camden11 (approved):
Oooh good catch. LGTM + tested and works as expected

PR #1081, opened 2024-05-20
chore: Rename tests and mock logger to remove logs from test ouput

joe-yeager:
## Description and Context

- Standardize the test file names to all end in `.test.js`.  I looked at `local-dev-lib` as a model for what format to use for these files, either `.spec.js` or `.test.js` and I couldn't find a standard,  so I went with `.test.js` because I see we use that in a few places.
- Mock the logger so the tests don't log to `stderr`

### Before
Screenshot 2024-05-20 at 4 09 02 PM

### After

This spinnies output is still there, but it's not as in your face as the error logs
Screenshot 2024-05-20 at 4 09 27 PM


packages/cli/lib/__tests__/projects.test.js
```diff
@@ -1,43 +1,39 @@
...
     let projectDir;
     let exitMock;
-    let errorSpy;
 
     beforeAll(() => {
       projectDir = fs.mkdtempSync(path.join(os.tmpdir(), 'projects-'));
       fs.mkdirSync(path.join(projectDir, 'src'));
 
       realProcess = process;
-      errorSpy = jest.spyOn(console, 'error');
```
joe-yeager, line 19: Instead of spying on `console.error`, I just mocked the `@hubspot/local-dev-lib/logger` so this test file doesn't need to be concerned with the implementation details of `logger.error`.  This also has the benefit for keeping the logs from hitting `stderr`

PR #1082, opened 2024-05-21
fix: Reject promise if taskStatus is nullable

joe-yeager:
## Description and Context

- Refactor unclear async behavior
- Reject the promise if `taskStatus` is undefined since it is essential for the rest of the method to function

## TODO

- Address the `TODO` comments added in this commit https://github.com/HubSpot/hubspot-cli/pull/1082/commits/48000af34a66a51f0ce2a643de74d3cd77d8c9bd


packages/cli/lib/projects.js
```diff
@@ -699,9 +699,24 @@ const makePollTaskStatusFunc = ({
 
     return new Promise((resolve, reject) => {
       const pollInterval = setInterval(async () => {
-        const taskStatus = await statusFn(accountId, taskName, taskId).catch(
-          reject
```
joe-yeager, line 703: So I think this was the problematic line.  Even if the `catch` block calls the `reject` methods the rest of the function will execute with a nullable `taskStatus`

packages/cli/lang/en.lyaml
```diff
@@ -1028,6 +1028,7 @@ en:
         componentCount: "Found {{ numComponents }} components in this project"
         successStatusText: "DONE"
         failedStatusText: "FAILED"
+        errorFetchingTaskStatus: "Error fetching task status"
```
joe-yeager, line 1031: I'm not the best at error messages, and I'm still learning my way around this project, so I'm all 👂s if anyone has a better idea for this message.

packages/cli/lib/projects.js
camden11: This is a little more complicated than most other cases in the repo due to the polling, but usually we try to surface errors from the backend using `logApiErrorInstance`. Probably worth giving that a shot and seeing if we at least get a helpful error. You could maybe try using `getAxiosErrorWithContext` from `local-dev-lib` and using that for the promise rejection.

If the backend error _isn't_ helpful, I like your solution!

packages/cli/lang/en.lyaml
camden11: If we do end up using this error message, could be good to specify "build" or "deploy" since task doesn't necessarily mean anything to users

joe-yeager: Good point, it looks like the `i18n` module supports interpolation and I can make this more specific

packages/cli/lib/projects.js
```diff
@@ -699,9 +699,51 @@ const makePollTaskStatusFunc = ({
...
+          taskStatus = await statusFn(accountId, taskName, taskId);
+        } catch (e) {
+          logApiErrorInstance(
+            e,
+            new ApiErrorContext({
+              accountId,
+              projectName: taskName,
+            })
+          );
+          return reject(
```
brandenrodgers, line 713: I'm curious how this error handling will impact the downstream code. Do we ever try to catch the errors thrown by this build/deploy polling util?

joe-yeager: Great question.  I was assuming that since this method was calling `reject` previously that it should be handled, but that wasn't always the case.  I'm going to push up a couple places that aren't handling the rejection with some TODO comments and address.

packages/cli/lib/projects.js
```diff
@@ -405,13 +405,19 @@ const pollProjectBuildAndDeploy = async (
...
+  try {
+    buildStatus = await pollBuildStatus(
+      accountId,
+      projectConfig.name,
+      buildId,
+      null,
+      silenceLogs
+    );
+  } catch (e) {
+    //TODO: Figure out how we want to actually handle this
```
kemmerle, line 418: I'm going to echo Camden's suggestion for `logApiErrorInstance`. That function checks whether the error is returned from `axios`, and if it's not, defaults to `logErrorInstance`, which works with generic errors. It's served me well in migration work.

packages/cli/lang/en.lyaml
```diff
@@ -1030,6 +1030,7 @@ en:
         componentCount: "Found {{ numComponents }} components in this project"
         successStatusText: "DONE"
         failedStatusText: "FAILED"
+        errorFetchingTaskStatus: "Error fetching {{ taskType }} status"
```
kemmerle, line 1033: I think this is fine for now, assuming that we're not failing the build/deploy when we error out while polling for build/deploy status. It might be useful to add a `logger.debug` with the full error message, so it's easier to debug any issues.

packages/cli/lib/projects.js
joe-yeager: > I'm curious how this error handling will impact the downstream code. Do we ever try to catch the errors thrown by this build/deploy polling util?


It looks like we don't explicitly catch the errors off of any of the polling methods.  I added that commit with the todos, but ended up reverting it because I couldn't think of anything else to do in those situations other than log, and I centralized that in this method.

PR #1084, opened 2024-05-29
Consolidate account creation helpers

adamawang:
## Description and Context

Enhancement to consolidate helper functions for creating sandboxes and developer test accounts. Both flows shared a large amount of duplicate code across a number of files, this PR cleans that up and consolidates it into a single `buildNewAccount` helper function that is used in `hs sandbox create` and `hs project dev`. I've moved error handling into their own respective helper files since there were fairly large differences between the two flows, you'll see those in `handleSandboxCreateError` and `handleDeveloperTestAccountCreateError`. 

There is no difference in behavior for any of the commands listed above, you should expect sandbox and developer test account creation behavior to be exactly the same as before. Please let me know if you see any differences. 

I've also updated a few lang keys to be more consistent with the rest of the CLI, eg: `sandboxName` and `name` are now `accountName`, `sandboxType` is now `accountType`, `sandboxHubId` is now `accountId`. 

### Testing

`hs project dev` testing:
- Create a new private app project. Set your default account to a STANDARD account (production) and run `hs project dev` within the project folder. You should be able to create a new development sandbox like before with all the appropriate messaging and correct account type naming. 
- Create a new public app project. Set your default account to a DEVELOPER account (app developer) and run `hs project dev`. You should be able to create a new developer test account like before with all the appropriate messaging and correct account type naming. 

`hs sandbox create` testing:
- Create a new standard sandbox. Set your default account to a STANDARD account (production) and run `hs sandbox create`. Select standard sandbox and continue. You should be able to create a new standard sandbox like before with all the appropriate messaging and correct account type naming. 
- Create a new development sandbox. Set your default account to a STANDARD account (production) and run `hs sandbox create`. Select development sandbox and continue. You should be able to create a new development sandbox like before with all the appropriate messaging and correct account type naming. 

## Screenshots

No UI changes

## Who to Notify

@shannon-lichtenwalter


camden11 (approved):
Nice, this really cleans things up. LGTM + tested and works

PR #1085, opened 2024-05-30
Add `hs project clone-app` command

kemmerle:
## Description and Context
In this PR, I'm adding support for public apps cloning, including new endpoints. 

There is a [sister PR](https://github.com/HubSpot/hubspot-local-dev-lib/pull/143) in local-dev-lib that has been merged and released.

## Screenshots

Happy path 😄 :

Screenshot 2024-05-29 at 8 30 50 PM

Sad path 😢  (Ignore the TypeError--that's expected because I'm not passing a real axios error):

Screenshot 2024-05-29 at 8 32 04 PM

## TODO

- [x] Update the local-dev-lib PR to hit the actual endpoint(s) when they're complete
- [x] Address any feedback

## Who to Notify


packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,171 @@
...
+        useCommand: uiCommandReference('hs accounts use'),
+        authCommand: uiCommandReference('hs auth'),
+      })
+    );
+    uiLine();
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  const { appId } =
+    'appId' in options
```
brandenrodgers, line 68: Is options an array here? Isn't it typically an object with keys?

packages/cli/lang/en.lyaml
```diff
@@ -529,6 +529,28 @@ en:
...
+          enterLocation: "[--location] Where should the cloned project be created?"
+          cloneStatus:
+            inProgress: "Cloning app configuration to {{#bold}}public-app.json{{/bold}} component definition ..."
+            done: "Cloning app configuration to public-app.json component definition ... DONE"
+            success: "Your cloned project was created in {{ location }}"
+            failure: "Cloning app configuration to public-app.json component definition ... FAILED"
+          errors:
+            invalidAccountTypeTitle: "{{#bold}}Developer account not targeted{{/bold}}"
+            invalidAccountTypeDescription: "Only public apps created in a developer account can be converted to a project component. Select a connected developer account with {{useCommand}} or {{authCommand}} and try again."
+            invalidAppId: "[--appId] Could not find appId {{ appId }}. Please choose another public app."
```
brandenrodgers, line 550: No need for the mention of the `[--appId]` flag here. We typically only do this for prompt options

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,171 @@
...
+      logger.success(i18n(`${i18nKey}.cloneStatus.success`, { location }));
+      logger.log('');
+      process.exit(EXIT_CODES.SUCCESS);
+    }
+  } catch (error) {
+    SpinniesManager.fail('cloneApp', {
+      text: i18n(`${i18nKey}.cloneStatus.failure`),
+      failColor: 'white',
+    });
+    if (error.errors) {
```
brandenrodgers, line 142: When does `error.errors` exist? Should we have handling for these in our apiErrors util in local-dev-lib? I have a similar gut feeling here that shrinking the size of the try/catch blocks will make error handling more straightforward. Some of the errors thrown will be filesystem errors and some will be api errors. We're calling `logApiErrorInstance` in the catch block, but what happens if it was the `extractZipArchive ` util that failed?

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,171 @@
...
+        });
+
+  try {
+    SpinniesManager.init();
+
+    SpinniesManager.add('cloneApp', {
+      text: i18n(`${i18nKey}.cloneStatus.inProgress`),
+    });
+
+    const cloneResponse = await cloneApp(accountId, appId);
```
brandenrodgers, line 112: You could destructure the responses from cloneApp and poll inline to shorten this section

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,171 @@
...
+    const { exportId } = cloneResponse;
+    const pollResponse = await poll(checkCloneStatus, accountId, exportId);
+    const { status } = pollResponse;
+    if (status === 'SUCCESS') {
```
brandenrodgers, line 116: Do we need an else case here?

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,171 @@
...
+  downloadClonedProject,
+} = require('@hubspot/local-dev-lib/api/projects');
+const { getCwd, isValidPath } = require('@hubspot/local-dev-lib/path');
+const { logger } = require('@hubspot/local-dev-lib/logger');
+const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
+const { extractZipArchive } = require('@hubspot/local-dev-lib/archive');
+
+const i18nKey = 'commands.project.subcommands.cloneApp';
+
+exports.command = 'clone-app';
```
brandenrodgers, line 42: Should this be `clone-app [--appId] [--location]`?

packages/cli/commands/project/cloneApp.js
kemmerle: Created an [issue](https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/600) for this, so we can follow up.

packages/cli/lang/en.lyaml
kemmerle: I've addressed that issue in this [PR](https://github.com/HubSpot/hubspot-cli/pull/1111). Going to leave as is to prevent merge conflicts.

kemmerle: Merged in these changes here: https://github.com/HubSpot/hubspot-cli/pull/1085/commits/9d7e99a5965e912bdad29344fc2265453c60f7a7

packages/cli/commands/project/cloneApp.js
kemmerle: The migrations endpoints (that includes migration and cloning) return a response object that looks like this:

```
{
id: 12234,
status: 'PENDING',
errors?: [Error, Error] //I know this is bad pseudo code, apologies
}
```

I don't believe that we need to account for this in our error handling in local-dev-lib, given that this is a special case (Projects can have multiple reasons why migration/cloning fails). I've accounted for other error types (as in the error by `extractZipArchive` with the conditional statement. I'm going to leave as is.

packages/cli/commands/project/cloneApp.js
kemmerle: No, we don't. In the `poll` function, we reject any failing status or error that occurs (the `poll` function is wrapped in a try catch block on your suggestion). That means the error is automatically thrown to the `catch` in this try catch block.

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1085/commits/80217fc0b34dc2fdf1b979b78f2c19967105fbbe

packages/cli/commands/project/cloneApp.js
kemmerle: Added a note directly in the code about this: https://github.com/HubSpot/hubspot-cli/pull/1085/commits/80217fc0b34dc2fdf1b979b78f2c19967105fbbe

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
...
+const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
+const {
+  fetchPublicAppMetadata,
+} = require('@hubspot/local-dev-lib/api/appsDev');
+const { extractZipArchive } = require('@hubspot/local-dev-lib/archive');
+
+const i18nKey = 'commands.project.subcommands.cloneApp';
+
+exports.command = 'clone-app [--appId] [--location]';
+exports.describe = null; // uiBetaTag(i18n(`${i18nKey}.describe`), false);
```
camden11, line 45: Should this be uncommented or are we purposefully hiding for now?

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
...
+    uiLine();
+    logger.error(i18n(`${i18nKey}.errors.invalidAccountTypeTitle`));
+    logger.log(
+      i18n(`${i18nKey}.errors.invalidAccountTypeDescription`, {
+        useCommand: uiCommandReference('hs accounts use'),
+        authCommand: uiCommandReference('hs auth'),
+      })
+    );
+    uiLine();
+    process.exit(EXIT_CODES.ERROR);
```
camden11, line 66: We're not the most consistent in how we use these exit codes, but I think we want this to be `SUCCESS`, since nothing is actually going wrong here. That's how I thought about it for local dev stuff at least. Might be something we should try to make some more concrete rules for

cc @brandenrodgers @joe-yeager

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
...
+    const appIdResponse =
+      'appId' in options
+        ? { appId: options.appId }
+        : await selectPublicAppPrompt({
+            accountId,
+            accountName,
+            options,
+            migrateApp: false,
+          });
+    appId = appIdResponse.appId;
```
camden11, line 82: Nit but this is a little difficult to read IMO. I'd suggest something like

```
let appId = options.appId;

if (!appId) {
  const appIdResponse = await selectPublicAppPrompt({
            accountId,
            accountName,
            options,
            migrateApp: false,
          });
  appId = appIdResponse.appId
}
```

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
...
+              }
+              if (fs.existsSync(input)) {
+                return i18n(`${i18nKey}.errors.invalidLocation`);
+              }
+              if (!isValidPath(input)) {
+                return i18n(`${i18nKey}.errors.invalidCharacters`);
+              }
+              return true;
+            },
+          });
```
camden11, line 106: I know this isn't a huge prompt but might be worth moving to its own file due to the validation function being pretty lengthy. In this case, rather than using the `'location' in options` pattern, you could use the `when` field in `promptUser` to avoid running the prompt when location already exists (see `projectNamePrompt` for example.)

I would have suggested that for `selectPublicAppPrompt` too, but that prompt makes a request so I think it makes sense to skip calling it completely when we can.

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
...
+const { logger } = require('@hubspot/local-dev-lib/logger');
+const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
+const {
+  fetchPublicAppMetadata,
+} = require('@hubspot/local-dev-lib/api/appsDev');
+const { extractZipArchive } = require('@hubspot/local-dev-lib/archive');
+
+const i18nKey = 'commands.project.subcommands.cloneApp';
+
+exports.command = 'clone-app [--appId] [--location]';
```
joe-yeager, line 44: Do we want to include the positional arguments here?  I think the plan is to move away from this in favor of flags

packages/cli/commands/project/cloneApp.js
camden11: If any status other than `SUCCESS` throws an error, you can probably just skip checking status entirely

packages/cli/commands/project/cloneApp.js
joe-yeager, line 45: Just checking that this is intentionally `null` with the commented code

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
...
+      process.exit(EXIT_CODES.SUCCESS);
+    }
+  } catch (error) {
+    SpinniesManager.fail('cloneApp', {
+      text: i18n(`${i18nKey}.cloneStatus.failure`),
+      failColor: 'white',
+    });
+    // Migrations endpoints return a response object with an errors property. The errors property contains an array of errors.
+    if (error.errors) {
+      error.errors.forEach(logApiErrorInstance);
```
camden11, line 149: Should these include an `ApiErrorContext` with `accountId` as well?

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
...
+      logger.log('');
+      process.exit(EXIT_CODES.SUCCESS);
+    }
+  } catch (error) {
+    SpinniesManager.fail('cloneApp', {
+      text: i18n(`${i18nKey}.cloneStatus.failure`),
+      failColor: 'white',
+    });
+    // Migrations endpoints return a response object with an errors property. The errors property contains an array of errors.
+    if (error.errors) {
```
joe-yeager, line 148: This could be `isArray(error.errors)` so we are also checking that `error.errors` is an array before iterating it

packages/cli/commands/project/cloneApp.js
```diff
@@ -0,0 +1,177 @@
+const fs = require('fs');
```
joe-yeager, line 1: Can we add some tests for this command so we can assure that future changes maintain the behavior?

packages/cli/commands/project/cloneApp.js
joe-yeager: Whoops, I hit submit after Camden reviewed so this is a dupe.  Resolving

kemmerle: Purposefully hiding for now 👍 Going to resolve.

packages/cli/commands/project/cloneApp.js
kemmerle: The dashes in front of these mark them as flags. Going to mark this as resolved.

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1085/commits/247b941649560ed561734257bca012ddf2d65f02

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1085/commits/247b941649560ed561734257bca012ddf2d65f02

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1085/commits/247b941649560ed561734257bca012ddf2d65f02

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1085/commits/247b941649560ed561734257bca012ddf2d65f02

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1085/commits/247b941649560ed561734257bca012ddf2d65f02

camden11:
All looking good on my end! Agree with joe that adding some tests could be good though

packages/cli/commands/project/cloneApp.js
kemmerle: Going to implement tests in a separate PR, so I can add tests for `migrate-app` at the same time. Also so that we can hopefully test this command in tomorrow's bug bash.

packages/cli/commands/project/cloneApp.js
joe-yeager: I noticed this weird behavior when I was working on the deploy command, they get treated as positional args when included here.  

Screenshot 2024-07-18 at 8 59 36 AM

joe-yeager (approved):
LGTM!  I just unresolved the comment about the positional args and added a screenshot to show what I was talking about

packages/cli/commands/project/cloneApp.js
kemmerle: Looked over the `yargs` docs, and they don't even mention flags in relation to the `command` method: https://github.com/yargs/yargs/blob/main/docs/advanced.md#commands. I'll remove them. Thanks for catching this, Joe🥇

PR #1086, opened 2024-05-30
Consolidate account name prompts for HubSpot portals

adamawang:
## Description and Context

Small enhancement to consolidate `developerTestAccountNamePrompt` and `sandboxNamePrompt` which were doing almost the same thing into one `hubspotAccountNamePrompt` which takes an accountType to determine what messaging to show. I've also removed a few lines of duplicate text from the lang file for input error handling and re-used the ones under `prompts.enterAccountNamePrompt`. 

Similar to the last consolidation PR, there are no observable changes in the CLI's behavior. Account name prompts for both sandboxes and developer test accounts are exactly the same as before. 

This PR does not touch `enterAccountNamePrompt` since that is more specific to CLI Config account names which are different in format.  

### Testing

- Developer test accounts can only be created through `hs project dev`. Create a public app project and set your default account to an App Developer account. Run `yarn hs project dev` and select the `<Test on a new developer test account>` option. You should be prompted to enter a name with appropriate messaging and default naming (eg: "Name your developer test account: (Developer test account 5)")
- Development sandboxes can be created two ways. 
  - The first will be through a private app project (create one if none exist). Set your default account to a STANDARD (production) account. Run `yarn hs project dev` from within the project folder and select the `<Test on a new development sandbox>` option. You should be prompted to enter a name with appropriate messaging and default naming (eg: "Name your development sandbox: (New development sandbox)")
  - The second will be through the sandbox command. Set your default account to a STANDARD (production) account. Run `yarn hs sandbox create` and select `Development Sandbox`. You should be prompted to enter a name with appropriate messaging and default naming (eg: "Name your development sandbox: (New development sandbox)")
- For standard sandboxes, it will be the same as the development sandbox step running `yarn hs sandbox create` but selecting `Standard sandbox`. You should be prompted to enter a name with appropriate messaging and default naming (eg: "Name your standard sandbox: (New standard sandbox)")

## Screenshots

No visual changes 

## Who to Notify

@shannon-lichtenwalter


camden11 (approved):
LGTM

packages/cli/lib/prompts/enterAccountNamePrompt.js
```diff
@@ -27,7 +31,53 @@ const enterAccountNamePrompt = defaultName => {
...
+          : true;
+      },
+      default: defaultName,
+    },
+  ]);
+};
+
 module.exports = {
   accountNamePrompt,
   enterAccountNamePrompt,
```
camden11, line 81: Sorry - thought of this after already approving but it i feel like it make make sense to rename these existing methods to something like `cliAccountNamePrompt` to clarify how they're different from `hubspotAccountNamePrompt`

camden11: I can handle this since this was already merged! My bad for not thinking of it sooner

adamawang: Agh merged too soon, but yes that would def clarify things. was going to say I can PR it but I can leave it to you if you'd like!

camden11: yeah np i got it!

PR #1090, opened 2024-06-06
Filter listed apps in the `hs project migrate-app` command

kemmerle:
## Description and Context

In this PR, I've added a filter to exclude listed public apps from the `migrate-app` command.

However, this may require a bit of discussion. We have two ways we could check for listed apps:

1) We could simply block customers from migrating listed apps at this time.

2) We could add a warning and a `yes/no` prompt before customers can migrate listed apps. 

Basically, it depends on how much control we'd like to give the customer at this time. I've coded both options, with option 2 being commented out. I am leaning toward option 2, given our Slack conversations about the flow we want customers to follow. @mindspin311 said that we would allow customers to migrate the app but require them to contact HubSpot before the second upload. 

UPDATE: In the cross-team sync, we decided to go with a third option: To list all the public apps in the portal and disable the choices for listed apps. @markhazlewood, I didn't have control over the styling here, so it does not match your mocks completely:

Screenshot 2024-06-06 at 8 57 15 PM

## TODO

- [x] Address initial feedback. 
- [x] Address secondary feedback for Option 3. 

## Who to Notify

@mendel-at-hubspot @markhazlewood @brandenrodgers @camden11 @joe-yeager

### DIFF
--- a/packages/cli/lib/prompts/selectPublicAppPrompt.js
+++ b/packages/cli/lib/prompts/selectPublicAppPrompt.js
@@ -14,7 +14,7 @@ const fetchPublicAppOptions = async (accountId, accountName) => {
   try {
     const publicApps = await fetchPublicAppsForPortal(accountId);
     const filteredPublicApps = publicApps.filter(
-      app => !app.projectId && !app.sourceId
+      app => !app.projectId && !app.sourceId && !app.listingInfo
     );
 
     if (!filteredPublicApps.length) {
--- a/packages/cli/lib/prompts/selectPublicAppPrompt.js
+++ b/packages/cli/lib/prompts/selectPublicAppPrompt.js
@@ -14,7 +14,7 @@ const fetchPublicAppOptions = async (accountId, accountName) => {
   try {
     const publicApps = await fetchPublicAppsForPortal(accountId);
     const filteredPublicApps = publicApps.filter(
-      app => !app.projectId && !app.sourceId && !app.listingInfo
+      app => !app.projectId && !app.sourceId
     );
 
     if (!filteredPublicApps.length) {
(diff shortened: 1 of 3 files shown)


packages/cli/lib/prompts/selectPublicAppPrompt.js
```diff
@@ -14,7 +14,7 @@ const fetchPublicAppOptions = async (accountId, accountName) => {
   try {
     const publicApps = await fetchPublicAppsForPortal(accountId);
     const filteredPublicApps = publicApps.filter(
-      app => !app.projectId && !app.sourceId
+      app => !app.projectId && !app.sourceId && !app.listingInfo
```
mindspin311, line 17: This should work for now. I'll update the app object to have a `preventProjectMigrations` in the future which should land later this month. I'll put it on the app object rather than the `listingInfo` just to future proof this such that we could prevent certain apps from being migrated regardless of marketplace-listing status.

mindspin311 (commented):
I think we said we'd prevent picking certain apps and show some "warning" rather than just hide them. So will let you update that.

Rest looks good though.

markhazlewood42:
For marketplace apps: I think we can just list those apps in a disabled state with a tag indicating why they're disabled. We can make sure documentation (external and CLI help text) explains fully why certain apps aren't able to be migrated. I believe we have disabled list items as a pattern elsewhere. Let me know if that's not the case and this isn't feasible.

image

kemmerle:
FYI: The `disabled` property for the `list` type in the `inquirer` prompt is not well documented. I had to dig through test cases to find it: https://github.com/SBoudrias/Inquirer.js/commit/3c466a2e250831e6ce5f528c227e203e75e6c31d

joe-yeager (approved):
LGTM

PR #1098, opened 2024-06-13
Refactor errors in `hs project migrate-app` command

kemmerle:
## Description and Context

@mendel-at-hubspot refactored errors returned while migrating apps. Now the error response will have an `errors` property that is an array of errors. This allows us to return multiple errors to the user. NOTE: We will still keep the older `error` property on the response object for folks on older versions of the CLI.

## Screenshots

New error handling (Doesn't appear very different with only one error returned)
Screenshot 2024-06-13 at 1 09 32 PM

## TODO

- [ ] Address any feedback

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/migrateApp.js
+++ b/packages/cli/commands/project/migrateApp.js
@@ -185,10 +185,11 @@ exports.handler = async options => {
       text: i18n(`${i18nKey}.migrationStatus.failure`),
       failColor: 'white',
     });
-    logApiErrorInstance(
-      error.error || error,
-      new ApiErrorContext({ accountId })
-    );
+    if (error.errors) {
+      error.errors.map(logApiErrorInstance);
+    } else {
+      logApiErrorInstance(error, new ApiErrorContext({ accountId }));
+    }
 
     process.exit(EXIT_CODES.ERROR);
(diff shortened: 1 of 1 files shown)


packages/cli/commands/project/migrateApp.js
```diff
@@ -185,10 +185,11 @@ exports.handler = async options => {
       text: i18n(`${i18nKey}.migrationStatus.failure`),
       failColor: 'white',
     });
-    logApiErrorInstance(
-      error.error || error,
-      new ApiErrorContext({ accountId })
-    );
+    if (error.errors) {
+      error.errors.map(logApiErrorInstance);
```
joe-yeager, line 189: Nit: Since we are ignoring the result we can probably use `forEach`

joe-yeager (approved):
LGTM, just one nit.

packages/cli/commands/project/migrateApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1098/commits/5bac1b437a2995c326ecc9b67d5f9794d9554c7a

PR #1099, opened 2024-06-13
Scope change reinstall prompt

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/589

Previously, users that had outdated scopes on an already installed public app would see the same prompt as if they hadn't installed at all, which was confusing. This adds new language specific to scope mismatches, and also allows users to _not_ reinstall the app, since scopes won't necessarily make a difference with CRM cards. (If users without the app installed decline the prompt, it exits).

## Screenshots
Screenshot 2024-06-13 at 1 46 46 PM

## Questions
Do we want to allow users to proceed anyway or should we force them to reinstall?

## Who to Notify
@brandenrodgers @kemmerle @joe-yeager @tamara-at-hubspot

### DIFF
--- a/packages/cli/lib/LocalDevManager.js
+++ b/packages/cli/lib/LocalDevManager.js
@@ -298,16 +298,20 @@ class LocalDevManager {
 
   async checkPublicAppInstallation() {
     const {
-      isInstalledWithScopeGroups: isInstalled,
+      isInstalledWithScopeGroups,
+      previouslyAuthorizedScopeGroups,
     } = await this.getActiveAppInstallationData();
 
-    if (!isInstalled) {
+    const isReinstall = previouslyAuthorizedScopeGroups.length > 0;
+
+    if (!isInstalledWithScopeGroups) {
       await installPublicAppPrompt(
         this.env,
         this.targetAccountId,
         this.activePublicAppData.clientId,
         this.activeApp.config.auth.requiredScopes,
-        this.activeApp.config.auth.redirectUrls
+        this.activeApp.config.auth.redirectUrls,
+        isReinstall
       );
     }
   }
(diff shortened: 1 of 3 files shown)


packages/cli/lib/LocalDevManager.js
```diff
@@ -298,16 +298,20 @@ class LocalDevManager {
 
   async checkPublicAppInstallation() {
     const {
-      isInstalledWithScopeGroups: isInstalled,
+      isInstalledWithScopeGroups,
+      previouslyAuthorizedScopeGroups,
     } = await this.getActiveAppInstallationData();
 
-    if (!isInstalled) {
+    const isReinstall = previouslyAuthorizedScopeGroups.length > 0;
```
brandenrodgers, line 305: Can you explain what the `previouslyAuthorizedScopeGroups` and `isInstalledWithScopeGroups` contain? I just want to make sure I'm understanding how this works now.

brandenrodgers:
Code in here LGTM, but I had one question about the new fields we're using.

packages/cli/lib/LocalDevManager.js
camden11: @brandenrodgers `previouslyAuthorizedScopeGroups` is the list of scopes you authorized the last time you installed the app. The field having a length of more than zero means that at some point you installed the app in the past. `isInstalledWithScopeGroups` is true when you've authorized the _current_ set of required scopes for the app. This can be false if the required scopes have changes since you last installed

brandenrodgers: Thanks, the logic here looks perfect then 👌

brandenrodgers (approved):

PR #1100, opened 2024-06-13
fix: Improve deploy error messages.  Make validation consistent.

joe-yeager:
## Description and Context

- Adds an alias `build` that can be used interchangeably with `buildId`
- Replicate the same `buildId` validations that existed on the prompt to the positional arguments
- Fix http client error handling that was leading to generic error messages being logged
- Add tests for `deploy.js`

#### Before
Screenshot 2024-06-17 at 1 57 46 PM
#### After
Screenshot 2024-06-17 at 1 59 56 PM

#### Before
Screenshot 2024-06-17 at 2 02 01 PM
Screenshot 2024-06-17 at 2 06 00 PM
#### After
Screenshot 2024-06-17 at 2 00 51 PM
Screenshot 2024-06-17 at 2 05 14 PM

#### Before
Screenshot 2024-06-17 at 2 03 32 PM
Screenshot 2024-06-17 at 2 08 39 PM
#### After

Screenshot 2024-06-18 at 2 16 45 PM

 
## Who to Notify

@brandenrodgers @kemmerle @camden11


packages/cli/commands/project/deploy.js
```diff
@@ -129,13 +167,14 @@ exports.builder = yargs => {
       type: 'string',
     },
     buildId: {
+      alias: ['build'],
```
joe-yeager, line 170: Add the alias `build`

packages/cli/commands/project/deploy.js
```diff
@@ -59,27 +79,46 @@ exports.handler = async options => {
...
+          )
       );
-
       buildIdToDeploy = buildIdPromptResponse.buildId;
     }
 
     if (!buildIdToDeploy) {
       logger.error(i18n(`${i18nKey}.errors.noBuildId`));
-      process.exit(EXIT_CODES.ERROR);
+      return process.exit(EXIT_CODES.ERROR);
```
joe-yeager, line 121: These `return process.exit(...)` are a hack to improve the testability of this method because these calls were killing the test process.  Since the `process.exit` method will kill the process so we don't actually `return` anything in a non test environment.  But when these methods are mocked, we are able to test only up until `process.exit` method is called.  I'm not crazy about adding code just for the sake of testing, but the other options I could come up with were to mock `process.exit` to throw an exception, which then triggered our `catch` block, or to mock `process.exit` to just return null, which meant the rest of the method just kept running.  

This felt like the lesser evil

packages/cli/lang/en.lyaml
```diff
@@ -1264,10 +1266,7 @@ en:
         bugPrompt: "Create an issue on Github in your browser?"
         generalPrompt: "Create an issue on Github in your browser?"
       buildIdPrompt:
-        enterBuildId: "[--build] Deploy which build?"
-        errors:
-          buildIdDoesNotExist: "Build {{ buildId }} does not exist for project {{ projectName }}."
-          buildAlreadyDeployed: "Build {{ buildId }} is already deployed."
+        enterBuildId: "[--buildId] Deploy which build?"
```
joe-yeager, line 1269: Changed this to `--buildId` to be consistent

packages/cli/commands/project/deploy.js
```diff
@@ -30,6 +30,26 @@ const { uiCommandReference, uiAccountDescription } = require('../../lib/ui');
...
+    });
+  }
+  if (Number(buildId) === deployedBuildId) {
+    return i18n(`${i18nKey}.errors.buildAlreadyDeployed`, {
+      buildId: buildId,
+    });
+  }
+  return true;
+};
+
```
camden11, line 52: Non blocking but could see this making sense living in `lib/projects`

packages/cli/commands/project/deploy.js
camden11: Works for me! Don't see any downside to returning when everything is exiting anyway

camden11 (approved):
Going to run through and test real quick, but code LGTM

camden11:
Screenshot 2024-06-18 at 12 40 04 PM
@joe-yeager One thing - I'm not seeing the default build argument when I test. Everything else working as expected

joe-yeager:
> Screenshot 2024-06-18 at 12 40 04 PM @joe-yeager One thing - I'm not seeing the default build argument when I test. Everything else working as expected

@camden11 I was able to reproduce this and I think this might be a pre-existing behavior.  We don't use a default value [when the latest build and the deployed build match](https://github.com/HubSpot/hubspot-cli/pull/1100/files#diff-38db40be30e01631bbf07e6415c323e75e34398fec8283b27228d3f036dbe0bcR14)

packages/cli/commands/project/deploy.js
joe-yeager: Yeah, I was wondering about that myself.  I can do it as a follow up PR if that works

camden11:
Ohh yeah i think I actually may have worked on that a while ago 🤦 . Don't mind me! Looks good to go

camden11 (approved):

PR #1101, opened 2024-06-14
Address feedback for the 6/14 bug bash

kemmerle:
## Description and Context
In this PR, I've addressed three separate issues that arose during the 6/14 bug bash:

1) I added a check for whether an app is a marketplace app when a user runs the command with the `--appId` flag:

Screenshot 2024-06-14 at 2 01 08 PM

2) I sanitized the project name when we create the default `location` option in the `createProjectPrompt`:

Screenshot 2024-06-14 at 1 55 47 PM
(This will also improve the `hs project create` command)
Screenshot 2024-06-14 at 2 01 08 PM

3) I accounted for the `REVERTED` state while polling for migration status. Now there are three ending states for migrated apps:

`SUCCESS`: The app has been succesfully migrated.
`REVERTED`: Something has gone wrong during migrated, but all the migration assocations, etc. have been successfully reverted.
`FAILURE`: Devastating error, associations were not reverted correctly, and we don't know the state of the user's app. They will need to contact a HubSpot rep for help.

Each state will return a detailed error message from the BE that we display automatically in the CLI.

## TODO

- [x] Address feedback

## Who to Notify

@mendel-at-hubspot @robrown-hubspot @brandenrodgers @camden11 @joe-yeager


packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -79,7 +100,10 @@ const createProjectPrompt = async (
       message: i18n(`${i18nKey}.enterLocation`),
       when: !promptOptions.location,
       default: answers => {
-        return path.resolve(getCwd(), answers.name || promptOptions.name);
+        const projectName = sanitizeProjectName(
```
joe-yeager, line 103: Should we apply the sanitization to the validate method as well?  Like if `sanitizeProjectName(input) !== input` we display an error letting them know it's invalid?

joe-yeager (approved):
LGTM, just a non-blocking question about validation

packages/cli/lib/prompts/createProjectPrompt.js
kemmerle: I would really love to validate the project's name, but we're trying to mimic the limits that the BE team has set on project names, and they have literally none. You can technically name a project just emojis 😭

joe-yeager: Isn't the `validate` method below for the `location` on the filesystem?

packages/cli/lang/en.lyaml
```diff
@@ -528,6 +528,7 @@ en:
             invalidAccountTypeDescription: "Only public apps created in a developer account can be converted to a project component. Select a connected developer account with {{useCommand}} or {{authCommand}} and try again."
             projectAlreadyExists: "A project with name {{ projectName }} already exists. Please choose another name."
             invalidAppId: "[--appId] Could not find appId {{ appId }}. Please choose another public app."
+            invalidMarketplaceApp: "[--appId]: Could not migrate appId {{ appId }}. Marketplace apps cannot be migrated at this time. Please choose another public app."
```
brandenrodgers, line 531: Do we need to include the `[--appId]` flag here? That's not typically a pattern that we use in our error messages. We normally only show those flag examples inside of the prompt questions themselves as ways for users to discover how to bypass the prompt next time.

Also small nit, but there's a colon in this message and there isn't one in the `invalidAppId` message

packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -16,6 +16,27 @@ const i18nKey = 'lib.prompts.createProjectPrompt';
 
 const PROJECT_PROPERTIES = ['name', 'label', 'path', 'insertPath'];
 
+// Based on the node-sanitize-filename package: https://github.com/parshap/node-sanitize-filename/blob/master/index.js
```
brandenrodgers, line 19: Thanks for adding comments here 🙏

packages/cli/lib/prompts/createProjectPrompt.js
brandenrodgers: +1 to Joe's question here. From what I can tell, this new location prompt will sanitize the initial default value that the CLI shows you, but it won't stop you from changing it to something that we don't want to allow. Does it make sense to move this over to the validation method for location?

kemmerle: Ok, some clarification:

-Sorry, @joe-yeager, I misunderstood your initial ask, I thought you were talking about project name. So we can't simply check with `sanitizeProjectName(input) !== input`. The customer can put *anything* into the location field, we just supply the default value. We would need to perform two separate functions: one to sanitize the project name for the default location value, and one to validate the final location that the user provides.  
-To answer your second question, @brandenrodgers, we need to sanitize the project name to create the `default` location; validating the final location the user provides is a separate issue.

packages/cli/lang/en.lyaml
kemmerle: I can remove this in a separate PR--will put it out Monday.

PR #1102, opened 2024-06-17
Add location flag validation in `hs project migrate-app` (Sister PR in local-dev-lib)

kemmerle:
## Description and Context

This PR does two things:

1) It moves the functions `sanitizeFileName` and `isValidPath` to the hubspot-local-dev-lib library. I think this makes more sense, since these are reusable for any file or path. 

2) It adds validation for the path provided in the `location` flag to ensure that users can't input invalid characters in different file systems. 

There is a sister [PR](https://github.com/HubSpot/hubspot-local-dev-lib/pull/150) in local-dev-lib. 

To test, you'll need to follow the `yarn link` instructions in the README file to use a symlinked version of local-dev-lib. 

## Screenshots

We sanitize the project name before inserting it in the default path, and we validate for invalid characters/reserved names for the location. 
Screenshot 2024-06-17 at 1 31 44 PM

## TODO

- [x] Address feedback

## Who to Notify

### DIFF
--- a/packages/cli/lib/prompts/createProjectPrompt.js
+++ b/packages/cli/lib/prompts/createProjectPrompt.js
@@ -1,6 +1,10 @@
 const fs = require('fs');
 const path = require('path');
-const { getCwd } = require('@hubspot/local-dev-lib/path');
+const {
+  getCwd,
+  sanitizeFileName,
+  isValidPath,
+} = require('@hubspot/local-dev-lib/path');
 const { PROJECT_COMPONENT_TYPES } = require('../../lib/constants');
 const { promptUser } = require('./promptUtils');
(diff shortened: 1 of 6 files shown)


packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -1,6 +1,10 @@
 const fs = require('fs');
 const path = require('path');
-const { getCwd } = require('@hubspot/local-dev-lib/path');
+const {
+  getCwd,
+  sanitizeFileName,
```
kemmerle, line 5: I decided to rename this function when moving it to local-dev-lib, because it makes it more readable and reusable in the future. We're basically treating the project name as a directory/file name.

brandenrodgers (approved):

PR #1104, opened 2024-06-18
Generate private app tokens if possible for private app localdev

wung-hubspot:
## Description and Context

Private app user tokens are temporary Private App tokens scoped to the portal and user of the Personal Access Key. Using this user can avoid grabbing the real Private app token that is used for in the serverless instances.

Add usage to PrivateApplicationUserTokenManager and pass the getToken function to local dev server. 

## Scre
Screenshot 2024-06-18 at 11 17 12 AM

### DIFF
--- a/packages/cli/lib/LocalDevManager.js
+++ b/packages/cli/lib/LocalDevManager.js
@@ -15,6 +15,9 @@ const {
   getAccountId,
   getConfigDefaultAccount,
 } = require('@hubspot/local-dev-lib/config');
+const {
+  PrivateAppUserTokenManager,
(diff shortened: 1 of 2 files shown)


packages/cli/lib/LocalDevManager.js
```diff
@@ -502,10 +516,17 @@ class LocalDevManager {
 
   async devServerStart() {
     try {
-      await DevServerManager.start({
+      var args = {
```
joe-yeager, line 519: This `var` can be a `const`

kemmerle (approved):
Tentatively approving this PR, but I do agree with Branden that the corresponding local-dev-lib PR needs to be moved before this can be merged.

PR #1105, opened 2024-06-20
Unified Dev Server

jsines:
## Description and Context

Adds unified dev server dependency behind a gate, plan is to have both dev servers in this release and then once we ungate 100% we'll remove the other one. Removes a few command args that either aren't implemented / won't get used going forward.


tenas97 (approved):

TanyaScales (approved):

commands/theme/preview.ts
```diff
@@ -207,12 +239,6 @@ exports.builder = yargs => {
     type: 'string',
     requiresArg: true,
   });
-  yargs.option('notify', {
```
brandenrodgers, line 210: Technically removing this makes it a breaking change. We're queuing up a major release in the `next` branch. If you merge this into that branch it should make getting this out alongside everything else nice and easy.

brandenrodgers (approved):

PR #1106, opened 2024-06-20
fix: Update error handling from request to axios

joe-yeager:
## Description and Context

When the migration from `cli-lib` to `local-dev-lib` occurred, the underlying http library was changed from `request` to `axios` so these error handling situations need to be updated to the `axios` way of checking for status code.

## Who to Notify

@brandenrodgers @camden11 @kemmerle

### DIFF
--- a/packages/cli/commands/functions/deploy.js
+++ b/packages/cli/commands/functions/deploy.js
@@ -75,13 +75,13 @@ exports.handler = async options => {
     );
   } catch (e) {
     spinner && spinner.stop && spinner.stop();
-    if (e.statusCode === 404) {
+    if (e.response && e.response.status === 404) {
       logger.error(
         i18n(`${i18nKey}.errors.noPackageJson`, {
           functionPath,
         })
       );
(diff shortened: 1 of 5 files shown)


packages/cli/commands/functions/deploy.js
```diff
@@ -75,13 +75,13 @@ exports.handler = async options => {
...
+    if (e.response && e.response.status === 404) {
       logger.error(
         i18n(`${i18nKey}.errors.noPackageJson`, {
           functionPath,
         })
       );
-    } else if (e.statusCode === 400) {
+    } else if (e.response && e.response.status === 400) {
       logger.error(e.error.message);
     } else if (e.status === 'ERROR') {
```
joe-yeager, line 86: Whoops, missed this case.  I'll try to figure out what this needs to be

joe-yeager: Actually, this looks like a custom case and not related to `request`

camden11 (approved):
LGTM! Didn't realize there were more of these kicking around

PR #1107, opened 2024-06-20
Clean up package.json files

camden11:
## Description and Context
I was poking around and noticed we had some references to what I presume is the old name of this repo. Cleaned those up. Also wasn't sure what to use as the name for the monorepo package.json so went with `hubspot-local-dev-tools`. We could also just call it `hubspot-cli` too. It's not a published package so I don't think it matters a ton.

## Who to Notify
@joe-yeager @kemmerle @brandenrodgers

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "repository": {
     "type": "git",
-    "url": "https://github.com/HubSpot/hubspot-cli"
+    "url": "https://github.com/HubSpot/hubspot-cms-tools"
   },
   "dependencies": {
     "@hubspot/local-dev-lib": "1.6.0",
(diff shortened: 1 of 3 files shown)


packages/cli/package.json
```diff
@@ -40,8 +40,7 @@
     "node": ">=16"
   },
   "bin": {
-    "hs": "./bin/hs",
-    "hscms": "./bin/hscms"
```
joe-yeager, line 44: Do `CMS` devs not use this?  Maybe we could delete the `bin/hscms` file as well.  It seems to be identical to `bin/hs` except it doesn't have the `--no-deprecation` flag.

camden11: Ok actually fully did not realize that running commands with `hscms` was something you could still do - thought this was left over from an older setup.

We probably _should_ remove that functionality, but it would be a whole thing and require a major version bump, so I'll leave it out of this PR

joe-yeager (approved):

brandenrodgers (approved):

PR #1109, opened 2024-06-24
chore: Update CLI to utilize new error object.

joe-yeager:
## Description and Context

Depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/159

- Update the CLI to use the new util methods and error object
- Simplify the logging down to a single method

## Who to Notify

@brandenrodgers @camden11 @kemmerle


camden11 (approved):
Looks like this still needs a LDL version bump, but otherwise LGTM

kemmerle (approved):

packages/cli/lib/errorHandlers/apiErrors.js
```diff
@@ -3,7 +3,7 @@ const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
 const {
   isMissingScopeError,
   isApiUploadValidationError,
-  getAxiosErrorWithContext,
+  getHubSpotHttpErrorWithContext,
```
brandenrodgers, line 6: I know this is constantly in flux, but the current version of this in local-dev-lib is `getUserFriendlyHttpErrorMessage`

packages/cli/lib/errorHandlers/apiErrors.js
```diff
@@ -53,9 +56,9 @@ function logValidationErrors(error, context) {
  * @param {ApiErrorContext}          context
  */
 function logApiErrorInstance(error, context) {
-  if (error.isAxiosError) {
+  if (isHubSpotHttpError(error)) {
     if (overrideErrors(error)) return;
-    const errorWithContext = getAxiosErrorWithContext(error, context);
+    const errorWithContext = getHubSpotHttpErrorWithContext(error, context);
     logger.error(errorWithContext.message);
```
brandenrodgers, line 62: Just noting b/c I ran into this during my testing, `getUserFriendlyHttpErrorMessage` returns the error message so there's no need to use `.message` here

joe-yeager:
I bumped this back to a draft.  I'm going to update this to use all of the changes from the final LDL refactor PR over [here](https://github.com/HubSpot/hubspot-local-dev-lib/pull/159)

packages/cli/lib/errorHandlers/index.js
```diff
@@ -0,0 +1,86 @@
...
+  isSystemError,
+  isFileSystemError,
+} = require('@hubspot/local-dev-lib/errors/index');
+const { shouldSuppressError } = require('./suppressError');
+const { i18n } = require('../lang');
+const util = require('node:util');
+
+const i18nKey = 'lib.errorHandlers.index';
+
+function logError(error, context = {}) {
```
joe-yeager, line 13: This is now the singular `logError` method that will handle the logging.  So we can call this and let it determine how to log the output.

packages/cli/lib/errorHandlers/index.js
```diff
@@ -0,0 +1,86 @@
...
+  debugError(error, context);
+
+  if (shouldSuppressError(error)) {
+    return;
+  }
+
+  if (isHubSpotHttpError(error) || isFileSystemError(error)) {
+    logger.error(error.toString());
```
joe-yeager, line 21: The `HubSpotHttpError` class has a `toString` method that [returns a friendly string representation of the contents](https://github.com/HubSpot/hubspot-local-dev-lib/pull/159/files#diff-ae25cee36b980aa393de912585da6d585d226617b9595f4a65e6e3215c2a1925R65).

I am add a similar method for the `FileSystemError` class so this method can be naive about how to format the messages for that class.

PR #1112, opened 2024-07-11
Sandbox command updates

adamawang:
## Description and Context

Related issue: https://git.hubteam.com/HubSpot/sandboxes-ui/issues/2455

This PR updates the `hs sandbox create` command flow with the following changes:

- Removal of a prefilled default sandbox name (users must decide on a name which will be far more unique than the default)
- User is no longer prompted to sync assets, that is now mandatory and run in the background. 
- User is only prompted to sync contact records for standard sandboxes, development sandboxes no longer have that option. This means creating a new dev sandbox through `hs project dev` also no longer has a contact records prompt. 
- Removal of the multi-progress bar, we now just show a link to the UI for sync progress. 
- Content changes to be more consistent with guidelines and tone
- Deprecation tag and warning message (currently only used for `hs sandbox sync`)
- Error tag and disabled command message (currently only used for `hs sandbox sync`)

Launch entry: https://tools.hubteam.com/launch/feature-rollouts/128885/info
Deprecation plan is in the works with DevRel here: https://hubspot.slack.com/archives/C03HPEU48L8/p1721246518840469

## Screenshots

Development sandbox create
Screenshot 2024-07-18 at 11 59 17

Standard sandbox create
Screenshot 2024-07-18 at 11 59 21

`hs project dev` development sandbox create
Screenshot 2024-07-18 at 12 01 20

Deprecation warning tag and description
Screenshot 2024-07-18 at 13 46 13
Screenshot 2024-07-18 at 13 45 59

Disabled error that fires when an API 404's. Not sure what the best way to display this is, I only put it for one command after the endpoint 404's for `hs sandbox sync`
Screenshot 2024-07-18 at 14 53 38

## TODO

For the last screenshot, I'm not too sure how best to handle these "API disabled" situations where an endpoint finally gets sunset. This "error" banner pops up in between a bunch of other messages so it feels a bit weird. @sandracoleto


sandracoleto:
minor request @adamawang in hs sandbox sync, could we remove the lines above and bellow the deprecated message so it doesn't look like an alert. figma below
image

adamawang:
@sandracoleto 
Screenshot 2024-07-18 at 13 56 41

packages/cli/lib/sandboxSync.js
```diff
@@ -157,53 +154,39 @@ const syncSandbox = async ({
...
+        {
+          url: uiLink(
+            i18n(`${i18nKey}.info.syncStatusDetailsLinkText`),
+            syncStatusUrl
+          ),
+        }
+      ),
+      true
+    );
+    uiLine();
```
adamawang, line 188: Replaced polling UI with a short info message

packages/cli/commands/sandbox/create.js
```diff
@@ -124,34 +117,19 @@ exports.handler = async options => {
...
-          sandboxName,
-        }),
-      },
-    ]);
-    sandboxSyncPromptResult = sandboxSyncPrompt;
-    // We can prompt for contact records before fetching types since we're starting with a fresh sandbox in create
-    if (sandboxSyncPrompt) {
+
+    // Prompt to sync contact records for standard sandboxes only
+    if (!isDevelopmentSandbox) {
```
brandenrodgers, line 127: Would it help to future proof this if we inverted this logic? I'm thinking about the scenario where we add new sandbox types way down the road (may not actually happen, but maybe still be worth guarding against?).

packages/cli/lang/en.lyaml
```diff
@@ -1046,6 +1046,14 @@ en:
...
         header: "{{#yellow}}***************************** WARNING ****************************{{/yellow}}"
         footer: "{{#yellow}}******************************************************************{{/yellow}}"
+      infoTag: "{{#bold}}[INFO]{{/bold}}"
+      deprecatedTag: "{{#bold}}[DEPRECATED]{{/bold}}"
+      errorTag: "{{#bold}}[ERROR]{{/bold}}"
+      deprecatedMessage: "The `{{#orange}}{{ command }}{{/orange}}` command is deprecated and will be disabled soon. {{ url }}"
+      deprecatedDescription: "{{ message }}. The `{{#orange}}{{ command }}{{/orange}}` command is deprecated and will be disabled soon. {{ url }}"
+      deprecatedUrlText: 'Learn more.'
+      disabledMessage: "The `{{#orange}}{{ command }}{{/orange}}` command is disabled. Run `{{#orange}}npm i -g @hubspot/cli@latest{{/orange}}` to update to the latest HubSpot CLI version. {{ url }}"
+      disabledUrlText: "See all HubSpot CLI commands here."
```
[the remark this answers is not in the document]
brandenrodgers: I like keeping all of this here 👌 

Just a note though, the error and info tags should already be natively supported by the logger [here](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/lib/logger.ts#L34). You can use them via `logger.error()` and `logger.info()`. This is a pattern we probably want to keep because there are some nuances to the behavior of logging via `.log()` vs. `.error()`. Logging as an error will log out via process.stderr, and will respect the console log level visibility controls

packages/cli/lib/ui/index.js
```diff
@@ -139,13 +139,114 @@ const uiBetaTag = (message, log = true) => {
...
+  });
+};
+
+const uiDeprecatedMessage = (command, url = undefined, message = undefined) => {
+  const i18nKey = 'lib.ui';
+
+  const tag =
+    message ||
+    i18n(`${i18nKey}.deprecatedMessage`, {
+      command,
```
brandenrodgers, line 228: We might want to leverage the `uiCommandReference` util when logging out the command here, just for consistency

packages/cli/lib/sandboxSync.js
```diff
@@ -157,53 +154,39 @@ const syncSandbox = async ({
           account: uiAccountDescription(accountId),
         })
       );
+    } else if (
+      isSpecifiedError(err, {
+        statusCode: 404,
+      })
+    ) {
+      uiCommandDisabledBanner(
```
brandenrodgers, line 162: Just to clarify, this won't be throwing 404's yet right? This is just prepping the CLI to be able to gracefully handle them when it does?

packages/cli/lib/ui/index.js
```diff
@@ -139,13 +139,114 @@ const uiBetaTag = (message, log = true) => {
...
+  uiLine();
+  uiErrorTag(tag, true);
+  uiLine();
+  logger.log();
+};
+
+const uiDeprecatedDescription = (message, command, url = undefined) => {
+  const i18nKey = 'lib.ui';
+
+  return i18n(`${i18nKey}.deprecatedDescription`, {
```
brandenrodgers, line 215: If you wanted, this could probably also return `uiDeprecatedTag` similar to how uiDeprecatedMessage does it. Just to save a little on LOC when calling

brandenrodgers:
Not a blocker, but at some point we might want to revisit the coloring that we're using for various things in the CLI. For example, our deprecation banner and our warnings kind of blend together since they're the same color now:
image

I wonder if we want to introduce some new colors from the HubSpot pallet (like purple, green, blue)

packages/cli/lib/sandboxSync.js
adamawang: Yep exactly. The endpoint will end up sticking around until we go GA with v2 sandboxes which will be quite a while from now. The command should be removed long before the endpoint gets decommissioned so this may never get to run

packages/cli/lib/ui/index.js
adamawang: Ahh nice, didn't realize this was here!

adamawang:
Yeah good point about the colors... let me check in with sandra and see if we can update that here

kemmerle (approved):
Reviewed the code and tested this out locally, and everything is looking good to me. I think this is good to be merged, after you receive feedback from Sandra about the color of the deprecation tags and where we want to place deprecation warnings (I agree, the placement is a bit awkward right now).

adamawang:
Sandra is going to spend some time coming up with a more permanent solution for colors so it won't make it into this PR - will merge this one in and we can track the color work in this issue here: https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/607

PR #1114, opened 2024-07-19
Fix bug app migration restrictions

kemmerle:
## Description and Context

I was incorrectly determining how to restrict public apps from being migrated. I was only using the `preventProjectMigration` property to determine whether to allow migration, when I should have been using the `preventProjectMigration` property in tandem with the `listingInfo` property. 

The `listingInfo` property exists for apps that are currently posted to the marketplace.
The `preventProjectMigrations` property returns `true` if the apps team has not added this app to their allowlist config. This extra buffer is in place to allow us to review apps before they get migrated and posted to the marketplace.

## Testing

You can JITA into my test account (Will provide details on Slack) to test out the PR. 

## TODO

- [x] Address any feedback
- [x] TEST

## Who to Notify

@brandenrodgers @camden11 @robrown-hubspot

### DIFF
--- a/packages/cli/lib/prompts/selectPublicAppPrompt.js
+++ b/packages/cli/lib/prompts/selectPublicAppPrompt.js
@@ -10,7 +10,11 @@ const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 
 const i18nKey = 'lib.prompts.selectPublicAppPrompt';
 
-const fetchPublicAppOptions = async (accountId, accountName, migrateApp) => {
+const fetchPublicAppOptions = async (
+  accountId,
+  accountName,
+  migrateApp = false
+) => {
   try {
     const publicApps = await fetchPublicAppsForPortal(accountId);
     const filteredPublicApps = publicApps.filter(
@@ -20,7 +24,9 @@ const fetchPublicAppOptions = async (accountId, accountName, migrateApp) => {
     if (
       !filteredPublicApps.length ||
       (migrateApp &&
-        !filteredPublicApps.find(app => !app.preventProjectMigrations))
+        !filteredPublicApps.find(
+          app => !app.preventProjectMigrations || !app.listingInfo
+        ))
     ) {
       const headerTranslationKey = migrateApp
         ? 'noAppsMigration'
@@ -52,7 +58,7 @@ const selectPublicAppPrompt = async ({
   const publicApps = await fetchPublicAppOptions(
     accountId,
     accountName,
-    (migrateApp = false)
+    migrateApp
   );
   const translationKey = migrateApp ? 'selectAppIdMigrate' : 'selectAppIdClone';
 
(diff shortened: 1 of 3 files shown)


packages/cli/lib/prompts/selectPublicAppPrompt.js
```diff
@@ -20,7 +24,9 @@ const fetchPublicAppOptions = async (accountId, accountName, migrateApp) => {
     if (
       !filteredPublicApps.length ||
       (migrateApp &&
-        !filteredPublicApps.find(app => !app.preventProjectMigrations))
+        !filteredPublicApps.find(
```
brandenrodgers, line 27: What exactly is this chunk of logic doing? It it confirming that there are valid items in the list? If so we may want to use `array.some()` here

packages/cli/lib/prompts/selectPublicAppPrompt.js
```diff
@@ -64,7 +70,8 @@ const selectPublicAppPrompt = async ({
       }),
       type: 'list',
       choices: publicApps.map(app => {
-        if (migrateApp && app.preventProjectMigrations) {
+        const { preventProjectMigrations, listingInfo } = app;
+        if (migrateApp && preventProjectMigrations && listingInfo) {
```
brandenrodgers, line 74: Do we need the `migrateApp` argument because you're also using this prompt for the clone app flow?

kemmerle: Yes, exactly.

packages/cli/lib/prompts/selectPublicAppPrompt.js
kemmerle: Yes, it's confirming there are valid items in the list, in response to your feedback that we shouldn't display a dropdown menu where everything is disabled. I can change to `some`.

packages/cli/lib/prompts/selectPublicAppPrompt.js
brandenrodgers: Would you mind renaming it to something along the lines of `isMigratingApp`? Just so the logic is a little easier to follow

kemmerle: No prob!

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1114/commits/575fa8706bfc21cc7c13a98b76aee15e4fb411f3

packages/cli/lib/prompts/selectPublicAppPrompt.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1114/commits/575fa8706bfc21cc7c13a98b76aee15e4fb411f3

brandenrodgers (approved):

PR #1115, opened 2024-07-22
Refactor: Create project directory when cloning app

kemmerle:
## Description and Context
@robrown-hubspot had a chance to test out the `hs project clone-app` on Friday afternoon, and he clarified that when cloning the app files, we also need to create a project directory with an `hsproject.json` file. This PR implements those changes. 

## Screenshots

Now we prompt for a project name:
Screenshot 2024-07-22 at 11 31 06 AM

And return the app files in the correct project folder structure:

Screenshot 2024-07-22 at 11 31 27 AM

## TODO

- [ ] Address feedback

## Who to Notify

@robrown-hubspot @brandenrodgers @camden11 @joe-yeager


packages/cli/commands/project/cloneApp.js
```diff
@@ -101,15 +98,26 @@ exports.handler = async options => {
...
-        { includesRootDir: true, hideLogs: true }
-      );
+      // Ensure correct project folder structure exists
+      const baseDestPath = path.resolve(getCwd(), location);
+      const absoluteDestPath = path.resolve(baseDestPath, 'src', 'app');
+      fs.mkdirSync(absoluteDestPath, { recursive: true });
+
+      // Extract zipped app files and place them in correct directory
+      const zippedApp = await downloadClonedProject(accountId, exportId);
+      await extractZipArchive(zippedApp, name, path.resolve(absoluteDestPath), {
```
brandenrodgers, line 108: Do you need `path.resolve(absoluteDestPath)` here or could you just use `absoluteDestPath `?

packages/cli/commands/project/cloneApp.js
```diff
@@ -101,15 +98,26 @@ exports.handler = async options => {
...
+      });
+
+      // Create hsproject.json file
+      const configPath = path.join(baseDestPath, 'hsproject.json');
+      const configContent = {
+        name,
+        srcDir: 'src',
+        platformVersion: '2023.2',
+      };
+      writeProjectConfig(configPath, configContent);
```
brandenrodgers, line 120: You may want to use separate try/catch blocks for these file writes rather than doing all of this logic inside of the same block. Right now the logging util that you're calling in the catch block is `logApiErrorInstance`, but there are a lot of other non-api utils in there that could throw errors (i.e. `extractZipArchive`, `fs.mkdirSync`, and `writeProjectConfig`).

Note: I just noticed that `writeProjectConfig` will never throw, but it also will never tell us that the write failed. We may want to revisit the shape of that util since it's not very informative. Ideally we could know if the config was successfully created or not

kemmerle: I can definitely rewrite `writeProjectConfig` with a try...catch block. I don't think we need separate track...catch blocks here, since `logApiErrorInstance` includes logic, where it falls back to `logErrorInstance` for non-api errors. That saves us a fair bit of code we need to maintain.

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1115/commits/43079c7c4627f466ab6fd739ceb2c418a72d62b1

packages/cli/commands/project/cloneApp.js
kemmerle: Partly addressed in https://github.com/HubSpot/hubspot-cli/pull/1115/commits/43079c7c4627f466ab6fd739ceb2c418a72d62b1. My question is: How do you want to confirm that the write failed? Currently, we do have a `logger.debug` statement in the try block that states when the write is successful. I would assume that the message would not log otherwise, so to me, that gives enough info. But is there something I'm missing?

brandenrodgers: > How do you want to confirm that the write failed?

fwiw I don't think you need to address it in this PR. I just noticed it while poking around and figured I'd mention it. It's not a blocker, but the `writeProjectConfig` doesn't return or throw any sort of errors when the write fails. We log a debug message, but there's currently no way for us to programmatically know whether the write succeeded or failed. For example, what happens if the write fails? Our code will continue to execute as if it had succeeded because the util doesn't tell us that it failed.

packages/cli/lib/projects.js
```diff
@@ -60,7 +61,7 @@ const writeProjectConfig = (configPath, config) => {
     fs.writeFileSync(configPath, JSON.stringify(config, null, 2));
     logger.debug(`Wrote project config at ${configPath}`);
   } catch (e) {
-    logger.error(`Could not write project config at ${configPath}`);
+    logErrorInstance(e);
```
brandenrodgers, line 64: hmm, it might be nice to log out the nice-to-read message by default. I would actually be ok with keeping this error as it was

packages/cli/commands/project/cloneApp.js
kemmerle: > the writeProjectConfig doesn't return or throw any sort of errors when the write fails.

I'm a little confused about this, because I changed the `catch` block of the `writeProjectConfig` to add a `logErrorInstance` statement specifically in case of system errors. That will return a more detailed message about why the write failed. But in your next comment, you write that I can remove that. 

Wouldn't it be better to add a `process.exit` in the `catch` block of `writeProjectConfig`, in case the config write fails? We can do this without breaking anything, because `writeProjectConfig` isn't currently being used elsewhere in the CLI.

brandenrodgers: I think there are a few layers to the misalignment here. There are two concepts. The first is the error message that we surface to the user when this `writeProjectConfig` util fails. Ideally this message should tell the user that we were not able to write to their config file. The second piece of this is that it would be nice for us to be able to programmatically determine whether or not the write failed. This part is separate from the user-facing message.

> I'm a little confused about this, because I changed the catch block of the writeProjectConfig to add a logErrorInstance statement specifically in case of system errors. That will return a more detailed message about why the write failed. But in your next comment, you write that I can remove that.

You are correct that it will return a more detailed error message about which specific part of `writeProjectConfig` failed, but that information might not be as useful to the end user as the former version of the log message was. The former log message was very clear, because it told the user that that the write to the config failed. This new version that uses `logErrorInstance()` will give a very specific error message that might not mean anything to the user.

For example, this is the error that I get when I try to call `fs.writeFileSync()` on a file that doesn't exist:
`ENOENT: no such file or directory, open '/brodgers/my-project'`

IMO that ^ error is not as informative to the user as `[Error] Could not write project config at /brodgers/my-project`

---
> Wouldn't it be better to add a process.exit in the catch block of writeProjectConfig, in case the config write fails? We can do this without breaking anything, because writeProjectConfig isn't currently being used elsewhere in the CLI.

You're on the right track, but I don't think that exiting the process from this util function is the right solution because a util should not determine whether the process exits or not. Consider the scenario where a command wants to do some cleanup or log some special messages before exiting. If we exit from a util, the command would never get that chance.

This is where I think the confusion is coming from. Regardless of everything I said in the top section of this comment, I actually don't think this `writeProjectConfig` function should log anything at all (note: maybe it could leverage `logger.debug()` just to help us debug if we ever need to). I think there are two reasonable options for how this util could handle failures:

### Option 1: Return a boolean
```js
const writeProjectConfig = () => {
  let success = true;
  try {
    // logic
  } catch(e) {
    logger.debug(e)
    success = false;
  }
  return success;
}

const someOtherFunction = () => {
  const success = writeProjectConfig()
  if (!success) {
    logger.error(`failed to write project config ${configPath}`);
  }
}
```
#### Option 2: Throw an error
```js
const writeProjectConfig = () => {
  try {
    // logic
  } catch(e) {
    logger.debug(e)
    throw new Error(`failed to write project config ${configPath}`)
  }
}

const someOtherFunction = () => {
  try {
    writeProjectConfig()
  } catch(e) {
    logger.error(e)
  }
}
```

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1115/commits/4045486aa0478e4b985436849d5418aba6119cc3

packages/cli/lib/projects.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1115/commits/4045486aa0478e4b985436849d5418aba6119cc3

packages/cli/lib/projects.js
```diff
@@ -55,13 +55,16 @@ const SPINNER_STATUS = {
...
 const writeProjectConfig = (configPath, config) => {
+  let success = true;
   try {
     fs.ensureFileSync(configPath);
     fs.writeFileSync(configPath, JSON.stringify(config, null, 2));
     logger.debug(`Wrote project config at ${configPath}`);
   } catch (e) {
-    logger.error(`Could not write project config at ${configPath}`);
+    logger.debug(e);
+    success = false;
```
joe-yeager, line 65: Nit: we could `return false` here and `return  true` on line 67 and eliminate the `success` variable

packages/cli/commands/project/cloneApp.js
```diff
@@ -101,20 +98,37 @@ exports.handler = async options => {
...
+
+      // Extract zipped app files and place them in correct directory
+      const zippedApp = await downloadClonedProject(accountId, exportId);
+      await extractZipArchive(zippedApp, name, absoluteDestPath, {
+        includesRootDir: true,
+        hideLogs: true,
+      });
+
+      // Create hsproject.json file
+      const configPath = path.join(baseDestPath, 'hsproject.json');
```
joe-yeager, line 114: Nit: I believe we have a constant for `hsproject.json`

joe-yeager (approved):
LGTM, just a couple of nits

packages/cli/lib/projects.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1115/commits/1031c18d71632dead15b6082e22b6366be615501

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1115/commits/1031c18d71632dead15b6082e22b6366be615501

PR #1116, opened 2024-07-22
Fix project commands declaration

brandenrodgers:
## Description and Context

We were not following the pattern defined in our other commands. We define subcommands using the pattern `yargs.command(subcommand).demandCommand(1, '')`.

The result of this is a more useful CLI output when the user runs `hs project` without specifying any subcommands. The empty string is necessary because the `demandCommand()` function requires two arguments. Using an empty string forces it to fall back to logging out the `--help` output.

### Before
image

### After
image

### DIFF
--- a/packages/cli/commands/project.js
+++ b/packages/cli/commands/project.js
@@ -23,19 +23,20 @@ exports.builder = yargs => {
   addConfigOptions(yargs);
   addAccountOptions(yargs);
 
-  // TODO: deploy must be updated
-  yargs.command(create).demandCommand(0, '');
-  yargs.command(add).demandCommand(0, '');
-  yargs.command(watch).demandCommand(0, '');
-  yargs.command(dev).demandCommand(0, '');
-  yargs.command(upload).demandCommand(0, '');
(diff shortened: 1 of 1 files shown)


packages/cli/commands/project.js
```diff
@@ -23,19 +23,20 @@ exports.builder = yargs => {
...
-  yargs.command(dev).demandCommand(0, '');
-  yargs.command(upload).demandCommand(0, '');
-  yargs.command(deploy).demandCommand(1, '');
-  yargs.command(logs).demandCommand(1, '');
-  yargs.command(listBuilds).demandCommand(0, '');
-  yargs.command(download).demandCommand(0, '');
-  yargs.command(open).demandCommand(0, '');
-  yargs.command(migrateApp).demandCommand(0, '');
-  yargs.command(cloneApp).demandCommand(0, '');
+  yargs
```
joe-yeager, line 26: 🎉

joe-yeager (approved):

camden11 (approved):
huge

PR #1117, opened 2024-07-24
fix: Remove default exports

joe-yeager:
## Description and Context

Depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/165/files

## Who to Notify

@camden11 @brandenrodgers @kemmerle


kemmerle (approved):
🚢

PR #1123, opened 2024-08-05
Refactor(config): Sister PR in local-dev-lib

kemmerle:
## Description and Context
We are updating our HubSpot CLI configuration file, so that it will be created the root directory in a hidden folder: `~/.hubspot/config.yml`. This will encourage users to take advantage of the `init`, `auth`, and `accounts` commands to update their config and help prevent config-related bugs. 

We will also be supporting the old configuration, which means that power users will still be able to use nested configs for different projects.

Sister PR: https://github.com/HubSpot/hubspot-local-dev-lib/pull/172

## QUESTIONS:

For Sandra, how would we like to alert customers about the new config? Currently, I have a prompt that asks them whether they'd like to use the new vs. old, but I'm not sure that's the best solution.

## Screenshots

`hs init` (with new config):

Screenshot 2024-08-05 at 11 34 19 AM

`hs init` (Erroring out because config exists):

This isn't the greatest, because the prompt for new vs old config still appears.

Screenshot 2024-08-05 at 11 39 57 AM

`hs auth` (Adding another account):

Screenshot 2024-08-05 at 11 34 59 AM

## TODO

- [ ] Need to fix minor bug where choose default account prompt still appears when re-authing a default account.
- [ ] Need to check for both old and new config paths to prevent users from using both config types. 

## Who to Notify

### DIFF
--- a/packages/cli/commands/auth.js
+++ b/packages/cli/commands/auth.js
@@ -76,14 +76,14 @@ exports.handler = async options => {
   setLogLevel(options);
   logDebugInfo(options);
 
+  const env = qa ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
+  loadConfig(configPath);
+  checkAndWarnGitInclusion(getConfigPath());
+
   if (!getConfigPath()) {
     logger.error(i18n(`${i18nKey}.errors.noConfigFileFound`));
     process.exit(EXIT_CODES.ERROR);
   }
-
-  const env = qa ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
-  loadConfig(configPath);
-  checkAndWarnGitInclusion(getConfigPath());
 
   trackCommandUsage('auth');
   trackAuthAction('auth', authType, TRACKING_STATUS.STARTED);
--- a/packages/cli/commands/init.js
+++ b/packages/cli/commands/init.js
@@ -1,10 +1,10 @@
-const path = require('path');
 const fs = require('fs-extra');
 const {
   getConfigPath,
   createEmptyConfigFile,
   deleteEmptyConfigFile,
   updateDefaultAccount,
+  loadConfig,
 } = require('@hubspot/local-dev-lib/config');
 const { addConfigOptions } = require('../lib/commonOpts');
 const { handleExit } = require('../lib/process');
@@ -31,7 +31,6 @@ const {
   getAccessToken,
   updateConfigWithAccessToken,
 } = require('@hubspot/local-dev-lib/personalAccessKey');
-const { getCwd } = require('@hubspot/local-dev-lib/path');
 const { toKebabCase } = require('@hubspot/local-dev-lib/text');
 const { trackCommandUsage, trackAuthAction } = require('../lib/usageTracking');
 const { setLogLevel, addTestingOptions } = require('../lib/commonOpts');
@@ -105,10 +104,14 @@ exports.describe = i18n(`${i18nKey}.describe`, {
 exports.handler = async options => {
   const {
     auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
-    c,
     account: optionalAccount,
   } = options;
-  const configPath = (c && path.join(getCwd(), c)) || getConfigPath();
+  const { useNewConfig } = await promptUser({
+    name: 'useNewConfig',
+    type: 'confirm',
+    message: i18n(`${i18nKey}.useNewConfig`),
+  });
+  const configPath = getConfigPath(useNewConfig);
   setLogLevel(options);
   logDebugInfo(options);
   trackCommandUsage('init', {
@@ -127,7 +130,8 @@ exports.handler = async options => {
   }
 
   trackAuthAction('init', authType, TRACKING_STATUS.STARTED);
-  createEmptyConfigFile({ path: configPath });
+  createEmptyConfigFile({ path: configPath }, useNewConfig);
+  loadConfig(configPath, options);
   handleExit(deleteEmptyConfigFile);
 
   try {
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -415,6 +415,7 @@ en:
         configFileUpdated: "Connected account \"{{ account }}\" using \"{{ authType }}\" and set it as the default account"
       logs:
         updateConfig: "To update an existing config file, use the \"hs auth\" command."
+      useNewConfig: "Would you like to use the new version of the HubSpot config file? It will be located in the root directory"
       errors:
         configFileExists: "The config file {{ configPath }} already exists."
     lint:
--- a/packages/cli/lib/errorHandlers/standardErrors.js
+++ b/packages/cli/lib/errorHandlers/standardErrors.js
@@ -51,6 +51,7 @@ function debugErrorAndContext(error, context) {
       context: util.inspect(context, false, null, true),
     })
   );
+  logger.debug(error.stack);
 }
 
 /**


packages/cli/commands/auth.js
```diff
@@ -76,15 +76,15 @@ exports.handler = async options => {
   setLogLevel(options);
   logDebugInfo(options);
 
+  const env = qa ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
+  loadConfig(configPath);
+  checkAndWarnGitInclusion(getConfigPath());
+
   if (!getConfigPath()) {
```
camden11, line 83: Curious why this was moved. Does it still makes sense to check for a config path after trying to load the config?

kemmerle: The `loadConfig` function calls `CLIConfiguration.init()`, which then sets `CLIConfiguration.isActive()`, which alerts us that the customer is using the new config. Had to move it up in the file, so that `getConfigPath()` works.

packages/cli/commands/init.js
```diff
@@ -105,10 +104,14 @@ exports.describe = i18n(`${i18nKey}.describe`, {
 exports.handler = async options => {
   const {
     auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
-    c,
     account: optionalAccount,
   } = options;
-  const configPath = (c && path.join(getCwd(), c)) || getConfigPath();
+  const { useNewConfig } = await promptUser({
```
brandenrodgers, line 109: This prompt may be confusing to users who are new to the CLI and have no idea what new or old versions of the config are.

One way we could slowly and safely roll this out could be to initially support a  hidden `--useHomeConfig` flag (or something along those lines) that routes to the new config logic when true. That way we won't need to add a prompt.

Then, later down the road and after tome validation testing, we could invert that flag so that the default behavior would be to use the new home config. We could then support some sort of `--useLegacyConfig` so devs who aren't ready to use the new version of config can still have access to the old version.

brandenrodgers: The major pro of a pattern like this would be that we can roll it out without it being a major breaking change.

kemmerle: Will take the suggestion, thanks! Just added the prompt because I knew we needed some kind of user input, but didn't think of a flag 🤦‍♀️

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/2296e84cb12ad2d78c96ce5e924f861afdfa5200

packages/cli/commands/auth.js
joe-yeager: If we use the flag idea @brandenrodgers suggested, can we make a call to `CLIConfiguration.init()` directly if that flag is provided?

packages/cli/commands/init.js
```diff
@@ -107,8 +109,11 @@ exports.handler = async options => {
     auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
     c,
     account: optionalAccount,
+    useRootConfig,
   } = options;
-  const configPath = (c && path.join(getCwd(), c)) || getConfigPath();
+  const configPath =
```
camden11, line 114: Might make sense to error if the `-c` flag and `--useRootConfig` are both passed in

packages/cli/commands/init.js
```diff
@@ -125,17 +130,22 @@ exports.handler = async options => {
...
+  loadConfig(configPath, options);
   handleExit(deleteEmptyConfigFile);
 
   try {
     const { accountId, name } = await CONFIG_CREATION_FLOWS[authType](
       env,
       optionalAccount
     );
-    const configPath = getConfigPath();
+    const configPath = getRootOrDeprecatedConfigPath();
```
camden11, line 148: Do we need to get config path a second time? Might make sense to rename if it has a different value

packages/cli/commands/init.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/7f9053a5b2f5659aad7c4203bd05278e774ed862

packages/cli/commands/init.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/7f9053a5b2f5659aad7c4203bd05278e774ed862

camden11 (approved):
Code looks good to me, just one comment about copy. Also haven't had a chance to test yet so should make sure we all do that before merging

joe-yeager (approved):

packages/cli/commands/init.js
```diff
@@ -107,8 +109,11 @@ exports.handler = async options => {
     auth: authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
     c,
     account: optionalAccount,
+    useHiddenConfig,
   } = options;
-  const configPath = (c && path.join(getCwd(), c)) || getConfigPath();
+  const configPath =
+    (c && path.join(getCwd(), c)) ||
```
brandenrodgers, line 115: It might make this easier to read if we use the `config` variable directly rather than the alias `c`.

brandenrodgers: Actually I take that back. I wonder why we didn't just call it `configPath` 🤔. You can ignore this initial comment.

packages/cli/lang/en.lyaml
```diff
@@ -82,6 +82,7 @@ en:
       describe: "Configure authentication for a HubSpot account. Supported authentication protocols are {{ supportedProtocols }}."
       errors:
         noConfigFileFound: "No config file was found. To create a new config file, use the \"hs init\" command."
+        bothConfigFilesNotAllowed: "Two config files are not allowed. Please delete one if necessary and try again."
```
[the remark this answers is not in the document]
kemmerle: Linked this comment to Sandra, so she can take a look 👀

packages/cli/commands/init.js
kemmerle: Going to resolve for now, feel free to reopen if necessary

kemmerle:
Update 8/16 (I am about to go on vacation next week):

TODOS to get this PR into a testable state:

- [ ] Add a check in the `init` command to use the new config if one exists. 
- [ ] Address feedback from Branden and the resulting changes to tests.
- [ ] Change copy based on Sandra's feedback. 

Phase 1 Proposal:

After I've gotten this PR into a testable state, I'd like to merge these changes into `main`. Then internal teams can test the new hidden config; the `--useHiddenConfig` flag allows us to do so without impacting workflows for existing customers. I'll be inviting members from the 1) DPG Dev Tooling team, 2) CPG Dev Tooling team, 3) JS Rendering team, 4) Developer Assets, and 5) Developer Relations. 

These teams are most familiar with the CLI and can best imitate the workflows of different HubSpot users, including "power users" who take advantage of nested configs. During this internal testing period, we can gain confidence in the new configuration setup, discover and address any bugs, and determine whether we need a `defaultAccountOverrides` property in the new config file.

Important to note: 

1) In this first phase, we will **not** be supporting overriding accounts with nested configs or with a specified property in the config file. It's not clear whether this is the best solution for power users. I would suggest that we explore the alternative workflow, where we encourage folks to use the `hs accounts use` command to switch default accounts. 

2) Also in this phase, we will only be supporting the "happy path", in which new users create a brand new hidden config file. We have not implemented a migration path at this time, but we will definitely discuss possible solutions. Sandra had an intriguing suggestion that we could write a command that automatically copies an old config file to the new, which seems feasible. 

I will be returning from vacation August 26, at which time I hope to finalize this PR and get this out for testing.

packages/cli/lang/en.lyaml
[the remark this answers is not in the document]
kemmerle: Meeting with Sandra about this case today.

kemmerle:
Returning to this PR after an interruption by Inbound work:

I have refactored this PR based on @brandenrodgers' feedback after its initial approval, so it will need a re-review and more testing, if possible. I tested the following scenarios for both the `hs init` and `hs auth` command:

`hs init`
1. Create new config from scratch 
2. Create old config from scratch 
3. Try to create new config when new config already exists 
4. Try to create new config when old config already exists 
5. Try to create new config when both configs already exist 
6. Try to create old config when old config already exists 
7. Try to create old config when new config already exists 
8. Try to create old config when both configs exist 

`hs auth`
1. Auth with a new config 
2. Auth with old config
3. Auth with both configs 

I will be meeting with Sandra this afternoon (9/10) to finalize copy.

commands/init.ts
```diff
@@ -168,24 +185,30 @@ exports.handler = async options => {
...
+        default: PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
+        defaultDescription: i18n(`${i18nKey}.options.auth.defaultDescription`, {
+          defaultType: PERSONAL_ACCESS_KEY_AUTH_METHOD.value,
+        }),
+      },
+      account: {
+        describe: i18n(`${i18nKey}.options.account.describe`),
+        type: 'string',
+      },
+      useHiddenConfig: {
```
joe-yeager, line 206: When we did the command audit, did we land on using `kebab-case` for the options?

kemmerle: We did! Will change that up.

lang/en.lyaml
```diff
@@ -414,13 +414,16 @@ en:
           defaultDescription: "\"{{ defaultType }}\": \nAn access token tied to a specific user account. This is the recommended way of authenticating with local development tools."
         account:
           describe: "HubSpot account to authenticate"
+        useHiddenConfig:
+          describe: "Use the new HubSpot configuration file located in a hidden file in the user's home directory (~)"
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/ce0d196390dbb88f81f8b820385c6d0dfb12e783

commands/sandbox/delete.ts
```diff
@@ -39,13 +44,15 @@ exports.describe = exports.describe = uiBetaTag(
 exports.handler = async options => {
   await loadAndValidateOptions(options, false);
 
-  const { account, force } = options;
+  // We don't want to auto inject the account flag from middleware.
+  // --providedAccountId preserves the original --account and --portal flags.
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/ce0d196390dbb88f81f8b820385c6d0dfb12e783

commands/project/dev.ts
```diff
@@ -108,12 +107,12 @@ exports.handler = async options => {
     (!hasPublicApps && isSandbox(accountConfig));
 
   // The account that the project must exist in
-  let targetProjectAccountId = options.account ? accountId : null;
+  let targetProjectAccountId = derivedAccountId || null;
   // The account that we are locally testing against
-  let targetTestingAccountId = options.account ? accountId : null;
+  let targetTestingAccountId = derivedAccountId || null;
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/ce0d196390dbb88f81f8b820385c6d0dfb12e783

commands/init.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/ce0d196390dbb88f81f8b820385c6d0dfb12e783

bin/cli.js
```diff
@@ -134,9 +143,28 @@ const setRequestHeaders = () => {
   addUserAgentHeader('HubSpot CLI', pkg.version);
 };
 
+const loadConfigMiddleware = async options => {
+  if (configFileExists(true)) {
+    loadConfig('', options);
+  }
+
+  if (options.config) {
```
kemmerle, line 151: We discussed changing this to load the config for all commands apart from `init`, but the tests were breaking with that change. Going to do some tests with/without the `--config` flag to see if this breaks anything.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/cli",
-  "version": "6.3.0",
+  "version": "6.3.0-experimental.0",
```
joe-yeager, line 3: We need to change this back before merging

bin/cli.js
joe-yeager: Did the tests go well on this?  Or did this need to be updated

package.json
brandenrodgers: Good call. Seems like this is a slightly annoying side-effect of the experimental releases, but it's not awful

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
```diff
@@ -1,36 +1,25 @@
...
+import { describe, it, expect, afterEach, beforeEach } from 'vitest';
 import { TestState } from '../../lib/TestState';
 import { getInitPromptSequence, ENTER } from '../../lib/prompt';
 
 describe('Account Management Flow', () => {
   let testState: TestState;
   const accountName = 'test-account';
 
-  beforeAll(async () => {
+  beforeEach(async () => {
```
brandenrodgers, line 9: Curious about this switch to `beforeEach`. If I'm interpreting this correctly, I think it kind of negates the intent of these "workflow" tests. The idea of the workflow tests is for them to sequentially run each step of the flow the same way that a CLI user would run them. This gives us the opportunity to catch bugs that could disrupt workflows, rather than just testing that each command in isolation works. Typically, devs don't run `hs init` before every account-related command (like `hs accounts list`).

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
```diff
@@ -68,17 +56,9 @@ describe('Account Management Flow', () => {
...
-  });
-
-  describe('hs accounts list', () => {
-    it('should not list the removed authenticated account', async () => {
-      const output = await testState.cli.executeWithTestConfig([
-        'accounts',
-        'list',
-      ]);
 
-      expect(output).not.toContain(accountName);
```
joe-yeager, line 81: Why did we remove this?

bin/cli.js
brandenrodgers: You could put both these if statements in the same block if you wanted to (i.e. no need to nest).

It might be useful to add a comment in here to explain that we need to load the config when `options.config` exists so that `getAccountIdFromConfig()` in `injectAccountIdMiddleware` reads from the right config.

It might also be useful to prevent users from trying to use `--config` while on the new, hidden config since they're incompatible with each other. I think we'd want to prevent both of these scenarios:
- `configFileExists(true) === true && options.config` (i.e. new config exists, but the user is trying to override the config path)
- `options.useHiddenConfig && options.config` (i.e. the user is opting into new config, but also trying to override the default location)

kemmerle: @joe-yeager, the tests went well, when I tested out commands with the `--config` flag. I did not discover any bugs, but I would appreciate if you or other members of the team could continue to test things out on the experimental branch. 

@brandenrodgers, I can add in those comments. I can also add the check for `configFileExists(true) === true && options.config`. We already check for `useHiddenConfig` and `config` with this [line](https://github.com/HubSpot/hubspot-cli/blob/44d4e2eee12276783dae40ce8e44a9d4fa8b4d03/commands/init.ts#L230).

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
kemmerle: Gotcha, I misunderstood the nature of the tests. I can change back.

package.json
camden11: Yeah, I was thinking this could be something we just catch in PRs. But I could also update the script to revert the version locally after you publish an experimental release

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
kemmerle: I misunderstood the nature of the tests 😬 I can change back. Insert the Homer Simpson disappearing GIF here please.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/4d29cfa5d53a3cd02049061caea52a460d5ef11f

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/4d29cfa5d53a3cd02049061caea52a460d5ef11f

package.json
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/4d29cfa5d53a3cd02049061caea52a460d5ef11f

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1123/commits/4d29cfa5d53a3cd02049061caea52a460d5ef11f

brandenrodgers (approved):
LGTM 🎉 (just a reminder to switch to `next` before merging)

PR #1125, opened 2024-08-13
fix: Fix a couple logging errors, update some failing tests

joe-yeager:
## Description and Context

- Update logError method to call `updateContext` if it makes sense to get the more detailed log messages we are expecting
- Fix a couple tests that were failing on the `next` branch

### DIFF
--- a/packages/cli/lib/errorHandlers/index.js
+++ b/packages/cli/lib/errorHandlers/index.js
@@ -20,6 +20,10 @@ function logError(error, context = {}) {
     return;
   }
 
+  if (isHubSpotHttpError(error) && context) {
+    error.updateContext(context);
+  }
+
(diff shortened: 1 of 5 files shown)


packages/cli/lib/errorHandlers/index.js
```diff
@@ -20,6 +20,10 @@ function logError(error, context = {}) {
     return;
   }
 
+  if (isHubSpotHttpError(error) && context) {
+    error.updateContext(context);
+  }
```
joe-yeager, line 25: We now call `updateContext` to update the messages if it makes sense to do so.

camden11 (approved):

brandenrodgers (approved):
👌

PR #1126, opened 2024-08-13
chore: Switch developerTestAccount & sandbox usages from `lib` to `api`

joe-yeager:
## Description and Context

- Switches to use `api/developerTestAccounts.ts` instead of `lib/developerTestAccounts.ts` because we are deprecating `lib/developerTestAccounts.ts` because it is a very thin wrapper and we can directly use `api/developerTestAccounts.ts`
- Switches usages of `lib/sandboxes.ts` to `api/sandboxHubs` and `api/sandboxSync`

Companion PR: https://github.com/HubSpot/hubspot-local-dev-lib/pull/178

## Who to Notify

@kemmerle @adamawang @camden11 @brandenrodgers

### DIFF
--- a/packages/cli/commands/sandbox/delete.js
+++ b/packages/cli/commands/sandbox/delete.js
@@ -10,7 +10,7 @@ const { trackCommandUsage } = require('../../lib/usageTracking');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const { logError, debugError } = require('../../lib/errorHandlers/index');
 const { isSpecifiedError } = require('@hubspot/local-dev-lib/errors/index');
-const { deleteSandbox } = require('@hubspot/local-dev-lib/sandboxes');
+const { deleteSandbox } = require('@hubspot/local-dev-lib/api/sandboxHubs');
 const { i18n } = require('../../lib/lang');
 const { deleteSandboxPrompt } = require('../../lib/prompts/sandboxesPrompt');
 const {
--- a/packages/cli/lib/developerTestAccounts.js
+++ b/packages/cli/lib/developerTestAccounts.js
@@ -5,7 +5,7 @@ const { getAccountId, getConfig } = require('@hubspot/local-dev-lib/config');
 const { i18n } = require('./lang');
 const {
   fetchDeveloperTestAccounts,
-} = require('@hubspot/local-dev-lib/developerTestAccounts');
+} = require('@hubspot/local-dev-lib/api/developerTestAccounts');
 const {
   isMissingScopeError,
   isSpecifiedError,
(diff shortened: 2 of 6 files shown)


packages/cli/lib/developerTestAccounts.js
```diff
@@ -32,31 +32,31 @@ const getHasDevTestAccounts = appDeveloperAccountConfig => {
...
+          limit,
+        })
+      );
+    } else {
+      throw new Error(
+        i18n('lib.developerTestAccount.create.failure.limit', {
+          accountName: accountConfig.name || accountId,
+          limit,
+        })
+      );
```
joe-yeager, line 56: This isn't related, but I just flipped the if statement and returned `null` up front to reduce a layer of nesting

adamawang (approved):
Looks good with the local-dev-lib PR 👍

adamawang:
Should I look into doing something similar with the [sandboxes](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/lib/sandboxes.ts) file? Since that doesn't really add much either

joe-yeager:
> Should I look into doing something similar with the [sandboxes](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/lib/sandboxes.ts) file? Since that doesn't really add much either

@adamawang Oh, good to know.  I'll just add it to these PRS since I'm already in here and we are trying to build up some breaking changes for a v2 of LDL

packages/cli/commands/sandbox/delete.js
```diff
@@ -10,7 +10,7 @@ const { trackCommandUsage } = require('../../lib/usageTracking');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const { logError, debugError } = require('../../lib/errorHandlers/index');
 const { isSpecifiedError } = require('@hubspot/local-dev-lib/errors/index');
-const { deleteSandbox } = require('@hubspot/local-dev-lib/sandboxes');
+const { deleteSandbox } = require('@hubspot/local-dev-lib/api/sandboxHubs');
```
joe-yeager, line 13: The call is on line 141 and the return value is being ignored, that's why there are no other changes here.

camden11 (approved):
LGTM. From what I can tell, this doesn't depend on any upcoming changes in LDL either so you could safely merge into main rather than next if you wanted to

joe-yeager:
> LGTM. From what I can tell, this doesn't depend on any upcoming changes in LDL either so you could safely merge into main rather than next if you wanted to

I believe it's depending on the changes you made to the API calls returning the full promise

camden11:
🤦 It absolutely is, i'm losing track of what's on `next` and what's not haha

brandenrodgers (approved):

PR #1127, opened 2024-08-14
Chore: Add detailed usage tracking to migration commands

kemmerle:
## Description and Context
We are adding more detailed usage tracking to the `migrate-app` and `clone-app` commands. In both commands, I have added three tracking events:

1) When the command is initially run, I've added an event after the `appId` is selected. **I am wondering whether I should add any other metadata here.** 

2) When the command succeeds, I've added an event where I've included the project name, app id, and status.

3) When the command fails, I've added an event where I've included the project name, app id, status, and error.

## TODO

- [x] Get feedback from Product folks (Laurie) 

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/cloneApp.js
+++ b/packages/cli/commands/project/cloneApp.js
@@ -6,7 +6,7 @@ const {
   getAccountId,
   addUseEnvironmentOptions,
 } = require('../../lib/commonOpts');
-const { trackCommandUsage } = require('../../lib/usageTracking');
+const { trackCommandMetadataUsage } = require('../../lib/usageTracking');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const { i18n } = require('../../lib/lang');
 const {
@@ -52,8 +52,6 @@ exports.handler = async options => {
   const accountConfig = getAccountConfig(accountId);
   const accountName = uiAccountDescription(accountId);
 
-  trackCommandUsage('clone-app', {}, accountId);
-
   if (!isAppDeveloperAccount(accountConfig)) {
     uiLine();
     logger.error(i18n(`${i18nKey}.errors.invalidAccountTypeTitle`));
@@ -81,6 +79,7 @@ exports.handler = async options => {
       });
       appId = appIdResponse.appId;
     }
+    trackCommandMetadataUsage('clone-app', { appId }, accountId);
 
     const projectResponse = await createProjectPrompt('', options, true);
     name = projectResponse.name;
@@ -120,6 +119,12 @@ exports.handler = async options => {
       };
       const success = writeProjectConfig(configPath, configContent);
 
+      trackCommandMetadataUsage(
+        'clone-app',
+        { projectName: name, appId, status },
+        accountId
+      );
+
       SpinniesManager.succeed('cloneApp', {
         text: i18n(`${i18nKey}.cloneStatus.done`),
         succeedColor: 'white',
@@ -137,6 +142,12 @@ exports.handler = async options => {
       process.exit(EXIT_CODES.SUCCESS);
     }
   } catch (error) {
+    trackCommandMetadataUsage(
+      'clone-app',
+      { projectName: name, appId, status: 'FAILURE', error },
+      accountId
+    );
+
     SpinniesManager.fail('cloneApp', {
       text: i18n(`${i18nKey}.cloneStatus.failure`),
       failColor: 'white',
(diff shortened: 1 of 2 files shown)


packages/cli/commands/project/cloneApp.js
```diff
@@ -81,6 +79,7 @@ exports.handler = async options => {
       });
       appId = appIdResponse.appId;
     }
+    trackCommandMetadataUsage('clone-app', { appId }, accountId);
```
kemmerle, line 82: Do we need more metadata here/is there a better place to put this tracking event?

packages/cli/commands/project/cloneApp.js
```diff
@@ -52,8 +55,6 @@ exports.handler = async options => {
   const accountConfig = getAccountConfig(accountId);
   const accountName = uiAccountDescription(accountId);
 
-  trackCommandUsage('clone-app', {}, accountId);
```
kemmerle, line 55: Should we include a tracking event early on just to mark if a customer attempts the command. Would we need to move it own after the selection of the `appId`?

packages/cli/commands/project/migrateApp.js
```diff
@@ -58,8 +58,6 @@ exports.handler = async options => {
   const accountConfig = getAccountConfig(accountId);
   const accountName = uiAccountDescription(accountId);
 
-  trackCommandUsage('migrate-app', {}, accountId);
```
kemmerle, line 61: See my other comment.

packages/cli/commands/project/cloneApp.js
brandenrodgers: Yes, we should still call `trackCommandUsage()` here without appId. This tracking method is intended to show us stats on every command run, so we always intentionally call it at the top of our commands. The metadata tracking command lets us pass along some additional data in follow-up requests so we can dig deeper if we want to.

packages/cli/commands/project/cloneApp.js
```diff
@@ -81,6 +82,15 @@ exports.handler = async options => {
       });
       appId = appIdResponse.appId;
     }
+    const selectedApp = await fetchPublicAppMetadata(appId, accountId);
+    // preventProjectMigrations returns true if we have not added app to allowlist config.
+    // listingInfo will only exist for marketplace apps
+    const { preventProjectMigrations, listingInfo } = selectedApp;
+    trackCommandMetadataUsage(
```
brandenrodgers, line 89: I think we can remove this one (per my comment above). So the three tracking calls in this file will be:
- trackCommandUsage (our typical command usage tracking, without appID)
- trackCommandMetadataUsage (for successful runs)
- trackCommandMetadataUsage (for failed runs)

kemmerle: Laurie asked for more information about tracking whether the apps are marketplace listed but specified that it's nice to have, not a must. Should I remove?

brandenrodgers: Ah ok. In that case can we include the listing info in the success and failure tracking requests? Also, what's the shape of `listingInfo`? Is is a string?

kemmerle: `listingInfo` is an object. Oh, I should use the spread operator then, correct?

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1127/commits/32cd50327d070cff90dc3dc3b07ec5a22bd993d8, let me know if I need to add the spread operator.

packages/cli/commands/project/cloneApp.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1127/commits/32cd50327d070cff90dc3dc3b07ec5a22bd993d8

brandenrodgers (approved):

packages/cli/commands/project/migrateApp.js
```diff
@@ -185,6 +191,12 @@ exports.handler = async options => {
         { includesRootDir: true, hideLogs: true }
       );
 
+      trackCommandMetadataUsage(
+        'migrate-app',
+        { projectName, appId, status, preventProjectMigrations, listingInfo },
```
brandenrodgers, line 196: iirc you're not able to include objects in tracking events. For listing info you might be better off just using a boolean like `isListed`

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1127/commits/b9f3bd85d7238c6c72af57fdb90fbdd283208361

PR #1128, opened 2024-08-14
Supporting a custom override for missing scope errors

brandenrodgers:
## Description and Context

The main focus of this was to support overriding missing scope errors. The typical errors returned by the BE aren't very useful, so this makes it easy for us to provide better messaging to the end user when they attempt to make a request that fails with a 403 status and `MISSING_SCOPE` category.

I removed all of the places where we were passing in `projectName` to the api error context because we don't need that anymore (I'm removing it [here](https://github.com/HubSpot/hubspot-local-dev-lib/pull/169/files#diff-9d5b4d724ee9d55d1fa4e316c00ad65c422dc4365c603388afd8f012d3327158L164)).

I also cleaned up a few things:
- I removed some dead translations that weren't being used anywhere anymore
- I slightly modified some code in overrideErrors to prevent a potential issue in the scenario where `err.response.data.context` doesn't exist
- I removed some code in the sandbox sync command that was manually handling the missing scope error

### DIFF
--- a/packages/cli/commands/project/download.js
+++ b/packages/cli/commands/project/download.js
@@ -6,7 +6,7 @@ const {
   addUseEnvironmentOptions,
 } = require('../../lib/commonOpts');
 const { trackCommandUsage } = require('../../lib/usageTracking');
-const { getCwd } = require('@hubspot/local-dev-lib/path');
+const { getCwd, sanitizeFileName } = require('@hubspot/local-dev-lib/path');
 const {
   logApiErrorInstance,
   ApiErrorContext,
@@ -95,7 +95,7 @@ exports.handler = async options => {
 
     await extractZipArchive(
       zippedProject,
-      projectName,
+      sanitizeFileName(projectName),
       path.resolve(absoluteDestPath),
       { includesRootDir: false }
     );
--- a/packages/cli/commands/project/download.js
+++ b/packages/cli/commands/project/download.js
@@ -108,7 +108,10 @@ exports.handler = async options => {
     );
     process.exit(EXIT_CODES.SUCCESS);
   } catch (e) {
-    logApiErrorInstance(e, new ApiErrorContext({ accountId, projectName }));
+    logApiErrorInstance(
+      e,
+      new ApiErrorContext({ accountId, action: 'project download' })
+    );
     process.exit(EXIT_CODES.ERROR);
   }
 };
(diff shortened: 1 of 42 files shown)


packages/cli/commands/project/download.js
```diff
@@ -108,7 +108,10 @@ exports.handler = async options => {
     );
     process.exit(EXIT_CODES.SUCCESS);
   } catch (e) {
-    logApiErrorInstance(e, new ApiErrorContext({ accountId, projectName }));
+    logApiErrorInstance(
+      e,
+      new ApiErrorContext({ accountId, action: 'project download' })
```
camden11, line 113: Looks like this should be `request` rather than `action`

packages/cli/commands/project/watch.js
```diff
@@ -146,7 +143,7 @@ exports.handler = async options => {
             result.uploadError,
             new ApiErrorContext({
               accountId,
-              projectName: projectConfig.name,
+              action: 'project upload',
```
camden11, line 146: same here

packages/cli/lib/errorHandlers/overrideErrors.js
```diff
@@ -1,38 +1,46 @@
...
-    [PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_NOT_SPECIFIED]:
-      'unspecified platformVersion',
-    [PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_RETIRED]:
-      errData.context.RETIRED_PLATFORM_VERSION,
-    [PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST]:
-      errData.context.PLATFORM_VERSION,
-  };
+function createPlatformVersionError(err, subCategory) {
+  let translationKey = 'unspecifiedPlatformVersion';
+  let platformVersion = '';
```
camden11, line 19: Any reason to default the `translationKey` but not the `platformVersion`? From what I could tell, you could set this to "unspecified platformVersion" and remove the check for `PLATFORM_VERSION_NOT_SPECIFIED` in the switch

camden11 (approved):
A few tiny things but looks good

packages/cli/lib/errorHandlers/overrideErrors.js
brandenrodgers: Ah, no not really. I updated in [1bfea4b](https://github.com/HubSpot/hubspot-cli/pull/1128/commits/1bfea4bfb993b0b7eba0b9a95387f6a9f784fe97)

PR #1133, opened 2024-08-16
feat: Add `hs project install-deps` command

joe-yeager:
## Description and Context

Adds a command `hs project install-deps` that has the following behavior:
- When called from outside of a project, it logs an error stating it must be ran within a project
- When called like this `hs project install-deps`, it walks directory structure of the project `srcDir` defined in `hsproject.json` looking for `package.json` files and installs the dependencies in all of those directories. 
- When called with a positional argument, like `hs project install-deps dependency` it will prompt the user with a list of locations for where they would like to install the dependency

### Install all dependencies
Kapture 2024-09-06 at 11 58 52

### Add a dependencies to subcomponents

Kapture 2024-09-06 at 12 00 18

## TODO

- [x] Tests
- [x] Get copy sign off from UX
- [x] Add gifs
## Who to Notify


packages/cli/commands/project/install.js
```diff
@@ -0,0 +1,175 @@
...
+
+const YARN = 'yarn';
+const NPM = 'npm';
+const PNPM = 'pnpm';
+
+const packageManagers = {
+  [NPM]: 'package-lock.json',
+  [PNPM]: 'pnpm-lock.yaml',
+  [YARN]: 'yarn.lock',
+};
```
banderson, line 27: I am still holding out hope that [corepack](https://nodejs.org/api/corepack.html) turns into the dream tool I thought it was going to be. Not saying we slow things down for that or anything, just we should keep an eye on that since that's part of node core and just feels like it HAS TO HAPPEN at some point, and become something people actually use.

That would eventually give us the benefit of enforcing the same package manager at both build time and local dev too, along with error messages if anyone tries to run a command from another package manager than the one selected for the project.

I'm praying this works for us someday!!

banderson (commented):
I only did a quick code review, so haven't dug in deep, but I LOVE where your head's at with this!! 😍 😍 😍

packages/cli/commands/project/install.js
joe-yeager: `corepack` looks cool and would really help to enforce consistency.  It's been lingering in `experimental` for a while.  I really hope it makes it to stable someday 🤞 

@brandenrodgers and I talked a little bit about making this configurable at the project or app level even.  We still need to think about what that would look like.  We could possibly use that information to mimic the validations that `corepack` is doing, but on `hs project upload`.  I.E. they have `yarn` set as their package manager for the project, but we detect a `pnpm` lockfile and log a warning.

packages/cli/lang/en.lyaml
```diff
@@ -487,6 +487,10 @@ en:
             parentAccountNotConfigured: "To develop this project locally, run {{ authCommand }} to authenticate the App Developer Account {{ accountId }} associated with {{ accountIdentifier }}."
           examples:
             default: "Start local dev for the current project"
+          install-deps:
+            installingForProject: "Installing dependencies for the project"
+            installingForProjectSuccessful: "Dependency installation successful"
+            installingForProjectFailed: "Dependency installation failed"
```
joe-yeager, line 493: These are for the installation that happens in the `hs project dev` command.

packages/cli/lang/en.lyaml
```diff
@@ -707,6 +711,17 @@ en:
...
+          installLocationPrompt: "Choose where to add the dependencies:"
+          installLocationPromptRequired: "You must choose at least one location"
+          installationSuccessful: "Dependencies installed successfully"
+          installingDependencies: "Installing dependencies in {{location}}"
+          addingDependenciesToLocation: "Installing {{dependencies}} in {{location}}"
+          installingDependenciesFailed: "Installing dependencies for {{directory}} failed"
+          describe: "Install your deps"
+          noProjectConfig: "No project detected. Run this command again from a project directory."
+          noPackageJsonInProject: "Could not find any package.json files in the project"
+          npmNotInstalled: "This command depends on {{ packageManager }}, install {{#bold}}{{ link }}{{/bold}}"
```
joe-yeager, line 724: These are for `hs project install-deps`

packages/cli/commands/project/installDeps.js
```diff
@@ -0,0 +1,73 @@
...
+const path = require('path');
+const { i18n } = require('../../lib/lang');
+
+const i18nKey = `commands.project.subcommands.installDeps`;
+
+exports.command = 'install-deps [packages..]';
+// Intentionally making this null to hide command
+exports.describe = null;
+// exports.describe = uiBetaTag(i18n(`${i18nKey}.help.describe`), false);
+exports.builder = yargs => {
```
brandenrodgers, line 18: Big nit, but in all of our other commands we've been placing the builder at the bottom of the file. I'd vote to keep consistency in that to make things easier to quickly find

packages/cli/commands/project/installDeps.js
```diff
@@ -0,0 +1,73 @@
...
+    [
+      '$0 project install-deps dependency1 dependency2',
+      i18n(`${i18nKey}.help.addDepToSubComponentExample`),
+    ],
+  ]);
+};
+
+exports.handler = async ({ packages }) => {
+  try {
+    const projectConfig = await getProjectConfig();
```
brandenrodgers, line 30: Can you add a call to `trackCommandUsage` somewhere at the top of this?

joe-yeager: Good catch, are there any naming conventions for those? `project-install-deps` ok?

packages/cli/commands/project/installDeps.js
joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1133/commits/2a9e7834c1260fd0cd49ac2908f6547f61804361

packages/cli/commands/project/installDeps.js
joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1133/commits/2a9e7834c1260fd0cd49ac2908f6547f61804361

packages/cli/commands/project/installDeps.js
```diff
@@ -0,0 +1,78 @@
...
+const { getProjectConfig } = require('../../lib/projects');
+const { promptUser } = require('../../lib/prompts/promptUtils');
+const path = require('path');
+const { i18n } = require('../../lib/lang');
+const { trackCommandUsage } = require('../../lib/usageTracking');
+const { getAccountId } = require('../../lib/commonOpts');
+
+const i18nKey = `commands.project.subcommands.installDeps`;
+
+exports.command = 'install-deps [packages..]';
```
brandenrodgers, line 16: Are you missing a third period here after packages?

joe-yeager: I had the same thought!  But `yargs` uses `..` instead of `...` like the JS rest operator 


Docs: https://github.com/yargs/yargs/blob/main/docs/advanced.md#variadic-positional-arguments

PR #1136, opened 2024-08-20
feat: More noticeable update notifications

brandenrodgers:
## Description and Context

This is WIP because we still need a 👍 from design, but I wanted to showcase some of our options for reformatting the update notification that we show to our users. Right now the notification is easy to miss and isn't customized to look like the rest of our CLI output.

## Screenshots

image

### DIFF
--- a/packages/cli/bin/cli.js
+++ b/packages/cli/bin/cli.js
@@ -16,7 +16,7 @@ const { getIsInProject } = require('../lib/projects');
 const pkg = require('../package.json');
 const { i18n } = require('../lib/lang');
 const { EXIT_CODES } = require('../lib/enums/exitCodes');
-const { uiCommandReference } = require('../lib/ui');
+const { UI_COLORS, uiCommandReference } = require('../lib/ui');
 
 const removeCommand = require('../commands/remove');
 const initCommand = require('../commands/init');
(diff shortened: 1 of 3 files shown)


packages/cli/bin/cli.js
```diff
@@ -51,12 +52,34 @@ const notifier = updateNotifier({
...
-  '\n\nTo upgrade, run:\n\nnpm uninstall -g @hubspot/cms-cli\nand npm install -g @hubspot/cli';
+  `\n\nTo upgrade, uninstall ${chalk.bold(
+    CMS_CLI_PACKAGE_NAME
+  )} \nand then run ${uiCommandReference('{updateCommand}')}`;
+
+const CLI_UPGRADE_MESSAGE = `HubSpot CLI version ${chalk.cyan(
+  chalk.bold('{currentVersion}')
+)} is outdated. \nRun ${uiCommandReference(
+  '{updateCommand}'
+)} to upgrade to version ${chalk.cyan(chalk.bold('{latestVersion}'))}`;
```
joe-yeager, line 67: Should we `i18n` these messages?

joe-yeager (approved):
LGTM, just a question about `i18n`

brandenrodgers:
Here is the most recent version
image

packages/cli/bin/cli.js
brandenrodgers: Good call, done in [7bda1dd](https://github.com/HubSpot/hubspot-cli/pull/1136/commits/7bda1dda97007a17c58d0015e9308185bd1433b6)

camden11 (approved):
Ah man, I kind of liked the aggressively hubspot orange. But this is good too!

packages/cli/bin/cli.js
```diff
@@ -51,12 +52,30 @@ const notifier = updateNotifier({
...
-  '\n\nTo upgrade, run:\n\nnpm uninstall -g @hubspot/cms-cli\nand npm install -g @hubspot/cli';
+const CMS_CLI_PACKAGE_NAME = '@hubspot/cms-cli';
 
 notifier.notify({
-  message: pkg.name === '@hubspot/cms-cli' ? CLI_UPGRADE_MESSAGE : null,
+  message:
+    pkg.name === CMS_CLI_PACKAGE_NAME
+      ? i18n(`${i18nKey}.updateNotify.cmsUpdateNotification`, {
+          packageName: CMS_CLI_PACKAGE_NAME,
+          updateCommand: uiCommandReference('{updateCommand}'),
```
brandenrodgers, line 62: Just for posterity, I'm passing in `{updateCommand}` here instead of the actual command because update-notify handles some interpolation for certain strings. They mention them in the docs [here](https://www.npmjs.com/package/update-notifier#message).

packages/cli/lang/en.lyaml
```diff
@@ -1,6 +1,10 @@
 en:
   commands:
     generalErrors:
+      updateNotify:
+        notifyTitle: "Update available"
+        cmsUpdateNotification: "{{#bold}}The CMS CLI is now the HubSpot CLI{{/bold}}\n\nTo upgrade, uninstall {{#bold}}{{ packageName }}{{/bold}}\nand then run {{ updateCommand }}"
+        cliUpdateNotification: "HubSpot CLI version {{#cyan}}{{#bold}}{currentVersion}{{/bold}}{{/cyan}} is outdated.\nRun {{ updateCommand }} to upgrade to version {{#cyan}}{{#bold}}{latestVersion}{{/bold}}{{/cyan}}"
```
brandenrodgers, line 7: Again for posterity, I'm intentionally using single brackets for `{currentVersion}` and `{latestVersion}`. See update-notify docs [here](https://www.npmjs.com/package/update-notifier#message)

PR #1137, opened 2024-08-20
fix: Turn off warnings for hs command

joe-yeager:
## Description and Context

Disable warnings to silence the `punycode` deprecation warnings.

## TODO

- [ ] Test on macOs
- [ ] Test on windows
- [ ] Test on linux


joe-yeager:
Doesn't work on Windows, back to the drawing board.

PR #1138, opened 2024-08-21
test: Adding scheduled actions to test latest CLI release

brandenrodgers:
## Description and Context

Adding a cron github action that runs once per hour to pull the latest CLI release and test out a command. The idea here is to catch any major issues in releases as soon as possible. I'm also configuring a slack notifications bot to post in our team's channel whenever the github actions fail.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Build Failure Slack Report
-        if: ${{ github.ref == 'refs/heads/main' }}
+        if: ${{ github.ref == 'refs/heads/br/scheduled-cli-tests' }}
         uses: ravsamhq/notify-slack-action@master
         with:
           status: ${{ job.status }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,7 @@
 on: [push]
-
 jobs:
   build:
     runs-on: ubuntu-latest
-
     steps:
       - name: Build Failure Slack Report
         if: ${{ github.ref == 'refs/heads/main' }}
@@ -11,6 +9,8 @@ jobs:
         with:
           status: ${{ job.status }}
           notify_when: 'failure'
+          notification_title: '{workflow} has {status_message}'
+          message_format: '{emoji} *{workflow}* {status_message} in <{repo_url}|{repo}>'
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.ACTION_MONITORING_SLACK }}
       - uses: actions/checkout@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
           status: ${{ job.status }}
           notify_when: 'failure'
           notification_title: 'CLI build failure'
-          message_format: '{emoji} *Build* {status_message} in <{repo_url}|{repo}> on <{commit_url}|{commit_sha}>"'
+          message_format: '{emoji} *Build* {status_message} in <{repo_url}|{repo}> on <{commit_url}|{commit_sha}>'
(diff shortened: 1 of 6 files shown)


.github/workflows/testLatestPublish.yml
```diff
@@ -0,0 +1,38 @@
...
+        shell: bash
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+      - name: Install Deps
+        run: npm install
+      - name: Test Latest CLI Release
+        run: |
+          npm run test-cli --cliNPMVersion="5.0.2"
```
brandenrodgers, line 24: Purposefully using a bad CLI release on the first pass so we can watch this fail

.github/workflows/build.yml
```diff
@@ -46,7 +44,7 @@ jobs:
           status: ${{ job.status }}
           notify_when: 'failure'
           notification_title: 'CLI build failure'
-          message_format: '{emoji} *Build* {status_message} in <{repo_url}|{repo}> on <{commit_url}|{commit_sha}>"'
```
brandenrodgers, line 49: This was a typo by me from another PR

.github/workflows/testLatestPublish.yml
```diff
@@ -0,0 +1,38 @@
+on:
```
brandenrodgers, line 1: There are two ways this will run:
1. Every time a new tag is published
2. Every 5 hours

The sleep at the beginning of the action is to give the newly-published package some time to become available on npm. Otherwise `latest` may not target the most recent release.

acceptance-tests/run-tests
```diff
@@ -18,12 +18,17 @@ setArgsOverrides(
...
-      description: 'Account ID',
+      description: 'The account ID to use',
     })
     .option('cliPath', {
       alias: 'c',
       type: 'string',
-      description: 'CLI path',
+      description: 'The path to the CLI',
+    })
+    .option('cliNPMVersion', {
```
brandenrodgers, line 28: Open to suggestions for a better arg name

acceptance-tests/tests/helpers/cmd.js
```diff
@@ -12,21 +12,31 @@ const PATH = process.env.PATH;
...
-  // Ensure that path exists
-  if (!processPath || !existsSync(processPath)) {
-    throw new Error('Invalid process path');
-  }
+function createProcess(cliPath, cliNPMVersion, args = [], env = null) {
+  let processCommand;
 
-  args = [processPath].concat(args);
+  if (cliNPMVersion) {
+    processCommand = 'npx';
```
brandenrodgers, line 24: This will use npx to download and execute cli commands for each test. It ends up looking something like this:
```bash
npx --yes --package=@hubspot/cli@latest hs init
```

acceptance-tests/run-tests
camden11: I feel like just calling it `cliVersion` might be clearer? `cliNPMVersion` kind of makes it sound like its the version of `npm` you're going to use. But don't feel super strongly

acceptance-tests/run-tests
```diff
@@ -18,12 +18,17 @@ setArgsOverrides(
...
+      alias: 'npm',
+      type: "string",
+      description: "The npm version of the CLI to download and run"
```
camden11, line 31: Might make sense to say that specifying a version will cause the tests to run with `npx`, rather than "download and run" (even though npx technically does download"

.github/workflows/testLatestPublish.yml
camden11: Don't forget to update this to `latest` before merging!

camden11 (approved):
This looks great! Pleasantly surprised that it was this simple to get all this working. Just a few minor suggestions

PR #1139, opened 2024-08-21
Fix error messaging in hs cms convert-fields

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/630
We noticed during our CLI bug bash that not providing the `--src` flag to `hs cms conver-fields` leads to a misleading error message. I also noticed that providing an invalid `src` results in a missing translation and even worse error message. This fixes both of those things

## Screenshots
No `--src` before
Screenshot 2024-08-21 at 12 16 07 PM

No `--src` after
Screenshot 2024-08-21 at 12 16 25 PM

Invalid `--src` before
Screenshot 2024-08-21 at 12 21 02 PM

Invalid `--src` after
Screenshot 2024-08-21 at 12 25 06 PM

## Who to Notify
@brandenrodgers @joe-yeager @TanyaScales

### DIFF
--- a/packages/cli/commands/cms/convertFields.js
+++ b/packages/cli/commands/cms/convertFields.js
@@ -93,7 +93,7 @@ exports.builder = yargs => {
   yargs.option('src', {
     describe: i18n(`${i18nKey}.positionals.src.describe`),
     type: 'string',
-    demandOption: true,
+    demandOption: i18n(`${i18nKey}.errors.missingSrc`),
   });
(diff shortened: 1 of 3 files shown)


brandenrodgers:
Nice! 🎉  this lgtm, but I'll defer to @TanyaScales for the final approve

TanyaScales (approved):
LGTM - Thanks!

PR #1143, opened 2024-08-22
Fix bugs with hs project add --type flag

camden11:
## Description and Context
Noticed recently that the `--type` flag on the `hs project add` command was broken in a few ways:
* The command would never recognize that a `--type` was passed in, and would open the prompt even if the user specified a valid type
* After selecting a type from the prompt, the command would error if you had used the type flag.

This PR fixed both of these things

## Screenshots

Before
Screenshot 2024-08-22 at 11 36 46 AM

After
Screenshot 2024-08-22 at 11 36 42 AM

## Testing

To test, try running with a valid component type
```
yarn hs project add  --type="components/example-app" --name="test"
```

and an invalid component type
```
yarn hs project add  --type="components/not-a-real-type" --name="test"
```

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/add.js
+++ b/packages/cli/commands/project/add.js
@@ -75,6 +75,12 @@ exports.builder = yargs => {
   });
 
   yargs.example([['$0 project add', i18n(`${i18nKey}.examples.default`)]]);
+  yargs.example([
+    [
+      '$0 project add --name="my-component" --type="components/example-app"',
+      i18n(`${i18nKey}.examples.withFlags`),
+    ],
+  ]);
 
   return yargs;
 };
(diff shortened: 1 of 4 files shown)


camden11:
It's the path to the component in the hubspot project components repo. Not exactly the most intuitive, but that's why the prompt is there. Figure its mostly helpful if you ever needed to automate something like this.

That said, just noticed the `help` command doesn't make it clear what `--type` is supposed to be so I'll update that

camden11:
@brandenrodgers Updated with a better description of the `--type` flag and an example with flags

brandenrodgers (approved):

PR #1147, opened 2024-08-22
Clean up generic errors

camden11:
## Description and Context
This removes the "an error has occurred" text the prefaces many error messages, since it's redundant with the `[ERROR]` logged at the beginning of every error message. The intent of this log was to inform users that a specific _kind_ of error had occurred, but I think it's likely that that's not really relevant information for most people. Also, with the new error setup, most specific kinds of errors are already handled as system errors or hubspot HTTP errors.

See issue: https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/594

## Screenshots
Having a hard time finding a good example of this, since the vast majority of errors are HubSpot HTTP errors or system errors and are handled different. Here's an example from `hs project add` that I've since fixed.

Before
Screenshot 2024-08-22 at 4 08 04 PM

After
Screenshot 2024-08-22 at 4 08 22 PM

## Who to Notify
@brandenrodgers @joe-yeager

### DIFF
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -1420,8 +1420,6 @@ en:
         errorOccurred: "Error: {{ error }}"
         errorContext: "Context: {{ context }}"
         errorCause: "Cause: {{ cause }}"
-        systemErrorOccurred: "A system error has occurred: {{ errorMessage }}"
-        genericErrorOccurred: "A {{ name }} has occurred."
         unknownErrorOccurred: "An unknown error has occurred."
       suppressErrors:
         platformVersionErrors:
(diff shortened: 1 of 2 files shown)


packages/cli/lang/en.lyaml
```diff
@@ -1420,9 +1420,7 @@ en:
         errorOccurred: "Error: {{ error }}"
         errorContext: "Context: {{ context }}"
         errorCause: "Cause: {{ cause }}"
-        systemErrorOccurred: "A system error has occurred: {{ errorMessage }}"
```
brandenrodgers, line 1423: `systemErrorOcurred` looks like it's used in `standardErrors.js`. Do we need to keep this around?

brandenrodgers: Ah my bad. Yeah you're right 👌

brandenrodgers (approved):

PR #1151, opened 2024-08-23
[HOLD for September 10th] Remove deprecated sandbox sync command

adamawang:
## Description and Context

After deprecating the `hs sandbox sync` command in the August 16th release, we are planning on removing the command for the September 15th release pre-inbound. 

This PR removes the `/sandbox/sync.js` file and any usages of it, but keeps around the `/lib/sandboxSync.js` file since we are still runnings syncs in the background for `hs sandbox create` and `hs project dev`. That behavior will change later this year once our version 2 of sandboxes rolls out where syncs are run automatically and we'll no longer need to trigger them from the FE or CLI. 

The only UI changes that are detectable are `hs sandbox sync` no longer shows up in the menus when running `hs sandbox --help`, nor does it show up as a command. 

## Screenshots

Command with deprecation message:
Screenshot 2024-08-23 at 11 34 08

Now removed from `hs sandbox --help`
Screenshot 2024-08-23 at 11 17 21
Screenshot 2024-08-23 at 11 17 33

## TODO

- Should we be showing a message when someone tries to run `hs sandbox sync` that this command has been deprecated instead of showing that error message (above screenshot) where it says `Unknown argument: sync`?


camden11:
@adamawang Commented this on the local-dev-lib PR as well, but this PR can be based off the `next` branch rather than `main` since its going to go out as part of the next major release. All of our breaking changes are sitting on that branch currently

camden11:
LGTM otherwise though (assuming we are OK with not including a warning, which, based on the slack thread I think everyone was on board with)

adamawang:
Hrm, not sure why the build is failing a dependency check as no packages have changed
```
yarn run v1.22.22
$ node bin/depcheck.js
Checking for dependency issues with "cli"...
The package at "cli" is missing deps.
Missing the following:  [ 'node:util' ]
Checking for dependency issues with "serverless-dev-runtime"...
Checking for dependency issues with "webpack-cms-plugins"...
There were missing or unused dependencies in one or more packages.
error Command failed with exit code 1.
```

joe-yeager:
@adamawang It looks like the dependency check might not support the new syntax for the node standard libraries. PR to fix it on the next branch https://github.com/HubSpot/hubspot-cli/pull/1163

Also just an side, the tests will fail on the next branch build action because it is depending LDL features that are unreleased

joe-yeager:
@adamawang Fix merged, you might need to pull `next` to apply it

adamawang:
Closing this in favor of https://github.com/HubSpot/hubspot-cli/pull/1168

PR #1152, opened 2024-08-23
Fix bug with hs project dev --account flag on private apps

camden11:
## Description and Context
`hs project dev` does not support local development of private apps on developer accounts, but you can currently use the `--account` flag to bypass this. The CLI will then attempt to upload your project, which will fail. This fixes that and adds a new, more helpful error message.

Note: not sure if this will affect those rare accounts that are simultaneously developer and standard accounts? But IIRC we're already not supporting those properly.

## Screenshots
Before
Screenshot 2024-08-23 at 4 10 30 PM

After
Screenshot 2024-08-23 at 4 10 53 PM

## Who to Notify

### DIFF
--- a/packages/cli/lang/en.lyaml
+++ b/packages/cli/lang/en.lyaml
@@ -1017,10 +1017,10 @@ en:
     localDev:
       confirmDefaultAccountIsTarget:
         declineDefaultAccountExplanation: "To develop on a different account, run {{ useCommand }} to change your default account, then re-run {{ devCommand }}."
-      checkIfAppDevloperAccount: "This project contains a public app. Local development of public apps is only supported on developer accounts and developer test accounts. Change your default account using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
+      checkIfAppDevloperAccount: "This project contains a public app. Local development of public apps is only supported on developer accounts and developer test accounts. Change your default account using {{ useCommand }}, or link a new account with {{ authCommand }}."
       validateAccountOption:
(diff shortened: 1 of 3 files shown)


packages/cli/lang/en.lyaml
```diff
@@ -1018,8 +1018,9 @@ en:
...
       checkIfAppDevloperAccount: "This project contains a public app. Local development of public apps is only supported on developer accounts and developer test accounts. Change your default account using {{#bold}}`hs accounts use`{{/bold}}, or link a new account with {{#bold}}`hs auth`{{/bold}}."
-      checkIfDeveloperTestAccount: "This project contains a public app. The \"--account\" flag must point to a developer test account to develop this project locally. Alternatively, change your default account to an App Developer Account using {{#bold}}`hs accounts use`{{/bold}} and run {{#bold}}`hs project dev`{{/bold}} to set up a new Developer Test Account."
-      suggestRecommendedNestedAccount:
+      validateAccountOption:
+        invalidPublicAppAccount: "This project contains a public app. The \"--account\" flag must point to a developer test account to develop this project locally. Alternatively, change your default account to an App Developer Account using {{#bold}}`hs accounts use`{{/bold}} and run {{#bold}}`hs project dev`{{/bold}} to set up a new Developer Test Account."
```
kemmerle, line 1022: We have a helper function called `uiCommandReference` that you can use here to style the commands. Same goes for `invalidPrivateAppAccount` below.

kemmerle (approved):
Made one suggestion, but otherwise, the PR LGTM :shipit:

packages/cli/lang/en.lyaml
camden11: Oh good call! Always forget about that one

PR #1154, opened 2024-08-26
Fix(migrate-app): project names with slashes causing download issues

kemmerle:
## Description and Context

We discovered a bug with the `migrate-app` command, where we were failing to download projects with `/` in their names. To address this, I have sanitized project names when passing them to the `extractZipArchive` function. On Mitchell's request, I also ensured that we are encoding project urls correctly. 

## Screenshots

Before:

Screenshot 2024-08-26 at 12 36 30 PM

After: 

Screenshot 2024-08-26 at 12 36 40 PM

## TODO

- [ ] Test out migrating projects with problematic characters
- [ ] Test out the project details link in the `migrate-app` command (See screenshot above)
- [ ] Address any feedback 

## Who to Notify

@joe-yeager @brandenrodgers @camden11

### DIFF
--- a/packages/cli/commands/project/cloneApp.js
+++ b/packages/cli/commands/project/cloneApp.js
@@ -38,7 +38,7 @@ const {
   checkCloneStatus,
   downloadClonedProject,
 } = require('@hubspot/local-dev-lib/api/projects');
-const { getCwd } = require('@hubspot/local-dev-lib/path');
+const { getCwd, sanitizeFileName } = require('@hubspot/local-dev-lib/path');
 const { logger } = require('@hubspot/local-dev-lib/logger');
 const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
 const { extractZipArchive } = require('@hubspot/local-dev-lib/archive');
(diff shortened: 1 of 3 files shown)


packages/cli/commands/project/cloneApp.js
```diff
@@ -119,10 +119,15 @@ exports.handler = async options => {
 
       // Extract zipped app files and place them in correct directory
       const zippedApp = await downloadClonedProject(accountId, exportId);
-      await extractZipArchive(zippedApp, name, absoluteDestPath, {
-        includesRootDir: true,
-        hideLogs: true,
-      });
+      await extractZipArchive(
+        zippedApp,
+        sanitizeFileName(name),
```
joe-yeager, line 124: This isn't blocking, but I wonder if we should move this to the LDL side after we ship this and unblock the beta participant.

joe-yeager (approved):
LGTM!  Thanks @kemmerle

packages/cli/commands/project/cloneApp.js
kemmerle: You read my mind.

PR #1156, opened 2024-08-27
fix: Prevent js error during initial upload for watch

brandenrodgers:
## Description and Context

There's a scenario where handleProjectUpload can return undefined. It happens when the callbackFunc doesn't return anything. Now it will default to an empty object.

You can test this by running `hs project watch --initial-upload`. Currently it will throw this error: `[ERROR] A TypeError has occurred. Cannot read properties of undefined (reading 'uploadError')`

### DIFF
--- a/packages/cli/commands/project/watch.js
+++ b/packages/cli/commands/project/watch.js
@@ -132,7 +132,7 @@ exports.handler = async options => {
         startWatching
       );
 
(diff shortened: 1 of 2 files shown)


kemmerle (approved):
:shipit:

camden11 (approved):

PR #1158, opened 2024-08-27
test: Add Slack group id to test notification

brandenrodgers:
## Description and Context

Adding a group ID as per the docs for the notify action: https://github.com/ravsamhq/notify-slack-action


camden11 (approved):

joe-yeager (approved):
:shipit:

PR #1159, opened 2024-08-27
feat: Reduce the time for hs accounts clean to detect inactive accounts

joe-yeager:
## Description and Context

Depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/187

Forces the token to refresh so we can detect inactive accounts faster

## Todo
- [x] Publish LDL and bump version

## Who to Notify


adamawang (approved):

camden11 (approved):
Nice. Pretty big upgrade for like, 4 total lines changed

PR #1160, opened 2024-08-28
Exit local dev earlier for projects without components

camden11:
## Description and Context
You can't run local dev if your project doesn't have any apps, but right now `hs project dev` doesn't tell you that until after prompting you to select a sandbox to test with. The way that the `dev` command is set up now, the `LocalDevManager` is responsible for determining runnable components, which happens _after_ a lot of prompting. A more robust solution would involve determining runnable components earlier in the `dev` process, but I think this will work fine for now and we can make things more flexible when we rethink the dev servers/dev command.

## Screenshots

Before (you get prompted)
Screenshot 2024-08-28 at 12 55 11 PM

After (exits immediately)
Screenshot 2024-08-28 at 2 07 45 PM

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -86,15 +86,21 @@ exports.handler = async options => {
   validateProjectConfig(projectConfig, projectDir);
 
   const components = await findProjectComponents(projectDir);
-  const componentTypes = getProjectComponentTypes(components);
+  const runnableComponents = components.filter(component => component.runnable);
+  const componentTypes = getProjectComponentTypes(runnableComponents);
   const hasPrivateApps = !!componentTypes[COMPONENT_TYPES.privateApp];
   const hasPublicApps = !!componentTypes[COMPONENT_TYPES.publicApp];
 
-  if (hasPrivateApps && hasPublicApps) {
-    logger.error(i18n(`${i18nKey}.errors.invalidProjectComponents`));
+  if (runnableComponents.length === 0) {
+    logger.error(
+      i18n(`${i18nKey}.errors.noRunnableComponents`, {
+        projectDir,
+        command: uiCommandReference('hs project add'),
+      })
+    );
     process.exit(EXIT_CODES.SUCCESS);
-  } else if (!hasPrivateApps && !hasPublicApps) {
-    logger.error(i18n(`${i18nKey}.errors.noSupportedComponents`));
+  } else if (hasPrivateApps && hasPublicApps) {
+    logger.error(i18n(`${i18nKey}.errors.invalidProjectComponents`));
     process.exit(EXIT_CODES.SUCCESS);
   }
 
@@ -227,7 +233,7 @@ exports.handler = async options => {
   }
 
   const LocalDev = new LocalDevManager({
-    components,
+    runnableComponents,
     debug: options.debug,
     deployedBuild,
     isGithubLinked,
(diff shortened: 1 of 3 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -93,6 +93,9 @@ exports.handler = async options => {
   if (hasPrivateApps && hasPublicApps) {
     logger.error(i18n(`${i18nKey}.errors.invalidProjectComponents`));
     process.exit(EXIT_CODES.SUCCESS);
+  } else if (!hasPrivateApps && !hasPublicApps) {
+    logger.error(i18n(`${i18nKey}.errors.noSupportedComponents`));
+    process.exit(EXIT_CODES.SUCCESS);
```
joe-yeager, line 98: Should this be `EXIT_CODES.ERROR`?

joe-yeager: Also, I just saw the `if` statement above it is also a`EXIT_CODES.SUCCESS`, but it feels like that should also be an error

camden11: Not 100% sure if this is correct, but I've been using `SUCCESS` to indicate that the command functioned as intended. I.e. an invalid project state doesn't indicate any problems with how the dev command actually ran, so its still successful

packages/cli/commands/project/dev.js
```diff
@@ -93,6 +93,9 @@ exports.handler = async options => {
   if (hasPrivateApps && hasPublicApps) {
     logger.error(i18n(`${i18nKey}.errors.invalidProjectComponents`));
     process.exit(EXIT_CODES.SUCCESS);
+  } else if (!hasPrivateApps && !hasPublicApps) {
```
brandenrodgers, line 96: The `components` should all have a `runnable` field on them. That's what the local dev manager uses here to check if there's anything that can run locally in the project. It might make sense to leverage that field so this can scale with new runnable components more easily.

packages/cli/commands/project/dev.js
brandenrodgers: Interesting. Do we have this recorded in our guidelines anywhere? I've been following a similar pattern to what Camden mentioned. This seems like something we should have clear guidelines for though. Technically the command succeeded, but the project that you ran the command on wasn't configured in a way that let the command do everything it wanted to do. 

There's a breakdown of common exit codes [here](https://tldp.org/LDP/abs/html/exitcodes.html). None of them seem to capture this case, which makes me think that success is the right move here.

joe-yeager: Interesting, I've always thought of it as the `SUCCESS` code means that it made it all the way through the happy path, so I looked at what a couple of common cli utils do and they contradict

```
🚀🚀🚀 ➜ cat notReal; echo $?  
cat: notReal: No such file or directory
0

🚀🚀🚀 ➜ ls notReal; echo $?
ls: notReal: No such file or directory
1

```


I think this is something we might want to standardize in the long term, but that's no something we need to worry about in this PR!

I'll mark this resolved and ✅

joe-yeager (approved):

packages/cli/commands/project/dev.js
camden11: That works too! Can move that check out of the `LocalDevManager`. The rest of `dev` is already kind of set up assuming that private/public apps are the only runnable components (and will require some updates if that changes) which is why I just went with this

packages/cli/commands/project/dev.js
camden11: Yeah, it's not consistent at all within the CLI either. From the little research I've done into exit codes they seem fairly inconsequential (and i guess `cat` and `ls` working differently is evidence for this), but totally agree that we should decide on one approach and use it consistently

camden11:
Ok, I updated this to move the check for runnable components out of the `LocalDevManager` and use existing (and more flexible) error copy

kemmerle (approved):
LGTM! Love the improvement to user experience 🚢

PR #1164, opened 2024-08-29
hs project dev: don't allow selection of nonexitent dev test accounts + more cleanup

camden11:
## Description and Context
`hs project dev` uses your local `hs.config.json` as the source of truth for developer test account selection. This can be out of date and allow users to select previously authed dev test accounts that no longer exist. For sandboxes, there's a check for this, since the CLI checks to see if the project exists in the sandbox. But for dev test accounts, the project lives in the parent account, and we never make this check. Currently, users can go all the way through the dev flow with a nonexistent accounts.

Since the dev test account prompt already fetches developer test account data to list un-authed dev test accounts as choices, we can use your live list of test accounts from the `integrators/test-portals/v2` as the main source of truth instead. The prompt now cross references the response from the endpoint with your local config to determine which accounts already exist locally and which need to be added config. Nonexistent accounts are no longer shown.

I also cleaned up a few other things here and there

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -123,9 +123,6 @@ exports.handler = async options => {
     }
   }
 
-  let createNewSandbox = false;
-  let createNewDeveloperTestAccount = false;
-
   // The user is targeting an account type that we recommend developing on
   if (!targetProjectAccountId && defaultAccountIsRecommendedType) {
     targetTestingAccountId = accountId;
@@ -153,6 +150,9 @@ exports.handler = async options => {
   } else if (!targetProjectAccountId && hasPublicApps) {
     checkIfAppDeveloperAccount(accountConfig);
   }
+
+  let createNewSandbox = false;
+  let createNewDeveloperTestAccount = false;
 
   if (!targetProjectAccountId) {
     const {
--- a/packages/cli/lib/localDev.js
+++ b/packages/cli/lib/localDev.js
@@ -129,13 +129,6 @@ const suggestRecommendedNestedAccount = async (
         `${i18nKey}.validateAccountOption.publicAppNonDeveloperTestAccountWarning`
       )
     );
-  } else if (isAppDeveloperAccount(accountConfig)) {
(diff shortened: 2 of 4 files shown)


packages/cli/commands/project/dev.js
```diff
@@ -154,6 +151,9 @@ exports.handler = async options => {
     checkIfAppDeveloperAccount(accountConfig);
   }
 
+  let createNewSandbox = false;
+  let createNewDeveloperTestAccount = false;
+
```
camden11, line 156: Just moving these declarations closer to where they're actually used

packages/cli/lib/localDev.js
```diff
@@ -129,13 +129,6 @@ const suggestRecommendedNestedAccount = async (
         `${i18nKey}.validateAccountOption.publicAppNonDeveloperTestAccountWarning`
       )
     );
-  } else if (isAppDeveloperAccount(accountConfig)) {
-    logger.error(
-      i18n(
-        `${i18nKey}.validateAccountOption.privateAppInAppDeveloperAccountError`
-      )
-    );
-    process.exit(EXIT_CODES.ERROR);
```
camden11, line 138: This check was redundant with another check for the same thing made earlier in the dev flow

packages/cli/lib/prompts/projectDevTargetAccountPrompt.js
```diff
@@ -109,13 +108,8 @@ const selectDeveloperTestTargetAccountPrompt = async (
...
   }
 
-  const devTestAccounts = accounts
-    .reverse()
-    .filter(
-      config =>
-        isDeveloperTestAccount(config) &&
-        config.parentAccountId === defaultAccountId
-    );
+  const accountIds = accounts.map(account => account.portalId);
```
joe-yeager, line 111: Nit: It seems like we are only using this in one `if` statement, so it could just be within that statement.

joe-yeager (approved):
Great catch!  LGTM, just one non blocking nit

PR #1165, opened 2024-08-30
test: Add acceptance tests

brandenrodgers:
## Description and Context

Adding a few very simple acceptance tests. I found and fixed some bugs while looking into this. I'll call them out below. I also added a new pattern for tests that makes it easy to start fresh with a reliable config file without having to make a request to fetch a new access token for each test.


acceptance-tests/tests/helpers/auth.js
```diff
@@ -0,0 +1,48 @@
+const yaml = require('js-yaml');
+const rimraf = require('rimraf');
+const { existsSync, readFileSync, writeFileSync } = require('fs');
+const cmd = require('./cmd');
+const { CONFIG_FILE_NAME } = require('../../lib/constants');
+
+let PARSED_CONFIG_YAML;
+
+async function initializeAuth() {
```
brandenrodgers, line 9: This initializes, parses, and then removes the cli config file so we can use the values to easily write new config files later in the tests.

packages/cli/lib/prompts/createProjectPrompt.js
```diff
@@ -59,14 +59,20 @@ const createProjectPrompt = async (
   skipTemplatePrompt = false
 ) => {
   let projectTemplates = [];
+  let selectedTemplate;
```
brandenrodgers, line 62: The `hs project create` command wasn't using the `--template` option correctly. That option takes in a string, but all of the following utils are expecting an object that has a shape like [this](https://github.com/HubSpot/hubspot-project-components/blob/main/config.json#L10-L15). Now we'll check to see if the template option matches any of the available templates in the repo's config.

camden11 (commented):
Overall looks good but looks like the `hs list` test is failing

camden11 (approved):

PR #1171, opened 2024-09-10
Handle case where account flag test account parent isn't authed on hs project dev

camden11:
## Description and Context
We were only confirming that a developer test account's parent account was authed in the CLI if it was the default account. This adds that check for developer test accounts passed in via the `--account` flag

## Screenshots
Before
Screenshot 2024-09-10 at 3 42 42 PM

After
Screenshot 2024-09-10 at 3 43 16 PM

## Who to Notify

### DIFF
--- a/packages/cli/lib/localDev.js
+++ b/packages/cli/lib/localDev.js
@@ -106,7 +106,7 @@ const checkIfParentAccountIsAuthed = accountConfig => {
         ),
       })
     );
-    process.exit(EXIT_CODES.ERROR);
+    process.exit(EXIT_CODES.SUCCESS);
   }
 };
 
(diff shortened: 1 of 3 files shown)


packages/cli/lib/localDev.js
```diff
@@ -95,16 +95,33 @@ const checkIfAppDeveloperAccount = accountConfig => {
...
     );
     process.exit(EXIT_CODES.SUCCESS);
+  }
+};
+
+// Confirm the default account is a developer account if developing public apps
+const validateAccountOption = (accountConfig, hasPublicApps) => {
+  if (hasPublicApps) {
+    if (!isDeveloperTestAccount) {
+      logger.error(
```
brandenrodgers, line 117: should we exit the process after this one?

camden11: good catch! yes

brandenrodgers (approved):

PR #1172, opened 2024-09-11
chore: Update linter so we can leverage modern JS features

joe-yeager:
## Description and Context

Update the linter to use `es2022` since node 18 fully supports `es2022`

https://node.green/#ES2022

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "private": true,
   "devDependencies": {
     "depcheck": "1.3.1",
-    "eslint": "7.32.0",
+    "eslint": "^8.56.0",
     "husky": "^1.3.1",
     "lerna": "^6.6.1",
     "lint-staged": "^10.5.4",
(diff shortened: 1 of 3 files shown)


camden11 (approved):

package.json
```diff
@@ -6,7 +6,7 @@
   "private": true,
   "devDependencies": {
     "depcheck": "1.3.1",
-    "eslint": "7.32.0",
+    "eslint": "^8.56.0",
```
kemmerle, line 9: OOC, why this version of eslint? Is it just the latest version from 2022?

kemmerle (approved):
:shipit:

package.json
joe-yeager: Good question.  V7 didn't seem to have support for 2022, and V9 introduced some fairly large changes to the config files.  It dropped support for the `.eslintignore`, renamed `eslintrc.js` to `eslint.config.js`, and significantly changed the format of `eslint.config.js`.

PR #1173, opened 2024-09-12
chore: Update node image version in dockerfile

joe-yeager:
## Description and Context

Update dockerfile to use node 18 image


brandenrodgers (approved):
Good find 🙏

PR #1174, opened 2024-09-12
fix: Use exec cwd option instead of npm prefix for Windows support

joe-yeager:
## Description and Context

`--prefix` is not functional on `npm` in a Windows environment.  Instead of using `prefix`, we can set the `cwd` of the child process via an option

## Who to Notify


camden11 (approved):
LGTM

PR #1177, opened 2024-09-17
Refactor config p.2

kemmerle:
## Description and Context
Base branch: #1123. 
Depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/172

This is the second part of the work to refactor the CLI config file. I decided to split the work up to make it easier to review and test. This PR enables the `hs accounts` bucket commands and accounts for the new `defaultAccount` and `accounts` properties in the new config. 

## TODO

- [ ] TEST. We need to test out the `hs auth`, `hs init`, and `hs accounts` commands with the old config, new config, and both configs in place (where we want the new config to override any old configs). In addition, we need to test as many CI commands as possible to ensure that these changes don't impact commands in unexpected ways. 
- [ ] Address any feedback. 

## Who to Notify


camden11:
@kemmerle Haven't gone all the way through this yet, but one big thing that I think could simplify this a lot. We have this [getAccountIdentifier](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/utils/getAccountIdentifier.ts#L8) util in local-dev-lib that takes any object with either a portalId or accountId and returns the one that exists. I think you could use this here, but also maybe even write another util that takes an object, gets the accountIdentifier, and then calls `getAccountId` on that

packages/cli/commands/accounts/clean.js
```diff
@@ -35,7 +35,8 @@ exports.handler = async options => {
 
   trackCommandUsage('accounts-clean', null);
 
-  const filteredTestAccounts = config.portals.filter(p =>
+  const accountsList = config.accounts || config.portals;
```
camden11, line 38: there's also a `getAccounts` util that could help you here

packages/cli/commands/accounts/list.js
```diff
@@ -97,7 +102,9 @@ exports.handler = async options => {
 
   logger.log(i18n(`${i18nKey}.configPath`, { configPath }));
   logger.log(
-    i18n(`${i18nKey}.defaultAccount`, { account: config.defaultPortal })
+    i18n(`${i18nKey}.defaultAccount`, {
+      account: config.defaultPortal || config.defaultAccount,
```
camden11, line 106: and a `getDefaultAccount`!

camden11 (commented):
Haven't tested yet, but ran through to understand the changes. I think those utils will help you a lot, but this is generally looking good

brandenrodgers:
+1 to Camden's suggestion about leveraging utils here to abstract away the different field names on the config object. Using utils to access fields on the config object will make it easier for us to port from "portal" to "account".

camden11:
Screenshot 2024-09-20 at 12 37 46 PM
When I ran `hs init` without the `useHiddenConfig` flag, the location of the config file didn't get logged correctly

camden11:
Screenshot 2024-09-20 at 12 41 13 PM
When you run `hs auth` but don't set the new account as your default account, it doesn't log the existing default account's name correctly

camden11:
Screenshot 2024-09-20 at 1 12 22 PM

This is kind of a crazy one. If you run `hs accounts rename` on a developer account, it will turn it into a standard account in the config. Not sure if this happens for other account types as well. Thought this might be an unrelated bug, but it's not happening in the live version of the CLI

packages/cli/commands/accounts/clean.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1177/commits/f9a718f1de95eac66bd6d91f19cf0a3ef74dbce3

packages/cli/commands/accounts/list.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1177/commits/a4d1c22ea6ac6b4368f6c9339bab7ae00bbebb72

kemmerle:
SUMMARY OF WORK (9/27):

1) On @camden11 and @brandenrodgers' feedback, I made use of the `getAccountIdentifier`, `getAccounts`, and `getDefaultAccount` utils. This required that I create an export for the `utils` folder in local-dev-lib. **The tests will continue to fail until I merge the corresponding local-dev-lib PR.**

2) I also fixed the bugs that @camden11 pointed out in the screenshots above, so that the correct config paths and account names appear in the `init` and `auth` commands. I also addressed a bug with the `hs accounts rename` command with a change in the corresponding local-dev-lib PR.

3) It would be great if we could test out this PR once again to search for any bugs. We need to test out the `hs auth`, `hs init`, and `hs accounts` commands with the old config, new config, and both configs in place (where we want the new config to override any old configs). In addition, we need to test as many CLI commands as possible to ensure that these changes don't impact commands in unexpected ways.

brandenrodgers:
> This required that I create an export for the utils folder in local-dev-lib. The tests will continue to fail until I merge the corresponding local-dev-lib PR.

We actually shouldn't be exporting that folder. The utils in there are intended to be for internal repo use only. Everything that gets exported should come from `lib/` or another top level folder

packages/cli/commands/accounts/clean.js
```diff
@@ -35,7 +42,8 @@ exports.handler = async options => {
 
   trackCommandUsage('accounts-clean', null);
 
-  const filteredTestAccounts = config.portals.filter(p =>
+  const accountsList = getAccounts(config);
```
brandenrodgers, line 45: Looks like `getAccounts()` doesn't take in any args anymore

packages/cli/lib/prompts/sandboxesPrompt.js
```diff
@@ -13,8 +20,8 @@ const mapSandboxAccountChoices = portals =>
     .filter(p => isSandbox(p))
     .map(p => {
       return {
-        name: uiAccountDescription(p.portalId, false),
-        value: p.name || p.portalId,
+        name: uiAccountDescription(getAccountIdentifier(p), false),
+        value: p.name || p.portalId || p.accountId,
```
brandenrodgers, line 24: Are we able to use the new `getAccountIdentifier()` util here?

packages/cli/lib/sandboxes.js
```diff
@@ -71,8 +77,10 @@ function getSandboxLimit(error) {
 
 // Fetches available sync types for a given sandbox portal
 async function getAvailableSyncTypes(parentAccountConfig, config) {
-  const parentPortalId = getAccountId(parentAccountConfig.portalId);
-  const portalId = getAccountId(config.portalId);
+  const parentId = getAccountIdentifier(parentAccountConfig);
+  const parentPortalId = getAccountId(parentId);
+  const id = getAccountIdentifier(config);
+  const portalId = getAccountId(id);
```
brandenrodgers, line 83: Is `getAccountId()` redundant here if we already have the id?

packages/cli/commands/init.js
```diff
@@ -126,27 +130,44 @@ exports.handler = async options => {
     process.exit(EXIT_CODES.ERROR);
   }
 
+  const doesOtherConfigFileExist = configFileExists(!useHiddenConfig);
+  if (doesOtherConfigFileExist) {
+    const path =
+      (c && path.join(getCwd(), c)) || getConfigPath('', !useHiddenConfig);
```
brandenrodgers, line 136: I think we may want to omit the `(c && path.join(getCwd(), c))` portion of this. If this IF block triggers, it means that  `configFileExists()` returned true. That util doesn't account for the `--c` flag. So ideally we'd be showing the user the exact config file that triggered the error, which (I think) would be the one returned by `getConfigPath()`.

This also got me wondering, should we allow devs to use the `--c` and `--useHiddenConfig` flags simultaneously?

kemmerle: I've omitted `(c && path.join(getCwd(), c))`. 

We do include a check for whether the c and useHiddenConfig flags both exist:

https://github.com/HubSpot/hubspot-cli/blob/2f261c2a8b0baa6e00e1537ccd4dbb76ddc43758/packages/cli/commands/init.js#L140-L143

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1177/commits/44cb0a0a0b10961680f9baf9372ab0b04ab65389

packages/cli/lib/prompts/sandboxesPrompt.js
kemmerle: https://github.com/HubSpot/hubspot-cli/pull/1177/commits/fe36cf45f176212ab281ac82be4080809b85b48d

packages/cli/commands/accounts/clean.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1177/commits/70875323cba89bb8f18ea7252527a2b6a008e6df

packages/cli/lib/sandboxes.js
kemmerle: `getAccountId()` is not redundant for the deprecated config. It processes non-numerical account ids:

https://github.com/HubSpot/hubspot-local-dev-lib/blob/2d681abf5d087afd71a6e6a6f2bbbc15e21dbbe9/config/config_DEPRECATED.ts#L396-L436

packages/cli/commands/init.js
```diff
@@ -189,6 +209,10 @@ exports.builder = yargs => {
       describe: i18n(`${i18nKey}.options.account.describe`),
       type: 'string',
     },
+    useHiddenConfig: {
+      describe: i18n(`${i18nKey}.options.useHiddenConfig.describe`),
+      type: 'boolean',
+    },
```
joe-yeager, line 215: Should we add a `.conflicts('config', 'useHiddenConfig')` since they are at odds with each other?

joe-yeager (approved):
LGTM!  Just one minor comment about leveraging `.conflicts` to prevent users from using conflicting options

packages/cli/commands/init.js
kemmerle: That sounds great! I'm manually checking for the conflict right now, but this is so much easier 🥳

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1177/commits/b5d40ded978b7c7d0858448a30cd65b2badd1448

brandenrodgers (approved):
lgtm!

PR #1180, opened 2024-09-20
Fix account validation in hs project dev

camden11:
## Description and Context
At some point through all of the recent `hs project dev` updates, I accidentally removed the check for developer accounts when developing private apps. This restores that check, and tries to streamline the account check logic a bit by putting it in the same place for default accounts and accounts passed in through the account flag. The whole process is still messy, but I think this is an improvement.

I also changed the error message from the previous one to make it more consistent for the public app error message.

Would appreciate if everyone could test and make sure this works as expected!
* For private apps: Developer accounts should fail, all other accounts should work
* For public accounts: Only developer accounts and developer test accounts should work

## Screenshots
The bug:
Screenshot 2024-09-20 at 2 01 02 PM

Fixed:
Screenshot 2024-09-20 at 2 02 14 PM

## Who to Notify

### DIFF
--- a/packages/cli/commands/project/dev.js
+++ b/packages/cli/commands/project/dev.js
@@ -40,13 +40,13 @@ const {
 const {
   confirmDefaultAccountIsTarget,
   suggestRecommendedNestedAccount,
-  checkIfAppDeveloperAccount,
+  checkIfDefaultAccountIsSupported,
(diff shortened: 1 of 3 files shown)


brandenrodgers (approved):
This lgtm 👍

PR #1181, opened 2024-09-23
chore: Convert ATs to Vitest

joe-yeager:
## Description and Context

- Updates the Acceptance tests to use `vitest` 
- Refactors the acceptance tests to be self contained so they can be safely ran in parallel
- Updates the test suites to run in parallel
- Updates the acceptance tests to no longer open a browser when `open` is called

## Who to Notify


acceptance-tests/README.md
```diff
@@ -4,7 +4,7 @@ This project tests our CLI's `hs <*>` commands as if they were being used by an
 
 ## Getting Started
 
-The main test execution is kicked off by running `yarn test-cli` from the root of `hubspot-cli`. This will run the `run-tests` script which itself is a CLI, so to see the available options, you can `run-tests --help`. This was done so that the tests can be installed against any given version of the CLI and tested against it. In addition, it was done to keep the test logic separated from the CLI library.
+The main test execution is kicked off by running `yarn test-cli` from the root of `hubspot-cli`. This will run the `setup.ts` script which itself is a CLI, so to see the available options, you can `setup.ts --help`. This was done so that the tests can be installed against any given version of the CLI and tested against it. In addition, it was done to keep the test logic separated from the CLI library.
```
joe-yeager, line 7: TODO: Fix these, these were updated by refactoring

package.json
```diff
@@ -25,7 +25,7 @@
     "publish-release": "branch=$(git rev-parse --abbrev-ref HEAD) && [ $branch = main ] && yarn lerna publish --conventional-graduate --exact || echo 'Error: New release can only be published on main branch' >&2",
     "publish-prerelease": "branch=$(git rev-parse --abbrev-ref HEAD) && [ $branch = main ] && yarn lerna publish prerelease --preid beta --dist-tag next --exact || echo 'Error: New release can only be published on main branch' >&2",
     "test": "jest",
-    "test-cli": "cd acceptance-tests && yarn test-headless --stream",
+    "test-cli": "yarn --cwd 'acceptance-tests' test-ci",
```
joe-yeager, line 28: The `--cwd` sets the `cwd` of the `yarn` command so we don't need to `cd`

acceptance-tests/vite.config.ts
```diff
@@ -0,0 +1,12 @@
+/// <reference types="vitest" />
+import { defineConfig } from 'vite';
+
+export default defineConfig({
+  test: {
+    environment: 'node',
+    include: ['./tests/**/*.spec.ts'],
+    exclude: ['./lib/**/*'],
+    testTimeout: 30000,
+    hookTimeout: 20000,
```
joe-yeager, line 10: I doubt the tests should take this long, I just went a little high to prevent false positives

acceptance-tests/lib/cmd.ts
```diff
@@ -37,53 +36,46 @@ function createProcess(cliPath, cliVersion, args = [], env = null) {
...
   if (!Array.isArray(inputs)) {
     opts = inputs;
     inputs = [];
   }
 
-  if (global.config.headless) {
-    opts.env = { BROWSER: 'none' };
-  }
+  // Prevent the browser from opening when `open` is called
+  opts.env = { BROWSER: 'none' };
```
joe-yeager, line 61: I removed the check for the `headless` config value, because I can't think of why we would want to open browser windows.  We don't plan on interacting with them from the tests and all it does is make your system borderline unusable for the duration of the tests.

acceptance-tests/lib/env.ts
```diff
@@ -66,7 +51,7 @@ const getTestConfig = () => {
   }
 
   if (!config.cliPath && !config.cliVersion) {
-    const defaultPath = path.join(process.cwd(), DEFAULT_CLI_PATH);
```
joe-yeager, line 69: `DEFAULT_CLI_PATH` was only used in one place and it's relative to the `cwd`, so it seem to make more sense to just inline it.

acceptance-tests/package.json
```diff
@@ -27,10 +30,10 @@
     "concat-stream": "^2.0.0",
     "cross-spawn": "^7.0.3",
     "dotenv": "^8.2.0",
-    "jasmine": "^3.7.0",
-    "jasmine-terminal-reporter": "^1.0.3",
     "js-yaml": "^4.0.0",
     "rimraf": "^3.0.2",
+    "uuid": "^10.0.0",
```
joe-yeager, line 35: Add the `uuid` lib to help with file name randomization

acceptance-tests/lib/testState.ts
```diff
@@ -0,0 +1,65 @@
+import { getTestConfig } from './env';
+import { createCli } from './cmd';
+import { CLI, TestConfig } from './types';
+import { getInitPromptSequence } from './prompt';
+import { existsSync, readFileSync, writeFileSync } from 'fs';
+import yaml from 'js-yaml';
+import rimraf from 'rimraf';
+import { v4 as uuidv4 } from 'uuid';
+
+export class TestState {
```
joe-yeager, line 10: This is a helper class to manage all of the state needed to be able to run the tests.  It helps to segment the tests so they all have their own copy of the test state and can run in isolation from one another.  It handles the generation of a config file with a semi-randomized name so file system collisions should not happen between tests.

.gitignore
```diff
@@ -12,3 +12,11 @@ npm-debug.log.*
...
 coverage
 .env
+
+# HubSpot config file
+hubspot.config.yml
+
+
+# Acceptance Tests
+acceptance-tests/my-project
+acceptance-tests/*.yml
```
brandenrodgers, line 22: Good call 👍

.gitignore
```diff
@@ -12,3 +12,11 @@ npm-debug.log.*
 .DS_Store
 coverage
 .env
+
+# HubSpot config file
+hubspot.config.yml
```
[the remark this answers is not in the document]
joe-yeager: I don't remember adding this 🤔 I'll remove it haha

acceptance-tests/lib/testState.ts
```diff
@@ -0,0 +1,65 @@
...
+  private parsedYaml: ReturnType<yaml.load>;
+
+  constructor() {
+    this.config = getTestConfig();
+    this.cli = createCli(this.config);
+    this.testConfigFileName = `hs-acceptance-test.config-${uuidv4()}.yml`;
+  }
+
+  getPAK(): string | undefined {
+    return this.config?.personalAccessKey;
```
brandenrodgers, line 23: I'm seeing a few TS errors for invocations of this because some utils require string and this returns string | undefined. One example is [here](https://github.com/HubSpot/hubspot-cli/pull/1181/files#diff-a585f4efe83b1715f0e5085c2e8bec8e25c2fb3a78d31ab21c5ace7cc24b210dR23).

joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1181/commits/e0704fd48bdcfcb72c5a1bc0bcfa92dfd98b6c1f

acceptance-tests/lib/env.ts
```diff
@@ -1,13 +1,9 @@
...
+import * as dotEnv from 'dotenv';
+import { existsSync } from 'fs';
+import { TestConfig } from './types';
 
-let dotEnvConfig;
-
-(() => {
-  dotEnvConfig = dotEnv.config({ path: path.join(__dirname, '../.env') });
-})();
+let dotEnvConfig = dotEnv.config({ path: path.join(__dirname, '../.env') });
```
camden11, line 6: nit: doesn't look like this gets reassigned anymore

acceptance-tests/tests/list.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    testState = new TestState();
+    await testState.withAuth();
+  });
+
+  afterAll(() => {
+    testState.cleanup();
+  });
+
+  it('should print the correct output', async () => {
+    let val = await testState.cli.execute([
```
camden11, line 18: this should also be `const`

camden11 (commented):
Nice! This seems like such a huge upgrade. Everything here looks good to me aside from a few tiny things. 

I'm having trouble getting these to run locally. Is there anything special you need to do? Ran `yarn` and `yarn test` but getting errors

```
Error: Invalid process path hs
```

camden11 (approved):

package.json
brandenrodgers: Should we make it so that we can also run the other variations of the acceptance tests from the root of the project? Like the QA and debug ones, for example

joe-yeager: Good call, I like that

brandenrodgers (approved):
This lgtm! We chatted about this already, but just a reminder that this will cause our "latest publish" tests to fail because it makes them run correctly and there's a bug in `hs project create`

PR #1182, opened 2024-09-24
fix: Fix project commands errors found in next branch bug bash

joe-yeager:
## Description and Context

- Fixes the error where `hs project logs` was broken in the next branch
- Fixes a bug where the `hs project` subcommands where encountering a NPE instead of logging an error

### Before - hs project upload  
Screenshot 2024-09-24 at 1 30 07 PM

### After - hs project uplaod
Screenshot 2024-09-24 at 1 34 38 PM

### Before - hs project logs
Screenshot 2024-09-24 at 1 31 34 PM

### After - hs project logs
Screenshot 2024-09-24 at 1 32 20 PM

### DIFF
--- a/packages/cli/lib/errorHandlers/index.js
+++ b/packages/cli/lib/errorHandlers/index.js
@@ -16,7 +16,10 @@ const i18nKey = 'lib.errorHandlers.index';
 function logError(error, context = {}) {
   debugError(error, context);
 
-  if (shouldSuppressError(error)) {
+  if (
+    shouldSuppressError(error, context) ||
+    shouldSuppressError(error, error.context)
+  ) {
     return;
   }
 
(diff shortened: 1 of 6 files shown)


packages/cli/lib/errorHandlers/index.js
```diff
@@ -16,7 +16,10 @@ const i18nKey = 'lib.errorHandlers.index';
 function logError(error, context = {}) {
   debugError(error, context);
 
-  if (shouldSuppressError(error)) {
+  if (
+    shouldSuppressError(error, context) ||
+    shouldSuppressError(error, error.context)
```
joe-yeager, line 21: This was the cause `hs project upload` errors, the context just wasn't getting passed in here.  I added the second call here because the error may have it's own `context` prop in the even of the `HubSpotHttpError`

camden11 (approved):
Nice, LGTM

PR #1183, opened 2024-09-25
Remove lerna

camden11:


.github/workflows/build.yml
```diff
@@ -13,21 +13,15 @@ jobs:
...
-      - name: Bootstrap packages
-        run: |
-          yarn lerna bootstrap
-      - name: Force local package symlinks
-        run: |
-          yarn lerna link --force-local
       - name: Verify package dependencies
         run: |
-          yarn check-deps
+          npx depcheck
```
camden11, line 18: Now that we don't have multiple packages in the repo, this is a bit simpler and we don't need the custom script. Figured it makes sense to just use `npx` (they suggest it on their npm docs https://www.npmjs.com/package/depcheck)

docs/PublishingReleases.md
```diff
@@ -1,20 +0,0 @@
...
-## Publishing a new version
-
-1. Run `yarn publish-release`. The publish command will handle bumping the version, tagging via git, and publishing the packages to NPM.
-1. If there are publishing errors, you can run again via `yarn lerna publish from-package` ([Docs][from-package]).
-
-## Publishing a prerelease
-
-1. Run `yarn publish-prerelease`. The publish command will handle bumping the version, tagging via git, and publishing the packages to NPM under the `next` dist-tag.
-1. If there are publishing errors, you can run again via `yarn lerna publish from-package --dist-tag next` ([Docs][from-package]).
-
```
camden11, line 20: I was going to update this with the new release commands, but wasn't sure if it even made a ton of sense to have release documentation on a public repo

lib/serverlessLogs.js
```diff
@@ -7,20 +7,16 @@ const { outputLogs } = require('./ui/serverlessFunctionLogs');
...
 const {
   SCOPE_GROUPS,
   PERSONAL_ACCESS_KEY_AUTH_METHOD,
-} = require('../../../../hubspot-local-dev-lib/dist/constants/auth');
-const {
-  getAccountConfig,
-} = require('../../../../hubspot-local-dev-lib/dist/config');
-const {
-  fetchScopeData,
-} = require('../../../../hubspot-local-dev-lib/dist/api/localDevAuth');
```
camden11, line 23: somehow these imports got messed up on the `next` branch

package.json
```diff
@@ -1,34 +1,63 @@
...
+    "circular-deps": "yarn madge --circular .",
+    "check-main": "branch=$(git rev-parse --abbrev-ref HEAD) && [ $branch = main ] || (echo 'Error: New release can only be published on main branch' && exit 1)",
+    "pub": "npm publish --tag latest",
+    "push": "git push --atomic origin main v$npm_package_version",
+    "release:major": "yarn check-main && yarn version --major && yarn pub && yarn push",
+    "release:minor": "yarn check-main && yarn version --minor && yarn pub && yarn push",
+    "release:patch": "yarn check-main && yarn version --patch && yarn pub && yarn push",
+    "prerelease:major": "yarn check-main && yarn version --premajor --preeid=beta && yarn pub && yarn push",
+    "prerelease:minor": "yarn check-main && yarn version --preminor --preeid=beta && yarn pub && yarn push",
+    "prerelease:patch": "yarn check-main && yarn version --prepatch --preeid=beta && yarn pub && yarn push",
```
camden11, line 58: These are our new release scripts. They match the ones in local-dev-lib but add 3 new ones to do prereleases

README.md
```diff
@@ -1,17 +1,160 @@
...
+
+```bash
+hs hubdb clear <id or name>
+```
+
+Delete a HubDB table
+
+```bash
+hs hubdb delete <id or name>
+```
```
kemmerle, line 138: Do we want to take the opportunity to give a rundown of project and sandbox commands here?

package.json
```diff
@@ -1,34 +1,64 @@
...
-    "prettier": "^1.19.1",
-    "sass": "~1.29.0"
+    "prettier": "^1.19.1"
   },
-  "workspaces": [
-    "packages/*",
-    "acceptance-tests"
-  ],
   "scripts": {
-    "check-deps": "node bin/depcheck.js",
```
kemmerle, line 22: Can you remind me what's happening with `depcheck`? Do we not need this command anymore?

camden11: Still using it, but we didn't need a custom script anymore so we're just using npx (see my comment above on build.yml)

docs/PublishingReleases.md
brandenrodgers: Yeah I'm fine with removing this 👍 we can keep it internal

acceptance-tests/lib/constants.js
```diff
@@ -1,5 +1,5 @@
 const CONFIG_FILE_NAME = 'hs-acceptance-test.config.yml';
-const DEFAULT_CLI_PATH = '../packages/cli/bin/hs';
+const DEFAULT_CLI_PATH = '../bin/hs';
```
joe-yeager, line 2: This `const` was removed and now an inline variable is used [here](https://github.com/HubSpot/hubspot-cli/blob/main/acceptance-tests/lib/env.ts#L55)

package.json
```diff
@@ -1,34 +1,65 @@
 {
-  "name": "hubspot-local-dev-tools",
-  "version": "1.0.0",
-  "description": "Local development tools for working with HubSpot",
+  "name": "@hubspot/cli",
+  "version": "6.2.0",
+  "description": "CLI for working with HubSpot",
```
joe-yeager, line 4: This might be a pending conflict, but I think we changed this to `The official CLI for developing on HubSpot`

package.json
```diff
@@ -1,34 +1,65 @@
...
+  "license": "Apache-2.0",
   "repository": "https://github.com/HubSpot/hubspot-cli",
-  "private": true,
+  "dependencies": {
+    "@hubspot/local-dev-lib": "2.0.0",
```
joe-yeager, line 8: This should be `2.0.1`

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
joe-yeager, line 1: 🙌 Excited to see this happening

acceptance-tests/package.json
```diff
@@ -17,10 +17,10 @@
...
   },
   "scripts": {
-    "test": "vitest run",
-    "test-debug": "DEBUG=true vitest run",
-    "test-ci": "CI=true vitest run",
-    "test-qa": "QA=true vitest run"
+    "test": "yarn && vitest run",
+    "test-debug": "DEBUG=true yarn && vitest run",
+    "test-ci": "CI=true yarn && vitest run",
+    "test-qa": "QA=true yarn && vitest run"
```
joe-yeager, line 23: I'm on the fence about prefixing all of these with `yarn` instead of running the install once before running these commands.  

Also I think the commands with env variables might be incorrect.  I think they would need to `yarn && ENV_VARIABLE second command` so the env variable applies to the second command.

package.json
```diff
@@ -1,36 +1,59 @@
...
-    "packages/*",
-    "acceptance-tests"
-  ],
   "scripts": {
-    "check-deps": "node bin/depcheck.js",
-    "lint": "eslint . && prettier --list-different packages/**/*.{js,json}",
-    "prettier:write": "prettier --write packages/**/*.{js,json}",
-    "publish-release": "branch=$(git rev-parse --abbrev-ref HEAD) && [ $branch = main ] && yarn lerna publish --conventional-graduate --exact || echo 'Error: New release can only be published on main branch' >&2",
-    "publish-prerelease": "branch=$(git rev-parse --abbrev-ref HEAD) && [ $branch = main ] && yarn lerna publish prerelease --preid beta --dist-tag next --exact || echo 'Error: New release can only be published on main branch' >&2",
+    "build": "rm -rf ./dist/ && tsc --rootDir . --outdir dist && yarn copy-files",
```
joe-yeager, line 44: Can we move the `rootDir` and `outdir` settings to the `tsconfig.json`?

scripts/test.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
joe-yeager, line 1: What is this file for?

acceptance-tests/package.json
camden11: Ah good to know, I didn't realize env variables only applied to the immediate following command. But yeah that makes sense. So just add a `yarn` to the gh workflow and then do it manually when you need to run locally?

package.json
camden11: I can't remember exactly why, but we ran into an issue with doing that in local-dev-lib and inlining it here solved it. Will look into that

scripts/test.sh
camden11: Oops, didn't mean to commit this one haha

acceptance-tests/package.json
joe-yeager: Yeah, that sounds good!

jest.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   testEnvironment: 'node',
-  projects: ['<rootDir>/packages/*'],
-  roots: ['packages'],
+  roots: ['commands', 'lib'],
   collectCoverage: true,
+  testPathIgnorePatterns: ['commands/functions/test.js'],
```
brandenrodgers, line 5: Is this intentionally here? Or just leftover from a test?

camden11: This is unchanged from the existing jest.config.js. But from what I can tell, it's not actually doing anything

scripts/release.sh
```diff
@@ -0,0 +1,111 @@
...
+  fi
+
+  echo
+}
+
+handle_options "$@"
+check_main
+validate_options
+
+[[ "$tag" = "latest" ]] && prefix="" || prefix="pre"
```
brandenrodgers, line 97: Would you mind including some comments in here (mostly starting on this line and below)? It's not immediately obvious what's happening in here, and what the various flags like prefix and preid are doing

scripts/release.sh
```diff
@@ -0,0 +1,111 @@
...
+[[ "$tag" = "experimental" ]] && preid="experimental" || preid="beta"
+next_version=$(yarn --silent semver $current_version -i $prefix$version_increment --preid $preid)
```
brandenrodgers, line 99: Does `--preid` get ignored here if prefix is empty?

scripts/release.sh
camden11: Just added some comments, let me know if that clarifies things.

joe-yeager (approved):
LGTM!  Thanks for taking this on, this is a huge step forward 🙌

brandenrodgers (approved):
lgtm! 🎉

PR #1193, opened 2024-10-02
Cleaning up some older dependencies

brandenrodgers:
## Description and Context

This bumps a few of our older dependencies and also removes our direct dependency on ora

### DIFF
--- a/packages/cli/commands/create/api-sample.js
+++ b/packages/cli/commands/create/api-sample.js
@@ -7,7 +7,6 @@ const {
 const { logger } = require('@hubspot/local-dev-lib/logger');
 const path = require('path');
 const fs = require('fs-extra');
-const ora = require('ora');
 const {
   fetchFileFromRepository,
   cloneGithubRepo,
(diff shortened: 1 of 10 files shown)


packages/cli/commands/create/api-sample.js
```diff
@@ -37,14 +36,13 @@ module.exports = {
         return;
       }
     }
-    const downloadSpinner = ora(i18n(`${i18nKey}.loading.apiSamples`));
```
brandenrodgers, line 40: This download happens so quickly that there's no point in us putting a loading spinner here.

packages/cli/commands/functions/deploy.js
```diff
@@ -52,17 +51,22 @@ exports.handler = async options => {
     })
   );
 
+  SpinniesManager.init();
+
+  SpinniesManager.add('loading', {
```
brandenrodgers, line 56: I tested this and it's effectively the same behavior as before

packages/cli/commands/functions/deploy.js
```diff
@@ -72,15 +76,19 @@ exports.handler = async options => {
...
+    });
+
     if (isHubSpotHttpError(e) && e.status === 404) {
       logger.error(
         i18n(`${i18nKey}.errors.noPackageJson`, {
           functionPath,
         })
       );
-    } else if (isHubSpotHttpError(e) && e.status === 400) {
-      logger.error(e.error.message);
```
brandenrodgers, line 83: I removed this because `logError()` should be enough. I tested and the error still looks good. The error is slightly shorter (not entirely sure why), but it still contains all of the important info.

joe-yeager (approved):
Nice!  LGTM

PR #1194, opened 2024-10-03
chore: add await to project build warning logs

brandenrodgers:
## Description and Context

This makes the build warning logs show above the "View deploy of build X in HubSpot" message where it belongs. We weren't waiting for the `displayWarnLogs` function to finish before moving on. I also removed the extra space after the warning logs (only matters when this logs above the "View deploy of build X in HubSpot" message).

### Before
image

### After
image


kemmerle (approved):
:shipit: Good catch!

PR #1195, opened 2024-10-04
fix: allow public app local dev for dev test account

brandenrodgers:
## Description and Context

Fixes an issue where devs are unable to initiate local dev for a public app when a developer test account is selected as the default account in the CLI


kemmerle (approved):
:shipit:

PR #1196, opened 2024-10-04
POC auto update the CLI

brandenrodgers:
## Description and Context

This is just a quick POC to see what it might look like for us to support auto updates for the CLI. There's a lot hard-coded in here that would need to change if we actually want to go this route:
- Installing major releases might not be safe b/c it could break existing behavior
- We'll need to verify how the CLI is installed on the dev's computer (global? npm?)
- We'd need to communicate this new behavior well before releasing it

## Screenshots

Oct-04-2024 14-38-11


brandenrodgers:
closing this because it was just a poc

PR #1197, opened 2024-10-07
Fix: clone-app bugs

kemmerle:
## Description and Context
We accidentally introduced some bugs when updating our API calls in the CLI. We needed to destructure the API responses differently. 

## TODO

- [ ] Address any feedback

## Who to Notify

@robrown-hubspot @brandenrodgers @camden11 @joe-yeager


joe-yeager (approved):
LGTM!  I know we are trying to get this out ASAP, but I think it would be good add some unit tests to cover these in a follow up PR 🙏

brandenrodgers (approved):

PR #1198, opened 2024-10-07
chore: Update the tests to write to an test-output directory

joe-yeager:
## Description and Context

Updates the acceptance tests so that they run in a directory `test-output` instead of dumping files and directories in the `acceptance-tests` directory.  Adds a setup and teardown method to clean up this directory afterwards


.gitignore
```diff
@@ -17,3 +17,4 @@ coverage
 acceptance-tests/my-project
 acceptance-tests/*.yml
 acceptance-tests/*.html
+acceptance-tests/test-output
```
joe-yeager, line 20: Ignore this new directory in the event it somehow doesn't get deleted.

acceptance-tests/globalSetup.ts
```diff
@@ -0,0 +1,29 @@
+import { existsSync, mkdirSync, rmSync } from 'fs';
```
joe-yeager, line 1: This is a global setup module, so the setup and teardown methods will only run one time a piece

acceptance-tests/lib/cmd.ts
```diff
@@ -49,6 +51,7 @@ export function createProcess(
       GITHUB_TOKEN: config.githubToken,
     },
     stdio: [null, null, null, 'ipc'], // This enables interprocess communication (IPC)
+    cwd: path.join(process.cwd(), config.testDir),
```
joe-yeager, line 54: This sets all of the commands we execute to run in the test directory.  So files generated should be placed within there.

acceptance-tests/lib/testState.ts
```diff
@@ -23,14 +28,26 @@ export class TestState {
...
 
+  getTestConfigFileName() {
+    return this.getPathWithinTestDirectory(this.testConfigFileName);
+  }
+
+  getPathWithinTestDirectory(filepath: string) {
+    return path.join(this.testOutputDir, filepath);
+  }
+
+  existsInProjectFolder(filepath: string) {
```
brandenrodgers, line 43: What's the word "project" referring to here? A generic project or a literal hubspot developer project?

acceptance-tests/lib/testState.ts
```diff
@@ -6,15 +6,20 @@ import { existsSync, readFileSync, writeFileSync } from 'fs';
...
+import * as path from 'node:path';
+
+export const testOutputDir = 'test-output';
 
 export class TestState {
   public config: TestConfig;
   public cli: CLI;
   private testConfigFileName: string;
   private parsedYaml: ReturnType<yaml.load>;
+  private testOutputDir = 'test-output';
```
brandenrodgers, line 18: Should this use the variable defined above? Just in case we ever change it

joe-yeager: Whoops, yep.  Good catch

joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1198/commits/e7b4c3d95d8a072afc89a652ff7853e8fd62a9c3

brandenrodgers (approved):
lgtm!

acceptance-tests/lib/testState.ts
joe-yeager: I went with `existsInTestOutputDirectory`

PR #1199, opened 2024-10-08
fix: Fix tests running on latest

joe-yeager:
## Description and Context

- Update the `lastestPublish` action to globally install the CLI
- Update the tests to no longer use `npx` and instead use the globally installed `hs` command.  `npx` was causing a ton of flakey behavior.
- Dockerize the tests
- Update the `test-docker` docker `npm` script to run the ATs
- Remove the `CLI_VERSION` env variable in favor of `USE_INSTALLED`.  
    - If you want to test a particular version you can do either: 
      - `npm install -g @hubspot/cli@version && npm run test-cli`
      - Update the version being installed in the Dockerfile and run `npm run test-docker`

### DIFF
--- a/acceptance-tests/tests/create.spec.ts
+++ b/acceptance-tests/tests/create.spec.ts
@@ -118,9 +118,6 @@ describe('hs create', () => {
     ).toBe(true);
   });
 
-  // For some reason, this test is getting tripped up on the creation.
-  // I verified it's just the test though, not the code, so
-  // instead we just check for some output to make sure the command runs
   it('api-sample', async () => {
     const out = await testState.cli.execute(
       ['create', FOLDERS.apiSample.name, FOLDERS.apiSample.name],
--- a/acceptance-tests/tests/create.spec.ts
+++ b/acceptance-tests/tests/create.spec.ts
@@ -1,6 +1,5 @@
 import { ENTER } from '../lib/prompt';
 
-import { existsSync } from 'fs';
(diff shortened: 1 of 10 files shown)


acceptance-tests/tests/create.spec.ts
```diff
@@ -16,22 +15,27 @@ const FOLDERS = {
   },
   websiteTheme: {
     folder: 'website-theme',
+    name: 'website-theme',
```
joe-yeager, line 18: I was in this file and noticed these were inconsistently used and used made it consistent.  This is pretty much unrelated to the rest of the changes.

package.json
```diff
@@ -28,7 +28,7 @@
     "test-cli": "yarn --cwd 'acceptance-tests' test-ci",
     "test-cli-debug": "yarn --cwd 'acceptance-tests' test-debug",
     "test-cli-qa": "yarn --cwd 'acceptance-tests' test-qa",
-    "test-docker": "yarn build-docker && docker container run -it --rm --name=hs-cli-container hs-cli-image yarn test",
+    "test-docker": "yarn build-docker && docker container run -it --rm --name=hs-cli-container hs-cli-image yarn test-cli",
```
joe-yeager, line 31: This was set to run the unit tests and I changed it to use the ATs instead.  I'm not sure why we would want to run the unit tests in a docker container, but I can add another command if we would like to.

brandenrodgers (approved):
🎉

PR #1200, opened 2024-10-09
Use workflows in acceptance tests

brandenrodgers:
## Description and Context

This switches up our pattern for acceptance tests from testing individual commands to testing workflows. This is a better way for us to be testing because most commands rely on certain assets existing in either the current working directory or somewhere in HubSpot. This pattern also makes our test more accurately follow the flows that our users will be taking while working with the CLI.

This makes it easier for us to run the tests against accounts without having to go into them and manually create the assets that the tests rely on.

I left in a folder for testing individual commands because I think that option will still be valuable in some scenarios.

Note: I removed the CLI_PATH env variable because it wasn't being used anywhere. We now support running the latest installed version of the CLI so we don't need this option anymore.


acceptance-tests/tests/commands/create.spec.ts
```diff
@@ -1,8 +1,7 @@
-import { ENTER } from '../lib/prompt';
-
 import { describe, beforeAll, afterAll, it, expect } from 'vitest';
 import rimraf from 'rimraf';
-import { TestState } from '../lib/testState';
+import { ENTER } from '../../lib/prompt';
+import { TestState } from '../../lib/TestState';
```
joe-yeager, line 4: The file name is `testState`, so this will work on macos since it's not case sensitive, but will fail on other OS

acceptance-tests/lib/cmd.ts
```diff
@@ -4,13 +4,12 @@
...
  */
 import { CLI, TestConfig } from './types';
-import { existsSync } from 'fs';
 import { constants } from 'os';
 import spawn from 'cross-spawn';
 import concat from 'concat-stream';
 import * as process from 'node:process';
 import * as path from 'node:path';
-import { testOutputDir } from './testState';
+import { testOutputDir } from './TestState';
```
joe-yeager, line 12: Same here

acceptance-tests/tests/workflows/cmsTemplateFlow.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, beforeAll, it, expect, afterAll } from 'vitest';
+import { TestState } from '../../lib/TestState';
```
joe-yeager, line 2: Same here

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
```diff
@@ -0,0 +1,111 @@
+import { describe, beforeAll, it, expect, afterAll } from 'vitest';
+import { TestState } from '../../lib/TestState';
```
joe-yeager, line 2: Same here

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+      expect(
+        testState.existsInTestOutputDirectory(
+          testState.getTestConfigFileNameRelativeToOutputDir()
+        )
+      ).toBe(false);
+    });
+  });
+
+  describe('hs init', () => {
+    it('should create a new config file', async () => {});
```
joe-yeager, line 28: Do we need this?

brandenrodgers: Good catch, removed in [aa8facf](https://github.com/HubSpot/hubspot-cli/pull/1200/commits/aa8facfdf6ade43b9dc39b3533031b715dadecad)

acceptance-tests/tests/commands/create.spec.ts
brandenrodgers: Ah right, I committed the file casing change which should fix this

acceptance-tests/lib/cmd.ts
```diff
@@ -64,12 +60,10 @@ function executeWithInput(
     inputs = [];
   }
 
-  // Prevent the browser from opening when `open` is called
-  opts.env = { BROWSER: 'none' };
```
brandenrodgers, line 68: I moved this into the `spawn()` call above because I think it makes more sense there. This will also prevent us from overriding any existing `opts.env` values that get passed into `executeWithInput()`

acceptance-tests/lib/cmd.ts
```diff
@@ -113,14 +107,14 @@ function executeWithInput(
...
-      console.log('error:', err.toString());
+      console.log('error:', String(err));
     }
   });
 
   // Get output from CLI for debugging
   childProcess.stdout.on('data', (data: unknown) => {
     if (config.debug) {
-      console.log('output:', data.toString());
+      console.log('output:', String(data));
```
brandenrodgers, line 117: These are all TS fixes

acceptance-tests/lib/env.ts
```diff
@@ -45,21 +44,13 @@ export const getTestConfig = (): TestConfig => {
     );
   }
 
-  if (config.cliPath && config.useInstalled) {
```
brandenrodgers, line 48: I removed `cliPath` so we don't need this logic anymore

acceptance-tests/tests/workflows/accountManagementFlow.spec.ts
```diff
@@ -0,0 +1,109 @@
+import { describe, beforeAll, it, expect, afterAll } from 'vitest';
+import { TestState } from '../../lib/TestState';
+import { getInitPromptSequence, ENTER } from '../../lib/prompt';
+
+describe('Account Management Flow', () => {
```
brandenrodgers, line 5: This is an example of a workflow. It tests:
1. the config file shouldn't exist initially
2. It creates a config file and checks that it exists
3. It lists out the accounts and checks to see that the auth'd account is in the list
4. It lists out info for the account
5. It removes the account and makes sure that succeeded
6. It lists the accounts again to confirm that the account doesn't exist
7. It authenticates the account again and makes sure that it exists in the config

acceptance-tests/tests/workflows/cmsTemplateFlow.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ENTER } from '../../lib/prompt';
+
+const TEMPLATE = {
+  name: 'test-template',
+  folder: 'test-template.html',
+};
+
+describe('CMS Template Flow', () => {
```
brandenrodgers, line 10: This flow tests:
1. creates a template locally
2. Uploads that template to the Design Manager
3. Lists the contents of the the Design Manager and makes sure that the uploaded file is in there
4. removes the file from the design manager
5. Lists the contents again to make sure the file is gone from Design Manager

so this tests `hs create template`, `hs upload`, `hs list`, and `hs remove`

brandenrodgers:
I just pushed up some new changes:
- Add a flow for the `hs secrets` commands
- Abstract away the need to include the `--c="configPath"` argument to every command in favor of a new `testState.cli.executeWithTestConfig()` function. IMO this helps with the readability of the tests, but let me know if you all don't like this and I can switch it back to how it was.

acceptance-tests/lib/TestState.ts
```diff
@@ -1,36 +1,36 @@
...
     this.testConfigFileName = `hs-acceptance-test.config-${uuidv4()}.yml`;
+    this.cli = createCli(this.config, this.testConfigFileName);
   }
 
   getPAK() {
     return this.config?.personalAccessKey;
   }
 
-  getTestConfigFileNameRelativeToOutputDir() {
+  getTestConfigPathRelativeToOutputDir() {
```
brandenrodgers, line 29: Switching from "FileName" to "Path" here just for naming consistency.

acceptance-tests/tests/commands/projectCreate.spec.ts
```diff
@@ -23,13 +23,12 @@ describe('hs project create', () => {
   });
 
   it('should create a project containing a private app', async () => {
-    await testState.cli.execute([
+    await testState.cli.executeWithTestConfig([
```
brandenrodgers, line 26: Here's an example of us calling the new cli method `executeWithTestConfig` to prevent us from having to pass in the config override argument.

acceptance-tests/tests/commands/config.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  afterAll(() => {
+    testState.cleanup();
+  });
+
+  describe('hs config set', () => {
+    it('should set the default mode to draft', async () => {
+      await testState.cli.executeWithTestConfig(
+        ['config', 'set'],
+        [ENTER, DOWN, ENTER]
+      );
```
camden11, line 22: Might be misinterpreting this, but is this actually checking anything? Or is the check just that this doesn't fail?

acceptance-tests/tests/workflows/secretsFlow.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { describe, beforeAll, it, expect, afterAll } from 'vitest';
+import { TestState } from '../../lib/TestState';
+import { ENTER } from '../../lib/prompt';
+
+const SECRET = {
+  name: 'KRABBY_PATTY_SECRET_FORMULA',
+  value: 'nice-try-plankton',
+};
```
camden11, line 8: lol

camden11 (commented):
The `Secrets Flow > hs secrets list > should not list the secret` test is pretty flaky for me locally. Only passed twice in 5 runs. Not sure if that's a huge problem. Otherwise, just one small comment.

brandenrodgers:
> The Secrets Flow > hs secrets list > should not list the secret test is pretty flaky for me locally. Only passed twice in 5 runs. Not sure if that's a huge problem. Otherwise, just one small comment.

Oh interesting, the tests didn't fail at all for me after several runs. Let me try running them a bunch more to see if I can get the same flaky behavior.

acceptance-tests/tests/commands/config.spec.ts
brandenrodgers: Ah good catch. I had actually forgotten to finish this one. I just added an actual check in here in my latest commit

joe-yeager (approved):

PR #1201, opened 2024-10-15
chore: Update test-latest-docker command to set the env variable USE_LATEST to true

joe-yeager:
## Description and Context

I realized this wasn't being explicitly set and was relying on the `.env` file


kemmerle (approved):
Tested out locally, and tests are passing for me 👍

PR #1202, opened 2024-10-16
feat: Add `hs doctor` command

joe-yeager:
## Description and Context

Adds a `hs doctor` command to aid users in self diagnosing problems with their CLI, CLI configuration, and Project configuration

#### Outside a project
Screenshot 2024-11-07 at 3 36 26 PM

#### Inside a project 
Screenshot 2024-11-07 at 3 38 37 PM

#### Inactive account

Screenshot 2024-11-07 at 3 40 57 PM

#### Missing dependency installs
Screenshot 2024-11-07 at 3 41 53 PM

#### Port 8080 in use
Screenshot 2024-11-07 at 3 48 02 PM


bmatto:
neat.

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           yarn install --cwd=acceptance-tests
       - name: Verify package dependencies
         run: |
-          npx depcheck --ignores=semver
+          npx depcheck --ignores="semver,@types/jest"
```
joe-yeager, line 21: I'm debating adding `@types/*` to the ignore because there are valid reasons to include types without ever importing them

commands/doctor.ts
```diff
@@ -0,0 +1,73 @@
...
+export const command = 'doctor';
+export const describe = i18n(`${i18nKey}.describe`);
+
+export const handler = async ({
+  outputDir,
+}: ArgumentsCamelCase<DoctorOptions>) => {
+  const doctor = new Doctor();
+
+  try {
+    trackCommandUsage('doctor', undefined, doctor.accountId);
```
brandenrodgers, line 26: The trackCommandUsage command itself should handle the try/catch and debugging. So you should be able to just call it directly without the wrapper.

commands/doctor.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  const outputFile = path.join(
+    outputDir,
+    `doctor-${new Date().toISOString()}.json`
+  );
+
+  try {
+    fs.writeFileSync(outputFile, JSON.stringify(output, null, 4));
+    logger.success(`Output written to ${outputFile}`);
```
brandenrodgers, line 54: Looks like some copy made it in here

lang/en.lyaml
```diff
@@ -1418,3 +1425,52 @@ en:
...
+      runningDiagnostics: "Running diagnostics..."
+      diagnosticsComplete: "Diagnostics complete"
+      accountChecks:
+        active: "Default account active"
+        inactive: "Default account isn't active"
+        inactiveSecondary: "Run {{ command }} to remove inactive accounts from your CLI config"
+        unableToDetermine: "Unable to determine if the portal is active"
+        pak:
+          invalid: "Personal access key is invalid"
+          valid: "Personal Access Key valid. {{ link }}"
```
brandenrodgers, line 1438: should this have "is"?

lang/en.lyaml
```diff
@@ -275,6 +275,13 @@ en:
               success:
                 update: "Your schema has been updated in account \"{{ accountId }}\""
                 viewAtUrl: "Schema can be viewed at {{ url }}"
+    doctor:
+      describe: "The doctor is in"
```
brandenrodgers, line 279: 😆 placeholder?

One suggestion is: "Run diagnostics to detect common CLI issues"

lang/en.lyaml
```diff
@@ -1418,3 +1425,52 @@ en:
...
+        success: "node v{{ nodeVersion }} is installed"
+      npmChecks:
+        notInstalled: "npm is not installed"
+        installed: "npm v{{ npmVersion }} is installed"
+        unableToDetermine: "Unable to determine if npm is installed"
+      projectDependenciesChecks:
+        missingDependencies: "missing dependencies in {{#cyan}}{{ dir }}{{/cyan}}"
+        missingDependenciesSecondary: "Run {{ command }} to install all project dependencies locally"
+        unableToDetermine: "Unable to determine if dependencies are installed {{ dir }}"
+        success: "App dependencies are installed an up to date"
```
brandenrodgers, line 1452: Typo on "and" here

lang/en.lyaml
```diff
@@ -1418,3 +1425,52 @@ en:
...
+      files:
+        invalidJson: "invalid JSON in {{#cyan}}{{ filename }}{{/cyan}}"
```
brandenrodgers, line 1454: iirc we don't typically use cyan for filepaths. We use bold. Was this from the mocks?

lang/en.lyaml
```diff
@@ -1418,3 +1425,52 @@ en:
...
+        inUse: "Port {{ port }} is in use"
+        inUseSecondary: "Make sure it is available if before running {{ command }}"
+        available: "Port {{ port }} available for local development"
+      diagnosis:
+        cli:
+          header: "HubSpot CLI install"
+        cliConfig:
+          header: "CLI configuration"
+          configFileSubHeader: "Config File: {{#cyan}}{{ filename }}{{/cyan}}"
+          defaultAccountSubHeader: "Default Account: {{#cyan}}{{ accountName }}[{{ accountType }}]({{ accountId }}) {{/cyan}}"
```
brandenrodgers, line 1466: We have a `uiAccountDescription` ui util for logging out accounts

lang/en.lyaml
```diff
@@ -1418,3 +1425,52 @@ en:
...
+        projectConfig:
+          header: "Project configuration"
+          projectDirSubHeader: "Project dir: {{#cyan}}{{ projectDir }}{{/cyan}}"
+          projectNameSubHeader: "Project name: {{#cyan}}{{ projectName }}{{/cyan}}"
+        counts:
+          errors: '{{#bold}}Errors{{/bold}}: {{ count }}'
```
brandenrodgers, line 1472: nit, but should the colons be included in the bold here?

lib/dependencyManagement.ts
```diff
@@ -150,9 +150,18 @@ async function getProjectPackageJsonLocations() {
   return packageParentDirs;
 }
 
+export async function packagesNeedInstalled(directory) {
```
brandenrodgers, line 153: Wonder if we could come up with a slightly more informative name for this one. Some suggestions are `areDependenciesUpToDate()`, `detectAvailableDependencyUpdates()`, `areNewDependenciesAvailable()`

lib/doctor/DiagnosticInfoBuilder.ts
```diff
@@ -0,0 +1,206 @@
...
+    accountId?: number | null;
+    scopeGroups?: AccessToken['scopeGroups'];
+    enabledFeatures?: AccessToken['enabledFeatures'];
+    accountType?: AccessToken['accountType'];
+    authType?: AuthType;
+  };
+  diagnosis?: string;
+}
+
+const configFiles = [
```
brandenrodgers, line 53: There are other files that can exist in projects too, like theme.json and cms-assets.json. I wonder though, rather than hardcode this list (which we'd have to continually update as new configs get added) could we instead just search the project for files with `.json` extensions and then run validation against all of them? The assumption being that any json file in the project is worth validating.

lib/doctor/DiagnosticInfoBuilder.ts
```diff
@@ -0,0 +1,206 @@
...
+      mainModule,
+    } = this.processInfo;
+
+    return {
+      platform,
+      arch,
+      path: mainModule?.path,
+      config: getConfigPath(),
+      versions: {
+        '@hubspot/cli': pkg.version,
```
brandenrodgers, line 104: Not a blocker, but we reference `@hubspot/cli` in enough places that it might be worthwhile to add it as a constant somewhere

lib/doctor/Doctor.ts
```diff
@@ -0,0 +1,314 @@
...
+        message: i18n(`${localI18nKey}.success`),
+      });
+    }
+  }
+
+  private async isValidJsonFile(filename: string) {
+    try {
+      const readFile = util.promisify(fs.readFile);
+      const fileContents = await readFile(filename);
+      JSON.parse(fileContents.toString());
```
brandenrodgers, line 263: Could this be shortened to `JSON.parse(fs.readFileSync(filePath, 'utf-8'))`?

brandenrodgers:
I deleted my hubspot.config.yml file and ran the command. It correctly threw an error, but I think we could give a more useful message here. This is what the output looks like:
image
It might be nice to make sure that the hubspot.config.yml file exists before continuing to the access token checks

lang/en.lyaml
joe-yeager: Yeah, all the copy is awaiting input from UX 😄 

I should have mentioned that

lib/dependencyManagement.ts
joe-yeager: I agree we could have a better name, but this isn't checking for updates, it is checking the any packages would be installed.  `update` implies that there may new versions available and that is not a part of the check.  Maybe something like `hasMissingPackages`

lib/doctor/Doctor.ts
joe-yeager: That will have performance ramifications.  I strayed away from all the `fs.*Sync` methods intentionally because they are blocking and don't allow for the checks to run async in parallel.

lib/doctor/Doctor.ts
```diff
@@ -0,0 +1,314 @@
...
+      this.diagnosis?.addProjectSection({
+        type: 'success',
+        message: i18n(`${i18nKey}.files.validJson`),
+      });
+    }
+  }
+
+  private async checkIfPortsAreAvailable() {
+    const localI18nKey = `${i18nKey}.port`;
+    if (await isPortManagerServerRunning()) {
```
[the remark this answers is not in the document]
joe-yeager: Yeah, I agree that this can be augmented to check if the service running is the actual port manager service.  I think that will be a good candidate for the next iteration!

lib/doctor/Doctor.ts
```diff
@@ -0,0 +1,314 @@
...
+import { isPortManagerServerRunning } from '@hubspot/local-dev-lib/portManager';
+import { PORT_MANAGER_SERVER_PORT } from '@hubspot/local-dev-lib/constants/ports';
+import { accessTokenForPersonalAccessKey } from '@hubspot/local-dev-lib/personalAccessKey';
+import { isSpecifiedError } from '@hubspot/local-dev-lib/errors/index';
+import { getHubSpotWebsiteOrigin } from '@hubspot/local-dev-lib/urls';
+import { uiCommandReference } from '../ui';
+const { i18n } = require('../lang');
+const { uiLink } = require('../ui');
+
+const minMajorNodeVersion = 18;
```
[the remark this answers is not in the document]
joe-yeager: Kind of.  I debated using the `engines.node` prop and stripping the non-numeric values, but `engines.node` supports syntax like `>=18 <=21`.  If we think we won't get in a place where we leverage the more complex syntax, I think reading it from there makes sense

lib/dependencyManagement.ts
```diff
@@ -150,9 +150,18 @@ async function getProjectPackageJsonLocations() {
...
+  const exec = util.promisify(execAsync);
```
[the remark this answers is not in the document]
joe-yeager: That is how I originally had it functioning, but it makes the testing near impossible 😞, so I went for the more testable approach.

lang/en.lyaml
joe-yeager: I just messaged Sandra to confirm

lib/doctor/DiagnosticInfoBuilder.ts
joe-yeager: Yeah, in the lack of a BE to provide these dynamically we kind of had to hardcode them 😞 

We could just validate all of the `json` files in the project if we wanted to.  That would mean that any user defined JSON files would also be validated, so it might be a lot in some of the larger projects.  But it scales with the addition of new files better.

joe-yeager:
> I deleted my hubspot.config.yml file and ran the command. It correctly threw an error, but I think we could give a more useful message here. This is what the output looks like: image It might be nice to make sure that the hubspot.config.yml file exists before continuing to the access token checks

Good catch @brandenrodgers!

I added a check to make sure the config file exists and if it doesn't log a message.  This content is just a place holder, I'll run the content by Sandra
Screenshot 2024-11-08 at 10 49 45 AM

lib/doctor/DiagnosticInfoBuilder.ts
joe-yeager: I changed it to validate all of the JSON files because it really doesn't add that much overhead.  I updated this list of have the two files you added so it adds them to the `configFiles` list which is just a QOL item for when users submit the output file

lang/en.lyaml
joe-yeager: Changed them to bold

lib/doctor/Doctor.ts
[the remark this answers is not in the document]
joe-yeager: Since we had an issue to update the portManager shutdown flow once [this LDL PR is merged](https://github.com/HubSpot/hubspot-local-dev-lib/pull/209), I can update this to check if the port is available, and then check if it is the port manager

lib/doctor/DiagnosticInfoBuilder.ts
```diff
@@ -0,0 +1,214 @@
...
+    const accountConfig = getAccountConfig(this.accountId!);
+    this.env = accountConfig?.env;
+    this.authType = accountConfig?.authType;
+    this.accountType = accountConfig?.accountType;
+    this.personalAccessKey = accountConfig?.personalAccessKey;
+    this.processInfo = processInfo;
+  }
+
+  async generateDiagnosticInfo(): Promise<DiagnosticInfo> {
+    this._projectConfig = await getProjectConfig(null);
```
camden11, line 88: nit: i think it would be a little clearer to just do `getProjectConfig()`. this is done a few other times in the repo

lib/doctor/DiagnosticInfoBuilder.ts
```diff
@@ -0,0 +1,214 @@
...
+import { getAccountConfig, getConfigPath } from '@hubspot/local-dev-lib/config';
+import { getAccessToken } from '@hubspot/local-dev-lib/personalAccessKey';
+import { walk } from '@hubspot/local-dev-lib/fs';
+import util from 'util';
+import { exec as execAsync } from 'child_process';
+import process from 'process';
+
+export type ProjectConfig = Awaited<ReturnType<typeof getProjectConfig>>;
+
+const hubspotCli = '@hubspot/cli';
```
camden11, line 23: I can't remember if I commented this last time, so ignore me if I did and you had a good reason not to. But could we import this from package.json?

camden11 (approved):
I have a few minor comments, but nothing blocking. Overall this looks great! Super exciting to finally be getting this out soon.

lib/doctor/DiagnosticInfoBuilder.ts
joe-yeager: Yep, I'll update it!  I'm using the pakage.name elsewhere, so not sure how this one hung around

lib/doctor/DiagnosticInfoBuilder.ts
joe-yeager: I'm getting a TS error if I omit the arg since it's not optional.  I can just add a `// @ts-expect-error` so when this method is type that can be deleted

lib/doctor/DiagnosticInfoBuilder.ts
joe-yeager: Oh, now I remember why.  This value is used in a type as a key, so I need to hardcode it in at least one place, but I'll update the rest

lib/doctor/DiagnosticInfoBuilder.ts
camden11: Ohh weird, I didn't realized TS had any opinions about stuff defined in `// @ts-nocheck` files. Fine by me either way then

lib/doctor/Doctor.ts
```diff
@@ -0,0 +1,385 @@
...
+    diagnosticInfoBuilder: DiagnosticInfoBuilder = new DiagnosticInfoBuilder(
+      process
+    )
+  ) {
+    SpinniesManager.init();
+    this.accountId = getAccountId();
+    this.diagnosticInfoBuilder = diagnosticInfoBuilder;
+  }
+
+  async diagnose() {
```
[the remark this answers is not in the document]
joe-yeager: Whoops, I did miss this.  I'm fine with `Promise<DiagnosticInfo>` 👍

PR #1204, opened 2024-10-16
Add registry to package.json

camden11:
## Description and Context
Our new release setup requires specifying the npm registry (otherwise, it will default to yarn registry, which I think is due to the presence of yarn.lock)

## Who to Notify
@joe-yeager @kemmerle @brandenrodgers


kemmerle (approved):
Thanks for the explanation in the notes!

PR #1205, opened 2024-10-16
Upgrade serverless-dev-runtime to v7.0.0

camden11:
## Description and Context
Upgrades to the latest serverless-dev-runtime, which uses LDL as a peerDependency to avoid conflicts.

## Who to Notify


kemmerle (approved):
:shipit:

PR #1206, opened 2024-10-16
Add build script sh

camden11:
## Description and Context
The previous build script was not copying the README and LICENSE, which should both be published to npm. This moves the build script to its own file to keep things clean.

## Who to Notify
@brandenrodgers

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
   },
   "scripts": {
     "build": "bash ./scripts/build.sh",
-    "copy-files": "cp -r lang dist/lang && cp bin/hs dist/bin/hs && cp bin/hscms dist/bin/hscms",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "typescript": "^5.6.2"
   },
   "scripts": {
-    "build": "rm -rf ./dist/ && tsc && yarn copy-files",
+    "build": "bash ./scripts/build.sh",
     "copy-files": "cp -r lang dist/lang && cp bin/hs dist/bin/hs && cp bin/hscms dist/bin/hscms",
```
brandenrodgers, line 45: Can we delete copy-files too?

brandenrodgers (approved):
I think we can maybe remove another script, but otherwise lgtm!

camden11:
Oh good call

PR #1207, opened 2024-10-16
chore: install the deps before running the acceptance tests

joe-yeager:
## Description and Context

Update the `latestPublish` action to explicitly run the `yarn install` in the `acceptance-tests` directory before running the tests

[Clean action run](https://github.com/HubSpot/hubspot-cli/actions/runs/11373179154/job/31639179633)


kemmerle (approved):
Thanks for the quick fix!

camden11 (approved):

PR #1209, opened 2024-10-16
chore: Update ATs to run against bundled code

joe-yeager:
## Description and Context

Run the ATs against the bundled code when not using latest


kemmerle (approved):
🚢

PR #1211, opened 2024-10-18
Initial pass at adding more command structure tests

brandenrodgers:
## Description and Context

This adds some command structure tests for the auth command. It tests the command export, the describe export, and the builder export. My plan is to add this type of test for all of the commands, and then move onto adding tests for the handler portion of the commands. Having these tests in place will prevent us from unintentionally releasing breaking command structure changes.

I also went through and made the existing tests consistently formatted. It's a little bit of a nit, but it'll be useful because there are going to be a lot of these.


commands/__tests__/auth.test.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('commands/auth', () => {
+  describe('command', () => {
+    it('should have the correct command structure', () => {
+      expect(authCommand.command).toEqual('auth [type] [--account]');
+    });
+  });
+
+  describe('describe', () => {
+    it('should provide a description', () => {
+      expect(authCommand.describe).toBeDefined();
```
brandenrodgers, line 20: I saw some of the existing tests are checking for exact values here. I'm actually leaning towards removing those tests in favor of just checking for the existence of a description. Changing a description isn't a breaking change. The two impactful variations of the description are whether or not it has a `[beta] and if it's undefined.

commands/__tests__/auth.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+  });
+
+  describe('builder', () => {
+    it('should support the correct positional arguments', () => {
+      authCommand.builder(yargs);
+
+      expect(yargs.positional).toHaveBeenCalledTimes(1);
+      expect(yargs.positional).toHaveBeenCalledWith('type', {
```
brandenrodgers, line 29: Similar to my point about command description, I'm also on the fence about whether or not it's valuable for us to be verifying the exact description copy for options and positionals. I'm considering updating this to only validate things like name and type for the options.

commands/__tests__/auth.test.ts
joe-yeager: I think the benefits of checking the exact value are that they should not be changing frequently, and in the rare event they do change, we confirm it by updating the test copy as well.  It blocks accidental changes from happening.

brandenrodgers: Ok I'm fine with keeping these in 👍. I think you're right that adding these for descriptions should be fairly harmless since they don't change often. I do want us to keep it in mind that there's a point where we start to get diminishing returns by increasing the granularity of our tests. Using tests to catch changes that would alter command functionality makes clear sense to me, but catching non-breaking copy changes feels to me like something that could also be handled at review stage. In the CLI, testing for exact copy output is essentially visual regression testing, which is useful in some areas but can get clunky pretty quickly if things change often.

camden11: Yeah honestly... I guess I don't feel super strongly but I'm not sure I see a ton of value in checking for exact strings here. They shouldn't change but also the odds of them accidentally being changes (and passing a review) seem very low

commands/__tests__/auth.test.ts
camden11: That makes sense to me

camden11 (approved):

PR #1212, opened 2024-10-22
Refactor account middleware

kemmerle:
## Description and Context

Currently, we fetch the user's account ID as needed in individual commands. This requires that we call the `getAccountId` function (found in the `lib/commonOpts` file) in nearly every CLI command. In this PR, I've moved the `getAccountId` to yargs middleware and refactored it, so that it automatically sets the `--account` flag from an environmental variable, user flag, or the default account set in the config. 

This allows us to streamline code in all our CLI commands, and it eliminates duplicates terms like `account`, `accountId`, and `accountName`, so that we only need to deal with the `account` flag itself. 

NOTE: Three commands--`hs auth`, `hs accounts remove`, and `hs sandbox delete`--require the original input for the `--account` flag, which I have preserved as `options.accountFlag`. Another solution that @joe-yeager suggested in [Slack](https://hubspot.slack.com/archives/C076QV345ME/p1728583164108989?thread_ts=1728569194.918219&cid=C076QV345ME) would be to blocklist commands we don't want to auto inject the the account flag. Would love some thoughts on this. 

## TODO

- [x] Make account names case insensitive (Actually addressed in LDL: https://github.com/HubSpot/hubspot-local-dev-lib/pull/172/commits/54d69c0ae9a6a8661a2e55dbdfb5ef6772c6f1d4).
- [x] Fix acceptance tests (It was an issue with my setup; needed to merge main and run `yarn install` and `yarn build` in the `acceptance-tests` folder)
- [x] Address feedback

## Who to Notify


packages/cli/commands/auth.js
```diff
@@ -71,7 +71,7 @@ exports.describe = i18n(`${i18nKey}.describe`, {
 });
 
 exports.handler = async options => {
-  const { type, config: c, qa, account } = options;
+  const { type, config: c, qa, accountFlag } = options;
```
joe-yeager, line 74: Interesting, is the creation of the `accountFlag` something we are getting for free from `yargs`?  It looks like it since the option remained intact.  I like this better than blocking particular commands 👍 

The only thing I think might be nice to add is comment for our futures selves to remember why we are using this version instead of `account`

kemmerle: Sounds good, will add comments now!

commands/mv.ts
```diff
@@ -32,24 +31,23 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
-  trackCommandUsage('mv', null, accountId);
+  trackCommandUsage('mv', null, account);
 
   try {
-    await moveFile(accountId, srcPath, getCorrectedDestPath(srcPath, destPath));
+    await moveFile(account, srcPath, getCorrectedDestPath(srcPath, destPath));
     logger.success(
       i18n(`${i18nKey}.move`, {
-        accountId,
+        account,
```
camden11, line 42: These should all be `accountId: account` (or update the lang file to use new keys)

commands/remove.ts
```diff
@@ -19,25 +18,23 @@ exports.command = 'remove <path>';
...
-  const accountId = getAccountId(options);
-
-  trackCommandUsage('remove', null, accountId);
+  trackCommandUsage('remove', null, account);
 
   try {
-    await deleteFile(accountId, hsPath);
-    logger.log(i18n(`${i18nKey}.deleted`, { accountId, path: hsPath }));
+    await deleteFile(account, hsPath);
+    logger.log(i18n(`${i18nKey}.deleted`, { account, path: hsPath }));
```
camden11, line 29: Should be `accountId: account` here and in the `ApiErrorContext` below

commands/upload.ts
```diff
@@ -157,20 +156,20 @@ exports.handler = async options => {
...
+      account,
       absoluteSrcPath,
       normalizedDest,
       getFileMapperQueryValues(mode, options)
     )
       .then(() => {
         logger.success(
           i18n(`${i18nKey}.success.fileUploaded`, {
-            accountId,
+            account,
```
camden11, line 167: Another one that should be `accountId: account`. There are a few other `i18n`s and `ApiErrorContexts` in this file as well that need to be updated

camden11 (commented):
Nice, this is a huge improvement! Just a few minor comments, but otherwise looks good to go.

commands/upload.ts
```diff
@@ -215,26 +214,24 @@ exports.handler = async options => {
...
-          logger.log(
-            i18n(`${i18nKey}.cleaning`, { accountId, filePath: dest })
-          );
+          await deleteFile(account, dest);
+          logger.log(i18n(`${i18nKey}.cleaning`, { account, filePath: dest }));
         } catch (error) {
           logger.error(
             i18n(`${i18nKey}.errors.deleteFailed`, {
-              accountId,
+              account,
```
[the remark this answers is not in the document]
joe-yeager: Sorry, missed Camden's comment about the same thing.  Resolving

commands/cms/lighthouseScore.ts
```diff
@@ -32,7 +31,7 @@ const DEFAULT_TABLE_HEADER = [
 ];
 
 exports.command = 'lighthouse-score [--theme]';
-exports.describe = false; // i18n(`${i18nKey}.describe`);
+exports.describe = i18n(`${i18nKey}.describe`);
```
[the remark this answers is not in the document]
kemmerle: I think I just un-hid it, because I was using the help menu to aid in testing. I'm going to re-hide it for now and alert the CMS team--it'd be better if they did it in a separate PR.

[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1212/commits/11c8c119479c484027f71dffa1ad97d64951d2d8

commands/mv.ts
kemmerle: Thanks for catching these! I was originally coding the `i18n` and `logger` functions incorrectly and then realized my mistake. I think these were some of the first files I was editing 🙏 

Addressed in https://github.com/HubSpot/hubspot-cli/pull/1212/commits/11c8c119479c484027f71dffa1ad97d64951d2d8

commands/remove.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1212/commits/11c8c119479c484027f71dffa1ad97d64951d2d8

commands/upload.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1212/commits/11c8c119479c484027f71dffa1ad97d64951d2d8

lib/commonOpts.ts
```diff
@@ -90,9 +90,15 @@ const getAccountId = (options = {}) => {
...
   if (options.useEnv && process.env.HUBSPOT_PORTAL_ID) {
+    options.account = parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
+    options.portal = parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
     return parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
   }
 
+  // Preserves the original --account and --portal flags for certain commands.
+  options.accountFlag = portal || account;
+  options.account = getAccountIdFromConfig(portal || account);
+  options.portal = getAccountIdFromConfig(portal || account);
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1212/commits/4de4f195430fa86d49d065a492c8a351f2ed16ff

lib/commonOpts.ts
```diff
@@ -90,9 +90,15 @@ const getAccountId = (options = {}) => {
   const { portal, account } = options;
 
   if (options.useEnv && process.env.HUBSPOT_PORTAL_ID) {
+    options.account = parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
+    options.portal = parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
     return parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
   }
 
+  // Preserves the original --account and --portal flags for certain commands.
+  options.accountFlag = portal || account;
```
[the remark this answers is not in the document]
kemmerle: I like that a lot! Addressed in https://github.com/HubSpot/hubspot-cli/pull/1212/commits/4de4f195430fa86d49d065a492c8a351f2ed16ff

camden11 (approved):
LGTM!

lib/commonOpts.ts
```diff
@@ -90,10 +90,18 @@ const getAccountId = (options = {}) => {
   const { portal, account } = options;
 
   if (options.useEnv && process.env.HUBSPOT_PORTAL_ID) {
-    return parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
+    const accountIdFromEnv = parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
+    options.account = accountIdFromEnv;
```
brandenrodgers, line 94: just a thought while we're in here making changes, but it might make sense for us to use `accountId` and `portalId` to reduce confusion in our commands. We can keep the user-facing option as `--account`, but my thinking is that `account` is already a somewhat loaded term so being explicit could make things easier on us. "account" could mean the id of the account, the name of the account in the cli config, the name of the account in hubspot, or the metadata for the account in the cli config.

Not a blocker, I just saw changes like [this one](https://github.com/HubSpot/hubspot-cli/pull/1212/files#diff-172a3366dd1c2700e6b704f0589a5a8815fb931c2b9c936da8711642d3076979R86) and my first thought was that I liked it better as `accountId`

lib/commonOpts.ts
```diff
@@ -90,10 +90,18 @@ const getAccountId = (options = {}) => {
   const { portal, account } = options;
 
   if (options.useEnv && process.env.HUBSPOT_PORTAL_ID) {
-    return parseInt(process.env.HUBSPOT_PORTAL_ID, 10);
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1212/commits/b831e5b2e302063bd8435cb3f1e39429fb2786e5

lib/commonOpts.ts
kemmerle: Resolved in:
https://github.com/HubSpot/hubspot-cli/pull/1212/commits/b831e5b2e302063bd8435cb3f1e39429fb2786e5
https://github.com/HubSpot/hubspot-cli/pull/1212/commits/c00709b062658e6e6331d7b732f6990aefd54396
https://github.com/HubSpot/hubspot-cli/pull/1212/commits/53b8a76d0c06d9cdf2d72165c7399bcf854c8d98
https://github.com/HubSpot/hubspot-cli/pull/1212/commits/e6030caf498302d61e1aa38cbfa5509f499aa185
https://github.com/HubSpot/hubspot-cli/pull/1212/commits/95e168b14d410e56f46f7d98be8cfcbbdfa00890

camden11 (approved):

PR #1215, opened 2024-10-23
Update husky and lint-staged for TS

camden11:
## Description and Context
All of our files are in TS now, but lint-staged was only looking for JS and never actually running. 

## Who to Notify


kemmerle (approved):
Thanks for fixing this 🧛‍♀️

Issue #1216, opened 2024-10-23
Is there a changelog and/or migration guide?


happycollision:
Since the last time I updated the CLI, you've moved from major version 5 to major version 6. I looked at [your release in GitHub](https://github.com/HubSpot/hubspot-cli/releases/tag/v6.0.0), but I don't see anything about breaking changes. I don't see anything of the sort in [your dev docs](https://developers.hubspot.com/docs/cms/developer-reference/local-development-cli) either.

brandenrodgers:
Hey @happycollision 👋 the best way to stay up to date is to check out the [Github Releases ](https://github.com/HubSpot/hubspot-cli/releases/tag/v6.0.0)(like you linked), and also follow our [developer changelog posts](https://developers.hubspot.com/changelog/september-2024-rollup) for an easier to read breakdown of major changes. Just a heads up, the tags in the changelog page don't always catch every mention of the CLI. I'd recommend just viewing the [latest posts](https://developers.hubspot.com/changelog) and cmd+f for "cli". 

To answer your question though, the reason for the major version bump is because we increased the minimum supported node version from 16 to 18 in v6.0.0

happycollision:
Thanks for the info.

PR #1217, opened 2024-10-24
Add change log info to readme

happycollision:
As answered in a recent issue.

https://github.com/HubSpot/hubspot-cli/issues/1216


brandenrodgers (approved):
Thanks for adding this! 👌

PR #1218, opened 2024-10-24
Update eslint config for TS + fix linting issues

camden11:
## Description and Context
This updates the ESlint config to use the standard typescript preset and brings the linting rules more or less in line with those of local-dev-lib. For now, continuing to allow `require` imports to support the `.ts` files that haven't been converted yet. Thankfully, eslint seems to not run any of its TS linting rules on files that have the `ts-nocheck` declaration 

This ended up adding a few rules that weren't present before, but I think they're all good ones to have so went through and made some updates to fix them. These include
* No using `let` for variables that don't get reassigned
* No unused variables
* No redundant `await`s

I had to change a few TS things in the acceptance tests to get these linting rules to pass. Everything still seems to work and tests pass as expected. @joe-yeager I think you worked on getting those into TS, if there's anything that needs to change back we can definitely figure something out

## Who to Notify
@joe-yeager @brandenrodgers @kemmerle

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,12 @@ module.exports = {
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/ban-ts-comment': 'off',
     '@typescript-eslint/no-require-imports': 'off',
+    '@typescript-eslint/no-unused-vars': [
+      'error',
+      {
+        caughtErrors: 'none',
+      },
+    ],
   },
 
   overrides: [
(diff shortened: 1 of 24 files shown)


joe-yeager:
>  @joe-yeager I think you worked on getting those into TS, if there's anything that needs to change back we can definitely figure something out

Some of the changes that were undone were to be changes @brandenrodgers made to the ATs, but they LGTM

joe-yeager (approved):
Thanks for doing this!  LGTM

PR #1219, opened 2024-10-25
Adding more command structure tests (pt. 2)

brandenrodgers:
## Description and Context

This adds more unit tests to validate command structure. That includes the command, describe, and builder exports for each command. This will prevent us from unintentionally making breaking changes to the command syntax.

**Note:**
I ignored the customObject and hubdb commands because there's some issue with Prettier. [This is the issue I was seeing](https://github.com/prettier/prettier/issues/15769). I can dig into resolving it, but I didn't want to include those changes in this PR.


lib/prompts/__tests__/downloadProjectPrompt.test.ts
```diff
@@ -1,21 +1,19 @@
 // @ts-nocheck
+const { getAccountId } = require('@hubspot/local-dev-lib/config');
```
brandenrodgers, line 2: I did a little bit of moving around of imports and files in the lib folder just to make things consistent. Some of these tests still accounted for the old folder structure of the repo where CLI was one of three packages

lib/__tests__/dependencyManagement.test.ts
```diff
@@ -205,16 +205,16 @@ describe('cli/lib/dependencyManagement', () => {
         throw new Error('OH NO');
       });
       util.promisify = jest.fn().mockReturnValue(execMock);
-      await expect(() => getProjectPackageJsonLocations()).rejects.toThrowError(
-        /This command depends on npm, install/
-      );
+      await expect(() =>
```
joe-yeager, line 208: Why are we removing the message checking?  Without the message checking these tests will pass for any error thrown an not just the specific error we are testing for, giving us a lesser degree confidence for regressions

brandenrodgers: iirc I removed these ones because they were failing from some other change I had made and it was kind of tedious to get them passing again. In my opinion there's a level of granularity with these tests where they become more trouble than they're worth. The majority of the value that we get from the tests is that they help to prevent us from unintentionally releasing unintentional breaking changes or bugs. Checking for exact copy in errors and logs like this is more similar to visual regression testing (i.e. not functional). The far extreme of this is that every 5 minute code change comes with a corresponding 10 minutes of trying to get broken tests to pass, when in reality we should be catching the changes to copy at review stage.

I had meant to chat irl this week about this, but maybe we can hold off on merging and chat about it in our next team sync.

commands/__tests__/accounts.test.ts
```diff
@@ -0,0 +1,75 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
brandenrodgers: I could, but I was under the assumption that we were starting with the leaves and then moving up the tree to make the port easier on us. It gets annoying when a file imports a non-ported file.

lib/__tests__/dependencyManagement.test.ts
joe-yeager: > I had meant to chat irl this week about this, but maybe we can hold off on merging and chat about it in our next team sync.

That sounds good to me.  I'm going to jot down some thoughts while it's fresh in my mind for when we circle back.

> In my opinion there's a level of granularity with these tests where they become more trouble than they're worth

That's fair, but I fear we might be swinging from too granular to not granular enough.  I feel there is a middle ground, especially since these methods can use regex to perform the match.

> The majority of the value that we get from the tests is that they help to prevent us from unintentionally releasing unintentional breaking changes or bugs

I think without an assertion to the contents of the error, these tests give us a false sense of security.  The error being thrown in all of these cases could be syntax errors, null pointer errors, etc and the tests would pass when the code is incorrect.  

> when in reality we should be catching the changes to copy at review stage

I have concerns with pushing an automatable task to a manual human task.  I don't trust myself to catch copy errors, especially in heavily interpolated strings using variable from multiple files.  And a lot of our messaging is using/reusing values from the `i18n` file as well as constants from around the code base, etc.  

> The far extreme of this is that every 5 minute code change comes with a corresponding 10 minutes of trying to get broken tests to pass

😅 I actually think this is fine.  Those 10 minutes of testing help to continue to preserve the behaviors we expect of our application.  

I also think of it as an investment that pays high dividends by removing the amount of manual testing we need to perform.   A recent example of this is the work we did with the error refactoring as well as the ongoing push to revamp the configuration.  If we had granular tests, it would certainly add some pain on the amount of tests that we need to update, but it would have reduced the amount of hours required in the regression testing phase, and allowed us a greater level of confidence in the change.

brandenrodgers: Ok that's compelling 😄 , I added the copy back in [65dab7e](https://github.com/HubSpot/hubspot-cli/pull/1219/commits/65dab7e9b6f7bd2c2a5beef4e00cef8cb2ff17ea)

lib/__tests__/ProjectLogsManager.ts
```diff
@@ -1,14 +1,14 @@
 // @ts-nocheck
```
joe-yeager, line 1: This file is missing the `.test` in the file name, would you mind updating it while we are in here? 🙏

joe-yeager (approved):
LGTM!  Just noticed one file name was in consistent

lib/__tests__/ProjectLogsManager.ts
brandenrodgers: ah good catch, done in [dd75ef5](https://github.com/HubSpot/hubspot-cli/pull/1219/commits/dd75ef54f0f87d557689dc2620071c3bc8ae6172)

PR #1221, opened 2024-10-25
Convert lib files to TS: Part 1

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/684

This converts the first batch of file in the lib directory to TS. This was thankfully really easy due to our work on LDL - every type I needed for this already existed.

A couple other minor changes (as we discussed, these are not breaking changes since the CLI's only exportable is the actual CLI bin file)
* I updated `hasFlag` to use named export to match every other file in lib
* I deleted `getFunctionsArray` and moved it inline to the only place it was being used.

## Who to Notify


lib/enums/exitCodes.ts
```diff
@@ -1,13 +1,10 @@
...
  */
-const EXIT_CODES = {
+export const EXIT_CODES = {
   SUCCESS: 0,
   ERROR: 1,
   WARNING: 2,
-};
-
-module.exports = { EXIT_CODES };
+} as const;
```
camden11, line 10: Thought about maybe moving this into `constants` and deleting this file, but its used in so many different places and didn't want to make this PR confusing

lib/regex.ts
```diff
@@ -1,6 +1 @@
-// @ts-nocheck
-const STRING_WITH_NO_SPACES_REGEX = /^\S*$/;
-
-module.exports = {
-  STRING_WITH_NO_SPACES_REGEX,
-};
+export const STRING_WITH_NO_SPACES_REGEX = /^\S*$/;
```
camden11, line 1: Thought about maybe removing this file too, but figured regexes are confusing and this makes things clearer even though it's only used in two places

joe-yeager (approved):
Let's gooooo!

lib/regex.ts
brandenrodgers: I actually just killed this file in my PR here: https://github.com/HubSpot/hubspot-cli/pull/1222 . So keeping it around in your Pr is fine

brandenrodgers (approved):
It begins! 🙌

PR #1222, opened 2024-10-25
Trimming library utils

brandenrodgers:
## Description and Context

Trimming some of the clutter thats built up over time. This should hopefully make it easier to port to TS and add tests because we're cutting down on the total number of files. I'll leave inline comments to explain all of the changes, but this is targeting unnecessary files and utils in the `lib/` folder.


lib/interpolationHelpers.ts
brandenrodgers, line 1: This was only used in interpolation.ts and was relatively small

lib/getFunctionArrays.ts
```diff
@@ -1,19 +0,0 @@
-// @ts-nocheck
```
brandenrodgers, line 1: This was only used in commands/functions/list.ts

lib/environment.ts
```diff
@@ -1,15 +0,0 @@
-// @ts-nocheck
-const getPlatform = () => {
```
brandenrodgers, line 2: This was only used in usageTracking.ts

lib/debugInfo.ts
brandenrodgers, line 1: This is an outdated debugging util that doesn't provide much value and was never rolled out into the majority of the commands. We have newer debugging utils coming in soon that will replace the need for this

lib/marketplaceValidate.ts
brandenrodgers, line 1: renaming lib utils to follow the same naming convention

lib/prompts/accountsPrompt.ts
```diff
@@ -28,23 +27,6 @@ const selectAccountFromConfig = async (config, prompt) => {
   return selectedDefault;
 };
 
-const selectAndSetAsDefaultAccountPrompt = async config => {
```
brandenrodgers, line 31: This prompt does exactly the same thing as the one above it. I added a call to `updateDefaultAccount()` in the two places that this was being used

lib/prompts/activeInstallConfirmationPrompt.ts
```diff
@@ -1,21 +0,0 @@
-// @ts-nocheck
-const { promptUser } = require('./promptUtils');
```
brandenrodgers, line 2: This can be replaced by `confirmPrompt()`

lib/prompts/createFunctionPrompt.ts
```diff
@@ -2,8 +2,6 @@
 const { promptUser } = require('./promptUtils');
 const { i18n } = require('../lang');
 
-const { STRING_WITH_NO_SPACES_REGEX } = require('../regex');
```
brandenrodgers, line 5: No need to use a regex to check for string spaces when we can do it easily with `indexOf()`

lib/prompts/deployBuildIdPrompt.ts
brandenrodgers, line 1: Moved this into the command that was using it because it's a fairly simple prompt and doesn't take up much space. My thinking is that we can keep these dedicated prompt files, but we should only move prompts into here when they grow unwieldy or can't be implemented using the exports from promptUtils.ts

lib/prompts/promptUtils.ts
```diff
@@ -1,24 +1,45 @@
-// @ts-nocheck
 const inquirer = require('inquirer');
```
camden11, line 1: Remember to switch this to `import` when you convert things to TS

lib/prompts/promptUtils.ts
```diff
@@ -1,24 +1,45 @@
...
 
-const promptUser = async promptConfig => {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const promptUser = async (promptConfig: any) => {
```
camden11, line 4: I think we could probably avoid using `any` here - at minimum could use `object`, but it also might be worth seeing if you can get a typed version of `inquirer` either with a newer version or `@types/inquirer`

lib/prompts/promptUtils.ts
```diff
@@ -1,24 +1,45 @@
...
   const prompt = inquirer.createPromptModule();
   return prompt(promptConfig);
 };
 
-const confirmPrompt = async (message, defaultAnswer = true) => {
+export const confirmPrompt = async (
```
camden11, line 9: Nit: we should use the `function` keyword over `const` when writing most TS functions. Pretty sure that was in the official HubSpot TS guidelines, although I can no longer find it. But it looks a lot cleaner too.

camden11 (commented):
Nice, this really cleans things up! A few minor comments - only blocking thing is some of the TS stuff in the prompt utils

lib/prompts/promptUtils.ts
brandenrodgers: done in [cc915d2](https://github.com/HubSpot/hubspot-cli/pull/1222/commits/cc915d22819ba914fcdfb84f88f28984e110cb07)

lib/prompts/promptUtils.ts
brandenrodgers: Ah good call. I made some bigger changes to this in [cc915d2](https://github.com/HubSpot/hubspot-cli/pull/1222/commits/cc915d22819ba914fcdfb84f88f28984e110cb07)

Inquirer has a `inquirer.prompt()` util that does exactly the same thing that this util was doing so I removed it. I also had to bump inquirer to v12 because our version (v8) wasn't an esm and didn't support imports

brandenrodgers: @camden11 scratch that, I ran into a ton of issues with this because inquirer v8 doesn't isn't an esm. I'm going to leave this as is for now because we can actually remove this export completely and directly leverage `inquirer.prompt()` everywhere once we bump the inquirer version to 12. I tried doing this in a way that didn't make the PR too crazy, but there are a ton of tests and files that would need to be updated.

commands/lint.ts
```diff
@@ -21,6 +19,33 @@ exports.command = 'lint <path>';
...
+    if (err.reason !== 'SYNTAX_ERROR') {
+      return;
+    }
+    ++count;
+    printHublValidationError(err);
+  });
+  logger.groupEnd(file);
+  return count;
+}
+
```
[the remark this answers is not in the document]
camden11: These functions have already been ported to TS - probably still worth moving the updated versions with types to this file even if you don't port the rest of `lint.ts`

camden11 (approved):
LGTM - `lint.ts` and `interpolation.ts` both have parts that have already been converted to TS so might be worth bringing over. But if it causes issues with the rest of those files, then nbd since porting them back alter will be super low effort

commands/lint.ts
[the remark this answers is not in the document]
brandenrodgers: 🤦 did this in [cc915d2](https://github.com/HubSpot/hubspot-cli/pull/1222/commits/cc915d22819ba914fcdfb84f88f28984e110cb07) and then reverted it

PR #1223, opened 2024-10-25
fix: Allow template name or label to be used for `template` flag

joe-yeager:
## Description and Context

Allow either the template name or the template label to be used to lookup the template


brandenrodgers (approved):
🚢

Issue #1224, opened 2024-10-27
The module which create by `hs create module`  will not be uploaded by error


kotto5:
## Description and Context

```
npx hs create module module_to_local
npx hs upload module_to_local module_name_to_remote
```
then
```
... (Some files will be uploaded)
Uploaded file "/Users/test/Develop/local-dev-tutorial/post_recent.module/fields.json" to "post_recent.module/fields.json"
Uploaded file "/Users/test/Develop/local-dev-tutorial/post_recent.module/module.js" to "post_recent.module/module.js"
Uploaded file "/Users/test/Develop/local-dev-tutorial/post_recent.module/module.html" to "post_recent.module/module.html"
[ERROR] Uploading file "post_recent.module" to "post_recent.module" failed
[ERROR] HubSpotHttpError: The post in account 20616822 was bad. PAGE is not a valid contentType. Must be one of [ANY, LANDING_PAGE, SITE_PAGE, BLOG_POST, BLOG_LISTING, EMAIL, KNOWLEDGE_BASE, QUOTE_TEMPLATE, CUSTOMER_PORTAL, WEB_INTERACTIVE, SMS, SUBSCRIPTION, MEMBERSHIP, CASE_STUDY, CASE_STUDY_LISTING]
```

This error's reason is setting invalid value for  "content_type". The value will be set as "Page", but this value is for previous key ("host_template_type")

I refer [this link](https://developers.hubspot.com/docs/cms/building-blocks/modules/configuration).
I'm new to hubspot (just two days) so maybe I take mistake.

my hs version is 6.2.1

### Steps to reproduce
As above.

## Expected behavior
As above.

## Screenshots
It is not frontend error. It will be occur in terminal so I think screenshots is not needed.

kotto5:
I submit PR for this issue!

I am new to Hubspot so welcome to any points and correction! Thanks for this community.

brandenrodgers:
Closing this as the new functionality should be available in CLI version >= `6.4.0`. Thanks for helping out! 😄

PR #1225, opened 2024-10-27
Modify `hs create module` command to correspondence to content_types (Previously host_template_types)

kotto5:
## Description and Context

Modify `hs create module` command's content_type checkbox to choose value for content_types.
Previously, you could choose a value for host_template_types.
It causes an error when you upload modules.

https://github.com/HubSpot/hubspot-cli/issues/1224

## Screenshots

Before
Screenshot 2024-10-27 at 23 38 03

After
Screenshot 2024-10-27 at 23 36 36

After version may have problem about UI. Choices are too many to choose and to display at once so you have to scroll to see all choices. It can make confuse user who create modules.

## TODO

- Discussion about UI and improvement may be needed.


kotto5:
Is this PR unimportant?

kotto5:
@graysky @also @timmfin @bbeaudreault 
Is this PR unimportant?

lib/prompts/createModulePrompt.ts
```diff
@@ -30,10 +30,18 @@ const CONTENT_TYPES_PROMPT = {
   message: i18n(`${i18nKey}.selectContentType`),
   default: ['PAGE'],
```
TanyaScales, line 31: Hi @kotto5 - one quick note that we should update the default here to point to the `ANY` option instead to support this change

TanyaScales:
@kotto5 - RE: this comment "After version may have problem about UI" and what to do here. I would be fine with this update moving through independently of UI updates to make this better (for the short term). We can discuss the proper pattern for long-form option lists over here on our side and make sure it is consistent across the commands.

lib/prompts/createModulePrompt.ts
kotto5: Thanks for review @TanyaScales 

I fixed it based on your point, and I'm glad to see that the UI thing is not a problem either.
Hope you will check the fixes.

kotto5: @TanyaScales 
How do I ask other reviewers for a review after I have been approved by you?

PR #1227, opened 2024-10-31
Convert widely used modules in lib folder

kemmerle:
## Description and Context
As part of our work to convert the CLI to TypeScript, I've migrated the following modules that are widely used in the `lib` folder:

- [x] `interpolation.ts`
- [x] `lang.ts`
- [x] `commonOpts.ts`
- [x] `process.ts`
- [x] `debugInfo.ts`

## TODO

- [x] Address any feedback 

## Who to Notify


lib/commonOpts.ts
```diff
@@ -1,44 +1,55 @@
...
+  getAccountId as getAccountIdFromConfig,
   getAccountConfig,
   getAndLoadConfigIfNeeded,
-} = require('@hubspot/local-dev-lib/config');
-const { i18n } = require('./lang');
+} from '@hubspot/local-dev-lib/config';
+import { i18n } from './lang';
+import { Argv, Arguments } from 'yargs';
+
+interface CustomOptions {
```
kemmerle, line 14: I use this type twice here and in the `logDebugInfo` file. Should I create a Types file in the `lib` folder? Wanted to check, if we already had plans about it.

camden11: I think you could even create a new `types` folder like we have in LDL at the top level. I'm not really sure what the best way to handle this specific type is though, since this theoretically represents anything that could be passed in via a flag. For right now I could see it making sense to have each function only specify what they need from the `Arguments` type, rather than sharing all of them. For example, in `getMode`, it could just take in `Arguments<{ mode?: string }>` or something like that. Feel like it'll become clearer how to handle this once we start converting commands

lib/commonOpts.ts
```diff
@@ -51,59 +62,59 @@ const addModeOptions = (yargs, { read, write }) => {
...
   }
-};
+}
 
-/**
- * Get command name from Yargs `argv`
- * @param {object} argv
- */
-const getCommandName = argv => (argv && argv._ && argv._[0]) || '';
+export function getCommandName(argv: Arguments<CustomOptions>): string {
```
camden11, line 93: Are you sure this type is right? My hunch is that it should take in `Argv`

lib/debugInfo.ts
```diff
@@ -1,8 +1,12 @@
-// @ts-nocheck
-const { version } = require('../package.json');
-const { getPlatform } = require('./environment');
+import { version } from '../package.json';
+import { getPlatform } from './environment';
+import { Arguments } from 'yargs';
 
-const logDebugInfo = ({ debug }) => {
+interface CustomOptions {
```
camden11, line 5: If we're going to be passing these into `Arguments` a lot, which it seems like we are, probably makes sense to give them their own unique names. `DebugInfoFlagOptions`? `DebugInfoOptions`? Not sure, but CustomOptions doesn't tell us a lot and will also make searching for a specific `CustomOptions` type more difficult

lib/interpolation.ts
```diff
@@ -11,32 +18,30 @@ const delimiters = {
...
   );
-};
+}
 
-const generateReplaceFn = (matchedText, startIndex, replacementString) => {
+function generateReplaceFn(
+  matchedText: string,
+  startIndex: number,
+  replacementString: string
+): (arg0: string) => string {
```
camden11, line 32: Curious about this `arg0` pattern you've been using for returned functions. In LDL we just used the variable name in the type (so `currentStringValue` here). Not opposed to doing it this way, just wondering if there was a reason you made the switch

lib/interpolation.ts
```diff
@@ -1,5 +1,12 @@
-// @ts-nocheck
-const helpers = require('./interpolationHelpers');
+import * as helpers from './interpolationHelpers';
+
+interface interpolationDataType {
```
camden11, line 3: I think this type could just be `InterpolationData`. Also just for consistency, I think this could be a `type` and not an `interface`

lib/interpolation.ts
```diff
@@ -1,5 +1,12 @@
...
+  [key: string]: string;
+}
+
+type HelperFunctions = typeof helpers;
+type HelperIdentifier = keyof HelperFunctions;
```
camden11, line 8: You can just do `keyof typeof` in one line

lib/lang.ts
```diff
@@ -1,17 +1,20 @@
...
-const fs = require('fs-extra');
-const yaml = require('js-yaml');
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const { interpolate } = require('./interpolation');
+import util from 'util';
+import path from 'path';
+// @ts-ignore
+import fs from 'fs-extra';
+// @ts-ignore
+import yaml from 'js-yaml';
```
camden11, line 6: Did you try installing `@typesfs-extra` and `@types/js-yaml`?

lib/lang.ts
```diff
@@ -66,10 +71,14 @@ const getTextValue = lookupDotNotation => {
...
+  if (typeof textValue !== 'string') {
+    return missingTextData;
+  }
+
   return textValue;
-};
+}
 
-const i18n = (lookupDotNotation, options = {}) => {
+export function i18n(lookupDotNotation: string, options = {}): string {
```
camden11, line 81: `options` here should have a type, see https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/utils/lang.ts#L111

lib/commonOpts.ts
kemmerle: Going to resolve for now but feel free to reopen.

lib/debugInfo.ts
kemmerle: Just changed this to `Arguments<{ debug?: boolean }>` based on our Huddle on Friday afternoon. Going to resolve but feel free to reopen if need be.

lib/interpolation.ts
kemmerle: It was GitHub Copilot's suggestion 😬 Per our Huddle on Friday, I'm changing `arg0` to something more descriptive.

lib/lang.ts
kemmerle: That worked; thanks, Camden!

lib/commonOpts.ts
kemmerle: Per our Huddle on Friday, I changed the code to have each function only specify what they need from the `Arguments` type. Going to resolve for now, but feel free to reopen.

kemmerle:
@camden11, I addressed your feedback in https://github.com/HubSpot/hubspot-cli/pull/1227/commits/e093f5560cc2a0dc2d8a5a784f6a399a7d1b7bef. This is now ready for a re-review.

lib/lang.ts
```diff
@@ -66,10 +69,17 @@ const getTextValue = lookupDotNotation => {
...
+    return missingTextData;
+  }
+
   return textValue;
-};
+}
 
-const i18n = (lookupDotNotation, options = {}) => {
+export function i18n(
+  lookupDotNotation: string,
```
joe-yeager, line 80: In LDL we have this typed to the [keys in the lang files](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/types/Lang.ts).  I wonder if it is worthwhile to invest the energy to repeat that over here?

kemmerle: @brandenrodgers @camden11, thoughts?

lib/process.ts
```diff
@@ -1,15 +1,20 @@
...
+interface KeyPress {
+  ctrl?: boolean;
+  shift?: boolean;
+  alt?: boolean;
+  meta?: boolean;
+  name?: string;
+}
+
+export function handleExit(
+  callback: (onTerminate?: { isSIGHUP: boolean }) => void
```
brandenrodgers, line 16: I think the `onTerminate` field should be required here instead of optional. The callback will always be passed `isSIGHUP` within this function, regardless of whether the calling code actually leverages it or not. I noticed this when trying to port a file that uses this `handleExit` util to TS and got an error.

brandenrodgers:
Just one small thing I noticed, but otherwise this all LGTM! 🎉

camden11 (approved):
LGTM

PR #1228, opened 2024-11-01
Improved release script

camden11:
## Description and Context
See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/682
This converts the build and release scripts from bash to node, which should make both running them and contributing them a much better experience. The release script has been substantially overhauled with a few major changes:
* `experimental` releases can no longer be released on the `main` branch, and have their own version continuity. The `experimental` branch uses the latest npm release as its source of truth for current version
* `next` and `latest` tag use the version in the package.json as the source of truth for current version
* all releases to `latest` now also update `next`
* Failed releases now clean up - no need to change version or delete tags manually anymore

## Screenshots
Screenshot 2024-11-04 at 3 34 28 PM
Screenshot 2024-11-04 at 3 34 45 PM

## TODO

- [x] Figure out the TS issue with inquirer types (this is currently breaking builds)
- [ ] Test this thoroughly
- [ ] Do an experimental release

## Who to Notify


scripts/build.ts
```diff
@@ -0,0 +1,27 @@
...
+  fs.rmSync('dist', { recursive: true, force: true });
+
+  // Build typescript
+  await new Promise(resolve => {
+    const childProcess = spawn('yarn', ['tsc'], {
+      stdio: 'inherit',
+    });
+
+    childProcess.on('close', () => {
+      resolve('');
```
joe-yeager, line 15: The callback for `close` should get the exit code as an argument.  Might be worthwhile to check it an reject if it is non-zero.  So something like:
```
    childProcess.on('close', (code) => {
      if(code != 0) {
        reject('Build failed');
      } else {
        resolve('');
      }
    });
```

joe-yeager (commented):
Thanks for taking this on!  Looking good

scripts/release.ts
```diff
@@ -0,0 +1,229 @@
...
+import { logger, setLogLevel, LOG_LEVEL } from '@hubspot/local-dev-lib/logger';
+import semver from 'semver';
+
+import { name as packageName, version as localVersion } from '../package.json';
+import { EXIT_CODES } from '../lib/enums/exitCodes';
+import { confirmPrompt } from '../lib/prompts/promptUtils';
+import { build } from './build';
+
+// TODO: Change to import when this is converted to TS
+const { uiLink, uiLine } = require('../lib/ui');
```
brandenrodgers, line 13: Can this be changed since the file is TS now?

scripts/release.ts
```diff
@@ -0,0 +1,229 @@
...
+}
+
+async function publish(tag: Tag): Promise<void> {
+  logger.log();
+  logger.log(`Publishing to ${tag}...`);
+  uiLine();
+  logger.log();
+
+  return new Promise((resolve, reject) => {
+    const childProcess = spawn('npm', ['publish', '--dry-run', '--tag', tag], {
```
brandenrodgers, line 77: Just a reminder to take the dry run out of this before merging. Maybe it could be a supported flag?

scripts/release.ts
```diff
@@ -0,0 +1,229 @@
...
+  await exec(`yarn version --new-version ${newVersion}`);
+  logger.success('Version updated successfully');
+
+  logger.log();
+  logger.log('Creating a new build...');
+  await build();
+  logger.success('Build successful');
+
+  try {
+    await publish(tag);
```
brandenrodgers, line 173: I wonder if we should try to enforce the QA --> PROD release flow here. We could check to see if the user has released these changes to the next branch, and encourage them to do it if they haven't.

For example, if the user attempts to increment from version 1.0.0 to version 1.0.1, we could first check to see if version 1.0.1-beta.<n> exists (where n doesn't matter). If it exists, then we can assume that the current local changes have already been released to the QA (next) tag. In that case, We can safely publish version 1.0.1, and we can still publish version 1.0.1-beta.<n+1>

Question: What should the version in the CLI's package.json be after this? 1.0.1 or 1.0.1-beta<n+1>?

If it doesn't exist (i.e. maybe only 1.0.0-beta.1 exists), we could prompt something like: `It looks like these changes have not been released to the next tag yet, we encourage you to release changes to the next tag before releasing them to the latest tag. Would you like to continue anyways?`. If they choose yes (Which IMO "no" should be the default option), we can continue to "dangerously" publish 1.0.1, and then we can also publish 1.0.1-beta.0 just to catch the next version up.

Here's an example of what the flow would look like:
1. [just to establish initial state] `yarn publish -v=major -tag=latest` --> publish version 1.0.0
2. `yarn publish -v=patch -tag=latest` --> We prompt to release these changes to next tag first, dev agrees
3. `yarn publish -v=patch -tag=next` --> publish version 1.0.1-beta.0
4. `yarn publish -v=patch -tag=latest` --> publish versions 1.0.1 and 1.0.1-beta.1

Alternatively...
1. [just to establish initial state] `yarn publish -v=major -tag=latest` --> publish version 1.0.0
2. `yarn publish -v=patch -tag=latest` --> We prompt to release these changes to next tag first, dev does not agree and continues to bypass QA (next) --> publish versions 1.0.1 and 1.0.1-beta.0

scripts/release.ts
```diff
@@ -0,0 +1,229 @@
...
+      'Invalid release: cannot increment prerelease number on latest tag.'
+    );
+  }
+
+  const {
+    next: currentNextTag,
+    experimental: currentExperimentalTag,
+  } = await getDistTags();
+
+  if (!isExperimental && currentNextTag !== localVersion) {
```
brandenrodgers, line 125: My suggestion [here](https://github.com/HubSpot/hubspot-cli/pull/1228/files#r1829439062) would probably make more sense to implement here (rather than where I commented it).

Just to enumerate the potential scenarios

**User is releasing to the experimental tag:** N/A nothing to prevent here

**User is releasing to next tag:** if local version contains a next tag, it should be equivalent to the `currentNextTag`. But it's possible that the local version won't contain one (i.e. the last publish was a publish to latest). In that case, the local version should have the same base as the `currentNextTag`.

**User is releasing to latest tag:** Check to see if local version is equal to `currentNextTag` AND that the `currentLatestTag` + `versionIncrement` would be equivalent to the base of the local version. This means that the local changes have already been released to the next tag, and can safely be released to latest. If not, we should prompt b/c it's dangerous to release changes directly to latest.

scripts/build.ts
camden11: I thought about this but figured it wasn't really necessary, since we're already directly logging the output from from `tsc`

scripts/release.ts
camden11: These unfortunately have to stay as `require` until these have been _actually_ converted to TS (and use `export` syntax)

scripts/build.ts
joe-yeager: Doesn't this mean that the `build` function will always succeed even `yarn tsc` fails?

camden11: Oooh. Yep. When the build fails it logs "Command failed with exit code 2" so I assumed we were good, but the actual exit code is zero due to `spawn`. Good catch

joe-yeager (approved):
🔥

brandenrodgers (approved):
Nice! I tested locally for a bunch of scenarios and this all seems to be WAD 🙌

PR #1230, opened 2024-11-04
chore: Add script to update local dev lib and lock version

joe-yeager:
## Description and Context

Add a script to update the local-dev-lib and lock it to whatever version is `latest` with the `--exact` flag

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,8 @@
     "build-docker": "docker image build --tag hs-cli-image . && docker image prune -f",
     "circular-deps": "yarn madge --circular .",
     "release": "bash ./scripts/release.sh",
-    "hs": "yarn build && node ./dist/bin/hs"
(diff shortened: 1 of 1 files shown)


camden11 (approved):
Nice this is a great idea!

package.json
```diff
@@ -56,7 +56,8 @@
     "build-docker": "docker image build --tag hs-cli-image . && docker image prune -f",
     "circular-deps": "yarn madge --circular .",
     "release": "bash ./scripts/release.sh",
-    "hs": "yarn build && node ./dist/bin/hs"
+    "hs": "yarn build && node ./dist/bin/hs",
+    "update-ldl": "yarn add --exact @hubspot/local-dev-lib@latest"
```
joe-yeager, line 60: The `--exact` flag will lock the version to the value of `@latest` and won't include `^` in the version

PR #1236, opened 2024-11-08
Fix bugs with build script and confirm prompt

camden11:
## Description and Context
This fixes two bugs that were affecting the release script
* The call to `build()` in the same file that the `build` function was exported from caused build to be run whenever it was imported. This moved the actual function to its own file to avoid redundant calls. It also adds logging to the build command so its more obvious if something like this happens again
* A recent change to the confirm prompt broke the prompts in the script and in one other place - this changes it to make the options object completely optional

## Who to Notify


kemmerle (approved):
Looked over the code, and tested it out. LGTM!

PR #1238, opened 2024-11-12
Fix bug with updating next tag on releases to latest

camden11:
## Description and Context
Currently, the release script errors out after publishing to `latest`, since it tries to republish the CLI to `next` when it should instead just be updating the dist tag to point to the latest published version. This fixes that.

Note: there does not appear to be any kind of `dry-run` functionality for the `dist-tag` command so there is unfortunately not a great way to fully test this without actually releasing. When the `dryRun` flag is passed in, the script will log the dist-tag command it would have run

## Who to Notify

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -199,6 +199,9 @@ async function handler({
   }
 
   logger.log();
+  if (dryRun) {
+    logger.log('DRY RUN');
+  }
   logger.log(`Current version: ${currentVersion}`);
   logger.log(`New version to release: ${newVersion}`);
 
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -101,7 +101,10 @@ async function publish(tag: Tag, isDryRun: boolean): Promise<void> {
   });
 }
 
-async function updateNextTag(newVersion: string): Promise<void> {
+async function updateNextTag(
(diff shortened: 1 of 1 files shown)


scripts/release.ts
```diff
@@ -101,6 +101,40 @@ async function publish(tag: Tag, isDryRun: boolean): Promise<void> {
...
+    TAG.NEXT,
+  ];
+
+  return new Promise((resolve, reject) => {
+    if (isDryRun) {
+      const distTagCommand = ['npm', ...commandArgs].join(' ');
+      logger.log(`Dry run: skipping run of \`${distTagCommand}\``);
+      resolve();
+    } else {
+      const childProcess = spawn('npm', commandArgs, { stdio: 'inherit' });
```
camden11, line 124: Using spawn here in case npm prompts for your two factor code again

scripts/release.ts
```diff
@@ -165,6 +199,9 @@ async function handler({
   }
 
   logger.log();
+  if (dryRun) {
+    logger.log('DRY RUN');
```
camden11, line 203: Just to make dry runs slightly less scary

joe-yeager (approved):

PR #1239, opened 2024-11-12
fix: Address bug where we destructure an nullable object

joe-yeager:
## Description and Context

Fixes this bug which just appeared on main
```
🚀🚀🚀 ➜ hs project install-deps --debug
[DEBUG] TypeError: Cannot destructure property 'portal' of 'options' as it is undefined.
    at getAccountId (/Users/jyeager/.nvm/versions/node/v18.20.4/lib/node_modules/@hubspot/cli/lib/commonOpts.js:79:13)
    at exports.handler
[ERROR] Cannot destructure property 'portal' of 'options' as it is undefined.
```

### DIFF
--- a/lib/commonOpts.ts
+++ b/lib/commonOpts.ts
@@ -92,9 +92,9 @@ export function getCommandName(argv: Arguments<{ _?: string[] }>): string {
 export function getAccountId(
   options: Arguments<{ portal?: number | string; account?: number | string }>
 ): number | null {
-  const { portal, account } = options;
+  const { portal, account } = options || {};
 
-  if (options.useEnv && process.env.HUBSPOT_PORTAL_ID) {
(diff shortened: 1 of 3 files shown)


lib/commonOpts.ts
```diff
@@ -92,9 +92,9 @@ export function getCommandName(argv: Arguments<{ _?: string[] }>): string {
 export function getAccountId(
   options: Arguments<{ portal?: number | string; account?: number | string }>
```
brandenrodgers, line 93: The reason this is failing for the install deps command is because options is a required field and it's not being passed in [here](https://github.com/HubSpot/hubspot-cli/blob/main/commands/project/installDeps.ts#L24). This will be easier to enforce once TS is rolled out everywhere, but we should keep the pattern of requiring options rather than allowing for it to not exist.

brandenrodgers (approved):

PR #1241, opened 2024-11-12
chore: Disable tracking in ATs

joe-yeager:
## Description and Context

Disable usage tracking in the ATs and throw an error in the event it is enabled

### DIFF
--- a/acceptance-tests/lib/TestState.ts
+++ b/acceptance-tests/lib/TestState.ts
@@ -64,6 +64,11 @@ export class TestState {
         cause: e,
       });
     }
+
+    // @ts-expect-error Non-existent field according to TS
+    if (this.parsedYaml?.allowUsageTracking) {
+      throw new Error('Usage tracking should be disabled');
+    }
   }
 
   async withAuth() {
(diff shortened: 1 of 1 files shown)


acceptance-tests/lib/TestState.ts
```diff
@@ -49,16 +49,26 @@ export class TestState {
...
-      // @ts-expect-error TypeScript thinks the cause doesn't exist
       throw new Error('Failed to initialize CLI config & authentication', {
         cause: e,
       });
     }
+
+    // @ts-expect-error Non-existent field according to TS
+    if (this.parsedYaml?.allowUsageTracking) {
+      throw new Error('Usage tracking should be disabled');
+    }
```
joe-yeager, line 71: I debated not adding this, but I think it might be worthwhile to fail the tests in the event usage tracking gets reenabled so we can quickly address it and prevent a sudden usage spike in the data.

brandenrodgers (approved):

acceptance-tests/lib/TestState.ts
kemmerle: I think that's wise 👍

PR #1242, opened 2024-11-13
Move jest to dev dep

brandenrodgers:
## Description and Context

Jest isn't needed in the production release so I moved it to a dev dep. This will likely fix some of the install deprecation warnings because jest includes outdated versions of glob and inflight.


package.json
```diff
@@ -20,7 +20,6 @@
     "findup-sync": "^4.0.0",
     "fs-extra": "^8.1.0",
     "inquirer": "8.2.0",
-    "jest": "^29.5.0",
```
joe-yeager, line 23: Interesting that this has been hiding in the prod deps all this time.  It looks like we also have a couple `@types` deps in there as well.  Would you be ok moving those in this PR as well?

brandenrodgers: done in [ad784bd](https://github.com/HubSpot/hubspot-cli/pull/1242/commits/ad784bd1e49be35f63a7c5a5eea61bc452d03d22)! I totally missed those

PR #1243, opened 2024-11-14
fix: Update portmanager shutdown flow

joe-yeager:
## Description and Context

Depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/209

Removes the check to see if there are any active servers, because this doesn't seem relevant.  We should instead be concerned with if the server is currently running, which is checked in the `stopPortManagerServer` function.

### DIFF
--- a/lib/DevServerManager.ts
+++ b/lib/DevServerManager.ts
@@ -1,5 +1,4 @@
 // @ts-nocheck
-
 const { logger } = require('@hubspot/local-dev-lib/logger');
 const { COMPONENT_TYPES } = require('./projectStructure');
 const { i18n } = require('./lang');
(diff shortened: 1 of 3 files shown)


lib/DevServerManager.ts
```diff
@@ -133,11 +132,7 @@ class DevServerManager {
         }
       });
 
-      const hasActiveServers = await portManagerHasActiveServers();
-
-      if (!hasActiveServers) {
-        await stopPortManagerServer();
-      }
```
joe-yeager, line 140: This check was seeing if any servers had registered with the `portManager` and hot if it was running.  The `stopPortManagerServer` method already checks if the server is running before stopping it.

kemmerle (approved):
LGTM

PR #1244, opened 2024-11-15
chore: Add flag to not track the hs init command

joe-yeager:
## Description and Context

I realized that tracking still occurs on the `hs init` command, so we will still get noise from the ATs

### DIFF
--- a/commands/init.ts
+++ b/commands/init.ts
@@ -133,7 +133,6 @@ exports.handler = async options => {
 
   createEmptyConfigFile({
     path: configPath,
-    allowUsageTracking: !options.disableTracking,
   });
 
   handleExit(deleteEmptyConfigFile);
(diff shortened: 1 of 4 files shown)


commands/init.ts
```diff
@@ -122,8 +127,12 @@ exports.handler = async options => {
     process.exit(EXIT_CODES.ERROR);
   }
 
-  trackAuthAction('init', authType, TRACKING_STATUS.STARTED);
+  if (!disableTracking) {
+    await trackAuthAction('init', authType, TRACKING_STATUS.STARTED);
+  }
+
   createEmptyConfigFile({ path: configPath });
```
joe-yeager, line 134: If we wanted to spend some more time on this, we could update this method to set the `allowUsageTracking` to false when the flag is a part of the command.  It seems like since the main reason we are adding this is to disable the tracking from ATs this might be overkill.  And then it might be strange if we support one of the `hs config set` items and not the others.  That's why I went with this approach

brandenrodgers (approved):
This makes sense to me! 🚢

camden11 (approved):

PR #1250, opened 2024-11-19
Port links utils to TS

brandenrodgers:
## Description and Context

Ports the Links utils to TS. I had to move some things around in the file to make the port easier.

### DIFF
--- a/lib/links.ts
+++ b/lib/links.ts
@@ -12,7 +12,7 @@ type SiteLink = {
   url?: string;
 };
 
-const SITE_LINKS = {
+const SITE_LINKS: { [key: string]: SiteLink } = {
   APPS_MARKETPLACE: {
     shortcut: 'apps-marketplace',
     alias: 'apm',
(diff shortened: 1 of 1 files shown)


lib/links.ts
```diff
@@ -116,12 +114,15 @@ const openLink = (accountId, shortcut) => {
...
+  const baseUrl = getHubSpotWebsiteOrigin(
+    getEnv() === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD
+  );
+
+  open(`${baseUrl}/${match.getUrl(accountId)}`, { url: true });
+  logger.success(`We opened ${match.getUrl(accountId)} in your browser`);
+}
 
 module.exports = {
-  getSiteLinks,
```
joe-yeager, line 124: Do we need to keep this around in case something was importing it with `require`?

joe-yeager (approved):

camden11 (approved):

PR #1251, opened 2024-11-20
Convert errorHandlers to TypeScript

kemmerle:
## Description and Context
I've converted the files located in the `lib/errorHandlers` folder. 

Problem with eslint as been addressed 👍 

## TODO

- [x] Address eslint error
- [x] Address other feedback 

## Who to Notify

### DIFF
--- a/lib/errorHandlers/index.ts
+++ b/lib/errorHandlers/index.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 const { logger } = require('@hubspot/local-dev-lib/logger');
 const {
   isHubSpotHttpError,
@@ -7,37 +6,39 @@ const {
   isValidationError,
   isMissingScopeError,
 } = require('@hubspot/local-dev-lib/errors/index');
+import { FileSystemError } from '@hubspot/local-dev-lib/models/FileSystemError';
+import { HubSpotHttpError } from '@hubspot/local-dev-lib/models/HubSpotHttpError';
+import { BaseError } from '@hubspot/local-dev-lib/types/Error';
 const { shouldSuppressError } = require('./suppressError');
 const { i18n } = require('../lang');
 const util = require('util');
-const { isAxiosError } = require('axios');
 
 const i18nKey = 'lib.errorHandlers.index';
 
-function logError(error, context = {}) {
+export function logError(error: unknown, context?: typeof ApiErrorContext) {
   debugError(error, context);
 
   if (
     shouldSuppressError(error, context) ||
-    shouldSuppressError(error, error.context)
+    shouldSuppressError(error, (error as HubSpotHttpError).context)
   ) {
     return;
   }
 
-  if (isHubSpotHttpError(error) && context) {
-    error.updateContext(context);
+  if (isHubSpotHttpError(error) && context instanceof ApiErrorContext) {
+    (error as HubSpotHttpError).updateContext(context);
   }
 
   if (isHubSpotHttpError(error) || isFileSystemError(error)) {
     if (isValidationError(error) || isMissingScopeError(error)) {
-      logger.error(error.formattedValidationErrors());
+      logger.error((error as HubSpotHttpError).formattedValidationErrors());
     } else {
-      logger.error(error.message);
+      logger.error((error as HubSpotHttpError | FileSystemError).message);
     }
   } else if (isSystemError(error)) {
-    logger.error(error.message);
-  } else if (error.message || error.reason) {
-    const message = [];
+    logger.error((error as BaseError).message);
+  } else if (isErrorWithMessageOrReason(error)) {
+    const message: string[] = [];
 
     [error.message, error.reason].forEach(msg => {
       if (msg) {
@@ -51,25 +52,17 @@ function logError(error, context = {}) {
   }
 }
 
-/**
- * Logs (debug) the error and context objects.
- *
- * @param {Error}  error
- * @param {ApiErrorContext} context
- */
-function debugError(error, context = {}) {
+export function debugError(error: unknown, context?: typeof ApiErrorContext) {
   if (isHubSpotHttpError(error)) {
-    logger.debug(error.toString());
+    logger.debug((error as HubSpotHttpError).toString());
   } else {
     logger.debug(i18n(`${i18nKey}.errorOccurred`, { error }));
   }
 
-  if (error.cause) {
+  if (isErrorWithCause(error)) {
     logger.debug(
       i18n(`${i18nKey}.errorCause`, {
-        cause: isAxiosError(error.cause)
-          ? error.cause
-          : util.inspect(error.cause, false, null, true),
+        cause: util.inspect(error.cause, false, null, true),
       })
     );
   }
@@ -82,21 +75,37 @@ function debugError(error, context = {}) {
   }
 }
 
-class ApiErrorContext {
-  constructor(props = {}) {
-    /** @type {number} */
+export class ApiErrorContext {
+  accountId?: number;
+  request?: string;
+  payload?: string;
+  projectName?: string;
+
+  constructor(
+    props: {
+      accountId?: number;
+      request?: string;
+      payload?: string;
+      projectName?: string;
+    } = {}
+  ) {
     this.accountId = props.accountId;
-    /** @type {string} */
     this.request = props.request || '';
-    /** @type {string} */
     this.payload = props.payload || '';
-    /** @type {string} */
     this.projectName = props.projectName || '';
   }
 }
 
-module.exports = {
-  logError,
-  debugError,
-  ApiErrorContext,
-};
+function isErrorWithCause(error: unknown): error is { cause: unknown } {
+  return typeof error === 'object' && error !== null && 'cause' in error;
+}
+
+function isErrorWithMessageOrReason(
+  error: unknown
+): error is { message?: string; reason?: string } {
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    ('message' in error || 'reason' in error)
+  );
+}
--- a/lib/errorHandlers/index.ts
+++ b/lib/errorHandlers/index.ts
@@ -1,42 +1,34 @@
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const {
+import { logger } from '@hubspot/local-dev-lib/logger';
+import {
   isHubSpotHttpError,
-  isSystemError,
-  isFileSystemError,
   isValidationError,
-  isMissingScopeError,
-} = require('@hubspot/local-dev-lib/errors/index');
-import { FileSystemError } from '@hubspot/local-dev-lib/models/FileSystemError';
+} from '@hubspot/local-dev-lib/errors/index';
 import { HubSpotHttpError } from '@hubspot/local-dev-lib/models/HubSpotHttpError';
-import { BaseError } from '@hubspot/local-dev-lib/types/Error';
-const { shouldSuppressError } = require('./suppressError');
-const { i18n } = require('../lang');
-const util = require('util');
+import { shouldSuppressError } from './suppressError';
+import { i18n } from '../lang';
+import util from 'util';
 
 const i18nKey = 'lib.errorHandlers.index';
 
-export function logError(error: unknown, context?: typeof ApiErrorContext) {
+export function logError(error: unknown, context?: ApiErrorContext): void {
   debugError(error, context);
 
-  if (
-    shouldSuppressError(error, context) ||
-    shouldSuppressError(error, (error as HubSpotHttpError).context)
-  ) {
+  if (shouldSuppressError(error, context)) {
     return;
   }
 
-  if (isHubSpotHttpError(error) && context instanceof ApiErrorContext) {
-    (error as HubSpotHttpError).updateContext(context);
+  if (isHubSpotHttpError(error) && 'context' in error) {
+    if (shouldSuppressError(error, error.context)) {
+      return;
+    }
   }
 
-  if (isHubSpotHttpError(error) || isFileSystemError(error)) {
-    if (isValidationError(error) || isMissingScopeError(error)) {
-      logger.error((error as HubSpotHttpError).formattedValidationErrors());
-    } else {
-      logger.error((error as HubSpotHttpError | FileSystemError).message);
-    }
-  } else if (isSystemError(error)) {
-    logger.error((error as BaseError).message);
+  if (isHubSpotHttpError(error) && context instanceof ApiErrorContext) {
+    error.updateContext(context);
+  }
+
+  if (isHubSpotHttpError(error) && isValidationError(error)) {
+    logger.error(error.formattedValidationErrors());
   } else if (isErrorWithMessageOrReason(error)) {
     const message: string[] = [];
 
@@ -52,14 +44,14 @@ export function logError(error: unknown, context?: typeof ApiErrorContext) {
   }
 }
 
-export function debugError(error: unknown, context?: typeof ApiErrorContext) {
+export function debugError(error: unknown, context?: ApiErrorContext): void {
   if (isHubSpotHttpError(error)) {
     logger.debug((error as HubSpotHttpError).toString());
   } else {
-    logger.debug(i18n(`${i18nKey}.errorOccurred`, { error }));
+    logger.debug(i18n(`${i18nKey}.errorOccurred`, { error: String(error) }));
   }
 
-  if (isErrorWithCause(error)) {
+  if (error instanceof Error) {
     logger.debug(
       i18n(`${i18nKey}.errorCause`, {
         cause: util.inspect(error.cause, false, null, true),
@@ -94,10 +86,6 @@ export class ApiErrorContext {
     this.payload = props.payload || '';
     this.projectName = props.projectName || '';
   }
-}
-
-function isErrorWithCause(error: unknown): error is { cause: unknown } {
-  return typeof error === 'object' && error !== null && 'cause' in error;
 }
 
 function isErrorWithMessageOrReason(
--- a/lib/errorHandlers/index.ts
+++ b/lib/errorHandlers/index.ts
@@ -22,7 +22,7 @@ export function logError(error: unknown, context?: ApiErrorContext): void {
     }
   }
 
-  if (isHubSpotHttpError(error) && context instanceof ApiErrorContext) {
+  if (isHubSpotHttpError(error) && context) {
     error.updateContext(context);
   }
 
--- a/lib/errorHandlers/index.ts
+++ b/lib/errorHandlers/index.ts
@@ -3,7 +3,6 @@ import {
   isHubSpotHttpError,
   isValidationError,
 } from '@hubspot/local-dev-lib/errors/index';
-import { HubSpotHttpError } from '@hubspot/local-dev-lib/models/HubSpotHttpError';
 import { shouldSuppressError } from './suppressError';
 import { i18n } from '../lang';
 import util from 'util';
@@ -46,7 +45,7 @@ export function logError(error: unknown, context?: ApiErrorContext): void {
 
 export function debugError(error: unknown, context?: ApiErrorContext): void {
   if (isHubSpotHttpError(error)) {
-    logger.debug((error as HubSpotHttpError).toString());
+    logger.debug(error.toString());
   } else {
     logger.debug(i18n(`${i18nKey}.errorOccurred`, { error: String(error) }));
   }
--- a/lib/errorHandlers/suppressError.ts
+++ b/lib/errorHandlers/suppressError.ts
@@ -1,7 +1,7 @@
-// @ts-nocheck
 const {
   isSpecifiedError,
   isMissingScopeError,
+  ApiErrorContext,
 } = require('@hubspot/local-dev-lib/errors/index');
 const { logger } = require('@hubspot/local-dev-lib/logger');
 const { PLATFORM_VERSION_ERROR_TYPES } = require('../constants');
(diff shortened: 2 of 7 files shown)


lib/errorHandlers/index.ts
```diff
@@ -7,37 +6,39 @@ const {
...
 const i18nKey = 'lib.errorHandlers.index';
 
-function logError(error, context = {}) {
+export function logError(error: unknown, context?: typeof ApiErrorContext) {
   debugError(error, context);
 
   if (
     shouldSuppressError(error, context) ||
-    shouldSuppressError(error, error.context)
+    shouldSuppressError(error, (error as HubSpotHttpError).context)
```
joe-yeager, line 23: If we only want to call this for `HubSpotHttpError` we could update this to `isHubSpotHttpError(error) && shouldSuppressError(error, error.context)`, of if we want to call it in all cases we can use `shouldSuppressError(error, 'context' in error ? error.context: undefined)` so we don't make assumptions about the errors type.

lib/errorHandlers/index.ts
```diff
@@ -7,37 +6,39 @@ const {
...
+      logger.error((error as HubSpotHttpError).formattedValidationErrors());
     } else {
-      logger.error(error.message);
+      logger.error((error as HubSpotHttpError | FileSystemError).message);
     }
   } else if (isSystemError(error)) {
-    logger.error(error.message);
-  } else if (error.message || error.reason) {
-    const message = [];
+    logger.error((error as BaseError).message);
```
joe-yeager, line 39: `isSystemError` should be type narrowing, is this assertion necessary?

lib/errorHandlers/suppressError.ts
```diff
@@ -1,7 +1,7 @@
-// @ts-nocheck
 const {
   isSpecifiedError,
   isMissingScopeError,
+  ApiErrorContext,
```
joe-yeager, line 4: I believe this failure is because you are importing the type from LDL and not the Class from `./index`

lib/errorHandlers/suppressError.ts
```diff
@@ -15,25 +15,43 @@ const {
...
       translationKey = 'nonExistentPlatformVersion';
-      if (errorContext && errorContext[subCategory]) {
-        platformVersion = errorContext[subCategory];
+      if (
+        errorContext &&
+        (errorContext as { [key: string]: string })[subCategory]
+      ) {
+        platformVersion = (errorContext as { [key: string]: string })[
+          subCategory
+        ];
```
joe-yeager, line 54: Could we use type predicates to remove the assertions?

lib/errorHandlers/suppressError.ts
```diff
@@ -59,7 +77,10 @@ function createPlatformVersionError(err, subCategory) {
   uiLine();
 }
 
-function shouldSuppressError(err, context = {}) {
+export function shouldSuppressError(
+  err: unknown,
+  context?: typeof ApiErrorContext
```
joe-yeager, line 82: We have an `ApiErrorContext` type in LDL.  I wonder if this would be a good time to abandon the class `ApiErrorContext` in favor of the type.

lib/ui/serverlessFunctionLogs.ts
```diff
@@ -82,20 +108,32 @@ function processLog(log, options) {
...
-    return logsResp.results
-      .map(log => {
+function isLogsResponse(
+  logsResp: LogsResponse | Log
+): logsResp is LogsResponse {
+  return (
+    'results' in logsResp &&
+    Array.isArray(logsResp.results) &&
+    logsResp.results !== undefined &&
+    logsResp.results.length > 0
```
joe-yeager, line 118: I think these last two checks can be removed.  They don't really determine the type are and more logic checks.  It seems like a `LogsResponse` with result = [] is still a valid logs response.  `Array.isArray` should also handle `undefined` [example](https://jsbin.com/betihisifu/edit?js,console)

lib/ui/serverlessFunctionLogs.ts
```diff
@@ -82,20 +108,32 @@ function processLog(log, options) {
...
+): string | void {
+  if (isLogsResponse(logsResp)) {
+    return logsResp
+      .results!.map(log => {
         return processLog(log, options);
       })
       .join('\n');
   }
-  return processLog(logsResp, options);
+  return processLog(logsResp as Log, options);
```
joe-yeager, line 133: Is this type assertion necessary?  The options are `LogsReponse | Log` and return in the if statement if it is a `LogsResponse`

lib/ui/spinniesUtils.ts
```diff
@@ -86,18 +116,31 @@ function turnToValidSpinner(spinner = {}) {
...
+}: SpinnerOptions): Partial<SpinnerOptions> {
+  const colors: Partial<SpinnerOptions> = {
+    color,
+    succeedColor,
+    failColor,
+    spinnerColor,
+  };
+  (Object.keys(colors) as Array<keyof SpinnerOptions>).forEach(key => {
+    if (!VALID_COLORS.includes(colors[key] as typeof VALID_COLORS[number])) {
+      delete colors[key as keyof SpinnerOptions];
```
joe-yeager, line 133: Can we get rid of these type assertions

lib/ui/serverlessFunctionLogs.ts
```diff
@@ -14,36 +13,63 @@ const LOG_STATUS_COLORS = {
...
+  log?: string;
+  error?: {
+    type: string;
+    message: string;
+    stackTrace?: string[];
+  };
+};
+
+type LogsResponse = {
+  results?: Log[];
```
joe-yeager, line 31: Is this supposed to always be defined?  We have a type predicate `isLogsResponse` that is checking that this is defined

lib/errorHandlers/index.ts
```diff
@@ -82,21 +75,37 @@ function debugError(error, context = {}) {
...
 }
 
-module.exports = {
-  logError,
-  debugError,
-  ApiErrorContext,
-};
+function isErrorWithCause(error: unknown): error is { cause: unknown } {
+  return typeof error === 'object' && error !== null && 'cause' in error;
+}
```
joe-yeager, line 101: The base `Error` class should have a `cause`, so we can potentially make this a more general `isError` predicate that looks something like:
```typescript
function isError(error: unknown): error is Error {
  return error instanceof Error;
}
```

lib/ui/supportHyperlinks.ts
```diff
@@ -62,15 +62,15 @@ function supportsHyperlink(stream) {
...
   return false;
 }
 
-module.exports = {
+export const supportsHyperlinkModule = {
   supportsHyperlink,
-  stdout: supportsHyperlink(process.stdout),
-  stderr: supportsHyperlink(process.stderr),
+  stdout: supportsHyperlink(process.stdout as NodeJS.WriteStream),
+  stderr: supportsHyperlink(process.stderr as NodeJS.WriteStream),
```
joe-yeager, line 75: Is the `supportsHyperlink` typed correctly if we need to assert these?

lib/errorHandlers/suppressError.ts
```diff
@@ -15,25 +15,43 @@ const {
 
 const i18nKey = 'lib.errorHandlers.suppressErrors';
 
-function createPlatformVersionError(err, subCategory) {
+function createPlatformVersionError(err: unknown, subCategory: string): void {
   let translationKey = 'unspecifiedPlatformVersion';
   let platformVersion = 'unspecified platformVersion';
   const errorContext =
-    err.response && err.response.data && err.response.data.context;
+    typeof err === 'object' &&
```
camden11, line 22: I think there are potentially a few problems with this. First - it looks like we always call `createPlatformVersionError` after confirming that an error is a `HubspotHttpError`, so we can update the type for `err`. Second, I think it might be possible that this code was never updated when we updated things to use `HubspotHttpError` - it looks `data` and `context` should exist on the error object

lib/errorHandlers/suppressError.ts
```diff
@@ -15,25 +15,43 @@ const {
...
 
   switch (subCategory) {
-    case [PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_RETIRED]:
+    case PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_RETIRED:
       translationKey = 'platformVersionRetired';
-      if (errorContext && errorContext[subCategory]) {
-        platformVersion = errorContext[subCategory];
+      if (
+        errorContext &&
+        (errorContext as { [key: string]: string })[subCategory]
```
camden11, line 39: I think if you sort out the above issue with `HubspotHttpError` you shouldn't have to cast here

lib/errorHandlers/suppressError.ts
camden11, line 82: Don't need typeof here

lib/errorHandlers/suppressError.ts
```diff
@@ -78,10 +99,12 @@ function shouldSuppressError(err, context = {}) {
       subCategory: PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_NOT_SPECIFIED,
     })
   ) {
-    createPlatformVersionError(
-      err.data,
-      PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_NOT_SPECIFIED
-    );
+    if (typeof err === 'object' && err !== null && 'data' in err) {
```
camden11, line 102: Once you update to use import instead of require, you should be able to get rid of these checks

camden11 (commented):
I think a good rule to follow in these Typescript PRs is to never cast unless you absolutely have to - 99% of the time there's a better way and not letting yourself cast will force you to find it! I'm always around if you have any questions about that kind of stuff.

Also, it looks like your commits from your last PR got included here, i'm assuming this is because we are now squashing commits and you need to rebase off `main` (unfortunately i haven't worked with squash commits in a bit and I forget exactly how to do that)

lib/commonOpts.ts
```diff
@@ -101,11 +102,11 @@ export function getAccountId(
   return getAccountIdFromConfig(portal || account);
 }
 
-export function getMode(options: Arguments<{ mode?: string }>): string {
+export function getMode(options: Arguments<{ mode?: Mode }>): Mode {
   // 1. --mode
   const { mode } = options;
   if (mode && typeof mode === 'string') {
-    return mode.toLowerCase();
+    return mode.toLowerCase() as Mode;
```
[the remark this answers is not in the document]
kemmerle: @joe-yeager, I accidentally included some outdated commits with this PR. I've since rebased to `main`, so going to resolve your comments apart from those on the `errorHandlers` files.

lib/errorHandlers/index.ts
```diff
@@ -7,37 +6,39 @@ const {
   isValidationError,
   isMissingScopeError,
 } = require('@hubspot/local-dev-lib/errors/index');
```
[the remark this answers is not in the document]
kemmerle: No, I just forget. Please insert Homer Simpson GIF receding into the bushes.

lib/errorHandlers/suppressError.ts
kemmerle: @joe-yeager, do you mean `HubSpotHttpErrorContext` here? There's no error type explicitly called `ApiErrorContext`.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/suppressError.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/suppressError.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/suppressError.ts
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/suppressError.ts
```diff
@@ -15,25 +15,43 @@ const {
...
+    err !== null &&
+    'response' in err &&
+    typeof err.response === 'object' &&
+    err.response !== null &&
+    'data' in err.response &&
+    typeof err.response.data === 'object' &&
+    err.response.data !== null &&
+    'context' in err.response.data
+      ? err.response.data.context
+      : undefined;
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/suppressError.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/suppressError.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/suppressError.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/146dcbafeabd6d865cb305cb61cf867207c31fbf

lib/errorHandlers/index.ts
```diff
@@ -82,21 +75,37 @@ function debugError(error, context = {}) {
...
     this.projectName = props.projectName || '';
   }
 }
 
-module.exports = {
-  logError,
-  debugError,
-  ApiErrorContext,
-};
+function isErrorWithCause(error: unknown): error is { cause: unknown } {
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
```diff
@@ -51,25 +52,17 @@ function logError(error, context = {}) {
...
 }
 
-/**
- * Logs (debug) the error and context objects.
- *
- * @param {Error}  error
- * @param {ApiErrorContext} context
- */
-function debugError(error, context = {}) {
+export function debugError(error: unknown, context?: typeof ApiErrorContext) {
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
```diff
@@ -7,37 +6,39 @@ const {
...
-    shouldSuppressError(error, error.context)
+    shouldSuppressError(error, (error as HubSpotHttpError).context)
   ) {
     return;
   }
 
-  if (isHubSpotHttpError(error) && context) {
-    error.updateContext(context);
+  if (isHubSpotHttpError(error) && context instanceof ApiErrorContext) {
+    (error as HubSpotHttpError).updateContext(context);
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
```diff
@@ -7,37 +6,39 @@ const {
...
+import { BaseError } from '@hubspot/local-dev-lib/types/Error';
 const { shouldSuppressError } = require('./suppressError');
 const { i18n } = require('../lang');
 const util = require('util');
-const { isAxiosError } = require('axios');
 
 const i18nKey = 'lib.errorHandlers.index';
 
-function logError(error, context = {}) {
+export function logError(error: unknown, context?: typeof ApiErrorContext) {
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
```diff
@@ -51,25 +52,17 @@ function logError(error, context = {}) {
...
- * Logs (debug) the error and context objects.
- *
- * @param {Error}  error
- * @param {ApiErrorContext} context
- */
-function debugError(error, context = {}) {
+export function debugError(error: unknown, context?: typeof ApiErrorContext) {
   if (isHubSpotHttpError(error)) {
-    logger.debug(error.toString());
+    logger.debug((error as HubSpotHttpError).toString());
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
```diff
@@ -7,37 +6,39 @@ const {
...
+    (error as HubSpotHttpError).updateContext(context);
   }
 
   if (isHubSpotHttpError(error) || isFileSystemError(error)) {
     if (isValidationError(error) || isMissingScopeError(error)) {
-      logger.error(error.formattedValidationErrors());
+      logger.error((error as HubSpotHttpError).formattedValidationErrors());
     } else {
-      logger.error(error.message);
+      logger.error((error as HubSpotHttpError | FileSystemError).message);
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1251/commits/9fc06a16fccfda6b8e5d060b6c4703e0a83a0d07

lib/errorHandlers/index.ts
```diff
@@ -1,43 +1,35 @@
...
   }
 
-  if (isHubSpotHttpError(error) && context) {
+  if (isHubSpotHttpError(error) && 'context' in error) {
+    if (shouldSuppressError(error, error.context)) {
+      return;
+    }
+  }
+
+  if (isHubSpotHttpError(error) && context instanceof ApiErrorContext) {
```
camden11, line 25: This can just be  `isHubSpotHttpError(error) && context` since we know that if `context` exists, its of type `ApiErrorContext`

camden11 (commented):
Nice, things are looking really good here! I just have a few small comments and then this should be good to go

lib/errorHandlers/index.ts
```diff
@@ -51,25 +43,17 @@ function logError(error, context = {}) {
...
- * @param {Error}  error
- * @param {ApiErrorContext} context
- */
-function debugError(error, context = {}) {
+export function debugError(error: unknown, context?: ApiErrorContext): void {
   if (isHubSpotHttpError(error)) {
     logger.debug(error.toString());
   } else {
-    logger.debug(i18n(`${i18nKey}.errorOccurred`, { error }));
+    logger.debug(i18n(`${i18nKey}.errorOccurred`, { error: String(error) }));
```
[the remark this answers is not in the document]
kemmerle: That throws an error that `error is of type unknown`. I'm going to keep the same for now.

lib/errorHandlers/index.ts
kemmerle: https://github.com/HubSpot/hubspot-cli/pull/1251/commits/38daa6fc28f0896853743c7df3b3312284c1420d

lib/errorHandlers/index.ts
```diff
@@ -51,25 +43,17 @@ function logError(error, context = {}) {
...
   }
 
-  if (error.cause) {
+  if (error instanceof Error) {
     logger.debug(
       i18n(`${i18nKey}.errorCause`, {
-        cause: isAxiosError(error.cause)
-          ? error.cause
-          : util.inspect(error.cause, false, null, true),
+        cause: util.inspect(error.cause, false, null, true),
```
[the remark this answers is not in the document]
kemmerle: Yes, it's actually an improvement! Because we are checking whether `error` is an instance of `Error` here, we know that `cause` will be defined. We could return `String(error.cause)` here, but `util.inspect` actually converts the `cause` object into a string that's useful for logging and debugging. I asked GitHub Copilot to give an explanation, it's pretty good:

```
The [util.inspect](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) function in Node.js is used to convert objects into a string representation that is useful for debugging and logging.

In the context of [util.inspect(error.cause, false, null, true)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html), here's what each parameter does:

[error.cause](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html): The object to be inspected. In this case, it is the [cause](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) property of the [error](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) object.
false: This parameter indicates whether to show hidden properties. false means hidden properties will not be shown.
null: This parameter sets the depth to which the object should be inspected. null means there is no limit to the depth.
true: This parameter indicates whether to enable color highlighting in the output. true means color highlighting is enabled.
So, [util.inspect(error.cause, false, null, true)](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) converts the [error.cause](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) object into a string with no depth limit, without showing hidden properties, and with color highlighting enabled. This string is then used for logging or debugging purposes.
```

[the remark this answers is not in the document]
kemmerle: Going to resolve this for now, feel free to reopen if there's something I've missed.

brandenrodgers (approved):

PR #1252, opened 2024-11-20
feat: Add translation to IR

joe-yeager:
## Description and Context

- Integrate with `@hubspot/projects-parsing-lib` and the new projects upload API
- Adds a `-use-v3` option to `hs project upload` and `hs project deploy` to use the v3 version of the projects APIs


commands/project/upload.ts
```diff
@@ -128,6 +131,11 @@ exports.builder = yargs => {
       type: 'string',
       default: '',
     },
+    translate: {
+      hidden: true,
+      type: 'boolean',
+      default: false,
+    },
```
joe-yeager, line 138: I added this to `hs project upload` but not `hs project watch` primarily because `hs project watch` gets almost no usage, and there were some other road bumps, such as `hs project watch` doesn't work unless your project already has a successful build, which is not currently achievable with the new system.

I can add the options to `hs project watch` if we think it is worthwhile, but the command won't work until build scheduling is in place and component teams have begun to implement their builds.

commands/project/upload.ts
```diff
@@ -43,6 +43,8 @@ exports.handler = async options => {
 
   const { projectConfig, projectDir } = await getProjectConfig();
 
+  trackCommandUsage('project-upload', { type: accountType }, derivedAccountId);
```
brandenrodgers, line 46: is this a duplicate tracking call?

joe-yeager: Good catch, removed https://github.com/HubSpot/hubspot-cli/pull/1252/commits/b776d671bad10894887f4fd1cf9a1834c5852f97

lib/projects/index.ts
```diff
@@ -312,14 +312,6 @@ export async function ensureProjectExists(
...
       }
     }
-    if (
-      isSpecifiedError(err, {
-        statusCode: 401,
-      })
-    ) {
-      logger.error(err.message);
-      process.exit(EXIT_CODES.ERROR);
-    }
```
joe-yeager, line 322: This was redundant, we are calling `logError` right below and that is the pattern we want to use instead of logging the whole `message` prop

lib/errorHandlers/index.ts
```diff
@@ -51,12 +50,10 @@ export function debugError(error: unknown, context?: ApiErrorContext): void {
     logger.debug(i18n(`${i18nKey}.errorOccurred`, { error: String(error) }));
   }
 
-  if (error instanceof Error && error.cause) {
+  if (error instanceof Error && error.cause && !isHubSpotHttpError(error)) {
```
joe-yeager, line 53: We don't want to log a `HubSpotHttpError` because it will contain the full axios error as the cause and will flood the terminal with irrelevant information.

brandenrodgers (approved):
lgtm!

PR #1256, opened 2024-11-22
Refactor function commands

brandenrodgers:
## Description and Context

This refactors the functions commands to conform to our guidelines. Major changes:

#### hs logs
- [x] Update description to: "View logs for a CMS serverless function."
- [x] The `endpoint` should be kept as a positional and should fall back to a prompt when not specified

#### hs functions
- [x] Update description to: "Commands for managing CMS serverless functions."
- [x] Fix duplicate argument issue for config and account in nested commands
- [x] The primary way to run this command should be `hs function`, but we should continue to support `hs functions`

#### hs functions list
- [x] Update description to: "List the currently deployed CMS serverless functions."


commands/function.ts
```diff
@@ -0,0 +1,23 @@
...
+
+exports.command = ['function', 'functions'];
+exports.describe = i18n(`${i18nKey}.describe`);
+
+exports.builder = yargs => {
+  yargs
+    .command({
+      ...list,
+      aliases: 'ls',
+    })
```
joe-yeager, line 17: Should we move this to the `command` field in the list subcommand?

joe-yeager (approved):
1 non blocking question about moving the list alias, but other than that LGTM!

commands/function.ts
brandenrodgers: Ah yeah I saw you did that in another PR. I'll do the same 👍

PR #1257, opened 2024-11-22
feat: Update secrets commands to match design guidelines

joe-yeager:
#### hs secrets
-  [x] Update description to: "Commands for managing secrets."
- [x] Change primary command to `hs secret`, but continue to support `hs secrets` as a plural alias

#### hs secrets list
- [x] Update description to: "List all secrets."

#### hs secrets add
- [x] Update description to: "Create a new secret."
- [x] Name input should fall back to an input prompt when not specified
- [x] Add a check to see if the secret already exists before prompting for value and creating

#### hs secrets update
- [x] Add a period to the end of the description
- [x] Name input should fall back to a list of available secrets when not specified
- [x] Add a check to see if the secret exists before prompting for value and attempting to update

#### hs secrets delete
- [x] Add a period to the end of the description
- [x] Name input should fall back to a list of available secrets when not specified
- [x] Prompt for confirmation before delete. Support `--force` flag to bypass this
- [x] Add a check to see if the secret exists before attempting to delete

### DIFF
--- a/lib/prompts/promptUtils.ts
+++ b/lib/prompts/promptUtils.ts
@@ -8,13 +8,13 @@ export async function confirmPrompt(
   message: string,
   options: { defaultAnswer?: boolean; when?: boolean | (() => boolean) } = {}
 ): Promise<boolean> {
-  const { defaultAnswer, when } = options;
+  const { defaultAnswer = true, when } = options;
   const { choice } = await promptUser([
     {
       name: 'choice',
       type: 'confirm',
       message,
-      default: defaultAnswer || true,
+      default: defaultAnswer,
       when,
     },
   ]);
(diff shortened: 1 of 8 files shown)


lib/prompts/promptUtils.ts
```diff
@@ -8,13 +8,13 @@ export async function confirmPrompt(
...
 ): Promise<boolean> {
-  const { defaultAnswer, when } = options;
+  const { defaultAnswer = true, when } = options;
   const { choice } = await promptUser([
     {
       name: 'choice',
       type: 'confirm',
       message,
-      default: defaultAnswer || true,
+      default: defaultAnswer,
```
joe-yeager, line 17: This is fixing a recently introduced bug where the `default` value was alway always see to true because `something || true` will always be `true`

brandenrodgers (approved):
🎉

PR #1258, opened 2024-11-22
Restructure filemanager commands

brandenrodgers:
## Description and Context

Refactoring the filemanager commands. This one is pretty small. Just removing the options that show when running `hs filemanager` with no sub command.

I also found a bug in the fetch command where we weren't respecting the `override` option. I don't think it was being used anywhere in the upload command so I didn't add it in there.

#### hs filemanager
- [x] Update description to: "Commands for managing files in the File Manager."
- [x] Fix duplicate argument issue for config and account in nested commands

#### hs filemanager upload
- [x] Update description to: "Upload a folder or file to the File Manager."
- [x] Remove `--version` flag because it's unnecessary

#### hs filemanager fetch
- [x] Update description to: "Fetch a folder or file from the File Manager."
- [x] Remove `--version` flag because it's unnecessary


commands/filemanager.ts
```diff
@@ -14,10 +9,6 @@ exports.command = 'filemanager';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.builder = yargs => {
-  addOverwriteOptions(yargs);
-  addConfigOptions(yargs);
-  addAccountOptions(yargs);
```
camden11, line 19: This is something that maybe we should find a consistent pattern for. Seems like there are a lot of redundant `addOptions` calls between the actual command files and the files for the command bucket. Do you think it makes sense to just add them individually? In this case it looks like there are a few shared ones between filemanager fetch and upload that we could add in the bucket.

I also took advantage of the buckets to not have to add `addGlobalOptions` to every command individually

brandenrodgers: I don't have a very strong opinion on this, but my reasoning for being redundant and moving all of these options to the commands is because it's easy to forget what's being declared at this level. We had an issue where we were calling `addAccountOptions(yargs)` in these command bucket files and also in the individual commands themselves.

I think ideally we just use middleware for stuff like this (where we need it to apply to every command), but in the meantime it feels slightly more straightforward to me to keep the definitions in each file just so we can quickly understand the full set of options that each command is subscribing to. But that's just my reasoning for why I did it this way. I could be convinced either way tbh.

joe-yeager: > I think ideally we just use middleware for stuff like this (where we need it to apply to every command), but in the meantime it feels slightly more straightforward to me to keep the definitions in each file just so we can quickly understand the full set of options that each command is subscribing to. But that's just my reasoning for why I did it this way. I could be convinced either way tbh.

+1 on this.  I do like the ease of adding them flags in the buckets, but putting the flags in the final commands is so much easier to reason about.

PR #1261, opened 2024-11-22
feat: Restructure custom-object commands to match design guidelines

joe-yeager:
#### hs custom-object
- [x] Update description to: "[BETA] Commands for managing custom objects."
- [x]  Fix duplicate argument issue for config and account in nested commands
- [x] Support the plural as an alias: `hs custom-objects`
- [x] Remove support for the `custom` alias (too generic)

#### hs custom-object schema
- [x] Update description to: "Commands for managing custom object schemas."
- [x] support custom-object schemas as an alias

#### hs custom-object schema list
- [x] Update description to: "List custom object schemas."

#### hs custom-object schema fetch
- [x] Update description to: "Fetch a custom object schema."
- [x] Name should fall back to a prompt when not specified
- [x] dest should fall back to a prompt when not specified

#### hs custom-object schema fetch-all
- [x] dest should fall back to a prompt when not specified

#### hs custom-object schema create
- [x] Add a period to the end of the description
- [x] definition should be renamed to `path` and should be moved to be a flag instead of a positional

#### hs custom-object schema update
- [x]  Add a period to the end of the description
- [x] definition should be renamed to `path` and should be moved to be a flag instead of a positional

#### hs custom-object schema delete
- [x] Add a period to the end of the description
- [x] Should list available schemas in a prompt when name is not included
- [x] should prompt for confirmation before deleting and support a `--force` flag to bypass the prompt

#### hs custom-object create
- [x] Add a period to the end of the description
- [x] definition should be renamed to `path` and should be moved to be a flag instead of a positional
- [x] Name should fall back to a prompt when not specified

### DIFF
--- a/commands/customObject/create.ts
+++ b/commands/customObject/create.ts
@@ -15,7 +15,7 @@ const { i18n } = require('../../lib/lang');
 const i18nKey = 'commands.customObject.subcommands.create';
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 
-exports.command = 'create <name> <definition>';
+exports.command = 'create [name]';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.handler = async options => {
@@ -33,6 +33,9 @@ exports.handler = async options => {
   }
 
   try {
+    if (!name) {
+      // TODO prompt for the name
+    }
     await batchCreateObjects(derivedAccountId, name, objectJson);
     logger.success(i18n(`${i18nKey}.success.objectsCreated`));
   } catch (e) {
@@ -46,13 +49,13 @@ exports.handler = async options => {
 };
 
 exports.builder = yargs => {
-  yargs.positional('name', {
-    describe: i18n(`${i18nKey}.positionals.name.describe`),
-    type: 'string',
-  });
-
-  yargs.positional('definition', {
-    describe: i18n(`${i18nKey}.positionals.definition.describe`),
-    type: 'string',
-  });
+  yargs
+    .positional('name', {
+      describe: i18n(`${i18nKey}.positionals.name.describe`),
+      type: 'string',
+    })
+    .option('path', {
+      describe: i18n(`${i18nKey}.options.path.describe`),
+      type: 'string',
+    });
 };
--- a/commands/customObject/schema/delete.ts
+++ b/commands/customObject/schema/delete.ts
@@ -1,4 +1,6 @@
 // @ts-nocheck
+import { EXIT_CODES } from '../../../lib/enums/exitCodes';
+
 const { logger } = require('@hubspot/local-dev-lib/logger');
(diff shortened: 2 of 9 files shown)


commands/customObject/schema/delete.ts
```diff
@@ -10,17 +14,37 @@ const { logError } = require('../../../lib/errorHandlers');
...
+      (await listPrompt(i18n(`${i18nKey}.selectSchema`), {
+        choices: schemaNames,
+      }));
+
+    const shouldDelete =
+      force ||
+      (await confirmPrompt(i18n(`${i18nKey}.confirmDelete`, { name })));
+
+    if (!shouldDelete) {
+      logger.success(i18n(`${i18nKey}.deleteCancelled`, { name }));
```
camden11, line 44: Feel like this might make more sense as a `log` or `info`

commands/customObject/create.ts
```diff
@@ -15,44 +15,48 @@ const { i18n } = require('../../lib/lang');
...
-  yargs.positional('definition', {
-    describe: i18n(`${i18nKey}.positionals.definition.describe`),
-    type: 'string',
-  });
+  yargs
+    .positional('name', {
+      describe: i18n(`${i18nKey}.positionals.name.describe`),
+      type: 'string',
+    })
+    .option('path', {
```
camden11, line 57: Should there be a prompt fort `path` since it's required? (I saw this wasn't mentioned in the issue so probably not, just double checking)

lib/prompts/promptUtils.ts
```diff
@@ -42,3 +42,22 @@ export async function listPrompt(
...
+  const { input } = await promptUser([
+    {
+      name: 'input',
+      type: 'input',
+      message,
+      when,
+    },
+  ]);
+  return input;
+}
```
camden11, line 63: we really just need to upgrade to the new inquirer haha

camden11 (approved):
A few comments/questions, nothing really blocking though

lib/prompts/promptUtils.ts
joe-yeager: Yeah haha.  The current version we are on its pretty unintuitive

commands/customObject/create.ts
joe-yeager: Probably a good call to add one.  I'll update it 👍

joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1261/commits/55b705e95243103a419bf87e4dafa3a0f3d64080

commands/customObject/schema/delete.ts
joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1261/commits/55b705e95243103a419bf87e4dafa3a0f3d64080

PR #1264, opened 2024-11-22
feat!: Restructure config commands to match design guidelines

joe-yeager:
#### hs config
- Update description to: "Commands for managing the CLI config file."
- Remove all flags except for the `--help` flag because they don't do anything

#### hs config set
- Add a period to the description
- flags should be kebab case
- Remove --version flag

### DIFF
--- a/commands/config.ts
+++ b/commands/config.ts
@@ -1,4 +1,6 @@
 // @ts-nocheck
+const { addConfigOptions } = require('../lib/commonOpts');
+
 const { i18n } = require('../lib/lang');
 const set = require('./config/set');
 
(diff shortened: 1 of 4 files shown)


brandenrodgers (approved):
Looks like a test is failing, but the changes LGTM 🙌

commands/config.ts
```diff
@@ -10,8 +11,6 @@ exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.builder = yargs => {
   addConfigOptions(yargs);
```
joe-yeager, line 13: The issue said to remove all the flags but removing the `--config` flag breaks the logic

PR #1265, opened 2024-11-22
feat!: Restructure account commands to meet the new design guidelines

joe-yeager:
#### hs accounts
- [x] Update description to: "Commands for managing configured accounts."
- [x] Support `hs account` as the primary and accounts as the plural alias

#### hs accounts list
- [x] Remove --version flag and --account flag

#### hs accounts rename
- [x] Description should be `Rename an account in the config.`
- [x] Remove --version flag and --account flag

#### hs accounts use
- [x] Remove --version flag and --account flag

#### hs accounts info
- [x] Description should be `Print information about the default account, or about the account specified with the
"account" option.`
- [x] Remove --version flag and --account flags

#### hs accounts remove
- [x] Remove --version flag and --account flag

#### hs accounts clean
- [x] Remove --version flag, use-env flag,  and --account flag

### DIFF
--- a/commands/__tests__/accounts.test.ts
+++ b/commands/__tests__/accounts.test.ts
@@ -14,6 +14,7 @@ jest.mock('../accounts/use');
 jest.mock('../accounts/info');
 jest.mock('../accounts/remove');
 jest.mock('../accounts/clean');
+jest.mock('../../lib/commonOpts');
 yargs.command.mockReturnValue(yargs);
 yargs.demandCommand.mockReturnValue(yargs);
 
@@ -56,6 +57,7 @@ describe('commands/accounts', () => {
     });
 
     it.each(subcommands)('should attach the %s subcommand', (name, module) => {
+      console.log(name, module);
       accountsCommand.builder(yargs);
       expect(yargs.command).toHaveBeenCalledWith(module);
     });
--- a/commands/__tests__/accounts.test.ts
+++ b/commands/__tests__/accounts.test.ts
@@ -6,7 +6,6 @@ import use from '../accounts/use';
 import info from '../accounts/info';
 import remove from '../accounts/remove';
 import clean from '../accounts/clean';
-import { addAccountOptions, addConfigOptions } from '../../lib/commonOpts';
 
 jest.mock('yargs');
 jest.mock('../accounts/list');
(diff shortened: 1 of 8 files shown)


commands/accounts/list.ts
```diff
@@ -112,10 +111,6 @@ exports.handler = async options => {
 };
 
 exports.builder = yargs => {
-  addConfigOptions(yargs);
```
joe-yeager, line 115: Whoops, got a little carried away. I'll the config options back

joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1265/commits/382facff635ca769953e6a5e1a402f6a56e4a491

commands/__tests__/accounts.test.ts
```diff
@@ -37,7 +35,7 @@ describe('commands/accounts', () => {
 
   describe('builder', () => {
     const subcommands = [
-      ['list', { ...list, aliases: 'ls' }],
+      ['list', list],
```
joe-yeager, line 38: I moved this to the list file `command` field to match the pattern we use elsewhere.

commands/__tests__/accounts.test.ts
```diff
@@ -52,22 +51,13 @@ describe('commands/accounts', () => {
...
-      expect(addAccountOptions).toHaveBeenCalledWith(yargs);
-    });
-
     it('should add the correct number of sub commands', () => {
       accountsCommand.builder(yargs);
       expect(yargs.command).toHaveBeenCalledTimes(subcommands.length);
     });
 
     it.each(subcommands)('should attach the %s subcommand', (name, module) => {
+      console.log(name, module);
```
brandenrodgers, line 60: Looks like a console log slipped in (unless intentional?)

commands/accounts/remove.ts
```diff
@@ -16,16 +16,16 @@ const { loadAndValidateOptions } = require('../../lib/validation');
 
 const i18nKey = 'commands.accounts.subcommands.remove';
 
-exports.command = 'remove [--account]';
+exports.command = 'remove <account>';
```
brandenrodgers, line 19: Does this make account required? I think it should be optional right since we fall back to a prompt?

commands/__tests__/accounts.test.ts
joe-yeager: Whoops, left over from debugging a failure

commands/accounts/remove.ts
joe-yeager: It sure does 😅, for some reason I thought we weren't prompting.  I'll revert this.

brandenrodgers (approved):

PR #1266, opened 2024-11-22
Restructure project commands to meet guidelines

brandenrodgers:
#### hs project
- [x] Update description to: "Commands for managing projects..."
- [x] Support `hs projects` as a plural alias

#### hs project create
- [x] Add a period to the description
- [x] use `--dest` instead of location
- [x] use kebab case in all flags
- [x] Remove `--version` flag because it's unnecessary

#### hs project add
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary

#### hs project watch
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] Remove path positional (this should be run from within a project)

#### hs project dev
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] Remove account positional

#### hs project upload
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] Remove path, forceCreate, and message positionals
- [x] Support `-m` alias for message
- [x] use kebab case for arguments

#### hs project deploy
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] use kebab case for flags
 
#### hs project logs
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary

#### hs project list-builds
- [x] Add a period to the description
- [x] Remove path positional. Support `--project` flag, but fall back to the project in the CWD when not provided
 
#### hs project download
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] project should be a flag, not a positional
- [x] use kebab for all flags

#### hs project open
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] project should be a flag, not a positional

#### hs project migrate-app
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] location should be switched to dest

#### hs project clone-app
- [x] Add a period to the description
- [x] Remove `--version` flag because it's unnecessary
- [x] location should be switched to dest


commands/project/cloneApp.ts
```diff
@@ -153,14 +153,14 @@ exports.handler = async options => {
...
-      logger.success(i18n(`${i18nKey}.cloneStatus.success`, { location }));
+      logger.success(i18n(`${i18nKey}.cloneStatus.success`, { dest }));
       logger.log('');
       process.exit(EXIT_CODES.SUCCESS);
     }
   } catch (error) {
     trackCommandMetadataUsage(
       'clone-app',
-      { projectName: name, appId, status: 'FAILURE', error },
+      { projectName, appId, status: 'FAILURE', error },
```
kemmerle, line 163: Are these properties tracked with the way that the usage tracking API is set up?

commands/project/download.ts
```diff
@@ -127,8 +131,9 @@ exports.builder = yargs => {
       describe: i18n(`${i18nKey}.options.dest.describe`),
       type: 'string',
     },
-    buildNumber: {
-      describe: i18n(`${i18nKey}.options.buildNumber.describe`),
+    build: {
+      describe: i18n(`${i18nKey}.options.build.describe`),
+      alias: ['build-id'],
```
kemmerle, line 136: Should we also leave `build-number` as an alias to make this a less breaking change?

commands/project/listBuilds.ts
```diff
@@ -29,22 +26,23 @@ const { isHubSpotHttpError } = require('@hubspot/local-dev-lib/errors/index');
...
 
-  const cwd = projectPath ? path.resolve(getCwd(), projectPath) : getCwd();
-  const { projectConfig, projectDir } = await getProjectConfig(cwd);
-
-  validateProjectConfig(projectConfig, projectDir);
+  let projectName = projectFlagValue;
 
-  logger.debug(`Fetching builds for project at path: ${projectPath}`);
+  if (!projectName) {
+    const { projectConfig, projectDir } = await getProjectConfig();
```
kemmerle, line 42: Does the user need to run this command from within a project folder? What happens if `getProjectConfig` fails?

commands/project/upload.ts
```diff
@@ -27,19 +27,19 @@ const { EXIT_CODES } = require('../../lib/enums/exitCodes');
...
 
-  const { forceCreate, path: projectPath, message, derivedAccountId } = options;
+  const { forceCreate, message, derivedAccountId } = options;
   const accountConfig = getAccountConfig(derivedAccountId);
   const accountType = accountConfig && accountConfig.accountType;
 
   trackCommandUsage('project-upload', { type: accountType }, derivedAccountId);
 
-  const { projectConfig, projectDir } = await getProjectConfig(projectPath);
+  const { projectConfig, projectDir } = await getProjectConfig();
```
kemmerle, line 42: Does this mean that the user needs to run the command from within a project folder? Should we add some help text to indicate that?

commands/project/watch.ts
```diff
@@ -88,11 +88,11 @@ const handleUserInput = (accountId, projectName, currentBuildId) => {
 exports.handler = async options => {
   await loadAndValidateOptions(options);
 
-  const { initialUpload, path: projectPath, derivedAccountId } = options;
+  const { initialUpload, derivedAccountId } = options;
 
   trackCommandUsage('project-watch', null, derivedAccountId);
 
-  const { projectConfig, projectDir } = await getProjectConfig(projectPath);
+  const { projectConfig, projectDir } = await getProjectConfig();
```
kemmerle, line 95: See my comment on the `upload` command.

kemmerle:
I left some comments with questions about whether command behavior will change. If we are changing command behavior, I think adding some comments in `--help` would be good.

The tests should also be updated; they are failing right now because `location` is `undefined` (probably to do with changes to the `createProjectPrompt`.

commands/project/cloneApp.ts
brandenrodgers: oh good call. They probably aren't even being tracked. I can remove them

commands/project/download.ts
brandenrodgers: All these audit changes are already going to be breaking so I'd personally prefer to just rip the bandaid here since we don't refer to `build number` anywhere else in projects

commands/project/upload.ts
brandenrodgers: Yeah I can look into making this more obvious via help text

commands/project/listBuilds.ts
brandenrodgers: So this command can be run from both inside and outside of a project. If the dev runs it from outside of a project they need to include the `[--project]` flag. If they don't include that flag, we'll throw an error and tell them that they need to run it from inside a project. If they run from inside a project it'll just default to the project they're in.

The ideal way to run this is from inside a project, but we do support running it with `--project` to target a specific project in your hubspot account

commands/project/upload.ts
brandenrodgers: Added better copy in [6d99f67](https://github.com/HubSpot/hubspot-cli/pull/1266/commits/6d99f67a0643d7d87a8c75fc8f866891f70fcfeb)

commands/project/cloneApp.ts
brandenrodgers: Fixed in [6d99f67](https://github.com/HubSpot/hubspot-cli/pull/1266/commits/6d99f67a0643d7d87a8c75fc8f866891f70fcfeb)

commands/project/watch.ts
brandenrodgers: Added better copy in [6d99f67](https://github.com/HubSpot/hubspot-cli/pull/1266/commits/6d99f67a0643d7d87a8c75fc8f866891f70fcfeb)

joe-yeager (approved):

commands/project/download.ts
```diff
@@ -116,6 +119,7 @@ exports.handler = async options => {
 };
 
 exports.builder = yargs => {
+  addAccountOptions(yargs);
```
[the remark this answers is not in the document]
brandenrodgers: good call, done in my latest changes 👍

PR #1267, opened 2024-11-22
chore: Refactor hubdb commands

kemmerle:
## Description and Context
As part of our CLI audit efforts, I have refactored the `hs hubdb` commands to conform to our CLI Design Guidelines. I'll list the major changes below for testing convenience:

`hs hubdb clear`:
-Falls back to prompt if tableId is not specified. 

`hs hubdb create`: 
-Removes the `src` positional in favor of a flag `--path`

`hs hubdb fetch`:
-Falls back to prompts for both `tableId` and `dest` if not specified

`hs hubdb delete`:
 -Falls back to prompt for table Id if not specified
-Prompts for confirmation before deleting and supports a `--force flag` to bypass it

## Screenshots

Screenshot 2024-11-22 at 4 39 56 PM

Screenshot 2024-11-22 at 4 40 20 PM

## TODO

- [ ] Add `addGlobalOptions` middleware to remove `--version` flag
- [ ] Address feedback 

## Who to Notify

### DIFF
--- a/commands/hubdb/create.ts
+++ b/commands/hubdb/create.ts
@@ -11,35 +11,31 @@ const {
   loadAndValidateOptions,
 } = require('../../lib/validation');
 const { trackCommandUsage } = require('../../lib/usageTracking');
-const {
-  addConfigOptions,
-  addAccountOptions,
-  addUseEnvironmentOptions,
-} = require('../../lib/commonOpts');
+const { addUseEnvironmentOptions } = require('../../lib/commonOpts');
 const { i18n } = require('../../lib/lang');
 
 const i18nKey = 'commands.hubdb.subcommands.create';
 const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 
-exports.command = 'create <src>';
+exports.command = 'create';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 exports.handler = async options => {
-  const { src, derivedAccountId } = options;
+  const { path: providedFilePath, derivedAccountId } = options;
 
   await loadAndValidateOptions(options);
 
   trackCommandUsage('hubdb-create', null, derivedAccountId);
 
   try {
-    const filePath = path.resolve(getCwd(), src);
+    const filePath = path.resolve(getCwd(), providedFilePath);
     if (!checkAndConvertToJson(filePath)) {
       process.exit(EXIT_CODES.ERROR);
     }
 
     const table = await createHubDbTable(
       derivedAccountId,
-      path.resolve(getCwd(), src)
+      path.resolve(getCwd(), providedFilePath)
     );
     logger.success(
       i18n(`${i18nKey}.success.create`, {
@@ -51,7 +47,7 @@ exports.handler = async options => {
   } catch (e) {
     logger.error(
       i18n(`${i18nKey}.errors.create`, {
-        src,
+        providedFilePath,
       })
     );
     logError(e);
@@ -59,12 +55,10 @@ exports.handler = async options => {
 };
 
 exports.builder = yargs => {
-  addAccountOptions(yargs);
-  addConfigOptions(yargs);
   addUseEnvironmentOptions(yargs);
 
-  yargs.positional('src', {
-    describe: i18n(`${i18nKey}.positionals.src.describe`),
+  yargs.options('path', {
+    describe: i18n(`${i18nKey}.options.path.describe`),
     type: 'string',
   });
 };
--- a/commands/hubdb/delete.ts
+++ b/commands/hubdb/delete.ts
@@ -4,27 +4,49 @@ const { logError } = require('../../lib/errorHandlers/index');
 const { deleteTable } = require('@hubspot/local-dev-lib/api/hubdb');
 const { loadAndValidateOptions } = require('../../lib/validation');
 const { trackCommandUsage } = require('../../lib/usageTracking');
-
+const { addUseEnvironmentOptions } = require('../../lib/commonOpts');
 const {
-  addConfigOptions,
-  addAccountOptions,
-  addUseEnvironmentOptions,
-} = require('../../lib/commonOpts');
+  selectHubDBTablePrompt,
+} = require('../../lib/prompts/selectHubDBTablePrompt');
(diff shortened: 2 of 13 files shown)


commands/hubdb/create.ts
```diff
@@ -51,20 +47,18 @@ exports.handler = async options => {
...
-        src,
+        providedFilePath,
       })
     );
     logError(e);
   }
 };
 
 exports.builder = yargs => {
-  addAccountOptions(yargs);
```
brandenrodgers, line 62: Rather than removing these utils from each command file, you can actually keep them here and remove them from the top level definition [here](https://github.com/HubSpot/hubspot-cli/blob/main/commands/hubdb.ts#L15).

The benefit of removing them there is that it prevents those flags from showing up when you run `hs hubdb --help` (where they're useless anyways and just clutter up the console output)

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1267/commits/db6a7e1dbb0beb080e3ef4bcdd26db551c69d7db

lang/en.lyaml
```diff
@@ -1322,6 +1326,17 @@ en:
         prompt: "Open hubspot.com to install this app?"
         reinstallPrompt: "Open hubspot.com to reinstall this app?"
         decline: "To continue local development of this app, install it in your target test account and re-run {{#bold}}`hs project dev`{{/bold}}"
+      selectHubDBTablePrompt:
+        selectTable: "Select a HubDB table:"
+        enterDest: "Enter the destination path:"
+        errors:
+          noTables: "No HubDB tables found in account {{ accountId }}"
+          allTablesEmpty: "All HubDB tables in account {{ accountId }} are empty"
```
brandenrodgers, line 1334: I'm able to successfully download empty tables in my current CLI so I think we might not need these empty messages.

lib/prompts/selectHubDBTablePrompt.ts
```diff
@@ -0,0 +1,93 @@
...
+      return;
+    }
+    if (tables.every(table => table.rowCount === 0)) {
+      logger.error(i18n(`${i18nKey}.errors.allTablesEmpty`, { accountId }));
+      process.exit(EXIT_CODES.ERROR);
+    }
+    return tables;
+  } catch (error) {
+    logError(error, { accountId });
+    logger.error(i18n(`${i18nKey}.errors.errorFetchingTables`, { accountId }));
```
brandenrodgers, line 29: Will this log two separate errors to the console?

commands/hubdb/fetch.ts
```diff
@@ -48,7 +58,7 @@ exports.builder = yargs => {
   addConfigOptions(yargs);
   addUseEnvironmentOptions(yargs);
 
-  yargs.positional('tableId', {
+  yargs.positional('table-id', {
```
brandenrodgers, line 61: Should this become an option now? `yargs.option()`

commands/hubdb/delete.ts
```diff
@@ -4,34 +4,61 @@ const { logError } = require('../../lib/errorHandlers/index');
...
+    if (!force) {
+      const { shouldDeleteTable } = await promptUser({
+        name: 'shouldDeleteTable',
+        type: 'confirm',
+        message: i18n(`${i18nKey}.shouldDeleteTable`, { tableId }),
+      });
+      process.stdin.resume();
+
+      if (!shouldDeleteTable) {
+        process.exit(EXIT_CODES.SUCCESS);
```
joe-yeager, line 50: Should we add a log stating we've canceled the delete?

commands/hubdb/delete.ts
```diff
@@ -4,34 +4,61 @@ const { logError } = require('../../lib/errorHandlers/index');
...
+            isDeleteCommand: true,
+          });
+
+    if (!force) {
+      const { shouldDeleteTable } = await promptUser({
+        name: 'shouldDeleteTable',
+        type: 'confirm',
+        message: i18n(`${i18nKey}.shouldDeleteTable`, { tableId }),
+      });
+      process.stdin.resume();
```
joe-yeager, line 47: Is this required for the prompt to function?

lib/prompts/selectHubDBTablePrompt.ts
kemmerle: Yes, it will. I wanted to add some context for the user that we were unable to fetch tables from their account apart from whatever the actual error is. Any suggestions for different formatting?

commands/hubdb/fetch.ts
kemmerle: It didn't say to turn it into an `option` in the issue, but does falling back to prompts imply that the positionals become options?

commands/hubdb/delete.ts
kemmerle: I think it's clear, because we exit immediately after the prompt, Are you sure you want to delete this table? And this is a pattern we use with the `migrate-app` command. Going to resolve for now

commands/hubdb/delete.ts
kemmerle: Nope, going to remove.

lang/en.lyaml
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1267/commits/78e18349a588f00f0104bb1b53ef5284563207cb

commands/hubdb/delete.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1267/commits/78e18349a588f00f0104bb1b53ef5284563207cb

lib/prompts/selectHubDBTablePrompt.ts
brandenrodgers: Ok thanks for the clarification. I think you could probably just keep the bottom (user-friendly) error and then switch the other one to a debug log for debugging.

brandenrodgers (approved):
Just one comment about switching a log to a debug statement, but otherwise lgtm!

PR #1270, opened 2024-11-22
feat!: Restructure theme commands to match new design guidelines

joe-yeager:
#### hs theme
- [x] Update description to: "Commands for managing themes."
- [x] Support `hs themes` as a plural alias

#### hs theme marketplace-validate
- [x] Add a period to the description
- [x] src should be path
- [x] Remove --version flag

#### hs theme generate-selectors
- [x] Add a period to the description
- [x] themePath should be path
- [x] Remove --version flag

### DIFF
--- a/commands/theme/preview.ts
+++ b/commands/theme/preview.ts
@@ -29,7 +29,7 @@ const { preview } = require('@hubspot/theme-preview-dev-server');
 const { hasFeature } = require('../../lib/hasFeature');
 const i18nKey = 'commands.theme.subcommands.preview';
 
-exports.command = 'preview [--src] [--dest]';
+exports.command = 'preview [--path] [--dest]';
 exports.describe = i18n(`${i18nKey}.describe`);
 
 const validateSrcPath = src => {
@@ -232,7 +232,7 @@ exports.builder = yargs => {
   addConfigOptions(yargs);
   addAccountOptions(yargs);
 
-  yargs.option('src', {
+  yargs.option('path', {
     describe: i18n(`${i18nKey}.options.src.describe`),
     type: 'string',
     requiresArg: true,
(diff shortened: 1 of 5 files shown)


commands/theme/preview.ts
kemmerle, line 1: Need to change `options.src` to `options.path` on this line: https://github.com/HubSpot/hubspot-cli/blob/d38c611522908dfb23415d19836292e115e38609/commands/theme/preview.ts#L77.

commands/theme/preview.ts
```diff
@@ -232,7 +232,7 @@ exports.builder = yargs => {
   addConfigOptions(yargs);
   addAccountOptions(yargs);
 
-  yargs.option('src', {
+  yargs.option('path', {
     describe: i18n(`${i18nKey}.options.src.describe`),
```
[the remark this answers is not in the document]
joe-yeager: Looks like I forgot the change the i18n and the `en.lyaml`, good find https://github.com/HubSpot/hubspot-cli/pull/1270/commits/1487fb1e99b616d6cd45b82fcad1a28a9a27fcb7

commands/theme/preview.ts
joe-yeager: Good catch https://github.com/HubSpot/hubspot-cli/pull/1270/commits/1487fb1e99b616d6cd45b82fcad1a28a9a27fcb7

commands/theme/preview.ts
```diff
@@ -29,7 +29,7 @@ const { preview } = require('@hubspot/theme-preview-dev-server');
 const { hasFeature } = require('../../lib/hasFeature');
 const i18nKey = 'commands.theme.subcommands.preview';
 
-exports.command = 'preview [--src] [--dest]';
+exports.command = 'preview [--path] [--dest]';
```
brandenrodgers, line 32: iirc we had opted to keep src and dest whenever both are present, and then use `path` instead of `src` if it's the only input.

brandenrodgers (approved):
One comment (mostly just a question because I'm not exactly sure where we landed on it), but otherwise lgtm 👌

commands/theme/preview.ts
joe-yeager: Revert the changes https://github.com/HubSpot/hubspot-cli/pull/1270/commits/331c932ccceb905beb3c917bd2eb84d4e6cfea57

PR #1272, opened 2024-11-25
chore: Tweak the secrets tests

joe-yeager:
## Description and Context

The secrets tests have been really flaky in the past couple days, this aims to address them by tweaking the timeout values so we poll for a secret every 2 seconds and stop after 20 seconds.  It also merges the `hs secret list` check into the `add`, `update`, and `delete` commands to check that the secrets exist when they should.


acceptance-tests/tests/workflows/secretsFlow.spec.ts
```diff
@@ -23,52 +37,39 @@ describe('Secrets Flow', () => {
...
-  describe('hs secrets list', () => {
-    it('should list the secret', async () => {
-      await expect
-        .poll(() => testState.cli.executeWithTestConfig(['secrets', 'list']), {
-          interval: 1000,
-          timeout: 10000,
-        })
-        .toContain(SECRET.name);
-    });
+    await waitForSecretsListToContainSecret(testState);
```
joe-yeager, line 46: I moved the check to the end of this test because it was had to tell if this test was failing because it had no `assertion`, so as long as the `hs secret add` exited successful it passed.  This will now check that the secret was added before passing the test.

acceptance-tests/tests/workflows/secretsFlow.spec.ts
```diff
@@ -23,52 +37,39 @@ describe('Secrets Flow', () => {
...
-    it('should update the existing secret', async () => {
-      await testState.cli.executeWithTestConfig(
-        ['secrets', 'update', SECRET.name],
-        ['a different secret value', ENTER]
-      );
-    });
-  });
+  it('should update the existing secret', async () => {
+    // Wait for the secret to exist before updating it
+    await waitForSecretsListToContainSecret(testState);
```
joe-yeager, line 51: This test kept failing because the secret didn't exist.  Adding a polling check in to make sure it exists before attempting to run the `hs secrets update`

acceptance-tests/tests/workflows/secretsFlow.spec.ts
```diff
@@ -23,52 +37,39 @@ describe('Secrets Flow', () => {
...
-      await expect
-        .poll(() => testState.cli.executeWithTestConfig(['secrets', 'list']), {
-          interval: 1000,
-          timeout: 10000,
-        })
-        .not.toContain(SECRET.name);
-    });
+  it('should delete the secret', async () => {
+    // Wait for the secret to exist before deleting it
+    await waitForSecretsListToContainSecret(testState);
```
joe-yeager, line 61: Same as update, make sure it exists before attempting to run the delete command

acceptance-tests/tests/workflows/secretsFlow.spec.ts
```diff
@@ -11,6 +11,20 @@ const SECRET = {
   value: 'an initial secret value',
 };
 
+const secretPollingOptions = {
+  interval: 2000,
+  timeout: 20000,
```
joe-yeager, line 16: Still polling 10 times, just spacing them out over a wider wide to give the secrets a better chance of existing

brandenrodgers (approved):

PR #1273, opened 2024-11-25
chore: Convert simple prompts to TypeScript

kemmerle:
## Description and Context
In this PR, I convert simple prompts to TypeScript. These include:

`accountNamePrompt.ts`
`cleanUploadPrompt.ts`
`createFunctionPrompt.ts`
`createModulePrompt.ts`
`createTemplatePrompt.ts`
`projectsLogsPrompt.ts`
`secretPrompt.ts`
`setAsDefaultAccountPrompt.ts`

IMPORTANT: I was having trouble importing inquirer when converting prompts to TS. We're currently using inquirer v.8.2.0, which is a CommonJS module. We need to bump to the current version (v.12.1.0) to import it using ES module syntax.

Currently, we're wrapping inquirer.createPromptModule() in our helper function, promptUser. My gut feeling is that we should convert promptUser to inquirer.prompt in one PR, once all the files that consume promptUser are converted to TS. Is that a reasonable plan?

I don't see how we can support two inquirer versions (one CommonJS andv one ES module syntax), but let me know if that's possible/preferable.

## TODO

- [x] Figure out how we're going to handle `inquirer`
- [ ] Address other feedback

## Who to Notify


lib/prompts/accountNamePrompt.ts
```diff
@@ -1,36 +1,52 @@
...
-      return i18n(`${i18nKey}.errors.nameRequired`);
-    } else if (val.indexOf(' ') >= 0) {
-      return i18n(`${i18nKey}.errors.spacesInName`);
-    }
-    return accountNameExistsInConfig(val)
-      ? i18n(`${i18nKey}.errors.accountNameExists`, { name: val })
-      : true;
-  },
-});
+type AccountNamePromptConfig = {
```
camden11, line 8: nit: calling this `Config` makes me think that its something you'd use to configure the prompt, when its really the data returned by the prompt. Maybe `AccountNamePromptData` or `AccountNamePromptResponse` would make more sense

lib/prompts/accountNamePrompt.ts
```diff
@@ -1,36 +1,52 @@
...
+  name: string;
+  message: string;
+  default: string;
+  validate(val: string): string | boolean;
```
camden11, line 12: This could be a good candidate to make a util type for (if inquirer doesn't already have one)`

Something like

`type PromptValidationFunction = (name: string) => string | boolean`

lib/prompts/accountNamePrompt.ts
```diff
@@ -1,36 +1,52 @@
...
   return promptUser(getCliAccountNamePromptConfig(defaultName));
-};
+}
 
-const hubspotAccountNamePrompt = ({ accountType, currentPortalCount = 0 }) => {
+export function hubspotAccountNamePrompt({
+  accountType,
+  currentPortalCount = 0,
+}: {
+  accountType: keyof typeof HUBSPOT_ACCOUNT_TYPES;
```
camden11, line 47: You can use the `AccountType` type from local-dev-lib

lib/prompts/accountNamePrompt.ts
camden11: Saw you're using `Response` in other places, i think that makes sense!

camden11 (commented):
Looking good! Just a few minor comments

lib/prompts/accountNamePrompt.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1273/commits/48cec69713afa4fca296bba1a1400534a838e662

lib/prompts/accountNamePrompt.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1273/commits/48cec69713afa4fca296bba1a1400534a838e662

lib/prompts/accountNamePrompt.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1273/commits/dc551a7bade92e90b7e5f5994c7b96a0085f785a

lib/prompts/accountNamePrompt.ts
```diff
@@ -56,7 +76,7 @@ const hubspotAccountNamePrompt = ({ accountType, currentPortalCount = 0 }) => {
     {
       name: 'name',
       message: promptMessageString,
-      validate(val) {
+      validate(val: string): string | boolean {
```
kemmerle, line 79: With the `PromptValidationFunction` type, it doesn't seem like I should need to type this, but I get an error otherwise. Is that ok?

camden11 (approved):
One tiny thing I noticed but otherwise LGTM

commands/account/remove.ts
```diff
@@ -59,12 +56,12 @@ exports.handler = async options => {
   );
 
   // Get updated version of the config
-  config = getConfig();
+  getConfig();
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1273/commits/54a3962107f0949148ca7d1aacbac2f3fbc582ee

PR #1275, opened 2024-11-26
chore: Improve unreleased changes script

joe-yeager:
## Description and Context

Make the script aware of multiple tags
Port to node


scripts/unreleasedChanges.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
```
joe-yeager, line 1: I was getting a lot of grief trying to run this in a worker using `ts-node` so I reverted back to javascript and made it an executable script

package.json
```diff
@@ -66,6 +66,7 @@
     "circular-deps": "yarn madge --circular .",
     "release": "yarn ts-node ./scripts/release.ts release",
     "hs": "yarn build && node ./dist/bin/hs",
+    "unreleased-changes": "./scripts/unreleasedChanges.js",
```
brandenrodgers, line 69: Is there any way that this could live in the `.github/` or `.github/workflows/` folders? It's not a blocker, but this isn't a script than can be run manually by us so it might be helpful to keep it separate.

joe-yeager: Not 100 percent certain, I'll give it a shot and see if it works out

joe-yeager: It took a lot of trial and error, but I moved the script into the `.github` directory and remove this line from the `package.json` and we are just running it using `node .github/scripts/unreleasedChanges.js`

brandenrodgers (approved):

PR #1277, opened 2024-11-26
fix: Make doctor aware of the next tag

joe-yeager:
## Description and Context

The doctor command will now see if the user is on the most up to date version of `latest` or `next`

### DIFF
--- a/lib/doctor/Doctor.ts
+++ b/lib/doctor/Doctor.ts
@@ -226,8 +226,12 @@ export class Doctor {
 
   private async checkCLIVersion(): Promise<void> {
     let latestCLIVersion;
+    let nextCliVersion;
+
     try {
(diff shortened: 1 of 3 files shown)


lib/doctor/Doctor.ts
```diff
@@ -245,14 +249,15 @@ export class Doctor {
       });
     }
 
-    if (latestCLIVersion !== pkg.version) {
+    if (latestCLIVersion !== pkg.version && nextCliVersion !== pkg.version) {
+      const onNextTag = pkg.version.includes('beta');
```
joe-yeager, line 253: This is the best way I know of to check if they are on the `next` tag, it doesn't always have latest, but that is usually when we publish to `latest` and the update the `next` tag to point to the same thing.

brandenrodgers (approved):

PR #1280, opened 2024-11-26
chore: Clean up buckets and command descriptions

camden11:
## Description and Context
This renames files and directories for the `account` and `secret` commands to be singular, matching the conventions we use everywhere else. It also fixes a few minor copy issues in `en.lyaml` and updates every command `describe` to match our new rules

See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/718

## Who to Notify


bin/cli.js
```diff
@@ -49,7 +49,7 @@ const projectCommands = require('../commands/project');
 const themeCommand = require('../commands/theme');
 const moduleCommand = require('../commands/module');
 const configCommand = require('../commands/config');
-const accountsCommand = require('../commands/accounts');
+const accountCommand = require('../commands/account');
```
brandenrodgers, line 52: nit, but can these be `accountCommands` and `secretCommands`? Helps identify them as command buckets

brandenrodgers (approved):
one nit, but not a blocker

PR #1282, opened 2024-12-03
Add prompt types

camden11:
## Description and Context
This adds some lightweight types to our prompt utils, since we're using a version of `inquirer` that doesn't come with types. These types are located in the `types/prompts` file in the new `types` dir. The prompt types aren't meant to be exhaustive - they just cover what we've converted to TS so far and may need to be modified in the future. These types confirm that `promptUser` is passed a valid `PromptConfig` and that the `name` in `PromptConfig` matches the expected returned fields in the prompt's response.

Note: merging this into @kemmerle's existing prompts branch, not `main` or `next`

## Who to Notify
@kemmerle


commands/upload.ts
```diff
@@ -218,7 +218,13 @@ exports.handler = async options => {
       //  If clean is true, will first delete the dest folder and then upload src. Cleans up files that only exist on HS.
       let cleanUpload = options.force;
       if (!options.force) {
-        cleanUpload = await cleanUploadPrompt(derivedAccountId, dest);
+        cleanUpload = await confirmPrompt(
```
brandenrodgers, line 221: Nice 👍 I'm a fan of cutting down on prompt utils that can be replaced by confirmPrompt()

brandenrodgers (approved):

kemmerle (approved):
Looks beautiful, thanks for the help 🙏

PR #1284, opened 2024-12-03
chore: convert usageTracking to TS

brandenrodgers:
## Description and Context

Converts usageTracking.ts to TS

### DIFF
--- a/lib/usageTracking.ts
+++ b/lib/usageTracking.ts
@@ -6,6 +6,7 @@ import {
 import { API_KEY_AUTH_METHOD } from '@hubspot/local-dev-lib/constants/auth';
 import { logger } from '@hubspot/local-dev-lib/logger';
 import { version } from '../package.json';
+import { debugError } from './errorHandlers';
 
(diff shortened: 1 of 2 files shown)


lib/usageTracking.ts
```diff
@@ -85,12 +87,14 @@ export function trackCommandUsage(command, meta = {}, accountId) {
       );
       logger.debug('Sent usage tracking command event: %o', usageTrackingEvent);
     } catch (e) {
-      logger.debug('Usage tracking failed: %s', e.message);
+      if (e instanceof Error) {
```
brandenrodgers, line 90: Do we have a standardized way to do this via util? Or is this the best way?

brandenrodgers: I was looking to add something like that and I realized we have `debugError` in errorHandlers/index.ts

I think I'll just use that instead here since the error returned by that util _should_ have all the right info in it

camden11 (approved):
Nice, LGTM

PR #1287, opened 2024-12-03
chore: Add a script to generate a list of all the commands and subcommands

joe-yeager:
Add a script to list all of the available (not hidden) cli commands because `yargs` doesn't have support for this


scripts/get-all-commands.ts
```diff
@@ -0,0 +1,53 @@
+import { exec } from 'child_process';
+import SpinniesManager from '../lib/ui/SpinniesManager';
+import util from 'util';
+
+const output = {};
+
+const execAsync = util.promisify(exec);
+
+async function extractCommands(
```
joe-yeager, line 9: This recurses over the commands running `--get-yargs-completions` and generates a JSON object to keep the hierarchy.  We can use this in the future to scrape help documents etc to auto generate docs from the help text in the CLI if we so choose.

brandenrodgers (approved):
Nice, I think we'll be able to do a lot with this!

PR #1290, opened 2024-12-04
Convert generateSelectors and marketplaceValidate to TS

camden11:
## Description and Context
This converts `lib/generateSelectors` and `lib/marketplaceValidate` to typescript. 

~Will need to merge and release this LDL PR before type checks will pass: https://github.com/HubSpot/hubspot-local-dev-lib/pull/218~ Done

## Who to Notify


lib/generateSelectors.ts
```diff
@@ -13,11 +13,11 @@ const i18nKey = 'commands.theme.subcommands.generateSelectors';
 
 let maxFieldsDepth = 0;
 
-function getMaxFieldsDepth() {
+export function getMaxFieldsDepth() {
```
joe-yeager, line 16: nit: Should we add a return type?

lib/marketplaceValidate.ts
```diff
@@ -75,72 +91,73 @@ const processValidationErrors = (i18nKey, validationResults) => {
...
+}
 
+export function displayValidationResults(
+  i18nKey: string,
+  validationResults: GetValidationResultsResponse
+) {
   Object.keys(validationResults.results).forEach(type => {
     logger.log(chalk.bold(i18n(`${i18nKey}.results.${type.toLowerCase()}`)));
-    displayResults(validationResults.results[type]);
+    displayResults(validationResults.results[type as ValidationType], i18nKey);
```
joe-yeager, line 160: Iterating objects in TS is super frustrating, I always think that there has to be a way to do it without asserting, and I'm always disappointed

package.json
```diff
@@ -7,7 +7,7 @@
   "dependencies": {
     "@hubspot/local-dev-lib": "3.0.0",
     "@hubspot/serverless-dev-runtime": "7.0.0",
-    "@hubspot/theme-preview-dev-server": "0.0.10",
+    "@hubspot/theme-preview-dev-server": "0.0.9",
```
joe-yeager, line 10: Might need to pull next, this seems unintentional

lib/marketplaceValidate.ts
camden11: Yeah `Object.keys` not being able to figure out the key types is super frustrating

package.json
```diff
@@ -52,7 +52,7 @@
     "typescript": "^5.6.2"
   },
   "optionalDependencies": {
-    "@hubspot/cms-dev-server": "^0.18.10"
+    "@hubspot/cms-dev-server": "^0.18.9"
```
joe-yeager, line 55: This is still downgrading the dep for some reason.  Since it has that `^` it is probably fine, but still seems unintentional.

joe-yeager (approved):
Approving, but the `@hubspot/cms-dev-server` version is still being downgraded for some reason.  The rest of it looks good

PR #1291, opened 2024-12-04
refactor!: Remove unused hidden commands

joe-yeager:
## Description and Context

Remove unused hidden CMS commands:
- `hs module marketplace-validate`
- `hs lint`
- `hs theme lighthouse-score`


TanyaScales (approved):

joe-yeager:
> LGTM. We should also probably think about removing related code from LDL. There is stuff in there for lighthouse score for example that now isn't used anywhere

@camden11 Good callout!  The reason I skipped removing those, is would mean another major release for LDL so I was hesitant.  I'll write up some issues though so we don't lose track of removing them in the major.

TheWebTech:
Hey uh I just noticed these. They're in our documentation, and probably unused because the help text didn't show them?

PR #1296, opened 2024-12-10
Fix: two bugs in `hs project dev` 

kemmerle:
## Description and Context
In the process of tracking down one bug in `hs project dev`, I discovered three:

1) Using the `--account` flag in any command that used the `addUseEnvironmentOptions` middleware resulted in the error: `Arguments use-env and account are mutually exclusive`. This was because we were setting both a `conflicts` and a `default` for `use-env`. 

2) We were not correctly handling the scenario where a customer was passing the `--account` flag. We no longer need to perform a separate check for the account flag value, because we already do so in the middleware. This allows us to eliminate and simplify code. 

## TODO

- [ ] Address any feedback 

## Who to Notify

### DIFF
--- a/lib/localDev.ts
+++ b/lib/localDev.ts
@@ -121,29 +121,6 @@ const checkIfParentAccountIsAuthed = accountConfig => {
   }
 };
 
-// Confirm the default account is a developer account if developing public apps
-const checkIfAccountFlagIsSupported = (accountConfig, hasPublicApps) => {
-  if (hasPublicApps) {
-    if (!isDeveloperTestAccount(accountConfig)) {
-      logger.error(
-        i18n(`${i18nKey}.validateAccountOption.invalidPublicAppAccount`, {
-          useCommand: uiCommandReference('hs accounts use'),
-          devCommand: uiCommandReference('hs project dev'),
-        })
-      );
-      process.exit(EXIT_CODES.SUCCESS);
-    }
-    checkIfParentAccountIsAuthed(accountConfig);
-  } else if (isAppDeveloperAccount(accountConfig)) {
-    logger.error(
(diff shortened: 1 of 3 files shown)


lib/localDev.ts
```diff
@@ -121,29 +121,6 @@ const checkIfParentAccountIsAuthed = accountConfig => {
   }
 };
 
-// Confirm the default account is a developer account if developing public apps
-const checkIfAccountFlagIsSupported = (accountConfig, hasPublicApps) => {
-  if (hasPublicApps) {
-    if (!isDeveloperTestAccount) {
-      logger.error(
-        i18n(`${i18nKey}.validateAccountOption.invalidPublicAppAccount`, {
```
brandenrodgers, line 129: Do we need to clean these up from the lang file?

brandenrodgers (approved):
I left a comment about cleaning up orphaned lang file entries, but otherwise lgtm! I tested all the permutations that I could think of and this seems to WAD 👍

camden11 (commented):
I haven't actually had the chance to test yet, but I think this changes the behavior of `hs project dev` in a way that we probably don't want. The separate `checkIfAccountFlagIsSupported` function exists because `hs project dev` behaves slightly differently when the account flag is passed in (for example, if fails if don't pass in a developer test account for public app local dev, rather than prompting you to select one)

kemmerle:
I'm going to close this issue. I actually discovered three bugs, one of which was caused by my account middleware changes. The other two bugs (including the `use-env` issue) needed to be fixed and merged on the `main` branch for a patch fix (https://github.com/HubSpot/hubspot-cli/pull/1298). 

I also misunderstood the intended behavior of the `dev` command, so I need to redo those changes. I think it makes more sense to simply close out this PR and start fresh.

PR #1300, opened 2024-12-11
Upgrade prettier and update style

camden11:
## Description and Context
This upgrades prettier to version 3 so it will recognize some of the new TS keywords, including `satisfies`. Updates the style of a few files to match prettier's new defaults

## Who to Notify


kemmerle (approved):
🚢

PR #1302, opened 2024-12-12
Chore: project structure ts port

brandenrodgers:
## Description and Context

Porting projectStructure to TS. I had to add types for project components which doesn't feel great, but I'm not sure that there's another option. Eventually we should offload this type of thing into the new package that will specialize in interacting with project schema, but in the meantime I think we need to keep this here.

I could also add these types to Local Dev Lib, but I think I'd just be adding more work for us down the road when we eventually start to use the translation lib.

### DIFF
--- a/lib/projectStructure.ts
+++ b/lib/projectStructure.ts
@@ -86,8 +86,8 @@ export const CONFIG_FILES: {
   [COMPONENT_TYPES.hublTheme]: 'theme.json',
 };
 
-function getTypeFromConfigFile(
-  configFile: ValueOf<typeof CONFIG_FILES>
+function getComponentTypeFromConfigFile(
+  configFile: string
 ): ComponentTypes | null {
   let key: ComponentTypes;
   for (key in CONFIG_FILES) {
@@ -190,11 +190,11 @@ export async function findProjectComponents(
       if (parsedConfig) {
         const isLegacy = getIsLegacyApp(parsedConfig, dir);
         const isHublTheme = base === CONFIG_FILES[COMPONENT_TYPES.hublTheme];
-        const type = getTypeFromConfigFile(base);
+        const componentType = getComponentTypeFromConfigFile(base);
 
-        if (type) {
+        if (componentType) {
           components.push({
-            type,
+            type: componentType,
             config: parsedConfig,
             runnable: !isLegacy && !isHublTheme,
             path: dir,
--- a/lib/projectStructure.ts
+++ b/lib/projectStructure.ts
@@ -1,24 +1,91 @@
-// @ts-nocheck
-const fs = require('fs');
-const path = require('path');
-const { walk } = require('@hubspot/local-dev-lib/fs');
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const { logError } = require('./errorHandlers/index');
+import * as fs from 'fs';
+import * as path from 'path';
+import { walk } from '@hubspot/local-dev-lib/fs';
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { logError } from './errorHandlers/index';
 
-const COMPONENT_TYPES = Object.freeze({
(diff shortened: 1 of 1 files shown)


lib/projectStructure.ts
```diff
@@ -63,7 +133,10 @@ function getAppCardConfigs(appConfig, appPath) {
   return cardConfigs;
 }
 
-function getIsLegacyApp(appConfig, appPath) {
+function getIsLegacyApp(
+  appConfig: PublicAppComponentConfigType | PrivateAppComponentConfigType,
+  appPath: string
+) {
```
kemmerle, line 139: Needs a return type

lib/projectStructure.ts
```diff
@@ -39,16 +106,19 @@ function loadConfigFile(configPath) {
   return null;
 }
 
-function getAppCardConfigs(appConfig, appPath) {
-  const cardConfigs = [];
+export function getAppCardConfigs(
+  appConfig: PublicAppComponentConfigType | PrivateAppComponentConfigType,
+  appPath: string
+) {
```
kemmerle, line 112: Needs a return type

lib/projectStructure.ts
```diff
@@ -1,36 +1,103 @@
...
+  for (key in CONFIG_FILES) {
     if (CONFIG_FILES[key] === configFile) {
       return key;
     }
   }
   return null;
 }
 
-function loadConfigFile(configPath) {
+function loadConfigFile(configPath: string) {
```
kemmerle, line 96: Needs a return type

kemmerle (commented):
Code looks good to me; there's no changes in behavior, etc. Just need to add some return types

lib/projectStructure.ts
```diff
@@ -1,36 +1,103 @@
...
-
-const COMPONENT_TYPES = Object.freeze({
+import * as fs from 'fs';
+import * as path from 'path';
+import { walk } from '@hubspot/local-dev-lib/fs';
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { logError } from './errorHandlers/index';
+
+type ComponentTypes = 'private-app' | 'public-app' | 'hubl-theme';
+type ValueOf<T> = T[keyof T];
```
camden11, line 8: We export `ValueOf` from `@hubspot/local-dev-lib/types/Utils` so you don't need to redefine it here

lib/projectStructure.ts
```diff
@@ -1,36 +1,103 @@
...
+} = {
   [COMPONENT_TYPES.privateApp]: 'app.json',
   [COMPONENT_TYPES.publicApp]: 'public-app.json',
   [COMPONENT_TYPES.hublTheme]: 'theme.json',
 };
 
-function getTypeFromConfigFile(configFile) {
-  for (const key in CONFIG_FILES) {
+function getTypeFromConfigFile(
+  configFile: ValueOf<typeof CONFIG_FILES>
```
camden11, line 85: It looks like this just resolves to `string`. Which... I think is probably fine but we can just type at that way directly. I messed around with adding an `as const` to  `CONFIG_FILES` to fix this and it caused its own set of problems

lib/projectStructure.ts
```diff
@@ -1,36 +1,103 @@
...
-const { logError } = require('./errorHandlers/index');
-
-const COMPONENT_TYPES = Object.freeze({
+import * as fs from 'fs';
+import * as path from 'path';
+import { walk } from '@hubspot/local-dev-lib/fs';
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { logError } from './errorHandlers/index';
+
+type ComponentTypes = 'private-app' | 'public-app' | 'hubl-theme';
```
camden11, line 7: Since you already declare the `COMPONENT_TYPES` object below, you could derive this type from that

`type ComponentTypes = ValueOf<typeof COMPONENT_TYPES>;`

camden11 (approved):

PR #1303, opened 2024-12-12
Fix bug with account middleware in `hs project dev`

kemmerle:
## Description and Context
This fixes some bugs I discovered with the new injected account middleware in `hs project dev.` We were not correctly checking and handling account flags and had changed the command's behavior accidentally, so that when a customer runs `hs project dev` with an app developer account set as default, they did not receive the menu to choose a developer test account. @camden11, you're the person who best understands the intricacies here.  

## TODO

- [ ] Address feedback

## Who to Notify

### DIFF
--- a/commands/project/dev.ts
+++ b/commands/project/dev.ts
@@ -13,7 +13,6 @@ const { logger } = require('@hubspot/local-dev-lib/logger');
 const {
   getConfigAccounts,
   getAccountConfig,
-  getAccountId: getAccountIdFromConfig,
   getEnv,
 } = require('@hubspot/local-dev-lib/config');
 const {
@@ -59,8 +58,6 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
 exports.handler = async options => {
   await loadAndValidateOptions(options);
   const { derivedAccountId, providedAccountId } = options;
-  const parsedProvidedAccountId = getAccountIdFromConfig(providedAccountId);
-  const providedAccountConfig = getAccountConfig(parsedProvidedAccountId);
   const accountConfig = getAccountConfig(derivedAccountId);
   const env = getValidEnv(getEnv(derivedAccountId));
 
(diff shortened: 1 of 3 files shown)


commands/project/dev.ts
```diff
@@ -107,16 +110,20 @@ exports.handler = async options => {
...
   // The account that we are locally testing against
-  let targetTestingAccountId = derivedAccountId;
+  let targetTestingAccountId = providedAccountId
+    ? parsedProvidedAccountId
+    : null;
 
   // Check that the default account or flag option is valid for the type of app in this project
-  if (derivedAccountId) {
-    checkIfAccountFlagIsSupported(accountConfig, hasPublicApps);
+  if (providedAccountId) {
```
brandenrodgers, line 122: Won't `providedAccountId` always be the same as `derivedAccountId` when the user passes in `--account`? Derived will first look for the account flag, and then fall back to using the default account if one isn't specified.

kemmerle: No, it won't. `derivedAccountId` will always be defined in commands, since we inject the default value into it in middleware. `providedAccountId` is only defined when a user passes in an account flag, and that's important here. We want different behavior depending on if the user passes in a flag.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1303/commits/1b2ade7d8896ec96bd1e86447be2d8c4a7ae269e

brandenrodgers: Ohhh right, because we bypass the prompts when the user passes in the flag. I forgot about that 👌

commands/project/dev.ts
```diff
@@ -107,12 +107,12 @@ exports.handler = async options => {
     (!hasPublicApps && isSandbox(accountConfig));
 
   // The account that the project must exist in
-  let targetProjectAccountId = derivedAccountId;
+  let targetProjectAccountId = providedAccountId ? derivedAccountId : null;
```
camden11, line 110: Theoretically, if `providedAccountId` exists, it's the same thing as `derivedAccountId` right? Could this just be
`let targetProjectAccountId = providedAccountId`?

camden11 (approved):
Tested the flow this fixes and it seems to work for me! I would just double check the other flows and to be extra sure this doesnt break anything. Also had one minor comment but doesn't need to be blocking

commands/project/dev.ts
brandenrodgers, line 110: Could you update the comment to explain why we need this to be `null` when the user doesn't pass in `--account`?

kemmerle: `providedAccountId` is not the same as `derivedAccountId`. `providedAccountId` is just the raw value that the user supplies in the flag, which can include the name in the config. The middleware takes the raw value from `providedAccountId` and spits out `derivedAccountId`, which will always be the account ID (type number). 

Going to resolve this but Slack me with any more questions.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1303/commits/0ade7877391e7585c8dc0364e36257ccd348c1d1

PR #1304, opened 2024-12-12
Fix bug in `hs fetch`

kemmerle:
## Description and Context
I found a spare `accountId` variable in `hs fetch` which was causing the command to fail. I replaced it with `derivedAccountId`. 

I performed a manual check of every command file where we refer to `accountId` to ensure that there were no more bugs, and I found none. We should be really careful not to introduce bugs before merging `next` into `main` next week. 

## TODO

- [ ] Address feedback 

## Who to Notify


brandenrodgers (approved):
good find 👍

PR #1307, opened 2024-12-17
Refactor `loadAndValidateOptions` p3: Add config validation 

kemmerle:
## Description and Context
Relies on: https://github.com/HubSpot/hubspot-local-dev-lib/pull/222

In this PR, I've added validation when loading the configuration file. This involves creating a `NO_VALIDATE_LIST` for certain commands, like `hs init`, that will fail upon validation. If there's a more elegant solution to the problem, I'm all ears! 

## TODO

- [ ] Address feedback 

## Who to Notify


bin/cli.js
```diff
@@ -169,11 +183,13 @@ const loadConfigMiddleware = async options => {
   if (options.config && fs.existsSync(options.config)) {
     const { config: configPath } = options;
     await loadConfig(configPath, options);
```
brandenrodgers, line 185: This load config shouldn't need `await`

bin/cli.js
```diff
@@ -149,6 +151,17 @@ const setRequestHeaders = () => {
 };
 
 const loadConfigMiddleware = async options => {
+  const validateConfig = () => {
+    const shouldValidate = options._.every(
+      item => !NO_VALIDATE_LIST.includes(item)
```
brandenrodgers, line 156: `NO_VALIDATE_LIST ` is only used in this file so I don't think it needs to live in constants.ts. That file is kind of a messy grab-bag of stuff. 

Looking at this list:
```
  'init',
  'clean',
  'list',
  'remove',
  'use',
  'delete',
```

Is that the list of commands that currently aren't running config validation? I wonder if we could remove every command from that list except for `init`. All of the other commands rely on a valid config existing in order to function properly.

bin/cli.js
```diff
@@ -149,6 +151,17 @@ const setRequestHeaders = () => {
 };
 
 const loadConfigMiddleware = async options => {
+  const validateConfig = () => {
```
brandenrodgers, line 154: If you name this something like `maybeValidateConfig` then you wouldn't need to rename the import for `validateConfig`. It gets a little confusing when we use the exact same name for functions.

bin/cli.js
```diff
@@ -149,6 +151,17 @@ const setRequestHeaders = () => {
...
+    const shouldValidate = options._.every(
```
brandenrodgers, line 155: I like your idea here, but I think this might trigger unintentionally.

For example, what if we had a `hs project delete` command. We'd want to run validation on it, but this would be checking for:
```js
['project', 'delete'].every(item => !['init'...'delete'].includes(item))
```
That would return false because of `delete`. Same thing for `hs project list` and `hs accounts use`.

I think we'll need to be checking for the full command structure to make sure we're not unintentionally skipping config validation for certain commands. A map might be useful here. Maybe something like this?

```js
const NO_CONFIG_VALIDATION = {
  init: { skip: true },
  account: {
    skip: false
    use: { skip: true }
  }
}
```

^ That might not work well once fleshed out, but maybe it helps w/ brainstorming.

Another question I have about this is how it would handle command aliases. Like, if we didn't want to validate config for `hs account use`, but the user runs it like `hs accounts use`.

bin/cli.js
kemmerle: Agreed! I think the only other command to add to the list is `auth`, since we want to allow folks to add an authorized account to an empty config.

bin/cli.js
kemmerle: Addressed in [24c78e2](https://github.com/HubSpot/hubspot-cli/pull/1307/commits/24c78e289fc7bf4e8832826b131a7ce821d8c462)

bin/cli.js
kemmerle: Addressed in [24c78e2](https://github.com/HubSpot/hubspot-cli/pull/1307/commits/24c78e289fc7bf4e8832826b131a7ce821d8c462)

bin/cli.js
kemmerle: Since we've eliminated all commands from the `NO_CONFIG_VALIDATION` apart from `init` and `auth`, I've gone with this structure:

```
const NO_CONFIG_VALIDATION = {
  init: { skip: true },
  auth: { skip: true },
};
``` 

We can revisit this later, if we need to add more commands. I'm going to resolve for now, but feel free to reopen!

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1307/commits/24c78e289fc7bf4e8832826b131a7ce821d8c462

bin/cli.js
```diff
@@ -148,7 +149,24 @@ const setRequestHeaders = () => {
...
 };
 
+const NO_CONFIG_VALIDATION = {
+  init: { skip: true },
+  auth: { skip: true },
+};
+
 const loadConfigMiddleware = async options => {
+  const maybeValidateConfig = () => {
+    const shouldValidate = options._.every(
```
brandenrodgers, line 159: Nice, I like this better. Just one more thought on it. I don't think `options._.every()` is doing anything for us here anymore. If we ever want to skip validation for something nested like `hs account auth` we'll probably have to switch to a different iteration pattern (maybe something recursive?). 

Also as it is now, we run the risk of accidentally opting certain commands out of validation. For example, if we added `list: {skip: true}` to the list with the intent of skipping validation for `hs list`. Using `options._.every()` here would also opt commands like `hs project list` out of validation.

Is `options._` guaranteed to exist? You could avoid accidentally triggering for other commands by simplifying to something like `if (NO_CONFIG_VALIDATION[options._[0]])`

lib/validation.ts
```diff
@@ -34,7 +33,7 @@ async function loadAndValidateOptions(options, shouldValidateAccount = true) {
     validAccount = await validateAccount(options);
   }
 
-  if (!(validateConfig() && validAccount)) {
+  if (!validAccount) {
```
brandenrodgers, line 36: Was this previously running account validation before config validation? Seems like this PR will also switch the flow to validate config before account, which feels like a better order IMO

bin/cli.js
```diff
@@ -161,19 +179,22 @@ const loadConfigMiddleware = async options => {
...
     return;
   }
 
   // We need to load the config when options.config exists,
   // so that getAccountIdFromConfig() in injectAccountIdMiddleware reads from the right config
   if (options.config && fs.existsSync(options.config)) {
     const { config: configPath } = options;
-    await loadConfig(configPath, options);
+    loadConfig(configPath, options);
+    maybeValidateConfig();
```
brandenrodgers, line 191: You could DRY this up a little and avoid calling `maybeValidateConfig` many times by switching to if/else blocks and removing the return calls.

bin/cli.js
brandenrodgers: Ignore my comment here. I just noticed your work on this in https://github.com/HubSpot/hubspot-cli/pull/1310/files#diff-ca900686fca4680d4692bf587dc5da1322879c344688c187b5511cda27902c18R153

bin/cli.js
```diff
@@ -148,7 +149,24 @@ const setRequestHeaders = () => {
...
+};
+
 const loadConfigMiddleware = async options => {
+  const maybeValidateConfig = () => {
+    const shouldValidate = options._.every(
+      command =>
+        !(NO_CONFIG_VALIDATION[command] && NO_CONFIG_VALIDATION[command].skip)
+    );
+    if (shouldValidate) {
+      if (!validateConfig()) {
```
joe-yeager, line 164: Nit: these can merged to `if (shouldValidate && !validateConfg())`

bin/cli.js
```diff
@@ -148,7 +149,24 @@ const setRequestHeaders = () => {
   addUserAgentHeader('HubSpot CLI', pkg.version);
 };
 
+const NO_CONFIG_VALIDATION = {
+  init: { skip: true },
+  auth: { skip: true },
+};
+
```
joe-yeager, line 156: Nit: we could compress this to:
```typescript
const NO_CONFIG_VALIDATION = {
  init: true,
  auth: true,
};
```

And simplify the check below from: 
`command => !(NO_CONFIG_VALIDATION[command] && NO_CONFIG_VALIDATION[command].skip)` 

to 

`command => !NO_CONFIG_VALIDATION[command]`

joe-yeager (commented):
LGTM, just a few nits

bin/cli.js
kemmerle: I'm changing this up in the next PR (https://github.com/HubSpot/hubspot-cli/pull/1310), so going to leave for now.

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1307/commits/c3fd8cc35cb4f896afc9f9af520107a6cd95732f

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1307/commits/c3fd8cc35cb4f896afc9f9af520107a6cd95732f

bin/cli.js
```diff
@@ -148,33 +147,40 @@ const setRequestHeaders = () => {
...
-  if (options.config && fs.existsSync(options.config)) {
+  // Load the new config and check for the conflicting config flag
+  if (configFileExists(true) && options.config) {
+    logger.error(
+      i18n(`${i18nKey}.loadConfigMiddleware.configFileExists`, {
+        configPath: getConfigPath(),
+      })
+    );
+    process.exit(EXIT_CODES.ERROR);
+  } else {
```
brandenrodgers, line 174: This could be cleaned up to `} else if (options._.includes('init')) {`. I also think you can get rid of the comments here since they're less relevant after the latest iteration.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1307/commits/24a38b52cc51d9d07959ff79fbb02e5078eb174b

PR #1309, opened 2024-12-17
port validation and configOptions to TS

brandenrodgers:
## Description and Context

This is dependent on https://github.com/HubSpot/hubspot-local-dev-lib/pull/223

This ports lib/validation.ts and lib/configOptions.ts to Typescript. I found some bugs while doing this that I'll call out inline.


lib/configOptions.ts
```diff
@@ -51,18 +56,21 @@ const setAllowUsageTracking = async ({ accountId, allowUsageTracking }) => {
...
 
-const selectCmsPublishMode = async () => {
-  const { cmsPublishMode } = await promptUser([
+async function selectCmsPublishMode(): Promise<CmsPublishMode> {
+  const { cmsPublishMode } = await promptUser<{
+    cmsPublishMode: CmsPublishMode;
+  }>([
     {
       type: 'list',
-      look: false,
```
brandenrodgers, line 65: `look` does not exist as a supported field ([search in repo](https://github.com/search?q=repo%3ASBoudrias%2FInquirer.js%20look&type=code))

lib/configOptions.ts
```diff
@@ -91,25 +102,24 @@ const setDefaultCmsPublishMode = async ({
     newDefault = defaultCmsPublishMode;
   } else {
     logger.error(
-      i18n(`${i18nKey}.defaultMode.errors`, {
-        mode: newDefault,
+      i18n(`${i18nKey}.defaultMode.error`, {
```
brandenrodgers, line 105: We were using the wrong lang key, and `newDefault` is undefined here

lib/configOptions.ts
```diff
@@ -134,13 +150,7 @@ const setHttpTimeout = async ({ accountId, httpTimeout }) => {
 
   updateHttpTimeout(newHttpTimeout);
 
-  return logger.success(
+  logger.success(
```
brandenrodgers, line 153: There's no need for any of these to return

lib/validation.ts
```diff
@@ -200,9 +199,9 @@ const fileExists = _path => {
   }
 
   return true;
-};
+}
 
-const checkAndConvertToJson = _path => {
+export function checkAndConvertToJson(_path: string): object | boolean | null {
```
brandenrodgers, line 204: This return type is pretty gross, but I didn't want to change the functionality of anything in this pass.

lib/configOptions.ts
```diff
@@ -91,25 +102,24 @@ const setDefaultCmsPublishMode = async ({
...
   } else {
     logger.error(
-      i18n(`${i18nKey}.defaultMode.errors`, {
-        mode: newDefault,
+      i18n(`${i18nKey}.defaultMode.error`, {
         validModes: commaSeparatedValues(ALL_CMS_PUBLISH_MODES),
       })
     );
-    newDefault = await selectCMsPublishMode();
+    newDefault = await selectCmsPublishMode();
```
brandenrodgers, line 109: We had a casing issue here that TS picked up 🎉

lib/validation.ts
camden11: It looks like the way this is used, you could just have it return `null` instead of `false` and at least clean up the return type a little bit

camden11 (approved):
One minor comment but overall LGTM

lib/validation.ts
brandenrodgers: Good call, fixed in [d625070](https://github.com/HubSpot/hubspot-cli/pull/1309/commits/d625070d6db3e8245d0ef57fa0ffcc74d6a52336)

PR #1310, opened 2024-12-18
Refactor `loadAndValidateOptions`: Add account middleware (Final part)

kemmerle:
## Description and Context
Depends on https://github.com/HubSpot/hubspot-cli/pull/1307 

In this PR, I've added middleware to validate accounts, and I've removed every instance of `loadAndValidateOptions` from the commands. 

One quick question: Should I delete the `loadAndValidateOptions` function entirely, or would that constitute a breaking change? It is exported. 

**NOTE:** The majority of these changes are simply removing `loadAndValidateOptions`. You should only need to give a good review to `bin/cli.js`

## TODO

- [ ] Address feedback 

## Who to Notify


bin/cli.js
```diff
@@ -149,18 +150,32 @@ const setRequestHeaders = () => {
   addUserAgentHeader('HubSpot CLI', pkg.version);
 };
 
+const shouldValidate = (options, validationList) => {
```
kemmerle, line 153: Should I love this function to the `lib/validation.ts` file?

brandenrodgers:
> Should I delete the loadAndValidateOptions function entirely, or would that constitute a breaking change? It is exported.

That would be a breaking change but all of these changes are going out with CLI v7.0.0 which is a major version bump anyways, so 🔪 away!

bin/cli.js
brandenrodgers: You could if you want to. Or you could make the utils slightly more generic and keep them here. Really what this is doing is telling us whether the user-entered command matches a command that we've opted into a list. This could potentially have value in middleware outside of validation.

Like, what if this was something like `isTargetedCommand(options, commandMap)`, and it takes in a similar object for commandMap, but just more generic:

```json
const SKIP_ACCOUNT_VALIDATION = {
  accounts: {
    target: false,
    subCommands: {
      clean: { target: true },
      list: { target: true },
      remove: { target: true },
    },
  },
  sandbox: {
    target: false,
    subCommands: {
      delete: { target: true },
    },
  },
};
```

One other aspect of this is that we need to support command aliases. i.e. `hs sandboxes delete` should also skip validation. I'm not sure if there's a graceful way to do it. We might just need to duplicate the entries in the object.

bin/cli.js
```diff
@@ -149,18 +150,32 @@ const setRequestHeaders = () => {
...
+  return options._.every(command => {
+    const mainCommand = command;
+    if (validationList[mainCommand]) {
+      if (validationList[mainCommand].skip) {
+        return false;
+      }
+      const subCommands = validationList[mainCommand].subCommands || {};
```
brandenrodgers, line 160: We'll need to account for arbitrarily-nested subcommands, which I think means recursion is the best option here

brandenrodgers:
Love to see a PR with more 🔴 than 🟢 ! 🙌

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1310/commits/dccf72fd6bd0bd7ea4b5a5b11374bc22179b0415. Making the `shouldTargetCommand` util recursive in another commit.

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1310/commits/312431ebc20d2fc2129c0f5d655fde96b4aff3e1

bin/cli.js
```diff
@@ -147,18 +148,38 @@ const setRequestHeaders = () => {
...
+      for (const subCommand in subCommands) {
+        if (!checkCommand(subCommand, subCommands)) {
+          return false;
+        }
+      }
+    }
+    return true;
+  };
+
+  return options._.every(command => checkCommand(command, commandMap));
```
brandenrodgers, line 169: We might be able to simplify this a little bit. From what I can tell this seems to work in most cases, but it gets there in a slightly roundabout way.

Kicking this off using `every()` means that you'll call `checkCommand()` for every command that gets passed in. So running `hs accounts remove` would result in us calling `checkCommand()` for `accounts` and then also for `remove`, which I'm pretty sure isn't necessary.

The logic flow currently looks something like this:
1. `checkCommand('accounts', SKIP_ACCOUNT_VALIDATION)`
2. The first if check skips because `SKIP_ACCOUNT_VALIDATION['accounts'].target` is false
3. We get subcommands and then iterate over them and recursively call checkCommand for each one like this: `checkCommand('clean', SKIP_ACCOUNT_VALIDATION['accounts'].subCommands)`
4. The first if check for "clean" would skip  because `SKIP_ACCOUNT_VALIDATION['clean'].target` is true but options._ doesn't include"clean". Eventually this returns true
5. We do this for "list", and "ls", and then we eventually get to "remove" where we return false
6.  This exits the for-in loop and returns false
7. This aborts the first `options._.every()` call because the first check returned `false`
8.  `shouldTargetCommand ` returns false

**A few takeaways:**
- The outcome would be equivalent here if you were to kick this off using `return checkCommand()` instead of using `every()`
- We'd be recursively calling `checkCommand()` for "clean", then "list", then "ls", and THEN "remove". In reality, all we care about is "remove" because that's the command that the user ran. If we had a third layer of nesting (like hs accounts list account) then we'd recur again before getting to "remove"

It looks like we're deciding to recursively call `checkCommand` based on the presence of `map[mainCommand].subCommands`, but this gets much simpler if we make the decision based off of the length of `options._`. If we make our decisions based off of the command that the user ran, it prevents us from doing extra cycles and makes the logic easier to follow. For example:
1. call `checkCommand(options, commandMap)`
2. grab `const currentCommand = options._[0]`
3. Check if `currentCommand` should be targeted or has subcommands. If there's no entry for it, return false. If it's targeted, then return true. If it isn't targeted but has subcommands, call `return checkCommand(options._.slice(1), commandMap[currentCommand])`

The benefit of doing it this way is that there aren't any unnecessary iterations because at most we call `checkCommand` once per command that the user entered.

The naming in here is also a confusing to me because the method is called `shouldTargetCommand`, and the value for `accounts.clean` is `{ target: true }`, but then we return false when the user runs `hs accounts clean`. If target is true, I'd personally expect `shouldTargetCommand` to return true for that command.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1310/commits/7fd67f3b76f8a0d7570611193f6e301d600ce7f8

bin/cli.js
```diff
@@ -147,18 +148,40 @@ const setRequestHeaders = () => {
...
+    if (commandMap[currentCommand].target) {
+      return true;
+    }
+
+    const subCommands = commandMap[currentCommand].subCommands || {};
+    if (options._.length > 1) {
+      return checkCommand({ _: options._.slice(1) }, subCommands);
+    }
+
+    return true;
```
brandenrodgers, line 168: Should this default to false instead of true?

lib/validation.ts
```diff
@@ -24,20 +24,8 @@ const {
   getExt,
 } = require('@hubspot/local-dev-lib/path');
 const { getAccountId, getCmsPublishMode } = require('./commonOpts');
-const { EXIT_CODES } = require('./enums/exitCodes');
 const { logError } = require('./errorHandlers/index');
 
-async function loadAndValidateOptions(options, shouldValidateAccount = true) {
```
brandenrodgers, line 30: 🎉 🔪

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1310/commits/9b093bb2f1bf781984fe332c411aa08a2bb6ed4b

brandenrodgers (approved):
Tested locally and LGTM! Good riddance to `loadAndValidateOptions` 🙌

PR #1312, opened 2024-12-20
Convert ProjectLogsManager to TS

camden11:
## Description and Context
Not much to report here. This converts the ProjectLogsManager to TS

## Who to Notify
@brandenrodgers @kecarroll @joe-yeager

### DIFF
--- a/lib/projects/ProjectLogsManager.ts
+++ b/lib/projects/ProjectLogsManager.ts
@@ -7,7 +7,7 @@ import { uiLink } from '../ui';
 
 const i18nKey = 'commands.project.subcommands.logs';
 
-class ProjectLogsManager {
+class _ProjectLogsManager {
   projectName: string | undefined;
   projectId: number | undefined;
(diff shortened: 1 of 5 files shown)


lib/projects/ProjectLogsManager.ts
```diff
@@ -133,12 +146,11 @@ class ProjectLogsManager {
 
     if (this.selectedFunction.deployOutput.endpoint) {
       this.endpointName = this.selectedFunction.deployOutput.endpoint.path;
-      this.method = this.selectedFunction.deployOutput.endpoint.method;
```
camden11, line 136: This was both trying to access a field that doesn't exist (it should be `methods`) _and_ not actually used anywhere so I just removed it

kemmerle: 😭 Well, it's good we caught it now

kemmerle (approved):
:shipit:

PR #1319, opened 2025-01-07
test: Increase the polling timeout

joe-yeager:
## Description and Context

Increase the polling timeout to 60 seconds to increase the test stability.  @brandenrodgers and I both observed that sometimes the secrets take up to 30 seconds to exist in the portal, so hopefully this will solve for the majority of the situations where latency is the cause of failure.


brandenrodgers (approved):
Hopefully this helps 🤞

PR #1321, opened 2025-01-07
Improve prompting behavior for `hs custom-object create`

camden11:
## Description and Context
We recently updated `hs custom-object create` to prompt users for a custom object's name and JSON definition path when not provided. When testing, I found the way we're currently prompting to be a bit confusing. The command currently asks for the path of the JSON definitions before asking what kind of object you want to create, which seems non intuitive. The prompt for the name of the custom object type is also misleading, asking what you'd like to name your schema when you're not actually creating a schema. I've fixed both of these things.

## Screenshots
Before:
Screenshot 2025-01-07 at 5 05 03 PM

After:
Screenshot 2025-01-07 at 5 03 57 PM

## Who to Notify

### DIFF
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -214,8 +214,8 @@ en:
               describe: "Schema name to add the object instance to"
           success:
             objectsCreated: "Objects created"
-          inputSchema: "What would you like to name the schema?"
-          inputPath: "[--path] Where is the JSON file containing the object definitions?"
+          inputName: "Enter the name of the schema for the custom object(s) you'd like to create:"
+          inputPath: "[--path] Enter the path to the JSON file containing the object definitions:"
         schema:
           describe: "Commands for managing custom object schemas."
           subcommands:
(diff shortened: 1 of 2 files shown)


lang/en.lyaml
```diff
@@ -214,8 +214,8 @@ en:
               describe: "Schema name to add the object instance to"
           success:
             objectsCreated: "Objects created"
-          inputSchema: "What would you like to name the schema?"
-          inputPath: "[--path] Where is the JSON file containing the object definitions?"
+          inputName: "Enter the name of the schema for the custom object(s) you'd like to create:"
```
joe-yeager, line 217: Should we prefix this with `[--name]`?

joe-yeager (dismissed):
LGTM!  Just one question on adding the flag to the prompt text

joe-yeager (approved):

brandenrodgers (approved):
This is much better 👌 good find

PR #1328, opened 2025-01-10
Chore: convert lib/sandboxes and sandboxSync to TS

camden11:
## Description and Context
This converts lib/sandboxes and lib/sandboxSync to TS

## Who to Notify

### DIFF
--- a/commands/sandbox/create.ts
+++ b/commands/sandbox/create.ts
@@ -141,12 +141,7 @@ exports.handler = async options => {
     const sandboxAccountConfig = getAccountConfig(result.sandbox.sandboxHubId);
     // For v1 sandboxes, keep sync here. Once we migrate to v2, this will be handled by BE automatically
     const handleSyncSandbox = async syncTasks => {
-      await syncSandbox({
-        accountConfig: sandboxAccountConfig,
-        parentAccountConfig: accountConfig,
-        env,
-        syncTasks,
-      });
+      await syncSandbox(sandboxAccountConfig, accountConfig, env, syncTasks);
     };
     try {
       let availableSyncTasks = await getAvailableSyncTypes(
--- a/commands/sandbox/create.ts
+++ b/commands/sandbox/create.ts
@@ -79,6 +79,7 @@ exports.handler = async options => {
 
   // Check usage limits and exit if parent portal has no available sandboxes for the selected type
   try {
+    console.log('validate usage limit');
     await validateSandboxUsageLimits(accountConfig, sandboxType, env);
   } catch (err) {
     if (isMissingScopeError(err)) {
--- a/commands/sandbox/create.ts
+++ b/commands/sandbox/create.ts
@@ -79,7 +79,6 @@ exports.handler = async options => {
 
   // Check usage limits and exit if parent portal has no available sandboxes for the selected type
   try {
-    console.log('validate usage limit');
     await validateSandboxUsageLimits(accountConfig, sandboxType, env);
   } catch (err) {
     if (isMissingScopeError(err)) {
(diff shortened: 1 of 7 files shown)


lib/sandboxes.ts
```diff
@@ -321,8 +306,8 @@ function handleSandboxCreateError({
       category: 'VALIDATION_ERROR',
       subCategory: 'SandboxErrors.NUM_STANDARD_SANDBOXES_LIMIT_EXCEEDED_ERROR',
     }) &&
-    err.error &&
-    err.error.message
+    'error' in err &&
+    err.error instanceof Error
```
joe-yeager, line 310: I think this if statement might be unreachable code.  [if the `isSpecifiedError` is true](https://github.com/HubSpot/hubspot-local-dev-lib/blob/2b22bad59719f3d3df34da0a52f1656eb2b7c5dd/errors/index.ts#L5-L20C11), it is a [`HubSpotHttpError` which doesn't have a `error` member on it](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/models/HubSpotHttpError.ts#L8)

camden11: Gotcha, I wasn't sure if we always knew the shape of `HubSpotHttpError`s

lib/sandboxes.ts
```diff
@@ -68,79 +64,46 @@ function getHasSandboxesByType(parentAccountConfig, type) {
...
   return syncTypes.map(t => ({ type: t.name }));
 }
 
-/**
- * @param {Object} accountConfig - Account config of sandbox portal
- * @param {Array} availableSyncTasks - Array of available sync tasks
- * @param {Boolean} skipPrompt - Option to skip contact records prompt and return all available sync tasks
- * @returns {Array} Adjusted available sync task items
- */
-const getSyncTypesWithContactRecordsPrompt = async (
```
camden11, line 100: This function wasn't being used anywhere

lib/sandboxes.ts
```diff
@@ -263,117 +225,15 @@ function handleSandboxCreateError({
...
     logger.error(
-      i18n('lib.sandboxes.create.failure.403Gating', {
+      i18n(`${i18nKey}.create.failure.403Gating`, {
         accountName: name,
         parentAccountName: uiAccountDescription(accountId),
         accountId,
       })
     );
     logger.log('');
-  } else if (
```
camden11, line 273: From what I can tell, all this code was unreachable. These cases were being handled ahead of time in `validateSandboxUsageLimits`

commands/sandbox/create.ts
```diff
@@ -74,11 +74,12 @@ exports.handler = async options => {
     }
   }
   const sandboxType = type
-    ? sandboxTypeMap[type.toLowerCase()]
+    ? SANDBOX_TYPE_MAP[type.toLowerCase()]
     : typePrompt.type;
 
   // Check usage limits and exit if parent portal has no available sandboxes for the selected type
   try {
+    console.log('validate usage limit');
```
joe-yeager, line 82: Leftover debugging log

joe-yeager (approved):

lib/sandboxes.ts
```diff
@@ -68,79 +64,46 @@ function getHasSandboxesByType(parentAccountConfig, type) {
...
+export async function validateSandboxUsageLimits(
+  accountConfig: CLIAccount,
+  sandboxType: AccountType,
+  env: Environment
+): Promise<void> {
   const id = getAccountIdentifier(accountConfig);
   const accountId = getAccountId(id);
+
+  if (!accountId) {
+    throw new Error(`${i18nKey}.create.failure.usageLimitFetch`);
```
brandenrodgers, line 99: This log needs the `i18n()` function

lib/sandboxes.ts
```diff
@@ -68,79 +64,46 @@ function getHasSandboxesByType(parentAccountConfig, type) {
...
+  }
+
   const {
     data: { usage },
   } = await getSandboxUsageLimits(accountId);
   if (!usage) {
-    throw new Error('Unable to fetch sandbox usage limits. Please try again.');
+    throw new Error(`${i18nKey}.create.failure.usageLimitFetch`);
```
brandenrodgers, line 106: Same here with `i18n()`

camden11: Ooh good catch on these!

Issue #1327, opened 2025-01-10
"hs project upload" stopped working in GitHub Actions after upgrading to 7.0.0


huwq1987:
## Description and Context
We have a UI extension project and we were using GitHub Actions to run `hs project upload` on PR merge. We've set secrets `HUBSPOT_ACCOUNT_ID` & `HUBSPOT_PERSONAL_ACCESS_KEY` in the repo, and the action works fine for a few months.

Today that action suddenly breaks, and we noticed that a new version of @hubspot/cli was published to npm just hours ago. We have to downgrade HubSpot CLI to version 6.4.0 to fix this issue.

michiel-koho:
We're running into the exact same issue on our end. Even when using the `--use-env` flag, the command throws:
`Error:  An account needs to be supplied either via "--account" or through setting a "defaultPortal"`.
This also seems to occur when using the `hs upload` command to upload a regular theme.

We also had to switch from using `hs project upload ${{ inputs.path }}` to `cd ${{ inputs.path }} && hs project upload` to avoid an unknown argument error.

Hopefully this gets resolved in an upcoming release!

brandenrodgers:
Thanks for reporting this!

I'm working with the team on a quick fix. In the meantime, if you want to use CLI v7 you should be able to get the `--use-env` option to work by setting both `HUBSPOT_PORTAL_ID` and `HUBSPOT_ACCOUNT_ID`. A fix should be incoming soon though.

For some context, we have been moving away from using the "portal" terminology, so this latest release removes the "portal" term from a few places. Mainly the `HUBSPOT_PORTAL_ID` variable and the `--portal` flags, which are now `HUBSPOT_ACCOUNT_ID` and `--account` respectively. We've been simultaneously supporting both for a while.

@michiel-koho the `hs project upload` changes were an intentional move to simplify the behavior of the project commands. The ability to run some of the project commands from outside of a project directory was behavior from an initial iteration of projects. Many of the other project commands require the user to be within the project directory. We've made a handful of syntax changes in CLI v7 to increase consistency across commands with the hope of making the behavior more intuitive and easier to learn.

That being said, we're always open to feedback! 😄

brandenrodgers:
@huwq1987 this should be fixed now in the latest CLI patch release: `7.0.1`. I'm going to close this, but feel free to re-open if you're still seeing issues.

PR #1329, opened 2025-01-10
Remove support for portal env variable

brandenrodgers:
## Description and Context

Issue: https://github.com/HubSpot/hubspot-cli/issues/1327

In CLI release 7.0.0 we've moved away from referring to hubspot accounts as "portals" in favor of "account". This is a move that we've been slowly moving towards for a while now.

There was an issue with `HUBSPOT_ACCOUNT_ID` not working correctly that's fixed by https://github.com/HubSpot/hubspot-local-dev-lib/pull/225

This also adds a message for folks who are still using `HUBSPOT_PORTAL_ID`

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -182,7 +182,7 @@ const handleDeprecatedEnvVariables = options => {
     !process.env.HUBSPOT_ACCOUNT_ID
   ) {
     uiDeprecatedTag(
-      i18n(`${i18nKey}.injectAccountIdMiddleware.portalEnvVarDeprecated`, {
+      i18n(`${i18nKey}.handleDeprecatedEnvVariables.portalEnvVarDeprecated`, {
         configPath: getConfigPath(),
       })
     );
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -8,6 +8,7 @@ const { logger } = require('@hubspot/local-dev-lib/logger');
 const { addUserAgentHeader } = require('@hubspot/local-dev-lib/http');
 const {
   loadConfig,
+  getAccountId,
   configFileExists,
   getConfigPath,
   validateConfig,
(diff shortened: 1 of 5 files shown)


brandenrodgers:
Ok Slight pivot, my latest changes make `HUBSPOT_PORTAL_ID` deprecated, but still supported. We have to continue to support this env variable because our HubSpot GH Deploy Action leverages it, and always uses the latest CLI version that's available. 


This is what will show
image

bin/cli.js
```diff
@@ -261,8 +284,8 @@ const argv = yargs
   .middleware([
     setLogLevel,
     setRequestHeaders,
-    loadConfigMiddleware,
     injectAccountIdMiddleware,
+    loadConfigMiddleware,
```
kemmerle, line 288: Need to move this above `injectAccountIdMiddleware` and add the order needs to be:

`loadConfigMiddleware`
`injectAccountIdMiddleware`
`validateAccountOptions`

joe-yeager (approved):
LGTM!

bin/cli.js
```diff
@@ -176,6 +173,39 @@ const SKIP_CONFIG_VALIDATION = {
...
+const handleDeprecatedEnvVariables = options => {
+  // HUBSPOT_PORTAL_ID is deprecated, but we'll still support it for now
+  // The HubSpot GH Deploy Action still uses HUBSPOT_PORTAL_ID
+  if (
+    options.useEnv &&
+    process.env.HUBSPOT_PORTAL_ID &&
+    !process.env.HUBSPOT_ACCOUNT_ID
+  ) {
+    uiDeprecatedTag(
+      i18n(`${i18nKey}.injectAccountIdMiddleware.portalEnvVarDeprecated`, {
```
kemmerle, line 185: nit: should change the i18n property from `injectAccountIdMiddleware` to `handleDeprecatedEnvVariables`

kemmerle (approved):
I like this approach to deprecating the portal ID environmental variable. LGTM!

PR #1331, opened 2025-01-10
Remove CLI getAccountId util

brandenrodgers:
## Description and Context

This still needs to be tested thoroughly, but I'm fairly certain that we can remove this getAccountId() util in the CLI in favor of the getAccountId in local-dev-lib because now we have the `injectAccountIdMiddleware` middleware that does everything this old util in the CLI was doing. I'll leave inline comments to explain my thinking.

### DIFF
--- a/lib/commonOpts.ts
+++ b/lib/commonOpts.ts
@@ -8,7 +8,7 @@ import {
 } from '@hubspot/local-dev-lib/constants/files';
 import { CmsPublishMode } from '@hubspot/local-dev-lib/types/Files';
 import {
-  getAccountId as getAccountIdFromConfig,
+  getAccountId,
   getAccountConfig,
   getAndLoadConfigIfNeeded,
 } from '@hubspot/local-dev-lib/config';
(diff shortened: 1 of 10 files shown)


lib/commonOpts.ts
```diff
@@ -102,31 +102,19 @@ export function getCommandName(argv: Arguments<{ _?: string[] }>): string {
   return (argv && argv._ && argv._[0]) || '';
 }
 
-/**
- * Obtains accountId using supplied --account flag or from environment variables
- */
-export function getAccountId(
```
brandenrodgers, line 108: This util is just a wrapper around the local-dev-lib's `getAccountId`, except it also checks for any account env variables.

Now that we have [injectAccountIdMiddleware](https://github.com/HubSpot/hubspot-cli/blob/main/bin/cli.js#L202) doing this same thing, we should be able to remove this util. Commands can get this same behavior by looking at `options.derivedAccountId`. 

The benefit of removing this is that it cuts down on the number of options for getting account ID's in the project.

lib/validation.ts
```diff
@@ -22,23 +23,24 @@ import {
   getCwd,
   getExt,
 } from '@hubspot/local-dev-lib/path';
-import { getAccountId, getCmsPublishMode } from './commonOpts';
+import { getCmsPublishMode } from './commonOpts';
 import { logError } from './errorHandlers/index';
 
 export async function validateAccount(
```
brandenrodgers, line 29: I noticed that these validation checks weren't working correctly. I'm actually not sure if we should even keep them around. Now that I fixed these I'm starting to see them showing up. For example, I'm getting an error if I have env variables set and I try to run a CLI command without using `--use-env`. But I don't think we need to show an error in that case. The command still works as expected.

I think these checks might be too aggressive, but curious to hear what others think.

camden11: Hmm yeah that seems unnecessary, you may still want to use your config file after setting env variables. The only thing we really need to check for is using `--account` and `--use-env` at the same time, which we already do

camden11 (commented):
Code all LGTM, lets definitely test this thoroughly before merging though

joe-yeager (approved):

camden11 (approved):

PR #1334, opened 2025-01-13
Port dependencyManagement to TS

brandenrodgers:
## Description and Context

Porting the lib/dependencyManagement file to TS

### DIFF
--- a/lib/dependencyManagement.ts
+++ b/lib/dependencyManagement.ts
@@ -1,21 +1,20 @@
-// @ts-nocheck
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const { getProjectConfig } = require('./projects');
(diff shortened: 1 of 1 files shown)


lib/dependencyManagement.ts
```diff
@@ -38,24 +37,36 @@ export async function isGloballyInstalled(command) {
...
-      await installPackagesInDirectory(packages, dir);
+      await installPackagesInDirectory(dir, packages);
     })
   );
 }
 
-async function installPackagesInDirectory(packages, directory) {
+async function installPackagesInDirectory(
+  directory: string,
+  packages?: string[]
```
brandenrodgers, line 68: had to switch the order of args since packages is optional

camden11 (approved):
LGTM

PR #1335, opened 2025-01-13
Port DevServerManager to TS

brandenrodgers:
## Description and Context

Porting the DevServerManager file to TS. I have some questions that I'll add inline.

### DIFF
--- a/lib/DevServerManager.ts
+++ b/lib/DevServerManager.ts
@@ -1,48 +1,86 @@
-// @ts-nocheck
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const { COMPONENT_TYPES } = require('./projects/structure');
-const { i18n } = require('./lang');
-const { promptUser } = require('./prompts/promptUtils');
-const { DevModeInterface } = require('@hubspot/ui-extensions-dev-server');
-const {
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { Environment } from '@hubspot/local-dev-lib/types/Config';
+import {
+  COMPONENT_TYPES,
+  ComponentTypes,
+  Component,
+} from './projects/structure';
+import { i18n } from './lang';
+import { promptUser } from './prompts/promptUtils';
+import { DevModeInterface as UIEDevModeInterface } from '@hubspot/ui-extensions-dev-server';
+import {
   startPortManagerServer,
   stopPortManagerServer,
   requestPorts,
-} = require('@hubspot/local-dev-lib/portManager');
-const {
+} from '@hubspot/local-dev-lib/portManager';
+import {
   getHubSpotApiOrigin,
   getHubSpotWebsiteOrigin,
-} = require('@hubspot/local-dev-lib/urls');
-const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
+} from '@hubspot/local-dev-lib/urls';
+import { getAccountConfig } from '@hubspot/local-dev-lib/config';
+import { ProjectConfig } from '../types/Projects';
 
 const i18nKey = 'lib.DevServerManager';
 
 const SERVER_KEYS = {
   privateApp: 'privateApp',
   publicApp: 'publicApp',
+} as const;
+
+type ServerKey = keyof typeof SERVER_KEYS;
+
+type DevServerInterface = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+  setup?: Function;
+  start?: (options: object) => Promise<void>;
+  fileChange?: (filePath: string, event: string) => Promise<void>;
+  cleanup?: () => Promise<void>;
+};
+
+type DevServer = {
+  componentType: ComponentTypes;
+  serverInterface: DevServerInterface;
+};
+
+type ComponentsByType = {
+  [key in ComponentTypes]: { [key: string]: Component };
 };
 
 class DevServerManager {
+  private initialized: boolean;
+  private started: boolean;
+  private componentsByType: ComponentsByType;
+  private devServers: { [key in ServerKey]: DevServer };
+
   constructor() {
     this.initialized = false;
     this.started = false;
-    this.componentsByType = {};
-    this.server = null;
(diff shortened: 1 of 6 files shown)


lib/DevServerManager.ts
```diff
@@ -1,48 +1,86 @@
...
 const SERVER_KEYS = {
   privateApp: 'privateApp',
   publicApp: 'publicApp',
+} as const;
+
+type ServerKey = keyof typeof SERVER_KEYS;
+
+type DevServerInterface = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-function-type
+  setup?: Function;
```
brandenrodgers, line 34: Let me know what you think about this. eslint doesn't like the generic Function type, but I was having some trouble with this. The UIE dev server defines its own type (they call it SetupArguments), but they use their own type interfaces for it and they're slightly different from what we have defined on our end. So I was having trouble [here](https://github.com/HubSpot/hubspot-cli/pull/1335/files#diff-2c6c8dea63a8a0b77cc903f016f011e50bcc5d79c276fb59944f6cf63e34c27eR66). I just ignored the error for now, but I'm happy to see if I can get things working with a more specific type if we want.

lib/DevServerManager.ts
```diff
@@ -1,48 +1,86 @@
...
-    const serverKeys = Object.keys(this.devServers);
+  async iterateDevServers(
+    callback: (
+      serverInterface: DevServerInterface,
+      compatibleComponents: {
+        [key: string]: Component;
+      }
+    ) => Promise<void>
+  ): Promise<void> {
+    const serverKeys: ServerKey[] = Object.keys(this.devServers) as ServerKey[];
```
brandenrodgers, line 83: Is there a better way to do this without having to type cast?

camden11: Unfortunately there is not

lib/projects/structure.ts
```diff
@@ -70,13 +63,20 @@ type GenericComponentConfig =
...
+};
+
 export const COMPONENT_TYPES = {
   privateApp: 'private-app',
   publicApp: 'public-app',
   hublTheme: 'hubl-theme',
 } as const;
 
-type ComponentTypes = ValueOf<typeof COMPONENT_TYPES>;
+export type ComponentTypes = ValueOf<typeof COMPONENT_TYPES>;
```
camden11, line 79: Might make sense to move these to `types/Projects`

lib/DevServerManager.ts
camden11: Messing around with this locally and I feel like there's probably a way to do this using generic types but i think either way will require some work in the UIE dev server repo so probably not worth it right now

lib/DevServerManager.ts
```diff
@@ -1,48 +1,86 @@
...
+  cleanup?: () => Promise<void>;
+};
+
+type DevServer = {
+  componentType: ComponentTypes;
+  serverInterface: DevServerInterface;
+};
+
+type ComponentsByType = {
+  [key in ComponentTypes]: { [key: string]: Component };
```
camden11, line 46: It looks like this is saying that _every_ component type will always be present in this object, which doesn't seem to be the case. I think you want `[key in ComponentTypes]?: { [key: string]: Component };`

lib/DevServerManager.ts
```diff
@@ -59,21 +97,37 @@ class DevServerManager {
...
       }
 
-      acc[component.type][component.config.name] = component;
+      if ('name' in component.config && component.config.name) {
+        acc[component.type][component.config.name] = component;
+      }
 
       return acc;
-    }, {});
+    }, {} as ComponentsByType);
```
camden11, line 111: If you make the change to `ComponentsByType` above, you can do this without casting

```
  arrangeComponentsByType(components: Component[]): ComponentsByType {
    return components.reduce<ComponentsByType>((acc, component) => {
      if (!acc[component.type]) {
        acc[component.type] = {};
      }

      if ('name' in component.config && component.config.name) {
        acc[component.type]![component.config.name] = component;
      }

      return acc;
    }, {});
  }
```

types/Projects.ts
```diff
@@ -49,3 +49,74 @@ export type ProjectTemplateRepoConfig = {
...
+    objectTypes: Array<{ name: string }>;
+  };
+};
+
+export type GenericComponentConfig =
+  | PublicAppComponentConfig
+  | PrivateAppComponentConfig
+  | AppCardComponentConfig;
+
+export enum ComponentTypes {
```
brandenrodgers, line 111: I moved all of these to the types/projects file, and I switched `ComponentTypes` to be an enum. It felt redundant to have both `ComponentTypes` and `COMPONENT_TYPES`. This let me replace `COMPONENT_TYPES` everywhere in favor of the `ComponentTypes` enum

lib/DevServerManager.ts
brandenrodgers: Ok, thanks for poking around and verifying. I'm going to leave this as is for now 👍

lib/DevServerManager.ts
brandenrodgers: done in [086af30](https://github.com/HubSpot/hubspot-cli/pull/1335/commits/086af30e65d2a8ef88ded01a818d1388441170dd)

lib/DevServerManager.ts
brandenrodgers: Done in [086af30](https://github.com/HubSpot/hubspot-cli/pull/1335/commits/086af30e65d2a8ef88ded01a818d1388441170dd)

lib/projects/structure.ts
brandenrodgers: Good idea. I moved it all over in [086af30](https://github.com/HubSpot/hubspot-cli/pull/1335/commits/086af30e65d2a8ef88ded01a818d1388441170dd), and I made some changes that let me get rid of `COMPONENT_TYPES`

camden11 (approved):

PR #1337, opened 2025-01-13
Convert lib/serverlessLogs and lib/buildAccount to TS

camden11:
## Description and Context
Converts 2 more files to TS. Only two more to go after this!
https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/684

## Who to Notify
@brandenrodgers @kecarroll @joe-yeager


lib/__tests__/serverlessLogs.test.ts
```diff
@@ -12,16 +12,10 @@ const ACCOUNT_ID = 123;
 describe('lib/serverlessLogs', () => {
   describe('tailLogs()', () => {
     let stdinMock;
-    let spinnies;
 
     beforeEach(() => {
       jest.spyOn(process, 'exit').mockImplementation(() => {});
       stdinMock = mockStdIn.stdin();
-      spinnies = {
```
camden11, line 20: This was being passed into tailLogs but not doing anything

lib/buildAccount.ts
```diff
@@ -93,105 +85,145 @@ async function saveAccountToConfig({
...
 
-async function buildNewAccount({
-  name,
-  accountType,
-  accountConfig,
-  env,
-  portalLimit, // Used only for developer test accounts
-  force = false,
-}) {
+export async function buildDeveloperTestAccount(
```
camden11, line 88: I decided to split `buildNewAccount` into two separate functions. It adds a few extra lines of code, but IMO is much easier to read. I don't think there was a ton of utility in combining them previously, since it only combined two account types and would have been super messy if we ever attempted to add a third. It also was more or less impossible to safely type

lang/en.lyaml
```diff
@@ -1499,7 +1499,7 @@ en:
         missingScopeError: "Couldn't execute the {{ request }} because the access key for {{ accountName }} is missing required scopes. To update scopes, run {{ authCommand }}. Then deactivate the existing key and generate a new one that includes the missing scopes."
     serverless:
       verifyAccessKeyAndUserAccess:
-        fetchScopeDataError: "Error verifying access of scopeGroup {{ scopeGroup }}: {{ error }}"
+        fetchScopeDataError: "Error verifying access of scopeGroup {{ scopeGroup }}:"
```
brandenrodgers, line 1502: Should we remove the trailing colon here?

brandenrodgers (approved):
tested locally and this lgtm!

lib/serverlessLogs.ts
```diff
@@ -57,14 +69,15 @@ async function verifyAccessKeyAndUserAccess(accountId, scopeGroup) {
...
+    const resp = await fetchScopeData(accountId, scopeGroup);
+    scopesData = resp.data;
   } catch (e) {
     logger.debug(
       i18n(`${i18nKey}.verifyAccessKeyAndUserAccess.fetchScopeDataError`, {
         scopeGroup,
-        error: e,
       })
     );
+    logger.debug(e);
```
brandenrodgers, line 80: Not a blocker, but could this use the debugError() util?

PR #1339, opened 2025-01-14
Convert localDev and LocalDevManager to TS

camden11:
## Description and Context
This converts the last remaining files in `lib`, `localDev` and `localDevManager` to TS

## TODO
~Still need to run through the local dev flows and make sure this doesn't break anything~ It works!

## Who to Notify


lib/localDev.ts
```diff
@@ -1,66 +1,77 @@
...
 
 const i18nKey = 'lib.localDev';
 
 // If the user passed in the --account flag, confirm they want to use that account as
 // their target account, otherwise exit
-const confirmDefaultAccountIsTarget = async accountConfig => {
+export async function confirmDefaultAccountIsTarget(
+  accountConfig: CLIAccount
+): Promise<void> {
+  if (!accountConfig.name || !accountConfig.accountType) {
```
brandenrodgers, line 66: I'm a little wary of adding new logs in a TS port since it could subtly change the functionality. Would it be possible to break the code additions out into a standalone PR to merge and test before this one?

lib/LocalDevManager.ts
```diff
@@ -110,9 +150,13 @@ class LocalDevManager {
...
     const activePublicAppData = portalPublicApps.find(
-      ({ sourceId }) => sourceId === this.activeApp.config.uid
+      ({ sourceId }) => sourceId === getComponentUid(this.activeApp)
     );
 
+    if (!activePublicAppData) {
+      return;
+    }
+
     // TODO: Update to account for new API with { data }
```
brandenrodgers, line 160: nit, but can this TODO be removed?

lib/LocalDevManager.ts
```diff
@@ -205,11 +249,11 @@ class LocalDevManager {
         getProjectDetailUrl(
```
brandenrodgers, line 249: I wonder why `getProjectDetailUrl` returns if the projectName doesn't exist. The projectName field is required

lib/localDev.ts
camden11: Any new logs added here are more or less impossible to actually reach. I added them to satisfy TS (throwing errors makes typing a lot easier than having to detail with returning `undefined`). I guess theoretically this would trigger if you went in and manually removed these fields from the config but I have to imagine this would be better than whatever the behavior was previously. If you still think its worth separating out I can do that though

brandenrodgers (commented):
tested and lgtm 👍

lib/LocalDevManager.ts
camden11: I think that was left over from before it was typed. Once we're sure everywhere in the CLI is passing in a string, we can probably remove that check

PR #1342, opened 2025-01-15
Porting the accounts info command to TS

brandenrodgers:
## Description and Context

Porting the `hs accounts info` command to TS. I'm only porting a single command in this so we can agree on a pattern for how the commands should look before I port more of them.


commands/account/info.ts
```diff
@@ -1,41 +1,47 @@
...
     logger.log(i18n(`${i18nKey}.scopeGroups`));
     logger.log(getTableContents(scopeGroups, { border: { bodyLeft: '  ' } }));
   } else {
     logger.log(i18n(`${i18nKey}.errors.notUsingPersonalAccessKey`));
   }
-};
+}
 
-exports.builder = yargs => {
+function builder(yargs: Argv): Argv {
```
joe-yeager, line 44: If I recall correctly, we can use the `BuilderCallback` type to further [type this function](https://github.com/HubSpot/hubspot-cli/blob/main/commands/doctor.ts#L77C23-L77C38)

commands/account/info.ts
```diff
@@ -1,41 +1,47 @@
...
   if (config && config.authType === 'personalaccesskey') {
     const { name, personalAccessKey, env } = config;
+    let scopeGroups: string[][] = [];
 
-    const response = await getAccessToken(
-      personalAccessKey,
-      env,
-      derivedAccountId
-    );
+    if (personalAccessKey) {
```
joe-yeager, line 23: Does this change the behavior of the command?

joe-yeager (commented):
LGTM, just a question about behavior and a suggestion for a different type on the builder

types/Yargs.ts
```diff
@@ -0,0 +1,10 @@
+import { Arguments } from 'yargs';
+
+export type CommonOptions<T = object> = Arguments<
```
joe-yeager, line 3: Since we decided to use camel case, how would you feel about `ArgumentsCamelCase`?  Example https://github.com/HubSpot/hubspot-cli/blob/main/commands/doctor.ts#L26C4-L26C22

brandenrodgers: Switched to this in [001438e](https://github.com/HubSpot/hubspot-cli/pull/1342/commits/001438e543c4042a85fb71987e3e596a2b2db25d)

commands/account/info.ts
brandenrodgers: I did a little digging on this and I _think_ there might be an alternative to BuilderCallback.

In my latest iteration I switched to this pattern:
```ts
type AccountInfoOptions = {
  config?: string;
  someNewField: number;
};

function handler(
  options: CommonArguments<AccountInfoOptions>
): Promise<void> {}

function builder(yargs: Argv): Argv<AccountInfoOptions> {
  addConfigOptions(yargs);
  return yargs;
}
```

As far as I can tell, this is essentially equivalent but it prevents us from having to switch builder over to an arrow function. This seems to catch all the same type issues that the BuilderCallback was catching. In this example, the builder function would show a TS error because `someNewField` hasn't been added as an option. Including something like `yargs.option('some-new-field'...)` would fix it.

I'm not entirely sure if/how we should be handling the optional options like `--config`. Should that option be represented in the `AccountInfoOptions` type? We're not really getting anything out of it since the option isn't referenced anywhere in the command.

commands/account/info.ts
brandenrodgers: Ah yes, it does. This is actually an unhandled edge case I noticed, but I can fix it in a follow up PR instead of in here. Just because the auth type is personalaccesskey does not mean that the personalaccesskey exists in the config (users could manually remove it).

commands/account/info.ts
joe-yeager: Nice, that sounds good to me 👍

commands/account/info.ts
```diff
@@ -1,41 +1,47 @@
...
     logger.log(i18n(`${i18nKey}.scopeGroups`));
     logger.log(getTableContents(scopeGroups, { border: { bodyLeft: '  ' } }));
   } else {
     logger.log(i18n(`${i18nKey}.errors.notUsingPersonalAccessKey`));
   }
-};
+}
 
-exports.builder = yargs => {
+export function builder(yargs: Argv): Argv<AccountInfoArgs> {
```
brandenrodgers, line 44: We went back and forth on some ideas for how to type the `builder` function in a more useful way, but ultimately we landed on this because it's the simplest option. We can revisit this later on if we feel like we're not getting any value out of it. The main value of TS in these command files will be in the handler methods. We have acceptance tests (coverage will be improving soon) that will catch breaking changes to the builder methods.

camden11 (dismissed):
A few minor comments, but this all looks good to me!

commands/account/info.ts
```diff
@@ -1,41 +1,47 @@
...
+  const { derivedAccountId } = args;
   const config = getAccountConfig(derivedAccountId);
   // check if the provided account is using a personal access key, if not, show an error
   if (config && config.authType === 'personalaccesskey') {
     const { name, personalAccessKey, env } = config;
+    let scopeGroups: string[][] = [];
 
     const response = await getAccessToken(
-      personalAccessKey,
+      personalAccessKey!,
```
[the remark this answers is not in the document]
brandenrodgers: I think we should, but I'm trying to avoid making any code changes as a part of this port. I can do that as a follow up PR though!

types/Yargs.ts
```diff
@@ -0,0 +1,17 @@
+import { Options } from 'yargs';
+
+export type CommonArgs = {
+  derivedAccountId: number;
+  providedAccountId?: number;
+  d: boolean;
+  debug: boolean;
+};
+
+export type ConfigOptions = {
```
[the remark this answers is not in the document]
brandenrodgers: I'm using "Args" in [380aa51](https://github.com/HubSpot/hubspot-cli/pull/1342/commits/380aa51c22496cb89db9eabdc30a05eb6df2ef8e)

PR #1344, opened 2025-01-16
Adjust `hs upload --clean` warning prompt language

jsines:
## Description and Context
@ruslanhubspot just added this warning to the description for this command - adding it for the prompt that happens when its run as well to try to prevent more foot-shotguns


TanyaScales (approved):
GOOD add - thank you

PR #1345, opened 2025-01-16
chore: Convert `lib/tests` to TypeScript

kemmerle:
## Description and Context
In this PR, I've converted the following files from `/lib/__tests__` to TypeScript:

`commonOpts.test.ts`
`dependencyManagement.test.ts`
`projectsLogManager.test.ts`
`projects.test.ts`

`serverlessLogs.test.ts` is the final file from `/lib/__tests__` that needs TS conversion, but the prerequisite files haven't been converted yet. I plan to make that its own PR. 

## TODO

- [x] Fix failing tests
- [x] Address feedback

## Who to Notify

### DIFF
--- a/lib/__tests__/ProjectLogsManager.test.ts
+++ b/lib/__tests__/ProjectLogsManager.test.ts
@@ -1,9 +1,7 @@
-// @ts-nocheck
-const { ProjectLogsManager } = require('../projects/ProjectLogsManager');
-const { getProjectConfig, ensureProjectExists } = require('../projects');
-const {
-  fetchProjectComponentsMetadata,
-} = require('@hubspot/local-dev-lib/api/projects');
+import { ProjectLogsManager } from '../projects/ProjectLogsManager';
+import { getProjectConfig, ensureProjectExists } from '../projects';
+import { fetchProjectComponentsMetadata } from '@hubspot/local-dev-lib/api/projects';
+import { SUBCOMPONENT_TYPES } from '../../../hubspot-local-dev-lib/dist/enums/build';
 
 jest.mock('../projects');
 jest.mock('@hubspot/local-dev-lib/api/projects');
@@ -26,10 +24,11 @@ describe('lib/projects/ProjectLogsManager', () => {
   const function1 = {
     componentName: 'function1',
     type: {
-      name: 'APP_FUNCTION',
+      name: SUBCOMPONENT_TYPES.APP_FUNCTION,
     },
     deployOutput: {
       appId,
+      appFunctionName: 'function1',
     },
   };
   const functions = [
@@ -37,10 +36,11 @@ describe('lib/projects/ProjectLogsManager', () => {
     {
       componentName: 'function2',
       type: {
-        name: 'APP_FUNCTION',
+        name: SUBCOMPONENT_TYPES.APP_FUNCTION,
       },
       deployOutput: {
         appId,
+        appFunctionName: 'function2',
       },
     },
   ];
@@ -48,9 +48,9 @@ describe('lib/projects/ProjectLogsManager', () => {
   beforeEach(() => {
     ProjectLogsManager.reset();
 
-    getProjectConfig.mockResolvedValue(projectConfig);
-    ensureProjectExists.mockResolvedValue(projectDetails);
-    fetchProjectComponentsMetadata.mockResolvedValue({
+    (getProjectConfig as jest.Mock).mockResolvedValue(projectConfig);
+    (ensureProjectExists as jest.Mock).mockResolvedValue(projectDetails);
+    (fetchProjectComponentsMetadata as jest.Mock).mockResolvedValue({
       data: {
         topLevelComponentMetadata: [
           {
@@ -81,7 +81,7 @@ describe('lib/projects/ProjectLogsManager', () => {
     });
 
     it('should throw an error if there is a problem with the config', async () => {
-      getProjectConfig.mockResolvedValue({});
+      (getProjectConfig as jest.Mock).mockResolvedValue({});
       await expect(async () =>
         ProjectLogsManager.init(accountId)
       ).rejects.toThrow(
@@ -99,7 +99,7 @@ describe('lib/projects/ProjectLogsManager', () => {
     });
 
     it('should throw an error if there is data missing from the project details', async () => {
-      ensureProjectExists.mockResolvedValue({});
+      (ensureProjectExists as jest.Mock).mockResolvedValue({});
       await expect(async () =>
         ProjectLogsManager.init(accountId)
       ).rejects.toThrow(/There was an error fetching project details/);
@@ -177,9 +177,11 @@ describe('lib/projects/ProjectLogsManager', () => {
       const functionToChoose = {
         componentName: 'function1',
         type: {
-          name: 'APP_FUNCTION',
+          name: SUBCOMPONENT_TYPES.APP_FUNCTION,
         },
         deployOutput: {
+          appId: 123,
+          appFunctionName: 'function1',
           endpoint: { path: 'yooooooo', methods: ['GET'] },
         },
       };
@@ -202,11 +204,11 @@ describe('lib/projects/ProjectLogsManager', () => {
 
   describe('reset', () => {
     it('should reset all the values', async () => {
-      ProjectLogsManager.someRandomField = 'value';
-      expect(ProjectLogsManager.someRandomField).toBeDefined();
+      ProjectLogsManager.projectName = 'value';
+      expect(ProjectLogsManager.projectName).toBeDefined();
 
       ProjectLogsManager.reset();
-      expect(ProjectLogsManager.isPublicFunction).toBeUndefined();
+      expect(ProjectLogsManager.projectName).toBeUndefined();
     });
   });
 });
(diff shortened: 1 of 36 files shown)


lib/__tests__/ProjectLogsManager.test.ts
```diff
@@ -1,9 +1,19 @@
...
-const { ProjectLogsManager } = require('../projects/ProjectLogsManager');
-const { getProjectConfig, ensureProjectExists } = require('../projects');
-const {
-  fetchProjectComponentsMetadata,
-} = require('@hubspot/local-dev-lib/api/projects');
+import { ProjectLogsManager } from '../projects/ProjectLogsManager';
+import { getProjectConfig, ensureProjectExists } from '../projects';
+import { fetchProjectComponentsMetadata } from '@hubspot/local-dev-lib/api/projects';
+
+const SUBCOMPONENT_TYPES = {
```
kemmerle, line 5: The `enums` folder is not exported from LDL, so I added `SUBCOMPONENT_TYPES` to this test. Is that ok? Too brittle?

lib/__tests__/ProjectLogsManager.test.ts
```diff
@@ -1,9 +1,19 @@
-// @ts-nocheck
```
brandenrodgers, line 1: nit, but can you move this test into `/lib/projects/__tests__/` just to be consistent with how the other tests are organized?

lib/__tests__/ProjectLogsManager.test.ts
brandenrodgers: I think eventually it would be nice to have a single source of truth for this, but seeing how we're updating the project schema anyways it might not be worth putting too much effort into cleaning this up now.

lib/__tests__/dependencyManagement.test.ts
```diff
@@ -8,21 +7,21 @@ jest.mock('fs', () => ({
   existsSync: jest.fn().mockReturnValue(true),
 }));
```
brandenrodgers, line 8: if you add `jest.mock('util')` here you can import `{promisify}` from util and then simplify all of the promisify overrides to:

```ts
(promisify as unknown as jest.Mock).mockReturnValue(execMock);
```

There might be an even better way to do it, but that's the cleanest option I could find from playing around with it.

lib/__tests__/projects.test.ts
```diff
@@ -23,15 +22,20 @@ describe('lib/projects', () => {
 
     beforeEach(() => {
       exitMock = jest.fn();
-      global.process = { ...realProcess, exit: exitMock };
+      global.process = {
```
brandenrodgers, line 25: I think all of this process.exit overriding can be simplified to this:
```ts
    let exitMock: jest.SpyInstance;

    beforeAll(() => {
      projectDir = fs.mkdtempSync(path.join(os.tmpdir(), 'projects-'));
      fs.mkdirSync(path.join(projectDir, 'src'));
    });

    beforeEach(() => {
      exitMock = jest.spyOn(process, 'exit').mockImplementation((): never => {
        throw new Error('process.exit() was called.');
      });
    });

    afterEach(() => {
      exitMock.mockRestore();
    });
```

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1345/commits/16aafdc05dcb920c6d8f463c3316e55ccfbee67b

lib/__tests__/dependencyManagement.test.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1345/commits/ee15a6b4d7c9cd9b947f871abaf6af859d949b98

lib/__tests__/ProjectLogsManager.test.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1345/commits/aad6d1b774389f1cc6f1e24108b61e8cc17505f8

lib/projects/__tests__/ProjectLogsManager.test.ts
```diff
@@ -1,9 +1,19 @@
...
+  PACKAGE_LOCK_FILE: 'PACKAGE_LOCK_FILE',
+  CRM_CARD_V2: 'CRM_CARD_V2',
+  CARD_V2: 'CARD_V2',
+  SERVERLESS_PKG: 'SERVERLESS_PKG',
+  SERVERLESS_ROUTE: 'SERVERLESS_ROUTE',
+  SERVERLESS_FUNCTION: 'SERVERLESS_FUNCTION',
+  APP_FUNCTION: 'APP_FUNCTION',
+  AUTOMATION_ACTION: 'AUTOMATION_ACTION',
+  REACT_EXTENSION: 'REACT_EXTENSION',
+} as const;
```
camden11, line 16: You can just grab these from LDL

camden11 (commented):
Nice, this is looking good! Just a few minor comments

lib/projects/__tests__/projects.test.ts
```diff
@@ -1,37 +1,34 @@
...
       fs.mkdirSync(path.join(projectDir, 'src'));
-
-      realProcess = process;
     });
 
     beforeEach(() => {
-      exitMock = jest.fn();
-      global.process = { ...realProcess, exit: exitMock };
+      exitMock = jest.spyOn(process, 'exit').mockImplementation((): never => {
+        throw new Error('process.exit() was called.');
```
camden11, line 22: Also, it looks like this is causing the tests to fail

lib/__tests__/commonOpts.test.ts
```diff
@@ -39,62 +55,76 @@ describe('lib/commonOpts', () => {
...
           );
-          getAccountId.mockReturnValue(accounts.PROD);
-          getAccountConfig.mockReturnValue(prodAccountConfig);
-          loadConfigFromEnvironment.mockReturnValue(undefined);
-          expect(getCmsPublishMode({ account: accounts.PROD })).toBe(
-            CMS_PUBLISH_MODE.draft
-          );
+          (getAccountId as jest.Mock).mockReturnValue(accounts.PROD);
+          (getAccountConfig as jest.Mock).mockReturnValue(prodAccountConfig);
+          (loadConfigFromEnvironment as jest.Mock).mockReturnValue(undefined);
```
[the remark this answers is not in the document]
brandenrodgers: Oh good call, I just made this same update in my test PR 👍

lib/__tests__/projects.test.ts
kemmerle: I am reverting this change, since throwing an error in the `beforeEach` block is breaking the tests. When I consulted GH copilot about the best way to fix it, it suggested the revert 😵‍💫

brandenrodgers: This works with:
```ts
    beforeEach(() => {
      exitMock = jest
        .spyOn(process, 'exit')
        .mockImplementation((): never => undefined as never);
    });
```

lib/projects/__tests__/projects.test.ts
brandenrodgers: I commented above, but it's because it's throwing the error and causing the tests to fail because of it. We can fix it by just returning undefined as never

lib/__tests__/commonOpts.test.ts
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1345/commits/37a009a43a91917e063212a65e5ba1a38ba49213

lib/__tests__/dependencyManagement.test.ts
```diff
@@ -34,8 +34,8 @@ describe('lib/dependencyManagement', () => {
 
   beforeEach(() => {
     execMock = jest.fn();
-    util.promisify = jest.fn().mockReturnValue(execMock);
-    getProjectConfig.mockResolvedValue({
+    (promisify as unknown as jest.Mock).mockReturnValue(execMock);
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1345/commits/37a009a43a91917e063212a65e5ba1a38ba49213

lib/__tests__/commonOpts.test.ts
```diff
@@ -1,15 +1,36 @@
...
+const mockedGetAccountConfig = getAccountConfig as jest.Mock;
+const mockedGetAndLoadConfigIfNeeded = getAndLoadConfigIfNeeded as jest.Mock;
+const mockedLoadConfigFromEnvironment = loadConfigFromEnvironment as jest.Mock;
+
+type CmsPublishModeArgs = {
+  cmsPublishMode?: CmsPublishMode;
+  account?: number | string;
+};
+
+function buildArguments(
```
brandenrodgers, line 25: Just a note, but I had to add something similar in my other PR. We may eventually want a util that does this so we don't have to do this everywhere. I can look into that in my PR after this is merged

lib/projects/__tests__/ProjectLogsManager.test.ts
kemmerle: I was getting an error when I attempted that--it didn't look to me like the `enums` folder where this is located had an export assigned in LDL's package.json. Am I missing anything?

camden11: Oh interesting, I didn't realize those aren't exported. That's fine then!

camden11 (approved):
LGTM!

PR #1347, opened 2025-01-21
feat: Update generic error message when repo doe not contain config.json file

joe-yeager:
## Description and Context

I noticed there was a generic error message that was being logged when the `--template-source` does not have ` config.json` file in the project root. Updating that message to give a more actionable error message instead of `[ERROR] An error occurred fetching JSON file.`

### Before

Screenshot 2025-01-21 at 1 43 49 PM

### After
Screenshot 2025-01-21 at 1 43 30 PM


kemmerle (approved):
:shipit:

PR #1349, opened 2025-01-22
Add error messages for default account override

kemmerle:
## Description and Context
Relies on https://github.com/HubSpot/hubspot-local-dev-lib/pull/228

In this PR, I'm implementing support for default account overrides with the new centralized configuration file. 

Now, customers can place a `.hs-account` file in any directory where they would like the default account overridden. The `.hs-account` file will contain a single numerical entry: `1234567`, or the account ID. Any changes made in the directory and its subdirectories will apply in the account stipulated in the `.hs-account` file. This will work very similarly to our nest configs currently do. 

The advantages are:

1) We will still have one centralized configuration file and therefore one single source of truth. 
2) We will create a command to auto-generate the `.hs-account` file, so that customers need never have to manually interact with the config. 
3) It's much more secure, if a customer accidentally uploads the file to GitHub--it contains no sensitive information besides an account ID. 

## TODO

- [ ] Implement the `hs accounts create-override` command 
- [ ] Address feedback 

## Who to Notify

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -13,6 +13,10 @@ const {
   getConfigPath,
   validateConfig,
 } = require('@hubspot/local-dev-lib/config');
+const {
+  DEFAULT_ACCOUNT_OVERRIDE_ERROR_INVALID_ID,
+  DEFAULT_ACCOUNT_OVERRIDE_ERROR_ACCOUNT_NOT_FOUND,
+} = require('@hubspot/local-dev-lib/constants/config');
 const { logError } = require('../lib/errorHandlers/index');
 const { setLogLevel, getCommandName } = require('../lib/commonOpts');
 const { validateAccount } = require('../lib/validation');
@@ -202,7 +206,27 @@ const injectAccountIdMiddleware = async options => {
   if (options.useEnv && process.env.HUBSPOT_ACCOUNT_ID) {
     options.derivedAccountId = parseInt(process.env.HUBSPOT_ACCOUNT_ID, 10);
   } else {
-    options.derivedAccountId = getAccountId(account);
+    try {
+      options.derivedAccountId = getAccountId(account);
+    } catch (error) {
+      logger.error(error.message || error);
+      if (error.cause === DEFAULT_ACCOUNT_OVERRIDE_ERROR_INVALID_ID) {
+        logger.log(
+          i18n(`${i18nKey}.injectAccountIdMiddleware.invalidAccountId`, {
+            overrideCommand: uiCommandReference('hs accounts create-override'),
+          })
+        );
+      }
+      if (error.cause === DEFAULT_ACCOUNT_OVERRIDE_ERROR_ACCOUNT_NOT_FOUND) {
+        logger.log(
+          i18n(`${i18nKey}.injectAccountIdMiddleware.accountNotFound`, {
+            configPath: getConfigPath(),
+            authCommand: uiCommandReference('hs accounts auth'),
+          })
+        );
+      }
+      process.exit(EXIT_CODES.ERROR);
+    }
   }
 };
 
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -11,6 +11,9 @@ en:
         portalEnvVarDeprecated: "The HUBSPOT_PORTAL_ID environment variable is deprecated. Please use HUBSPOT_ACCOUNT_ID instead."
       loadConfigMiddleware:
         configFileExists: "A configuration file already exists at {{ configPath }}. To specify a new configuration file, delete the existing one and try again."
+      injectAccountIdMiddleware:
+        invalidAccountId: "In the default override file (.hs-account), the account ID must be a number. Please delete the current file and generate a new one using {{ overrideCommand }}."
+        accountNotFound: "The account in the default override file (.hs-account) was not found in the centralized config file at {{ configPath }}. You can authorize this account using {{ authCommand }}."
     completion:
       describe: "Enable bash completion shortcuts for commands. Concat the generated script to your .bashrc, .bash_profile, or .zshrc file."
       examples:


bin/cli.js
```diff
@@ -202,7 +206,27 @@ const injectAccountIdMiddleware = async options => {
...
   } else {
-    options.derivedAccountId = getAccountId(account);
+    try {
+      options.derivedAccountId = getAccountId(account);
+    } catch (error) {
+      logger.error(error.message || error);
+      if (error.cause === DEFAULT_ACCOUNT_OVERRIDE_ERROR_INVALID_ID) {
+        logger.log(
+          i18n(`${i18nKey}.injectAccountIdMiddleware.invalidAccountId`, {
+            overrideCommand: uiCommandReference('hs accounts create-override'),
```
brandenrodgers, line 216: nit, but we prefer the singular usage now `hs account create-override`

bin/cli.js
```diff
@@ -202,7 +206,27 @@ const injectAccountIdMiddleware = async options => {
   if (options.useEnv && process.env.HUBSPOT_ACCOUNT_ID) {
     options.derivedAccountId = parseInt(process.env.HUBSPOT_ACCOUNT_ID, 10);
   } else {
-    options.derivedAccountId = getAccountId(account);
+    try {
+      options.derivedAccountId = getAccountId(account);
+    } catch (error) {
+      logger.error(error.message || error);
```
brandenrodgers, line 212: You can probably leverage the `logError` util here

bin/cli.js
```diff
@@ -202,7 +206,27 @@ const injectAccountIdMiddleware = async options => {
   if (options.useEnv && process.env.HUBSPOT_ACCOUNT_ID) {
     options.derivedAccountId = parseInt(process.env.HUBSPOT_ACCOUNT_ID, 10);
   } else {
-    options.derivedAccountId = getAccountId(account);
```
brandenrodgers, line 205: I'm realizing the implication of `getAccountId` potentially throwing an error. Before, the worst case scenario was that `getAccountId` could return null, but now it can throw an error. Does that mean we have to go through every usage of it in the app and wrap them all in try/catch blocks? I wonder if there's some way to get around having to do that 🤔 

Is it safe to just make the assumption that any potential thrown errors from `getAccountId()` would get caught here in the middleware, before any of the code in the commands can execute?

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (e instanceof Error) {
+      logger.error(i18n(`${i18nKey}.errors.writeFile`, { error: e.message }));
+    } else {
+      logger.error(i18n(`${i18nKey}.errors.writeFile`, { error: String(e) }));
+    }
+    process.exit(EXIT_CODES.ERROR);
+  }
+}
+
+export function builder(yargs: Argv): Argv<AccountInfoArgs> {
```
brandenrodgers, line 59: This should declare the `account` positional (like in account/use.ts)

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const i18nKey = 'commands.account.subcommands.createOverride';
+
+export const describe = null; // i18n(`${i18nKey}.describe`);
+
+export const command = 'create-override [account]';
+
+type AccountInfoArgs = CommonArgs &
+  ConfigArgs & {
+    account?: string | number;
```
brandenrodgers, line 21: Account can always be a string here, which you can declare when you define the positional

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+  args: ArgumentsCamelCase<AccountInfoArgs>
+): Promise<void> {
+  let overrideDefaultAccount = args.account;
+
+  if (!overrideDefaultAccount) {
+    overrideDefaultAccount = await selectAccountFromConfig();
+  } else if (
+    (typeof overrideDefaultAccount !== 'string' &&
+      typeof overrideDefaultAccount !== 'number') ||
+    !getAccountId(overrideDefaultAccount)
```
brandenrodgers, line 34: This can be simplified to just `!getAccountId()` since yargs will yell at the user if the input type of the positional doesn't match what we're expecting

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+    );
+    overrideDefaultAccount = await selectAccountFromConfig();
+  }
+  const accountId = getAccountId(overrideDefaultAccount);
+
+  try {
+    const overrideFilePath = path.join(getCwd(), '.hs-account');
+    await fs.writeFile(overrideFilePath, accountId!.toString(), 'utf8');
+    logger.success(i18n(`${i18nKey}.success`, { overrideFilePath }));
+  } catch (e: unknown) {
```
brandenrodgers, line 49: You can probably use `logError` here too

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+      i18n(`${i18nKey}.errors.accountNotFound`, {
+        configPath: getConfigPath() || '',
+      })
+    );
+    overrideDefaultAccount = await selectAccountFromConfig();
+  }
+  const accountId = getAccountId(overrideDefaultAccount);
+
+  try {
+    const overrideFilePath = path.join(getCwd(), '.hs-account');
```
brandenrodgers, line 46: Do you have a constant from LDL that we can use for the file name here?

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+    await fs.writeFile(overrideFilePath, accountId!.toString(), 'utf8');
+    logger.success(i18n(`${i18nKey}.success`, { overrideFilePath }));
```
brandenrodgers, line 48: I think you'll also want a process.exit for the success case

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    process.exit(EXIT_CODES.ERROR);
+  }
+}
+
+export function builder(yargs: Argv): Argv<AccountInfoArgs> {
+  addConfigOptions(yargs);
+
+  yargs.example([
+    ['$0 accounts create-override', i18n(`${i18nKey}.examples.default`)],
```
brandenrodgers, line 63: Same nit here about us preferring the singular case: `$0 account`

commands/account/createOverride.ts
```diff
@@ -0,0 +1,75 @@
...
+import { selectAccountFromConfig } from '../../lib/prompts/accountsPrompt';
+import { CommonArgs, ConfigArgs } from '../../types/Yargs';
+
+const i18nKey = 'commands.account.subcommands.createOverride';
+
+export const describe = null; // i18n(`${i18nKey}.describe`);
+
+export const command = 'create-override [account]';
+
+type AccountInfoArgs = CommonArgs &
```
brandenrodgers, line 19: Looks like a copy/paste leftover here 😄

bin/cli.js
kemmerle: Yes, we can safely assume that any potential errors from `getAccountId()` will be caught in `injectAccountIdMiddleware`. 

Apart from the `init` and `auth` commands, `injectAccountMiddleware` will run in the middleware before every command handler. Actually, the work you did to eliminate multiple `getAccountId` functions guarantees that--now we call the same function every time, so we know that if it fails in the command, it will also fail in middleware.

commands/account/createOverride.ts
kemmerle: Not quite sure what you mean--I based the type off of the types in the `hs account info` command but had to make some changes. Should I move this to `Types`, so we can reuse between commands?

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5b758d2e4b814cb8abc89bba8cb54cb3c25df7b1

bin/cli.js
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5b758d2e4b814cb8abc89bba8cb54cb3c25df7b1

commands/account/createOverride.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5b758d2e4b814cb8abc89bba8cb54cb3c25df7b1

commands/account/createOverride.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5b758d2e4b814cb8abc89bba8cb54cb3c25df7b1

commands/account/createOverride.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5b758d2e4b814cb8abc89bba8cb54cb3c25df7b1

commands/account/createOverride.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5b758d2e4b814cb8abc89bba8cb54cb3c25df7b1

commands/account/createOverride.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5b758d2e4b814cb8abc89bba8cb54cb3c25df7b1

commands/account/createOverride.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/3cb9c4b8c3e2f171fc96e70836a9c644a3df8f1f

lib/doctor/Doctor.ts
```diff
@@ -112,6 +116,26 @@ export class Doctor {
     return [this.checkIfAccessTokenValid()];
   }
 
+  private performDefaultAccountOverrideFileChecks(): Array<Promise<void>> {
```
brandenrodgers, line 119: I was a little confused why this has to return an empty array and it looks like the other functions in this file return `Array<Promise>` because they're all doing something asynchronous.  Returning the arrays of promises lets them be used up [here in the promise.all()](https://github.com/HubSpot/hubspot-cli/pull/1349/files#diff-e4ba2d37078a484aad98f6bb4d2e0f34fbf3d6e4c9d1ced40dec710ba9a55775R67-R72). This function isn't doing anything asynchronous so it doesn't need to follow that pattern. This function can have a return type of `void` and it can be removed from the promise.all chain above.

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1349/commits/5cdb2178041106ba8f1ea6733e535a3c005628e4

brandenrodgers (dismissed):
Looks like there's a merge conflict, but this lgtm 👌

lib/doctor/Doctor.ts
```diff
@@ -117,6 +122,25 @@ export class Doctor {
...
+        }),
+      });
+      this.diagnosis?.addDefaultAccountOverrideFileSection({
+        type: 'warning',
+        message: i18n(`${localI18nKey}.overrideAccountId`, {
+          overrideAccountId: getCWDAccountOverride(),
+        }),
+      });
+    }
+  }
```
joe-yeager, line 142: Do we want to do any checks on the validity of the account?  Should these be warnings by default?

joe-yeager (dismissed):
LGTM!  I have some questions about the doctor changes being made, but that doesn't need to hold up this PR and we can always fast follow changes if we decide to make them.

commands/account/auth.ts
```diff
@@ -0,0 +1,212 @@
...
+
+const i18nKey = 'commands.account.subcommands.auth';
+
+const TRACKING_STATUS = {
+  STARTED: 'started',
+  ERROR: 'error',
+  COMPLETE: 'complete',
+};
+
+async function createOrUpdateConfig(
```
camden11, line 45: Does this function actually create a config? I might be misunderstanding, but from what I can tell, it just updates existing configs and doesn't do anything if `doesConfigExist` is false

commands/account/auth.ts
```diff
@@ -0,0 +1,212 @@
...
+    trackCommandUsage('account-auth', {}, providedAccountId);
+    await trackAuthAction('account-auth', authType, TRACKING_STATUS.STARTED);
+  }
+
+  const env = args.qa ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD;
+  const configExists = configFileExists(true);
+
+  if (!configExists) {
+    if (!disableTracking) {
+      await trackAuthAction('account-auth', authType, TRACKING_STATUS.STARTED);
```
camden11, line 126: It looks like this could cause duplicate tracking of `TRACKING_STATUS.STARTED` if the config doesn't exist

commands/account/auth.ts
```diff
@@ -0,0 +1,212 @@
...
+    }
+    createEmptyConfigFile({}, true);
+  }
+  loadConfig('');
+
+  handleExit(deleteEmptyConfigFile);
+
+  const updatedConfig = await createOrUpdateConfig(
+    env,
+    configExists,
```
camden11, line 136: Ah, I think I see what's happening here. Is `configExists` used to tell `createOrUpdateConfig` if the config has newly been created? It might be a little clearer to name it `configIsEmpty` or something like that

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hubspot/cli",
-  "version": "7.1.1",
+  "version": "7.0.13-experimental.0",
```
camden11, line 3: Just flagging to make sure we update this before merging!

lang/en.lyaml
```diff
@@ -11,18 +11,45 @@ en:
...
+        createOverride:
+          describe: "Create a new default account override file (.hs-account) in the current working directory."
+          success: "Default account override file created at {{ overrideFilePath }}"
+          errors:
+            accountNotFound: "The specified account could not be found in the config file {{ configPath }}"
+          options:
+            account:
+              describe: "Name or ID of the account to create an override file for."
+          examples:
+            default: "Create a new default account override file (.hs-account) in the current working directory"
```
brandenrodgers, line 44: I think you can use the `hsAccountFileName` variables here too

joe-yeager (approved):
LGTM!

PR #1351, opened 2025-01-22
Port account commands to TS and add basic tests

brandenrodgers:
## Description and Context

This ports the following files to TS and adds (or updates) the units tests for each one:
- account/clean.ts
- account/list.ts
- account/remove.ts
- account/rename.ts
- account/use.ts
- account.ts


commands/__tests__/account.test.ts
```diff
@@ -0,0 +1,66 @@
+import yargs, { Argv } from 'yargs';
```
brandenrodgers, line 1: This is mostly just a rename (from accounts.test.ts), but I also had to make some changes in here to make TS happy. The main difference is this:
```ts

const commandSpy = jest
  .spyOn(yargs as Argv, 'command')
  .mockReturnValue(yargs as Argv);
const demandCommandSpy = jest
  .spyOn(yargs as Argv, 'demandCommand')
  .mockReturnValue(yargs as Argv);
```

commands/__tests__/account.test.ts
```diff
@@ -0,0 +1,66 @@
...
+jest.mock('../account/remove');
+jest.mock('../account/clean');
+jest.mock('../../lib/commonOpts');
+
+const commandSpy = jest
+  .spyOn(yargs as Argv, 'command')
+  .mockReturnValue(yargs as Argv);
+const demandCommandSpy = jest
+  .spyOn(yargs as Argv, 'demandCommand')
+  .mockReturnValue(yargs as Argv);
```
camden11, line 23: Too bad that you have to do all this casting. Looked into the yargs types though and looks like its the only way.

commands/account/clean.ts
```diff
@@ -51,10 +54,10 @@ exports.handler = async options => {
 
   for (const account of filteredTestAccounts) {
     try {
-      await accessTokenForPersonalAccessKey(
-        getAccountIdentifier(account),
-        true
-      );
+      const accountId = getAccountIdentifier(account);
+      if (accountId) {
```
camden11, line 58: I would just double check that this doesn't change any existing behavior - previously if there was no accountId, the catch block would be triggered

commands/account/list.ts
```diff
@@ -1,113 +1,131 @@
...
 
-exports.builder = yargs => {
+export function builder(yargs: Argv): Argv<AccountListArgs> {
   addConfigOptions(yargs);
+
   yargs.example([['$0 accounts list']]);
-  return yargs;
-};
+
+  return yargs as Argv<AccountListArgs>;
```
camden11, line 130: Fair if you don't think it's worth the effort, but I was able to make builders work without having to cast in my POC PR: https://github.com/HubSpot/hubspot-cli/pull/1346

I also noticed that we don't explicitly add the `qa` option here. Presumably because we don't want to show it in `help`? Might be worth adding a comment to clarify that

commands/account/remove.ts
```diff
@@ -50,26 +62,28 @@ exports.handler = async options => {
   );
 
   // Get updated version of the config
-  loadConfig(getConfigPath(), options);
+  loadConfig(getConfigPath()!, args as CLIOptions);
```
camden11, line 65: I think in this case you can probably just omit `args` from the `loadConfig` call - it looks like the fields that LDL are looking for in `CLIOptions` will never exist in this command

commands/__tests__/account.test.ts
brandenrodgers: Yeah I spent a little time trying to avoid having to do this and this was still the best pattern I could find

commands/account/clean.ts
brandenrodgers: Oh good call, I'll take a look to make sure

commands/account/list.ts
brandenrodgers: I think we intentionally don't add `qa` here because all this does is log out the configured accounts in the config file. The env of the account doesn't really matter too much for this one. We could maybe support it and then filter to just the qa env, but that would be net-new functionality.

>  I was able to make builders work without having to cast in my POC PR:

I actually went back and forth a few times when I first saw that POC. I went with this because the alternative of declaring the input as `Argv<AccountListArgs>` seems a little misleading to me. It's saying up front that yargs is of type `Argv<AccountListArgs>`, but in reality it is just a regular `Argv` when it gets passed into the builder func. It gets all of the options added inside of the builder.

That being said, I don't feel too strongly about this one so I can switch it up if we'd prefer to avoid casting.

commands/account/remove.ts
brandenrodgers: Nice catch 🙏

commands/account/clean.ts
brandenrodgers: Fixed this in my latest commit. I think it did change the behavior

joe-yeager (approved):

PR #1353, opened 2025-01-23
chore: unit tests for polling

brandenrodgers:
## Description and Context

Adding tests for lib/polling. I also made some changes to make the polling function more generic and easier to test. I'll add more context inline.

### DIFF
--- a/lib/__tests__/polling.test.ts
+++ b/lib/__tests__/polling.test.ts
@@ -1,0 +1,107 @@
+import { poll, DEFAULT_POLLING_STATES } from '../polling';
+import { DEFAULT_POLLING_DELAY } from '../constants';
+import { HubSpotPromise } from '@hubspot/local-dev-lib/types/Http';
+
+// Mock response types
+type MockResponse = {
+  status: string;
+};
+
+// Helper to create a mock polling callback
+const createMockCallback = (responses: MockResponse[]) => {
+  let callCount = 0;
+  return jest.fn((): HubSpotPromise<{ status: string }> => {
+    const response = responses[callCount];
+    callCount++;
+    return Promise.resolve({ data: response }) as HubSpotPromise<{
+      status: string;
+    }>;
+  });
+};
+
+describe('lib/polling', () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+  });
+
+  describe('poll()', () => {
+    it('should resolve when status is SUCCESS', async () => {
+      const mockCallback = createMockCallback([
+        { status: DEFAULT_POLLING_STATES.STARTED },
+        { status: DEFAULT_POLLING_STATES.SUCCESS },
+      ]);
+
+      const pollPromise = poll(mockCallback);
+
+      // Fast-forward through two polling intervals
+      jest.advanceTimersByTime(DEFAULT_POLLING_DELAY * 2);
+
+      const result = await pollPromise;
+      expect(result.status).toBe(DEFAULT_POLLING_STATES.SUCCESS);
+      expect(mockCallback).toHaveBeenCalledTimes(2);
+    });
+
+    it('should reject when status is ERROR', async () => {
+      const mockCallback = createMockCallback([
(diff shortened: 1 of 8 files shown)


lib/polling.ts
```diff
@@ -1,42 +1,54 @@
...
 type GenericPollingResponse = {
-  status: ValueOf<typeof POLLING_STATUS>;
+  status: string;
 };
 
-type PollingCallback<T extends GenericPollingResponse> = (
-  accountId: number,
-  taskId: number | string
-) => HubSpotPromise<T>;
+type PollingCallback<T extends GenericPollingResponse> =
```
brandenrodgers, line 25: I changed the signature of the callback so we could simplify the arguments in `poll`. This makes the polling function more generic so it can be more easily leveraged in other places.

lib/polling.ts
```diff
@@ -1,42 +1,54 @@
...
+
+export const DEFAULT_POLLING_STATES = {
+  STARTED: 'STARTED',
+  SUCCESS: 'SUCCESS',
+  ERROR: 'ERROR',
+  REVERTED: 'REVERTED',
+  FAILURE: 'FAILURE',
+} as const;
+
+const DEFAULT_POLLING_STATUS_LOOKUP = {
```
brandenrodgers, line 12: This lookup will enable this polling func to be used for other apis, even if their status enums are slightly different from the default ones here. The only requirement in the polling util now is that the callback response needs to include a `status` field.

lib/polling.ts
```diff
@@ -1,42 +1,54 @@
 import { HubSpotPromise } from '@hubspot/local-dev-lib/types/Http';
-import { ValueOf } from '@hubspot/local-dev-lib/types/Utils';
-import { POLLING_DELAY, POLLING_STATUS } from './constants';
+import { DEFAULT_POLLING_DELAY } from './constants';
+
+export const DEFAULT_POLLING_STATES = {
```
brandenrodgers, line 4: I moved this here because it's only relevant to this file

lib/__tests__/polling.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    it('should resolve when status is SUCCESS', async () => {
+      const mockCallback = createMockCallback([
+        { status: DEFAULT_POLLING_STATES.STARTED },
+        { status: DEFAULT_POLLING_STATES.SUCCESS },
+      ]);
+
+      const pollPromise = poll(mockCallback);
+
+      // Fast-forward through two polling intervals
+      jest.advanceTimersByTime(DEFAULT_POLLING_DELAY * 2);
```
brandenrodgers, line 41: Cursor came up with the idea to use `advanceTimersByTime`, but it's a cool way to test utils that leverage setTimeout. This gives us control over each execution of the setTimeout in poll(). So we can iterate over the mock responses one at a time.

commands/account/hello.ts
kemmerle, line 1: I think this might have been included accidentally--it's not referred to anywhere else.

kemmerle (dismissed):
Read through the code and tested this out by cloning an app and building/deploying a project. Besides the one comment on the `hello` command, LGTM!

commands/account/hello.ts
brandenrodgers: 🤦 haha thanks for catching this. I was trying to use Cursor to auto-generate cli commands.

camden11 (dismissed):
Nice, love the changes to make polling more flexible. One super minor comment but LGTM

lib/polling.ts
```diff
@@ -1,42 +1,54 @@
...
+  () => HubSpotPromise<T>;
 
 export function poll<T extends GenericPollingResponse>(
   callback: PollingCallback<T>,
-  accountId: number,
-  taskId: number | string
+  statusLookup?: { successStates: string[]; errorStates: string[] }
 ): Promise<T> {
+  if (!statusLookup) {
+    statusLookup = DEFAULT_POLLING_STATUS_LOOKUP;
```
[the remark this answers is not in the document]
brandenrodgers: done in [327b5f8](https://github.com/HubSpot/hubspot-cli/pull/1353/commits/327b5f82b44970c8ddb102faabe96bbb118a4caf)

PR #1354, opened 2025-01-23
Some auth flow bug fixes

brandenrodgers:
## Description and Context

This fixes two things related to the auth flow:
- Fixes the `--auth-type` field in `init.ts` which wasn't being respected
- Fixes an issue where we were logging the "config not found" message multiple times when running `hs auth` when no config file exists.

### Before
image

### After
image


kemmerle (approved):
Tested this out and everything is WAD 🚢

joe-yeager (approved):

PR #1357, opened 2025-01-27
chore: Pin versions on all prod deps

joe-yeager:
## Description and Context

Pin all production dependencies versions


brandenrodgers (approved):

.yarnrc
```diff
@@ -0,0 +1 @@
+--install.no-lockfile true
```
joe-yeager, line 1: This will prevent the creation/use of a lockfile

brandenrodgers (dismissed):

PR #1363, opened 2025-01-31
POC: Remove i18n files

joe-yeager:
## Description and Context

Pros:
- Removes the friction from having to edit large JSON/YAML files
- Code completion
- Type checking for interpolated strings
- Unit testable since the lookup no longer requires 
- Remove custom interpolation language, uses direct JavaScript
- Easy to tell if a key is not used
- Easy to tell exactly where a key is used
- The Co-Pilot suggestions were actually really useful when using the consts 

Cons:
- If we do eventually leverage i18n, we will have to walk back the pattern. 
- Porting to the new format will take time

### DIFF
--- a/lang/constants.ts
+++ b/lang/constants.ts
@@ -17,7 +17,7 @@ export const doctor = {
   diagnosticsComplete: `Diagnostics complete`,
   accountChecks: {
     active: `Default account active`,
-    inactive: 'Default account isn`t active',
+    inactive: `Default account isn't active`,
     inactiveSecondary: (command: string) =>
       `Run ${command} to remove inactive accounts from your CLI config`,
     unableToDetermine: `Unable to determine if the portal is active`,
--- a/lang/constants.ts
+++ b/lang/constants.ts
@@ -1,0 +1,104 @@
+import { bold } from 'chalk';
+export const doctor = {
+  command: {
+    describe:
+      'Retrieve diagnostic information about your local HubSpot configurations.',
+    options: {
+      outputDir: 'Directory to save a detailed diagnosis JSON file in',
+    },
+    errors: {
+      generatingDiagnosis: 'Error generating diagnosis',
+      unableToWriteOutputFile: (file: string, errorMessage: unknown) =>
+        `Unable to write output to ${bold(file)}, ${errorMessage}`,
+    },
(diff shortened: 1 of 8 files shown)


lang/constants.ts
```diff
@@ -0,0 +1,104 @@
...
+  command: {
+    describe:
+      'Retrieve diagnostic information about your local HubSpot configurations.',
+    options: {
+      outputDir: 'Directory to save a detailed diagnosis JSON file in',
+    },
+    errors: {
+      generatingDiagnosis: 'Error generating diagnosis',
+      unableToWriteOutputFile: (file: string, errorMessage: unknown) =>
+        `Unable to write output to ${bold(file)}, ${errorMessage}`,
```
joe-yeager, line 12: Instead of using the bespoke markdown language we could directly use JS functions and directly invoke whatever chalk methods we wish. We also get the benefits of type checking for these string generating methods

kemmerle: To me, this is one of the most convincing arguments for moving toward a constant file with JS functions to implement interpolations. The functions are more readable to me than our bespoke markdown language (love that phrasing).

kemmerle:
Joe, did you use a script to generate a JS file with all the constants in place? If so, how did that process work? Did you just plug in the en.lyaml entries for the `doctor` command? How scalable do you think it is for the rest of the `en.lyaml` file? 

So to summarize:

1) No other open-source HubSpot repo implements `i18n` with an `en.lyaml` file. 
2) The `lyaml` file is not human-friendly and prone to error.
3) We can script to convert `lyaml` entries to JS constants, so we can take advantage of JavaScript functions and TypeScript. 
4) At the end, the constants files would be human-friendly and less prone to error. 

@joe-yeager, you've convinced me. Let's discuss further in one of next week's syncs. Maybe we can start work on this after we complete the TypeScript migration?

joe-yeager:
> Joe, did you use a script to generate a JS file with all the constants in place? If so, how did that process work? Did you just plug in the en.lyaml entries for the `doctor` command? How scalable do you think it is for the rest of the `en.lyaml` file?

@kemmerle - I didn't script it on this first pass since I was just targeting a single command and not the whole file.  I'm fairly confident we could write a script that loads the file and dumps the whole JS object into a file and from there we can split it up as we wish. From there we can manually convert the interpolated strings to JS functions

PR #1364, opened 2025-01-31
Adding tests for project/structure

brandenrodgers:
## Description and Context

Adding unit tests for the `lib/project/structure` utils

### DIFF
--- a/lib/projects/__tests__/structure.test.ts
+++ b/lib/projects/__tests__/structure.test.ts
@@ -60,10 +60,6 @@ describe('lib/projects/structure', () => {
   });
 
   describe('loadConfigFile()', () => {
-    afterEach(() => {
-      jest.resetAllMocks();
-    });
-
     it('returns parsed JSON when file exists', () => {
       const mockConfig = { name: 'test-app' };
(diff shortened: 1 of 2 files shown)


camden11 (approved):

lib/projects/__tests__/structure.test.ts
```diff
@@ -0,0 +1,274 @@
...
+
+    it('returns null for unknown config file', () => {
+      expect(getComponentTypeFromConfigFile('unknown.json')).toBeNull();
+    });
+  });
+
+  describe('loadConfigFile()', () => {
+    afterEach(() => {
+      jest.resetAllMocks();
+    });
```
joe-yeager, line 65: Nit: This should be globally set to reset after every test in the jest config file.

joe-yeager (approved):
LGTM, just a nit about resetting mocks

lib/projects/__tests__/structure.test.ts
brandenrodgers: Cool, I didn't know we had that. I can probably remove other instances of this that I've added in other PR's

PR #1365, opened 2025-01-31
Feature: Implement `hs account auth` command

kemmerle:
## Description and Context
In this PR, I'm implementing the `hs account auth` command. This command is meant to replace `hs init` and `hs auth` as part of our centralized configuration file EPIC. With this single command, the customer can both create a configuration file and later add accounts to it. 

## Questions

1. **[RESOLVED]** Currently, when I run `hs account auth` to create a config (ie. running the command for a first time), I receive the warning `A hubspot.config.yml file could not be found. To create a new config file, use the "hs init" command.` 

Screenshot 2025-02-04 at 11 51 55 AM

This warning originates from the `injectAccountMiddleware` function in `bin/cli.js`. We need to load the config file before we can retrieve the account ID and inject it before every command. Stack trace below:

Screenshot 2025-02-04 at 11 52 34 AM

I'm tempted to remove this warning entirely, since it originates from the deprecated config ([here](https://github.com/HubSpot/hubspot-local-dev-lib/blob/19a88a63d4c284f3c6115b57fa6b8bd0acb21b67/config/config_DEPRECATED.ts#L282)). Would it be better to suppress it from `injectAccountMiddleware`? Or is there another solution I'm not seeing?

## TODO

- [x] Finish implementing the `auth` side of the command
- [x] Address questions above and in comments
- [x] Get feedback on copy from Jono 

## Who to Notify


commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (doesConfigExist) {
+      updateAccountConfig({
+        ...updatedConfig,
+        environment: updatedConfig.env,
+        tokenInfo: updatedConfig.auth!.tokenInfo,
+      });
+      writeConfig();
+    }
+
+    return updatedConfig as CLIAccount_NEW;
```
kemmerle, line 77: I'm casting `updatedConfig` as `CLIAccount_NEW` here, because otherwise, I'd receive a TypeScript error that `accountId` doesn't exist on type `CLIAccount`. This is because `CLIAccount` can be either `CLIAccount_NEW` or `CLIAccount_DEPRECATED`, and `accountId` doesn't exist on `CLIAccount_DEPRECATED`. 

Is this an acceptable solution, or am I missing anything? I'm assuming we don't want to make any major changes to types for the deprecated config.

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+export const command = 'auth';
+
+type AccountInfoArgs = CommonArgs & ConfigArgs;
+
+export async function handler(
+  args: ArgumentsCamelCase<AccountInfoArgs>
+): Promise<void> {
+  const { providedAccountId, disableTracking } = args;
+
+  if (!disableTracking) {
```
brandenrodgers, line 94: You shouldn't need this check for if tracking is disabled. trackCommandUsage calls isTrackingAllowed

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+      updateAccountConfig({
+        ...updatedConfig,
+        environment: updatedConfig.env,
+        tokenInfo: updatedConfig.auth!.tokenInfo,
+      });
+      writeConfig();
+    }
+
+    return updatedConfig as CLIAccount_NEW;
+  } catch (e) {
```
brandenrodgers, line 78: Does this need its own try/catch if `createOrUpdateConfig` is already wrapped in a try/catch in the handler? Seems like this would prevent the call in the handler from throwing, which would call this `logError()` and then also the `failedToUpdateConfig` statement. We might be better off just logging a single error, and then exposing the caught error as a debug log.

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+    return null;
+  }
+}
+
+export const describe = null; // i18n(`${i18nKey}.describe`);
+export const command = 'auth';
+
+type AccountInfoArgs = CommonArgs & ConfigArgs;
+
+export async function handler(
```
brandenrodgers, line 89: We'll need to include the "track auth action" behavior in here so we can identify what percentage of users are successfully making it though the flow

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+        })
+      );
+      logger.success(
+        i18n(`${i18nKey}.success.configFileUpdated`, {
+          account: name || accountId,
+        })
+      );
+    }
+
+    if (configExists) {
```
brandenrodgers, line 137: nit, but this could be an `else` block

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    const updatedConfig = await updateConfigWithAccessToken(
+      token,
+      personalAccessKey,
+      env,
+      name,
+      !doesConfigExist
+    );
+
+    if (!updatedConfig) return null;
```
brandenrodgers, line 62: Is it possible for this line to get hit? iirc `updateConfigWithAccessToken` will throw if it encounters an error

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    if (doesConfigExist && !updatedConfig.name) {
```
brandenrodgers, line 64: Is this line setting a name on an account if it already existed in the config, but didn't have a name yet?

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+      updatedConfig.name = (await cliAccountNamePrompt(defaultName)).name;
+    }
+
+    if (doesConfigExist) {
```
brandenrodgers, line 68: Could you explain why we only want to update the config and write if `doesConfigExist` is true?

commands/account/auth.ts
```diff
@@ -0,0 +1,183 @@
...
+      hidden: true,
+      default: false,
+    },
+  });
+
+  addConfigOptions(yargs);
+  addTestingOptions(yargs);
+  addGlobalOptions(yargs);
+
+  return yargs as Argv<AccountInfoArgs>;
```
brandenrodgers, line 182: Looks like a small copy/paste miss re: `AccountInfoArgs`

lib/prompts/accountNamePrompt.ts
```diff
@@ -12,7 +12,7 @@ type AccountNamePromptResponse = {
 };
 
 export function getCliAccountNamePromptConfig(
-  defaultName?: string
+  defaultName?: string | null
```
brandenrodgers, line 15: Would you still need this change if you defaulted to `undefined` [here](https://github.com/HubSpot/hubspot-cli/pull/1365/files#diff-f0f185d80a68501d80d813e663386e095d242d8b6bc4836d250897471ef14735R48)?

brandenrodgers:
> I'm tempted to remove this warning entirely, since it originates from the deprecated config ([here](https://github.com/HubSpot/hubspot-local-dev-lib/blob/19a88a63d4c284f3c6115b57fa6b8bd0acb21b67/config/config_DEPRECATED.ts#L282)). Would it be better to suppress it from injectAccountMiddleware? Or is there another solution I'm not seeing?

I'm curious, shouldn't `injectAccountMiddleware` also be failing when users run `hs init`? Why doesn't it fail for that command?

commands/account/auth.ts
kemmerle: I meant to ask about this--do we want to change the track auth action behavior? Currently, we track authType, which includes `oauth`.

commands/account/auth.ts
kemmerle: `updateConfigWithAccessToken` logs errors but does not throw them, so this can be hit.

commands/account/auth.ts
kemmerle: Yes. Basically, there is a complicated set of scenarios when users update the config file:

1) If the user is creating the config for the first time, we want to prompt them for an account name first thing.

2) If the user is re-authorizing an account that is already in the config, we do not want to prompt them for the account name.

3) If the user is authorizing an account that is NOT already in the config, we do want to prompt them for the account name. 

This conditional on line 64 covers scenarios 2 and 3. I'll see whether I can't simplify this and combine the conditional on line 68 with this one.

commands/account/auth.ts
kemmerle: Yes. Basically, there is a complicated set of scenarios when users update the config file:

1. If the user is creating the config for the first time, we want to prompt them for an account name first thing.

2. If the user is re-authorizing an account that is already in the config, we do not want to prompt them for the account name.

3. If the user is authorizing an account that is NOT already in the config, we do want to prompt them for the account name.

I am combining this conditional with the one above on line 64 to make it clear that we only want to update the config if we are naming an account that is not already in the config. So the code will look like this:

```
    if (doesConfigExist && !updatedConfig.name) {
      updatedConfig.name = (await cliAccountNamePrompt(defaultName)).name;
      updateAccountConfig({
        ...updatedConfig,
      });
      writeConfig();
    }
```

kemmerle:
> I'm curious, shouldn't injectAccountMiddleware also be failing when users run hs init? Why doesn't it fail for that command?

I skipped validation for this command, as we did with `hs init` and `hs auth`.

commands/account/auth.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/2acdfb8eb4d268b18e4e76619c9101f2aa19d8bc

commands/account/auth.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/2acdfb8eb4d268b18e4e76619c9101f2aa19d8bc

commands/account/auth.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/2acdfb8eb4d268b18e4e76619c9101f2aa19d8bc

commands/account/auth.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/2acdfb8eb4d268b18e4e76619c9101f2aa19d8bc

lib/prompts/accountNamePrompt.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/2acdfb8eb4d268b18e4e76619c9101f2aa19d8bc

commands/account/auth.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/5529392be0664cf948ca893bd6c95af2d373d9c1

commands/account/auth.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/5529392be0664cf948ca893bd6c95af2d373d9c1

commands/account/auth.ts
brandenrodgers: iirc we made a util to handle this situation. It's `getAccountIdentifier`, which can be used to safely read the id from an account regardless of the config type.

kemmerle: I'm a little unclear about how I would use `getAccountIdentifier` in this case, since we're returning all the account info and not just the ID. I don't think reassigning `updatedConfig.accountId` using `getAccountIdentifier` would help, correct? Because the issue is the property name.

brandenrodgers: Oh, I was assuming the error was coming from [here](https://github.com/HubSpot/hubspot-cli/pull/1365/files#diff-f0f185d80a68501d80d813e663386e095d242d8b6bc4836d250897471ef14735R139) where you're accessing `accountId` from the config object.

Is the error coming from the return type of this function instead? If so, could it be fixed by switching the return type to `CLIAccount` instead of `CLIAccount_NEW`?

kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1365/commits/c7dcc882173084d4594bea4cb6294fc4ab993d14

commands/account/auth.ts
```diff
@@ -102,9 +103,14 @@ type AccountAuthArgs = CommonArgs &
 export async function handler(
   args: ArgumentsCamelCase<AccountAuthArgs>
 ): Promise<void> {
-  const { providedAccountId, disableTracking } = args;
+  const { providedAccountId, disableTracking, config } = args;
   const authType = PERSONAL_ACCESS_KEY_AUTH_METHOD.value;
 
+  if (config) {
```
brandenrodgers, line 109: I should have been a little clearer when I mentioned the existing config thing. I was referring to the case where the user already has a `hubspot.config.yml` file, and then tries to run `hs account auth`. In that case we should log an error or some sort of message to explain what they can do to port their existing config to the new format.

This `--config` case is still useful to catch, but I think you can solve this one by just removing `addConfigOptions()` in the builder. Then this command won't support `--config` at all.

commands/account/auth.ts
```diff
@@ -0,0 +1,212 @@
...
+  } catch (e) {
+    if (!disableTracking) {
+      await trackAuthAction('account-auth', authType, TRACKING_STATUS.ERROR);
+    }
+    debugError(e);
+    return null;
+  }
+}
+
+export const describe = ''; // i18n(`${i18nKey}.describe`);
```
brandenrodgers, line 92: This needs to be `undefined` otherwise this command will show up when a user runs `hs account --help`

brandenrodgers:
I saw this error when running `hs account auth`. I think it happened because I didn't have a `~/.hscli/config.yml` file yet. It didn't show up when I ran the command to authenticate my second account.
image

kemmerle:
@brandenrodgers, I wrote this up in my description but am going to paste here for convenience:

1. Currently, when I run `hs account auth` to create a config (ie. running the command for a first time), I receive the warning `A hubspot.config.yml file could not be found. To create a new config file, use the "hs init" command.` 

Screenshot 2025-02-04 at 11 51 55 AM

This warning originates from the `injectAccountMiddleware` function in `bin/cli.js`. We need to load the config file before we can retrieve the account ID and inject it before every command. Stack trace below:

Screenshot 2025-02-04 at 11 52 34 AM

I'm tempted to remove this warning entirely, since it originates from the deprecated config ([here](https://github.com/HubSpot/hubspot-local-dev-lib/blob/19a88a63d4c284f3c6115b57fa6b8bd0acb21b67/config/config_DEPRECATED.ts#L282)). Would it be better to suppress it from `injectAccountMiddleware`? Or is there another solution I'm not seeing?

brandenrodgers (approved):
LGTM! 🎉 

This isn't a blocker, but in a follow up PR we could also make it so that running `hs auth` or `hs init` while a  `~/.hscli/config.yml` file exists will surface some sort of warning or error.

PR #1366, opened 2025-01-31
Unit tests for sandboxes.ts

brandenrodgers:
## Description and Context

Adding unit tests for lib/sandboxes.ts

### DIFF
--- a/lib/__tests__/sandboxes.test.ts
+++ b/lib/__tests__/sandboxes.test.ts
@@ -52,10 +52,6 @@ describe('lib/sandboxes', () => {
       env: 'qa' as Environment,
     };
 
-    beforeEach(() => {
-      jest.resetAllMocks();
-    });
-
     it('returns true when sandbox of specified type exists', () => {
       mockedGetAccountId.mockReturnValue(mockParentAccount.portalId);
       mockedGetConfigAccounts.mockReturnValue([
@@ -99,10 +95,6 @@ describe('lib/sandboxes', () => {
       portalId: 456,
     };
 
-    beforeEach(() => {
-      jest.resetAllMocks();
-    });
-
     it('returns available sync types when fetch is successful', async () => {
       const mockSyncTypes = [{ name: 'type1' }, { name: 'type2' }];
       mockedGetAccountId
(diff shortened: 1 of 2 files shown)


lib/__tests__/sandboxes.test.ts
```diff
@@ -0,0 +1,263 @@
...
+        },
+      });
+
+      await expect(
+        validateSandboxUsageLimits(
+          mockAccount,
+          HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX,
+          'qa'
+        )
+      ).rejects.toThrow();
```
joe-yeager, line 176: I think in all the places we are asserting that a function should throw via `toThrow`, we should also validate the contents of the error as well.  This test would pass if there was a runtime exception

lib/__tests__/sandboxes.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    });
+  });
+
+  describe('handleSandboxCreateError()', () => {
+    const mockEnv = 'qa' as Environment;
+    const mockName = 'Test Sandbox';
+    const mockAccountId = 123;
+
+    beforeEach(() => {
+      jest.resetAllMocks();
```
joe-yeager, line 186: We should have resetting the mocks globally enabled in the jest config, we should be able to remove this

lib/__tests__/sandboxes.test.ts
```diff
@@ -0,0 +1,263 @@
...
+              message: 'Missing scopes error',
+              category: 'MISSING_SCOPES',
+            },
+          },
+        },
+      });
+
+      expect(() =>
+        handleSandboxCreateError(error, mockEnv, mockName, mockAccountId)
+      ).toThrow();
```
joe-yeager, line 205: If we change this to `toThrow(error)` we get the benefit of making sure the correct error was thrown.

lib/__tests__/sandboxes.test.ts
brandenrodgers: You caught me being lazy 😅 updated in [1380930](https://github.com/HubSpot/hubspot-cli/pull/1366/commits/1380930867893625d088c394b4c0c43354980525)!

lib/__tests__/sandboxes.test.ts
brandenrodgers: Removed in [1380930](https://github.com/HubSpot/hubspot-cli/pull/1366/commits/1380930867893625d088c394b4c0c43354980525) 👍

joe-yeager (approved):

PR #1367, opened 2025-02-04
feat: add scope authorization check to hs doctor

b-ash:
## Description and Context

Adds scope authorization checking to `hs doctor`. The goal is to alert the user if there are potential issues we detect with their local configuration. In this situation, we're highlighting when the personal access key being used for auth has fewer scopes than the current user has access to. If so, we'll detail that in the doctor response.

I'll add some tests before merging but wanted to get some eyes on the shape of things ahead of that.

Depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/232

### DIFF
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -1520,7 +1520,7 @@ en:
           incomplete:
             one: "Personal access key is valid. {{ count }} scope available, but not included in your key."
             other: "Personal access key is valid. {{ count }} scopes available, but not included in your key."
-          incompleteSecondary: "To add the available scopes, run `hs auth` and re-authenticate your account with a new key that has those scopes. {{ link }}"
+          incompleteSecondary: "To add the available scopes, run {{ command}} and re-authenticate your account with a new key that has those scopes. {{ link }}"
           invalid: "Personal access key is invalid"
           invalidSecondary: "To get a new key, run {{ command }}, deactivate your access key, and generate a new one. Then use that new key to authenticate your account."
           valid: "Personal Access Key is valid. {{ link }}"
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -1517,7 +1517,7 @@ en:
         inactiveSecondary: "Run {{ command }} to remove inactive accounts from your CLI config"
         unableToDetermine: "Unable to determine if the portal is active"
         pak:
-          incomplete: "Personal access key is valid. There are more scopes available to your user, but not included in your key."
+          incomplete: "Personal access key is valid, but there are more scopes available to your user that are not included in your key."
           incompleteSecondary: "To add the available scopes, run {{ command }} and re-authenticate your account with a new key that has those scopes. Visit hubspot.com to view selected and available scopes for your personal access key. {{ link }}"
           invalid: "Personal access key is invalid"
           invalidSecondary: "To get a new key, run {{ command }}, deactivate your access key, and generate a new one. Then use that new key to authenticate your account."
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -1517,6 +1517,10 @@ en:
         inactiveSecondary: "Run {{ command }} to remove inactive accounts from your CLI config"
         unableToDetermine: "Unable to determine if the portal is active"
         pak:
+          incomplete:
+            one: "Personal access key is valid. {{ count }} scope available, but not included in your key."
+            other: "Personal access key is valid. {{ count }} scopes available, but not included in your key."
+          incompleteSecondary: "To add the available scopes, run `hs auth` and re-authenticate your account with a new key that has those scopes. {{ link }}"
           invalid: "Personal access key is invalid"
           invalidSecondary: "To get a new key, run {{ command }}, deactivate your access key, and generate a new one. Then use that new key to authenticate your account."
           valid: "Personal Access Key is valid. {{ link }}"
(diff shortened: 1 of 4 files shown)


lang/en.lyaml
```diff
@@ -1517,6 +1517,10 @@ en:
         inactiveSecondary: "Run {{ command }} to remove inactive accounts from your CLI config"
         unableToDetermine: "Unable to determine if the portal is active"
         pak:
+          incomplete:
+            one: "Personal access key is valid. {{ count }} scope available, but not included in your key."
+            other: "Personal access key is valid. {{ count }} scopes available, but not included in your key."
+          incompleteSecondary: "To add the available scopes, run `hs auth` and re-authenticate your account with a new key that has those scopes. {{ link }}"
```
kemmerle, line 1523: We have a helper function called `uiCommandReference` that will format `hs auth` correctly.

kemmerle: I'd also add some messaging here that the personal access key UI will have a full list of missing scopes, if we're not going to list them all out in the CLI.

lib/doctor/Doctor.ts
```diff
@@ -116,21 +120,51 @@ export class Doctor {
...
     try {
       await accessTokenForPersonalAccessKey(this.accountId!, true);
+      const pakScopes = new Set(await scopesOnAccessToken(this.accountId!));
+      const missingScopes = (
+        await authorizedScopesForPortalAndUser(this.accountId!)
+      )
+        .filter(
+          data => data.userAuthorized && !pakScopes.has(data.scopeGroup.name)
+        )
+        .map(data => data.scopeGroup.name);
```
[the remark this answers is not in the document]
b-ash: Aha, totally. I had this from the previous iteration where it was spitting out the scopes so the names were used.

lang/en.lyaml
b-ash: Not a bad idea, I added some extra copy. lmk what you think

lib/doctor/Doctor.ts
```diff
@@ -116,21 +120,45 @@ export class Doctor {
...
-          link: uiLink(
-            i18n(`${localI18nKey}.pak.viewScopes`),
-            `${getHubSpotWebsiteOrigin(
-              this.diagnosticInfoBuilder?.env || 'PROD'
-            )}/personal-access-key/${this.diagnosticInfo?.account.accountId}`
+
+      const linkToPakUI = uiLink(
+        i18n(`${localI18nKey}.pak.viewScopes`),
+        `${getHubSpotWebsiteOrigin(
+          this.diagnosticInfoBuilder?.env || 'PROD'
```
brandenrodgers, line 138: This can be simplified to `getHubSpotWebsiteOrigin(this.diagnosticInfoBuilder?.env)` b/c the util will default to the prod domain if no env is passed in. [util](https://github.com/HubSpot/hubspot-local-dev-lib/blob/main/lib/urls.ts#L8)

b-ash: is that true? the [util](https://github.com/HubSpot/hubspot-local-dev-lib/blob/e67718e153e33fb1929fecfe00e7892914c45277/lib/urls.ts#L8) it's calling requires it to be defined, so we'd either need to cast it with `!` or do that fallback. any preference?

joe-yeager (dismissed):

lib/doctor/Doctor.ts
brandenrodgers: Ah shoot, you're right. This looks like a typing mistake over in `getHubSpotWebsiteOrigin`. The env arg should be typed more similarly to the way that our CLI_Account.env is typed, which allows undefined b/c we always fall back to prod behavior.

You can keep this as is 👍

lang/en.lyaml
```diff
@@ -1517,6 +1517,10 @@ en:
         inactiveSecondary: "Run {{ command }} to remove inactive accounts from your CLI config"
         unableToDetermine: "Unable to determine if the portal is active"
         pak:
+          incomplete:
+            one: "Personal access key is valid. {{ count }} scope available, but not included in your key."
```
brandenrodgers, line 1521: What are your thoughts on something like this for copy?
"Personal access key is valid, but you have 11 available scopes that are not included in your key"

This might just be me, but "N scopes available, but not included in your key." reads a little awkwardly. I initially read it as "you have 11 scopes total"

lang/en.lyaml
```diff
@@ -1517,6 +1517,10 @@ en:
...
+            other: "Personal access key is valid. {{ count }} scopes available, but not included in your key."
+          incompleteSecondary: "To add the available scopes, run {{ command}} and re-authenticate your account with a new key that has those scopes. Visit hubspot.com to view selected and available scopes for your personal access key. {{ link }}"
```
brandenrodgers, line 1523: We refer to it as " visit HubSpot" instead of "visit hubspot.com" (small change but worth the consistency)

b-ash: Understood (I thought the hubspot.com was a little wacky too). The existing copy for when you need to paste in your key is `"Open hubspot.com to copy your personal access key?"` - should we change that also? My thinking is yes, and there are a few others as well. I followed convention with mine 🤷‍♂️

lang/en.lyaml
b-ash: Agreed, I reworded after removing the count and it's a lot smoother IMO

lang/en.lyaml
brandenrodgers: Good call, I didn't realize we had those inconsistencies tbh. But yeah it would be great to update that too if you don't mind! (not a blocker though)

brandenrodgers (dismissed):
Tested locally and this looks great! 🔥

PR #1368, opened 2025-02-04
Remove unsupported types from sandbox sync

adamawang:
## Description and Context

Supporting PR: https://github.com/HubSpot/hubspot-local-dev-lib/pull/233

Did some testing and realized the new endpoint (in LDL) is returning a few unsupported types for our v1 sandbox sync. Since we don't support v2 sandboxes in the CLI yet, we have to filter these options out so the sandbox sync succeeds. Without the filter, the sandbox sync hangs indefinitely due to unsupported sync type. 

I ran this locally with `yarn hs sandbox create` and with LDL running locally at the same time (hoping that is still the way to test!)

## Screenshots

n/a

### DIFF
--- a/lib/sandboxes.ts
+++ b/lib/sandboxes.ts
@@ -35,6 +35,13 @@ export const SANDBOX_API_TYPE_MAP = {
   [HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX]: 2,
 } as const;
 
+export const UNSUPPORTED_V1_SYNC_TYPES = [
+  'hub-trials',
+  'gates',
+  'lead-settings',
+  'object-schemas-full', // TODO: Remove object-schemas-full once we've migrated sandboxes to v2
+];
+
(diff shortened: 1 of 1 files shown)


brandenrodgers:
> Without the filter, the sandbox sync hangs indefinitely due to unsupported sync type.

Does this make the LDL change a breaking change?

adamawang:
Hmm I suppose so yeah, since merging LDL alone would break sandbox creation but only in the UI. CLI will complete without any errors

For some more context: V2 sandboxes required a necessary move over to PortabilitySync but since we are deprecating the old SandboxesSync, we had to move v1 sandboxes over in sandboxes-ui as well with some special handling (similar to what we're doing in this PR)

adamawang:
Just met with Branden, going to hold off on any deploys until we figure out a game plan. Will update here once I get info

adamawang:
Closing this out for now, will revisit in Q2 - thanks all!

PR #1370, opened 2025-02-05
feat: Remove option, upload to v3 based on platform version

joe-yeager:
## Description and Context

Remove the `--useV3` flag and route to v3 of the projects APIs based on platform version instead


brandenrodgers (approved):
Thanks for getting this in quickly 🙏

PR #1372, opened 2025-02-06
Chore: Convert `hubdb` commands to TypeScript

kemmerle:
## Description and Context
In this PR, I've converted the `hubdb` commands to TypeScript and added basic tests for each command. This includes:

#### Commands
- [x] `hubdb/clear.ts`
- [x] `hubdb/create.ts`
- [x] `hubdb/delete.ts`
- [x] `hubdb/fetch.ts`

#### Bucket
- [x] `hubdb.ts`

## TODO

- [ ] Address feedback 

## Who to Notify


commands/hubdb/delete.ts
```diff
@@ -1,37 +1,41 @@
...
 
-exports.command = 'delete [table-id]';
-exports.describe = i18n(`${i18nKey}.describe`);
+export const command = 'delete [table-id]';
+export const describe = i18n(`${i18nKey}.describe`);
 
-exports.handler = async options => {
-  const { force, derivedAccountId } = options;
+type HubdbDeleteArgs = CommonArgs &
+  ConfigArgs & { tableId?: number; dest?: string };
```
brandenrodgers, line 23: I don't think dest is used anywhere in this command, maybe just snuck its way in from copy/paste?

commands/hubdb/fetch.ts
```diff
@@ -1,37 +1,40 @@
...
 
 const i18nKey = 'commands.hubdb.subcommands.fetch';
 
-exports.command = 'fetch [table-id] [dest]';
-exports.describe = i18n(`${i18nKey}.describe`);
+export const command = 'fetch [table-id] [dest]';
+export const describe = i18n(`${i18nKey}.describe`);
+
+type HubdbFetchArgs = CommonArgs &
+  ConfigArgs & { tableId?: number; dest?: string };
```
brandenrodgers, line 21: Do we have a system to determine which args are optional and which ones are required? It looks like we have a mix. I think I had made them required when I did the types for the account commands, but I think it actually makes more sense to make them optional as long as we don't define a default value for the option down in the builder. Maybe I'll fix the account commands to follow this pattern.

It looks like the hubdb/create args could use `path?:` too

commands/hubdb/clear.ts
```diff
@@ -1,36 +1,39 @@
...
 
-exports.command = 'clear [table-id]';
-exports.describe = i18n(`${i18nKey}.describe`);
+export const command = 'clear [table-id]';
+export const describe = i18n(`${i18nKey}.describe`);
 
-exports.handler = async options => {
-  const { derivedAccountId } = options;
+type HubdbClearArgs = CommonArgs &
+  ConfigArgs & { tableId?: number; dest?: string };
```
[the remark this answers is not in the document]
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1372/commits/a33d8a0132a153f1dab2785447a92e7c312e4375

commands/hubdb/fetch.ts
kemmerle: Addressed in https://github.com/HubSpot/hubspot-cli/pull/1372/commits/a33d8a0132a153f1dab2785447a92e7c312e4375

brandenrodgers (approved):

PR #1377, opened 2025-02-10
Unit tests for usage tracking and process utils

brandenrodgers:
## Description and Context

Adding some more unit test PR's. I'm trying to do this in small batches to keep these PR's relatively small so they're easier to review

### DIFF
--- a/lib/__tests__/process.test.ts
+++ b/lib/__tests__/process.test.ts
@@ -1,6 +1,6 @@
 import readline from 'readline';
 import { logger, setLogLevel, LOG_LEVEL } from '@hubspot/local-dev-lib/logger';
-import { handleExit, handleKeypress } from '../process';
+import { handleExit, handleKeypress, TERMINATION_SIGNALS } from '../process';
 
 jest.mock('readline');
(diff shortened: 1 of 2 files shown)


lib/__tests__/process.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      handleExit(mockCallback);
+
+      // Get the callback function passed to process.on for SIGHUP
+      const sighupCallback = processOnSpy.mock.calls.find(
+        call => call[0] === 'SIGHUP'
+      )?.[1];
+
+      if (!sighupCallback) {
+        throw new Error('SIGHUP callback not found');
+      }
```
joe-yeager, line 46: Nit, instead of throwing we should just `expect(sighupCallback).toBeDefined()`

brandenrodgers: done in [50ae059](https://github.com/HubSpot/hubspot-cli/pull/1377/commits/50ae059b50dd10cdfbd03e26b7f9120273eb7ebd) 👍

joe-yeager (approved):

PR #1379, opened 2025-02-11
Add await to create test

brandenrodgers:
## Description and Context

This is async so we need to add an await here. This wasn't breaking the tests, but I noticed the warning when I was looking into a failed action execution


kemmerle (approved):
This just broke a test for me lol, good catch!

PR #1381, opened 2025-02-14
Chore: Change the min platform version for the API revamp

joe-yeager:
## Description and Context

Change the minimum required version from `2025.1` to either `2025.2` or `UNSTABLE`

### DIFF
--- a/lib/projects/buildAndDeploy.ts
+++ b/lib/projects/buildAndDeploy.ts
@@ -49,7 +49,7 @@ export function useV3Api(platformVersion?: string | null) {
   if (!platformVersion || typeof platformVersion !== 'string') {
     return false;
   }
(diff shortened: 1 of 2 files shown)


lib/projects/buildAndDeploy.ts
```diff
@@ -49,8 +49,11 @@ export function useV3Api(platformVersion?: string | null) {
   if (!platformVersion || typeof platformVersion !== 'string') {
     return false;
   }
+  if (platformVersion === 'UNSTABLE') {
```
kporter999, line 52: I think we are planning for this to be lowercase

kporter999 (approved):

kemmerle (approved):
Thanks for updating the tests, as well!

PR #1383, opened 2025-02-14
chore: Convert `sandbox` commands to TypeScript

kemmerle:
## Description and Context
In this PR, I've converted the `sandbox` commands to TypeScript and added basic tests. I also corrected the names on the `hubdb` and `secret` tests which I wrote in previous PRs. 

#### Commands
- [x] `sandbox/create.ts`
- [x] `sandbox/delete.ts`

#### Bucket
- [x] `sandbox.ts`

## TODO

- [x] Address feedback

## Who to Notify

### DIFF
--- a/commands/sandbox/create.ts
+++ b/commands/sandbox/create.ts
@@ -177,7 +177,7 @@ export async function handler(
     );
 
     const sandboxAccountConfig = getAccountConfig(result.sandbox.sandboxHubId);
-    // Check if account config exists
+    // Check if sandbox account config exists
     if (!sandboxAccountConfig) {
       logger.error(
         i18n(`${i18nKey}.failure.noSandboxAccountConfig`, {
--- a/commands/sandbox/create.ts
+++ b/commands/sandbox/create.ts
@@ -1,49 +1,84 @@
-// @ts-nocheck
-const {
+import { Argv, ArgumentsCamelCase } from 'yargs';
+import { getAccountConfig, getEnv } from '@hubspot/local-dev-lib/config';
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { isMissingScopeError } from '@hubspot/local-dev-lib/errors/index';
+import { getHubSpotWebsiteOrigin } from '@hubspot/local-dev-lib/urls';
+import {
+  HUBSPOT_ACCOUNT_TYPES,
+  HUBSPOT_ACCOUNT_TYPE_STRINGS,
+} from '@hubspot/local-dev-lib/constants/config';
+import { getValidEnv } from '@hubspot/local-dev-lib/environment';
+
+import {
   addAccountOptions,
   addConfigOptions,
   addUseEnvironmentOptions,
   addTestingOptions,
-} = require('../../lib/commonOpts');
-const { i18n } = require('../../lib/lang');
-const { EXIT_CODES } = require('../../lib/enums/exitCodes');
-const { getAccountConfig, getEnv } = require('@hubspot/local-dev-lib/config');
-const { uiFeatureHighlight, uiBetaTag } = require('../../lib/ui');
-const {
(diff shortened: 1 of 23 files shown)


commands/sandbox/create.ts
```diff
@@ -56,7 +89,7 @@ exports.handler = async options => {
           HUBSPOT_ACCOUNT_TYPE_STRINGS[
             HUBSPOT_ACCOUNT_TYPES[accountConfig.accountType]
           ],
-        accountName: accountConfig.name,
+        accountName: getAccountIdentifier(accountConfig) || '',
```
camden11, line 92: I think you probably want to keep this as name - getAccountIdentifier returns the accountId or portalId

commands/sandbox/create.ts
```diff
@@ -65,17 +98,16 @@ exports.handler = async options => {
...
     } else {
       logger.error(i18n(`${i18nKey}.failure.optionMissing.type`));
       process.exit(EXIT_CODES.ERROR);
     }
   }
-  const sandboxType = type
-    ? SANDBOX_TYPE_MAP[type.toLowerCase()]
-    : typePrompt.type;
+
+  const sandboxType = type ? SANDBOX_TYPE_MAP[type] : typePrompt!.type;
```
camden11, line 110: Noticed that you removed the `toLowerCase()` here but not above - I think this could cause a bug if type existed but wasn't in `SANDBOX_TYPE_MAP`

lib/prompts/sandboxesPrompt.ts
```diff
@@ -70,13 +71,19 @@ export async function sandboxTypePrompt(): Promise<SandboxTypePromptResponse> {
...
   promptParentAccount = false
-): Promise<DeleteSandboxPromptResponse> | void {
+): Promise<DeleteSandboxPromptResponse> {
   const accountsList = getConfigAccounts();
   const choices = promptParentAccount
     ? mapNonSandboxAccountChoices(accountsList)
     : mapSandboxAccountChoices(accountsList);
   if (!choices.length) {
-    return;
+    return Promise.reject(
```
brandenrodgers, line 80: I like the thought here to handle this case better, but you may be able to do this in a slightly simpler way. Throwing from here means that we'd need to wrap the call to `deleteSandboxPrompt` in a try/catch. It might be easier to return void here and catch this case with an `if ! parentAccountPrompt.account` somewhere around [here](https://github.com/HubSpot/hubspot-cli/pull/1383/files#diff-2e9c34f9182dc290b80774109b85c7614b0d947cf32d90028639bcec586ed947R93).

commands/sandbox/delete.ts
```diff
@@ -71,20 +80,30 @@ exports.handler = async options => {
 
   let parentAccountId;
   const accountsList = getConfigAccounts();
+  if (!accountsList) {
```
brandenrodgers, line 83: If you do `const accountsList = getConfigAccounts() || []` above, the `for (const portal of accountsList)` will skip, causing `parentAccountId` to remain undefined. This would trigger the if check beneath it. Then you'd only need to log the "no parent Id" message in one place.

commands/sandbox/delete.ts
```diff
@@ -195,6 +203,7 @@ export async function handler(
       logger.error(
         i18n(`${i18nKey}.failure.invalidKey`, {
           account: uiAccountDescription(parentAccountId),
+          authCommand: uiCommandReference('hs auth personalaccesskey'),
```
brandenrodgers, line 206: I think this would just be `hs auth` since personalaccesskey is the default (and it gets passed in via option if you want to manually set it)

brandenrodgers (approved):
Just one comment about removing a reference to personal access key in a command reference, but otherwise lgtm 👌

PR #1385, opened 2025-02-19
Updates to project create and add

brandenrodgers:
## Description and Context

This depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/241

This makes a few updates to the project add and create commands. These commands essentially do the same thing (i.e. download code from github), but they were doing it in different ways. Most of the work in here is just restructuring things so that both commands behave the same way. The rest I'll point out inline. The major changes are:
- The project add command will now prevent users from adding a private app to a public app, which isn't allowed
- The project create command won't let you create a project within an existing project. Previously we had a prompt to ask if you want to want to proceed with the create, but there's no scenario where a project nested within another project is valid
- I added some try/catch blocks to a handful of api calls to make things safer. This should hopefully also make debug logs more useful.
- I simplified the `createProjectPrompt` util to be consistent with the way that `projectAddPrompt` works. I moved the template list fetch outside of it, which makes it easier for other commands like cloneApp and migrateApp to use it. It also makes the logic a little easier to follow.

I have an upcoming PR in local-dev-lib that will enable us to hide the console logs coming from the clone repo utils. I noticed a bug in the existing logs where we're not correctly passing in a `type` to the clone util, which causes it to log like this:
image

IMO it'll be better to just ignore the logs completely and log out our own content from the CLI. 

## TODO

- [x] Test to make sure it all works
- [x] Add tests for the new utils

### DIFF
--- a/commands/create/api-sample.ts
+++ b/commands/create/api-sample.ts
@@ -6,11 +6,11 @@ const { confirmPrompt } = require('../../lib/prompts/promptUtils');
 const { logger } = require('@hubspot/local-dev-lib/logger');
 const path = require('path');
 const fs = require('fs-extra');
-const {
-  fetchFileFromRepository,
-  cloneGithubRepo,
-} = require('@hubspot/local-dev-lib/github');
+const { fetchRepoFile } = require('@hubspot/local-dev-lib/api/github');
+const { cloneGithubRepo } = require('@hubspot/local-dev-lib/github');
 const { i18n } = require('../../lib/lang');
+const { debugError } = require('../../lib/errorHandlers');
+const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 
 const i18nKey = 'commands.create.subcommands.apiSample';
 
@@ -39,23 +39,31 @@ module.exports = {
       }
     }
 
-    const samplesConfig = await fetchFileFromRepository(
-      'HubSpot/sample-apps-list',
-      'samples.json',
-      'main'
-    );
+    let samplesConfig;
+    try {
+      const { data } = await fetchRepoFile(
+        'HubSpot/sample-apps-list',
+        'samples.json',
+        'main'
+      );
+      samplesConfig = data;
+    } catch (err) {
+      debugError(err);
+    }
 
     if (!samplesConfig) {
       logger.error(i18n(`${i18nKey}.errors.noSamples`));
-      return;
+      process.exit(EXIT_CODES.ERROR);
     }
-    const { sampleType, sampleLanguage } = await createApiSamplePrompt(
-      samplesConfig
-    );
+
+    const { sampleType, sampleLanguage } =
+      await createApiSamplePrompt(samplesConfig);
+
     if (!sampleType || !sampleLanguage) {
       logger.error(i18n(`${i18nKey}.errors.noSamples`));
-      return;
+      process.exit(EXIT_CODES.ERROR);
     }
+
     logger.info(
       i18n(`${i18nKey}.info.sampleChosen`, {
         sampleType,
--- a/commands/project/add.ts
+++ b/commands/project/add.ts
@@ -73,6 +73,9 @@ exports.handler = async options => {
     }
   } catch (err) {
     debugError(err);
+  }
+
+  if (!latestRepoReleaseTag) {
     logger.error(i18n(`${i18nKey}.error.failedToFetchComponentList`));
     process.exit(EXIT_CODES.ERROR);
   }
@@ -85,29 +88,32 @@ exports.handler = async options => {
     process.exit(EXIT_CODES.ERROR);
   }
 
-  const { componentData, name } = await projectAddPrompt(components, options);
-  const componentName = name || options.name;
+  const projectAddPromptResponse = await projectAddPrompt(components, options);
 
   try {
     const componentPath = path.join(
       projectDir,
       projectConfig.srcDir,
-      componentData.insertPath,
-      componentName
+      projectAddPromptResponse.componentTemplate.insertPath,
+      projectAddPromptResponse.name
     );
 
     await cloneGithubRepo(
       HUBSPOT_PROJECT_COMPONENTS_GITHUB_PATH,
       componentPath,
       {
-        sourceDir: componentData.path,
+        sourceDir: projectAddPromptResponse.componentTemplate.path,
         tag: latestRepoReleaseTag,
         hideLogs: true,
       }
     );
 
     logger.log('');
-    logger.success(i18n(`${i18nKey}.success`, { componentName }));
+    logger.success(
+      i18n(`${i18nKey}.success`, {
+        componentName: projectAddPromptResponse.name,
+      })
+    );
   } catch (error) {
     debugError(error);
     logger.error(i18n(`${i18nKey}.error.failedToDownloadComponent`));
--- a/commands/project/add.ts
+++ b/commands/project/add.ts
@@ -1,19 +1,25 @@
 // @ts-nocheck
+const path = require('path');
 const { logger } = require('@hubspot/local-dev-lib/logger');
-const { logError } = require('../../lib/errorHandlers/index');
-const { fetchReleaseData } = require('@hubspot/local-dev-lib/github');
-
+const {
+  cloneGithubRepo,
+  fetchReleaseData,
+} = require('@hubspot/local-dev-lib/github');
+const { debugError } = require('../../lib/errorHandlers');
 const { trackCommandUsage } = require('../../lib/usageTracking');
 const { i18n } = require('../../lib/lang');
 const { projectAddPrompt } = require('../../lib/prompts/projectAddPrompt');
+const { getProjectConfig } = require('../../lib/projects');
 const {
-  createProjectComponent,
-  getProjectComponentsByVersion,
-} = require('../../lib/projects');
+  getProjectComponentListFromRepo,
+} = require('../../lib/projects/create');
+const { findProjectComponents } = require('../../lib/projects/structure');
+const { ComponentTypes } = require('../../types/Projects');
 const { uiBetaTag } = require('../../lib/ui');
 const {
   HUBSPOT_PROJECT_COMPONENTS_GITHUB_PATH,
 } = require('../../lib/constants');
+const { EXIT_CODES } = require('../../lib/enums/exitCodes');
 
 const i18nKey = 'commands.project.subcommands.add';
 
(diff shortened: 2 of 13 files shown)


brandenrodgers:
I added a bit more to this. I'll add inline comments, but the major changes were:
- No more usage of `fetchFileFromRepository` in the repo. I'm going to deprecate that util in LDL because it's a very thin wrapper around the fetchRepoFile api call. It was also causing our `debugError` util to log out the entire axios error. In general, I think we should try to get rid of the thin api wrapper pattern in LDL. They don't really add much value.
-  The `--type` option for `hs project add` now works with the componentData's `path` and also the `label`. Previously it only worked for path, which isn't discoverable unless you go into the repo and look at it.
- I updated some of the error logs to be slightly more informative. The standard usage of the `hs project create` command will probably not include the `--template-source` option b/c that's more of a power user feature. So I updated the logs to be more useful to the standard users.

commands/create/api-sample.ts
```diff
@@ -39,23 +39,31 @@ module.exports = {
       }
     }
 
-    const samplesConfig = await fetchFileFromRepository(
```
brandenrodgers, line 42: This change probably feels irrelevant, but this was the only other usage of `fetchFileFromRepository` in the repo so I figured it would be worthwhile to remove it. The debug logs are significantly better if you directly call `fetchRepoFile`

commands/project/add.ts
```diff
@@ -23,40 +29,91 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
 
-  const components = await getProjectComponentsByVersion(
-    projectComponentsVersion
+  if (!projectDir || !projectConfig) {
+    logger.error(i18n(`${i18nKey}.error.locationInProject`));
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  // We currently only support adding private apps to projects
+  let projectContainsPublicApp = false;
```
brandenrodgers, line 42: This is the check to determine whether or not we're in a project that contains a public app. I re-used the `findProjectComponents` util that we use for local dev

commands/project/add.ts
```diff
@@ -23,40 +29,91 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
+      componentName
+    );
+
+    await cloneGithubRepo(
+      HUBSPOT_PROJECT_COMPONENTS_GITHUB_PATH,
+      componentPath,
+      {
+        sourceDir: componentData.path,
+        tag: latestRepoReleaseTag,
+        hideLogs: true,
```
brandenrodgers, line 105: I'm exposing this `hideLogs` option over in my LDL pr

commands/project/cloneApp.ts
```diff
@@ -79,7 +79,7 @@ exports.handler = async options => {
       });
       appId = appIdResponse.appId;
     }
-    const { name, dest } = await createProjectPrompt('', options, true);
+    const { name, dest } = await createProjectPrompt(options);
```
brandenrodgers, line 82: This is the new syntax for `createProjectPrompt` if all we need is a name and dest from it. Alternatively, you can also pass it a list of available templates and it will prompt you to select one (see usage over in the project create command).

commands/project/create.ts
```diff
@@ -28,36 +37,92 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
 exports.handler = async options => {
   const { derivedAccountId } = options;
 
-  const hasCustomTemplateSource = Boolean(options.templateSource);
+  let latestRepoReleaseTag;
```
brandenrodgers, line 40: Most of the changes in this file are to make the patterns match the ones in the project add command file. Now both files will:
1. Determine the template source (aka GH repo). This command lets the user override the default
2. Fetch the list of available assets using the config.json file at the repo root
3. Prompt the user for info (name, dest, template)
4. Call `cloneGithubRepo` to download the contents

types/Projects.ts
```diff
@@ -8,12 +8,6 @@ export type ProjectTemplate = {
   insertPath: string;
 };
 
-export type ComponentTemplate = {
```
brandenrodgers, line 11: This was a duplicate of `ProjectAddComponentData`

lib/prompts/projectAddPrompt.ts
```diff
@@ -47,4 +57,13 @@ export async function projectAddPrompt(
       },
     },
   ]);
+
+  if (providedTypeIsValid) {
```
brandenrodgers, line 61: Doing this to match the pattern in createProjectPrompt. It makes it so this prompt will always return a value for `componentData`

lib/prompts/createProjectPrompt.ts
```diff
@@ -85,28 +28,20 @@ function findTemplate(
 }
 
 export async function createProjectPrompt(
```
brandenrodgers, line 30: The behavior of this util was a little confusing because it was doing a lot. I moved the template fetch out of it to simplify things a little. Now all this does is prompt for name, dest, and template if a list of templates is provided.

lib/projects/index.ts
```diff
@@ -325,43 +244,3 @@ export function logFeedbackMessage(buildId: number): void {
     logger.log(i18n(`${i18nKey}.logFeedbackMessage.feedbackMessage`));
   }
 }
-
-export async function createProjectComponent(
```
brandenrodgers, line 329: I removed `createProjectConfig` and `createProjectComponent` and put the functionality into the project create and add command files (respectively)

lib/projects/index.ts
```diff
@@ -74,7 +61,7 @@ export async function getProjectConfig(dir?: string): Promise<{
   projectDir: string | null;
   projectConfig: ProjectConfig | null;
 }> {
-  const configPath = await getProjectConfigPath(dir);
+  const configPath = getProjectConfigPath(dir);
```
brandenrodgers, line 64: From what I can tell, `getProjectConfig` doesn't actually need to be async. But I think I've probably already passed my scope creep limit for this PR 😅

lib/projects/create.ts
```diff
@@ -0,0 +1,81 @@
...
+  ProjectAddComponentData,
+  ProjectTemplateRepoConfig,
+} from '../../types/Projects';
+import { debugError } from '../errorHandlers/index';
+
+const i18nKey = 'lib.projects.create';
+
+const PROJECT_TEMPLATE_PROPERTIES = ['name', 'label', 'path', 'insertPath'];
+
+export async function getProjectComponentListFromRepo(
```
brandenrodgers, line 21: This was moved from `lib/projects/index`

lib/projects/create.ts
```diff
@@ -0,0 +1,81 @@
...
+    debugError(err);
+  }
+
+  if (config) {
+    return config[PROJECT_COMPONENT_TYPES.COMPONENTS] || [];
+  }
+  return [];
+}
+
+export async function getProjectTemplateListFromRepo(
```
brandenrodgers, line 43: This behavior was previously happening over in the createProjectPrompt

commands/project/create.ts
```diff
@@ -28,36 +37,92 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
+    } catch (err) {
+      logger.error(i18n(`${i18nKey}.error.failedToFetchProjectList`));
+      process.exit(EXIT_CODES.ERROR);
+    }
+  }
 
-  let githubRef = '';
+  const projectTemplates = await getProjectTemplateListFromRepo(
+    templateSource,
+    latestRepoReleaseTag || DEFAULT_PROJECT_TEMPLATE_BRANCH
```
camden11, line 59: Will this cause any issues if someone uses a custom template with a repo that uses something other than `main` (could see `master` still being common, for example)

commands/project/create.ts
```diff
@@ -28,36 +37,92 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
+    process.exit(EXIT_CODES.ERROR);
   }
 
   const { name, template, dest } = await createProjectPrompt(
-    githubRef,
-    options
+    options,
+    projectTemplates
   );
+  const projectName = name || options.name;
```
camden11, line 71: Same as my above comment - looks like `name` and `dest` are guaranteed to be returned

commands/project/create.ts
```diff
@@ -28,36 +37,92 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
+    fs.readFileSync(projectConfigPath).toString()
   );
 
+  writeProjectConfig(projectConfigPath, {
+    ...parsedConfigFile,
+    name: projectName,
+  });
+
+  // If the template is 'no-template', we need to manually create a src directory
+  if (template.name === 'no-template') {
```
camden11, line 119: Good candidate for a constant

camden11 (commented):
Nice, love to see a PR that cleans things up this much!

commands/project/add.ts
```diff
@@ -23,40 +29,91 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
-  if (!component) {
-    component = components.find(t => t.path === options.type);
+  if (!components.length) {
+    logger.error(i18n(`${i18nKey}.error.failedToFetchComponentList`));
+    process.exit(EXIT_CODES.ERROR);
   }
 
-  trackCommandUsage('project-add', null, derivedAccountId);
+  const { componentData, name } = await projectAddPrompt(components, options);
+  const componentName = name || options.name;
```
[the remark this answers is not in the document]
brandenrodgers: I tested this out and I don't think it's true. The prompts are using the `when` field to skip certain questions if they're passed in via `options`. So the values for these are undefined in certain cases. I updated the prompts in [c44ed7c](https://github.com/HubSpot/hubspot-cli/pull/1385/commits/c44ed7cdd215da43bb4fb73c921386cae534d6ed) to make them consistently return values for each prompt option though. I think we might want to update our other prompts to follow this pattern at some point. It makes them easier to use in the commands.

commands/project/create.ts
brandenrodgers: Good call, done in [6970a63](https://github.com/HubSpot/hubspot-cli/pull/1385/commits/6970a63c86ee06d3bb34307d3af05eae6d2a26a2)

commands/project/add.ts
```diff
@@ -23,40 +29,91 @@ exports.describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
...
+      latestRepoReleaseTag = repoReleaseData.tag_name;
+    }
+  } catch (err) {
+    debugError(err);
+    logger.error(i18n(`${i18nKey}.error.failedToFetchComponentList`));
+    process.exit(EXIT_CODES.ERROR);
+  }
 
-  name = name || options.name;
+  const components =
```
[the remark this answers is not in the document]
brandenrodgers: Fixed that in a recent commit.

Re: TS port. I'm going to keep this as is for now. I'd prefer to do the TS port as an isolated piece of work just to avoid mixing in functionality changes with it. In hindsight maybe a better strategy could be to port things like this to TS as an initial PR, and then make the changes in a follow up. I can try to do that next time

camden11 (approved):
LGTM!

joe-yeager (approved):

PR #1386, opened 2025-02-20
Chore: Convert `custom-object` commands to TS

kemmerle:
## Description and Context
Note: This PR depends on https://github.com/HubSpot/hubspot-local-dev-lib/pull/242. I have created an experimental release for easier testing. 

This PR converts the `custom-object` commands bucket to TypeScript. It also removes the `useCustomObjectHubfile` and the `file-transport` endpoints from the commands, given that they appear to have never been used. 

## TODO

- [x] Add basic tests
- [x] Convert the remaining `co` bucket commands

## Who to Notify


package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/HubSpot/hubspot-cli",
   "dependencies": {
-    "@hubspot/local-dev-lib": "3.3.1",
+    "@hubspot/local-dev-lib": "0.1.4-experimental.0",
```
kemmerle, line 8: Flagging to change this before merging this PR.

lib/validation.ts
```diff
@@ -187,7 +189,7 @@ export function fileExists(_path: string): boolean {
   return true;
 }
 
-export function checkAndConvertToJson(_path: string): object | null {
+export function checkAndConvertToJson(_path: string): Schema | null {
```
brandenrodgers, line 192: This generic util doesn't know which type of object it's parsing the JSON for. It could be a schema, a hubdb table, an object, or something else. This seems like a good place to leverage TS generic function typing.

commands/customObject/schema.ts
```diff
@@ -1,17 +1,22 @@
...
 
 const i18nKey = 'commands.customObject.subcommands.schema';
 
-exports.command = ['schema', 'schemas'];
-exports.describe = i18n(`${i18nKey}.describe`);
-exports.builder = yargs => {
+export const command = ['schema', 'schemas'];
+export const describe = i18n(`${i18nKey}.describe`);
+
+export function handler(): void {
```
kemmerle, line 15: Because `schema` is technically a subcommand of the `custom-object` command, it was throwing a TypeScript error because the `handler` property was missing from the `CommandModule` type. When I was researching solutions, I came up with three:

1) I could explicitly cast the `schema` command to `CommandModule` in `commands/customObject`. This did not resolve the TypeScript error.

2) I could adjust `schema`'s type to `Partial<CommandModule>` in exporting it, ie. 

```
const schemaCommand: Partial<CommandModule> = {
  command,
  describe,
  builder
}
```
That also didn't resolve the error entirely. 

3) I could provide an empty `handler`. This seems the cleanest solution to me that doesn't require casting or a `@ts-ignore` comment.

Please let me know if I'm missing anything.

lib/validation.ts
```diff
@@ -187,7 +188,7 @@ export function fileExists(_path: string): boolean {
   return true;
 }
 
-export function checkAndConvertToJson(_path: string): object | null {
+export function checkAndConvertToJson<T>(_path: string): T | null {
```
joe-yeager, line 191: This generic seems to be functioning as an implicit type assertion since it is only used to specify the return type. I.E. is basically the same as `checkAndConvertToJson(somePath) as T`, which bypasses the type system entirely.

I think ideally we should make this method return `unknown` and then the consuming pieces of code would need to type narrow on their end using type predicates.

commands/customObject/create.ts
```diff
@@ -50,16 +65,22 @@ exports.handler = async options => {
...
+
+  yargs.positional('name', {
+    describe: i18n(`${i18nKey}.positionals.name.describe`),
+    type: 'string',
+  });
+
+  yargs.option('path', {
+    describe: i18n(`${i18nKey}.options.path.describe`),
+    type: 'string',
+  });
```
joe-yeager, line 83: Nit: Do we need to split these or can we keep them chained?

joe-yeager: You can make the mock return itself.  Here's [an example](https://github.com/HubSpot/hubspot-cli/blob/main/commands/secret/__tests__/deleteSecret.test.ts#L16-L24) of how you did it in an earlier test

lib/validation.ts
joe-yeager: I see that you added the type predicates which is great 👍 

Could we make this method return `unknown | null` and remove the generic?

joe-yeager (dismissed):

PR #1387, opened 2025-02-24
Convert project dev command to ts

camden11:
## Description and Context
Wanted to have everything in TS before getting started on the work for unified apps local dev. This file honestly probably needs a complete re-write so this is more of a temporary bandaid solution to get all of the types working. In any case I think it'll make working on this file a little easier

## Who to Notify

### DIFF
--- a/commands/project/dev.ts
+++ b/commands/project/dev.ts
@@ -1,41 +1,39 @@
-// @ts-nocheck
-const {
+import {
   addAccountOptions,
   addConfigOptions,
   addUseEnvironmentOptions,
-  addTestingOptions,
-} = require('../../lib/commonOpts');
-const { trackCommandUsage } = require('../../lib/usageTracking');
-const { handleExit } = require('../../lib/process');
-const { i18n } = require('../../lib/lang');
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const {
+} from '../../lib/commonOpts';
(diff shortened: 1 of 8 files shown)


commands/project/dev.ts
```diff
@@ -99,6 +111,15 @@ exports.handler = async options => {
 
   const accounts = getConfigAccounts();
 
+  if (!accounts) {
+    logger.error(
+      i18n(`${i18nKey}.errors.noAccountsInConfig`, {
+        authCommand: uiCommandReference('hs auth'),
+      })
+    );
+    process.exit(EXIT_CODES.SUCCESS);
```
brandenrodgers, line 120: Should this be an error?

types/Projects.ts
```diff
@@ -127,3 +128,10 @@ export type Component<T = GenericComponentConfig> = {
   runnable: boolean;
   path: string;
 };
+
+export type ProjectDevTargetAccountPromptResponse = {
```
brandenrodgers, line 132: Should this live in the prompt file instead?

brandenrodgers (approved):
Just some nits, but nothing blocking

types/Projects.ts
camden11: I couldn't decide between the two, but if you vote for prompts lets go with that

PR #1388, opened 2025-02-26
WIP: Unified apps local dev first pass

camden11:


lib/DevServerManagerV2.ts
```diff
@@ -0,0 +1,139 @@
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { Environment } from '@hubspot/local-dev-lib/types/Config';
+import { i18n } from './lang';
+import { promptUser } from './prompts/promptUtils';
+import { DevModeInterface as UIEDevModeInterface } from '@hubspot/ui-extensions-dev-server';
```
camden11, line 5: @eaverhe This is where you can swap in the new `DevModeInterface`

lib/DevServerManagerV2.ts
```diff
@@ -0,0 +1,139 @@
...
+  private started: boolean;
+  private devServers: DevServerInterface[];
+
+  constructor() {
+    this.initialized = false;
+    this.started = false;
+    this.devServers = [UIEDevModeInterface];
+  }
+
+  async iterateDevServers(
```
kemmerle, line 40: An explanatory note would be helpful with this function. Is this handling if the user has multiple dev servers running?

lib/LocalDevManagerV2.ts
```diff
@@ -0,0 +1,571 @@
...
+  projectDir: string;
+  projectId: number;
+  debug?: boolean;
+  deployedBuild?: Build;
+  isGithubLinked: boolean;
+  intermediateRepresentation: IntermediateRepresentationLocalDev;
+  env: Environment;
+};
+
+class LocalDevManagerV2 {
```
kemmerle, line 65: I'm torn about `LocalDevManager2`. On the one hand, reading through it I understand what each function does taken on its own, and I appreciate the explanatory notes for stickier sections.

But I do get confused when I attempt to look at the file as a whole. For example, we have multiple `start()` methods. There's the `start` method defined on the `LocalDevManager2` class, which is then invoked as `LocalDev.start()`.`LocalDev.start()` calls `this.devServerStart()`. And then `this.devServerStart()` calls `DevServerManagerV2.start()`. 

Can we remove layers of abstraction here? Can we call `DevServerManagerV2.start()` from the original `LocalDev.start()`? I'm afraid this will make debugging very difficult. Combining functions like this would also eliminate some code, less for us to maintain in the future.

types/ProjectComponents.ts
```diff
@@ -0,0 +1,48 @@
...
+  APP_DISTRIBUTION_TYPES,
+  APP_AUTH_TYPES,
+} from '../lib/constants';
+import { ValueOf } from '@hubspot/local-dev-lib/types/Utils';
+
+type AppDistributionType = ValueOf<typeof APP_DISTRIBUTION_TYPES>;
+type AppAuthType = ValueOf<typeof APP_AUTH_TYPES>;
+
+type AppConfig = {
+  description: string;
```
kemmerle, line 15: Do we know which properties will be required/optional on the AppConfig? Do we need to speak with BE folks about it?

lib/DevServerManagerV2.ts
camden11: See comment below, but this is just repeated functionality from the existing `DevServerManager`. Basically, its set up to handle the case where there are multiple dev servers (even though there's only one right now) and call each lifecycle method on each of them

dennisedson:
proud of you, camden

lib/constants.ts
```diff
@@ -69,3 +69,18 @@ export const PLATFORM_VERSION_ERROR_TYPES = {
   PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST:
     'PlatformVersionErrorType.PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST',
 } as const;
+
+export const IR_COMPONENT_TYPES = {
```
brandenrodgers, line 73: Might be nice to get these from the parsing lib eventually. But keeping them here is fine for now 👍

types/ProjectComponents.ts
```diff
@@ -0,0 +1,48 @@
...
+  auth: {
+    type: AppAuthType;
+    redirectUrls: string[];
+    requiredScopes: string[];
+    optionalScopes: string[];
+    conditionallyRequiredScopes: string[];
+  };
+};
+
+type CardConfig = {
```
brandenrodgers, line 28: These types should also probably live in the parsing lib eventually, but not a blocker. We can move them later

lib/DevServerManagerV2.ts
brandenrodgers: I forget how exactly this manifests itself, but I'm pretty sure the way I initially built this causes errors thrown by the dev servers to get swallowed. No need to fix it in this PR, but we may want to eventually revisit it.

commands/project/dev/unifiedFlow.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  // Get IR
+  try {
+    const intermediateRepresentation = await translateForLocalDev({
+      projectSourceDir: path.join(projectDir, projectConfig.srcDir),
+      platformVersion: projectConfig.platformVersion,
+      accountId: targetProjectAccountId,
+    });
+
+    components = intermediateRepresentation.intermediateNodesIndexedByUid;
```
brandenrodgers, line 55: This is a little bit of a nit (not a blocker), but we might be able to clear up some potential confusion by referring to this as something separate from `components`. The IR shape isn't 1:1 with the user-facing component config, so it might be helpful to use a different name to help us remember that the shape of these components will be different. Maybe something like "projectNodes", "intermediateComponentsByUid", or we could even keep it as "intermediateNodesIndexedByUid".

lib/LocalDevManagerV2.ts
```diff
@@ -0,0 +1,577 @@
...
+    } = await fetchPublicAppProductionInstallCounts(
+      activePublicAppData.id,
+      this.targetProjectAccountId
+    );
+
+    this.activePublicAppData = activePublicAppData;
+    this.publicAppActiveInstalls = uniquePortalInstallCount;
+  }
+
+  async checkActivePublicAppInstalls(): Promise<void> {
```
brandenrodgers, line 170: Also not a blocker, but we may be able to set ourselves up for the future a little bit by trying to break some of the app-specific logic out into its own file(s)

camden11: Oh that's a good call. I'll see if there's anything I can do right now. Might need to be part of a bigger revamp since some of the logic is pretty intertwined with the LocalDevManager (for example, the idea of having an `activePublicApp` that's attached to the instance)

commands/project/dev/unifiedFlow.ts
```diff
@@ -0,0 +1,169 @@
...
+      projectSourceDir: path.join(projectDir, projectConfig.srcDir),
+      platformVersion: projectConfig.platformVersion,
+      accountId: targetProjectAccountId,
+    });
+
+    projectNodes = intermediateRepresentation.intermediateNodesIndexedByUid;
+
+    logger.debug(util.inspect(projectNodes, false, null, true));
+  } catch (e) {
+    if (isTranslationError(e)) {
```
brandenrodgers, line 59: not a blocker, but does our standard behavior for `logError` not handle this?

lang/en.lyaml
```diff
@@ -529,6 +529,7 @@ en:
             noAccountsInConfig: "No accounts found in your config. Run {{ authCommand }} to configure a HubSpot account with the CLI."
             invalidProjectComponents: "Projects cannot contain both private and public apps. Move your apps to separate projects before attempting local development."
             noRunnableComponents: "No supported components were found in this project. Run {{ command }} to see a list of available components and add one to your project."
+            invalidUnifiedAppsAccount: "You must target a Combined account to use Unified Apps Local Dev"
```
brandenrodgers, line 532: I know I initially wrote this, but I think we want to use different terminology here. This should probably be something like:
"Cannot start local dev. Target a standard HubSpot account or a Developer account and try again."

lib/LocalDevManagerV2.ts
brandenrodgers: In the short-term we only support a single app, which might make splitting this up easier. But we can dig more into this in a follow up PR

commands/project/dev/unifiedFlow.ts
camden11: Good question, I just pulled this logic from `hs project upload`. Assuming it's something specific to how project-parsing-lib formats errors. Maybe worth moving to `logError` though.

@joe-yeager Is there any reason for manually calling `toString` here?

lang/en.lyaml
camden11: Updated it to require a developer account only, since we require you to create a dev test account and you can't do that with a standard account at the moment

commands/project/dev/unifiedFlow.ts
joe-yeager: Sorry, missed this. The `toString` method gives a pretty printed message when we want to show more details. 
https://github.com/HubSpot/project-parsing-lib/blob/main/src/lib/errors.ts#L46-L58

We could update the `logError` method to do this if check if we what to centralize it there.

commands/project/dev/unifiedFlow.ts
```diff
@@ -0,0 +1,171 @@
...
+    logger.error(
+      i18n(`${i18nKey}.errors.noRunnableComponents`, {
+        projectDir,
+        command: uiCommandReference('hs project add'),
+      })
+    );
+    process.exit(EXIT_CODES.SUCCESS);
+  }
+
+  // @TODO Validate component types (i.e. previously you could not have both private and public apps)
```
joe-yeager, line 78: Future PR?

commands/project/dev/unifiedFlow.ts
```diff
@@ -0,0 +1,171 @@
...
+  } catch (e) {
+    if (isTranslationError(e)) {
+      logger.error(e.toString());
+    } else {
+      logError(e);
+    }
+    return process.exit(EXIT_CODES.ERROR);
+  }
+
+  // @TODO Do we need to do more than this or leave it to the dev servers?
```
joe-yeager, line 67: I vote for leave it to the dev servers to tell us what they are capable of

joe-yeager (approved):
LGTM, just a couple non-blocking comments

lib/constants.ts
joe-yeager: Yeah, we can get these from the parsing lib eventually.  We have a list of the valid component keys, and a method to translate them to the internal values which is what is shown in the IR.

PR #1390, opened 2025-02-28
chore: Add flag to bypass validation

joe-yeager:
## Description and Context
Add a hidden flag to bypass upload on v2 uploads


brandenrodgers (dismissed):
code lgtm 👍 but my slight preference would be to use `--skip-validation` as the flag name (not a blocker though)

joe-yeager:
> code lgtm 👍 but my slight preference would be to use `--skip-validation` as the flag name (not a blocker though)

Updated!

PR #1391, opened 2025-03-11
Surface httpUseLocalhost in the hs doctor check

brandenrodgers:
## Description and Context

This sets up the `hs doctor` command to be able to surface warnings when certain CLI settings are enabled. For now it's just surfacing a warning when `httpUseLocalhost` is true.

## Screenshots

image

### DIFF
--- a/lib/doctor/DiagnosticInfoBuilder.ts
+++ b/lib/doctor/DiagnosticInfoBuilder.ts
@@ -39,7 +39,7 @@ export interface DiagnosticInfo extends FilesInfo {
   path?: string;
   versions: { [hubspotCli]: string; node: string; npm: string | null };
   config: string | null;
-  configSettings: Record<string, unknown>;
+  configSettings: { [key: string]: unknown };
   project: {
     details?: Project;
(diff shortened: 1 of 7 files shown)


brandenrodgers:
The screenshot is slightly outdated. I updated the copy to: `The setting {{#bold}}httpUseLocalhost{{/bold}} is enabled`

lib/doctor/DiagnosticInfoBuilder.ts
```diff
@@ -36,6 +39,7 @@ export interface DiagnosticInfo extends FilesInfo {
   path?: string;
   versions: { [hubspotCli]: string; node: string; npm: string | null };
   config: string | null;
+  configSettings: Record<string, unknown>;
```
kemmerle, line 42: I think we're staying away from the `Record` pattern in favor of a plain object.

brandenrodgers: Ah good catch. I fixed this in one place and forgot to do it here too

joe-yeager (approved):

PR #1392, opened 2025-03-11
WIP: Handle LDL config Revamp changes

camden11:


commands/account/list.ts
```diff
@@ -115,7 +112,7 @@ export async function handler(
   logger.log(i18n(`${i18nKey}.configPath`, { configPath: configPath! }));
   logger.log(
     i18n(`${i18nKey}.defaultAccount`, {
-      account: getConfigDefaultAccount()!,
+      account: getConfigDefaultAccount().accountId,
```
brandenrodgers, line 115: `getConfigDefaultAccount ` can throw. Are we protected from it throwing here because we make sure that it won't throw in our middleware?

commands/account/remove.ts
```diff
@@ -50,7 +50,11 @@ export async function handler(
 
   trackCommandUsage('accounts-remove', undefined, account.accountId);
 
-  const currentDefaultAccount = getConfigDefaultAccount();
+  let currentDefaultAccount;
+
+  try {
+    currentDefaultAccount = getConfigDefaultAccount();
```
brandenrodgers, line 56: Not a blocker, but I wonder if it would be worthwhile to support a version of popular utils like this one where they don't throw. Like adding an argument to it that lets it return null instead of throwing for situations like this one.

commands/account/use.ts
```diff
@@ -22,31 +22,31 @@ interface AccountUseArgs extends CommonArgs {
...
-    newDefaultAccount = await selectAccountFromConfig();
+    newDefaultAccountIdentifier = await selectAccountFromConfig();
   }
 
-  trackCommandUsage(
-    'accounts-use',
-    undefined,
-    getAccountId(newDefaultAccount)!
-  );
+  const newDefaultAccount = getConfigAccountIfExists(
```
brandenrodgers, line 39: Technically `getConfigAccountIfExists` can throw now. Should we wrap this in a try/catch too? Maybe this is a +1 for allowing a way to call this without it throwing? I suggested adding an argument to these to allow for a "no throw" variation, but another option could be to export versions of these utils that catch errors and return null instead of throwing. Like `getConfigAccountOrNull`

commands/account/remove.ts
camden11: Yeah, definitely more instances where we don't want this to throw than I thought. I think I'm going to stick with this for now, but if it ends up being an issue I'll add a non throwing version. Want to keep the LDL exports as uncluttered as possible

commands/account/use.ts
camden11: `getConfigAccountIfExists` is a new function I added specifically for occasions like this where you don't want to throw if the account isn't there. Technically it can still throw if there's no config at all, but I think that's the probably the behavior we want.

commands/account/remove.ts
camden11: I've run into this a few more times so going to add a `getConfigDefaultAccountIfExists`. Might switch over to that in a different PR though

brandenrodgers (approved):

commands/account/clean.ts
```diff
@@ -34,9 +33,9 @@ export async function handler(
 
   trackCommandUsage('accounts-clean');
 
-  const accountsList = getConfigAccounts() || [];
-  const filteredTestAccounts = accountsList.filter(p =>
-    qa ? p.env === 'qa' : p.env !== 'qa'
+  const accountsList = getAllConfigAccounts();
```
kemmerle, line 36: Do we need to default to an empty array here?

commands/account/list.ts
```diff
@@ -112,12 +109,22 @@ export async function handler(
...
-  logger.log(
-    i18n(`${i18nKey}.defaultAccount`, {
-      account: getConfigDefaultAccount()!,
-    })
-  );
+
+  if (defaultAccount) {
+    logger.log(
+      i18n(`${i18nKey}.defaultAccount`, {
+        account: defaultAccount.name,
```
kemmerle, line 123: Assuming that in the new `getConfigDefaultAccount()` function you've written that `name` does exist, correct?

PR #1394, opened 2025-03-12
chore: Tagless experimental releases

joe-yeager:
## Description and Context

If do not generate git tags for experimental releases

### DIFF
--- a/scripts/release.ts
+++ b/scripts/release.ts
@@ -84,7 +84,8 @@ async function cleanup(newVersion: string): Promise<void> {
 async function publish(
   tag: Tag,
   otp: string,
-  isDryRun: boolean
+  isDryRun: boolean,
+  isExperimental: boolean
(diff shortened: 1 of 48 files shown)


scripts/release.ts
```diff
@@ -308,10 +312,20 @@ async function handler({
...
-  await exec(gitCommand);
-  logger.log(`Changes pushed successfully`);
+  if (isExperimental) {
+    logger.log();
+    logger.log(`Experimental release: Skipping push to Github`);
+
+    // Reset the version back to the local version
+    await exec(
+      `yarn version --no-git-tag-version --new-version ${localVersion}`
+    );
```
joe-yeager, line 322: If we don't reset this back to the `localVersion` the change will still be in the package.json

brandenrodgers (approved):

PR #1396, opened 2025-03-13
chore: Fix typescript papercuts in ATs

joe-yeager:
## Description and Context

- Add a `tsconfig.json` specifically for the ATs and provide the vitest types.
- Set `global: true` so we no longer need to import the vitest methods like `describe`, `expect`, etc and they will be provided on runtime.


scripts/get-all-commands.ts
```diff
@@ -43,7 +43,7 @@ async function extractCommands(
   return commands;
 }
 
-(async function() {
+(async function () {
```
joe-yeager, line 46: I ran `prettier:write` and this one came along for the ride.

camden11 (approved):

PR #1397, opened 2025-03-13
Update middleware for config revamp

camden11:
## Description and Context
This updates `bin/cli.js` with the new config methods from `local-dev-lib`, and makes a few other changes
* Sets environment variables when `use-env` and `config` flags so that LDL is aware of these settings
* Updates the `injectAccountIdMiddleware` to exit with a helpful message if the user provides an account ID that doesn't exist in their config. Right now this still fails, but the missing accountId goes uncaught until the `http` module tries to use it.

## TODO
We'll have to thoroughly test these changes, but won't be able to do that until the rest of the updates are made

## Who to Notify

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -7,12 +7,12 @@ const chalk = require('chalk');
 const { logger } = require('@hubspot/local-dev-lib/logger');
 const { addUserAgentHeader } = require('@hubspot/local-dev-lib/http');
 const {
-  loadConfig,
-  getAccountId,
-  configFileExists,
-  getConfigPath,
-  validateConfig,
+  globalConfigFileExists,
+  getConfigFilePath,
+  isConfigValid,
+  getConfigAccountIfExists,
 } = require('@hubspot/local-dev-lib/config');
+import { ENVIRONMENT_VARIABLES } from '@hubspot/local-dev-lib/constants/config';
 const { logError } = require('../lib/errorHandlers/index');
 const { setLogLevel, getCommandName } = require('../lib/commonOpts');
 const { validateAccount } = require('../lib/validation');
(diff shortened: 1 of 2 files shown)


bin/cli.js
```diff
@@ -183,30 +183,67 @@ const handleDeprecatedEnvVariables = options => {
...
       i18n(`${i18nKey}.handleDeprecatedEnvVariables.portalEnvVarDeprecated`, {
-        configPath: getConfigPath(),
+        configPath: getConfigFilePath(),
       })
     );
     process.env.HUBSPOT_ACCOUNT_ID = process.env.HUBSPOT_PORTAL_ID;
   }
 };
 
+function addConfigEnvironmentVariablesMiddleware(options) {
```
brandenrodgers, line 193: Not a blocker, but we might have enough middleware at this point to warrant moving it all into a dedicated middleware util file

bin/cli.js
```diff
@@ -183,30 +183,67 @@ const handleDeprecatedEnvVariables = options => {
...
 /**
  * Auto-injects the derivedAccountId flag into all commands
  */
 const injectAccountIdMiddleware = async options => {
-  const { account } = options;
+  const { account: accountOption } = options;
 
   // Preserves the original --account flag for certain commands.
-  options.providedAccountId = account;
+  options.providedAccountId = accountOption;
```
brandenrodgers, line 212: Is there any scenario where we'd want to support both the `--use-env` flag and also the `--account` option at the same time? I'm wondering if we should abort if that ever happens. Using `--use-env` is essentially the same thing as passing in the `--account` flag

bin/cli.js
```diff
@@ -183,30 +183,67 @@ const handleDeprecatedEnvVariables = options => {
...
+      );
+      process.exit(EXIT_CODES.ERROR);
+    }
+  }
+
+  // If no provided account or environment variable, use the default account
+  try {
+    const defaultAccount = getConfigDefaultAccount();
+    options.derivedAccountId = defaultAccount.accountId;
+  } catch (e) {}
```
brandenrodgers, line 243: Do we do nothing here because any errors will be thrown during the validation step? If so it might be worth adding a comment to explain it

bin/cli.js
```diff
@@ -183,30 +183,67 @@ const handleDeprecatedEnvVariables = options => {
...
+          accountIdentifier: accountOption,
+          authCommand: uiCommandReference('{authCommand}'),
+        })
+      );
+      process.exit(EXIT_CODES.ERROR);
+    }
+  }
+
+  // If no provided account or environment variable, use the default account
+  try {
```
brandenrodgers, line 240: We only want to do this if `options.derivedAccountId` is undefined right?

bin/cli.js
camden11: Agreed. Might wait until this is all merged though because that'll be one more thing that creates conflicts

bin/cli.js
camden11: Currently the CLI doesn't error out here if you don't have a default account, so wanted to preserve that behavior

bin/cli.js
camden11: Oh yes good catch

brandenrodgers (approved):
lgtm. Looks like the target branch is `main`, should that be `config-revamp-main`?

PR #1398, opened 2025-03-13
Feature: `hs config migrate` command

kemmerle:
## Description and Context
**UPDATED 3/21 WITH NEW SCREENSHOTS**
Depends on changes in LDL ([PR](https://github.com/HubSpot/hubspot-local-dev-lib/pull/247)). I've created an experimental release of LDL to help with testing. 

This PR adds support for the `hs config migrate` command to help users migrate from the deprecated config file (`hubspot.config.yml`) to the new centralized config located at the user's working directory. 

**Note**: To fix failing tests, I needed to migrate the `config` bucket of commands to TypeScript. 

## Screenshots
When only deprecated config file is detected:

Screenshot 2025-03-21 at 2 33 01 PM

When deprecated and global configs are both present (with conflicts):

Screenshot 2025-03-21 at 2 35 36 PM

When only global config is present (ie. migration is complete):

Screenshot 2025-03-18 at 10 21 35 PM

**NEW** With force flag when merging configs:

Screenshot 2025-03-21 at 2 37 33 PM

With config flag when only deprecated config file is present: 

Screenshot 2025-03-21 at 2 39 50 PM

With config flag when deprecated and global configs are both present (with conflicts):

Screenshot 2025-03-21 at 2 43 07 PM

**NEW** With config and force flags when deprecated and global configs are both present:

Screenshot 2025-03-21 at 2 44 14 PM

With invalid config flag:

Screenshot 2025-03-21 at 2 40 20 PM

**With `hs account auth` command**:

When only deprecated config is present:

Screenshot 2025-03-21 at 2 48 36 PM

When deprecated and global configs are both present (with conflicts):

Screenshot 2025-03-21 at 2 50 03 PM

## TODO

- [x] Add support for conflicting migrating `defaultPortal` and `defaultAccount` properties when merging
- [x] Get feedback from Jono on copy/format of copy

## Who to Notify


bin/cli.js
```diff
@@ -252,6 +253,10 @@ const SKIP_LOAD_CONFIG = {
   accounts: skipLoadConfigAccountSubCommands,
 };
 
+const SKIP_CONFIG_FLAG_VALIDATION = {
```
brandenrodgers, line 256: Nit, but the naming of this is a little misleading. This looks like it's being used to skip the error that typically gets logged when the user passes in the `--config` flag while a global config exists.

lib/configMigrate.ts
```diff
@@ -0,0 +1,72 @@
...
+    })
+  );
+  process.exit(EXIT_CODES.SUCCESS);
+}
+
+export async function handleMerge(configPath?: string) {
+  const { shouldMergeConfigs } = await promptUser({
+    name: 'shouldMergeConfigs',
+    type: 'confirm',
+    message: i18n(`${i18nKey}.mergeConfigsPrompt`, {
```
brandenrodgers, line 50: This is probably something more for Jono, but the messages in the prompts are pretty long. We might want to consider moving the majority of this copy into a log statement above the prompt to explain what will happen, and then shorten the prompt to something simple like "Continue with the merge?"

commands/config/migrate.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+  } else if (globalConfigExists && deprecatedConfigExists) {
+    try {
+      await handleMerge(configPath, force);
+      process.exit(EXIT_CODES.SUCCESS);
+    } catch (error) {
+      logError(error);
+      process.exit(EXIT_CODES.ERROR);
+    }
+  } else if (globalConfigExists && !deprecatedConfigExists) {
```
brandenrodgers, line 60: This could probably just check if `!deprecatedConfigExists`. If you do it before the above if statements, you'd be able to remove `deprecatedConfigExists` from the checks.

lang/en.lyaml
```diff
@@ -1249,6 +1259,13 @@ en:
           describe: "Use environment variable config"
         debug:
           describe: "Set log level to debug"
+    configMigrate:
+      migrateConfigPrompt: "Migrate config?\n From deprecated config file: {{ deprecatedConfigPath }}\n To new global config file: {{ globalConfigPath }}\n (This action will delete the deprecated config file).\n"
+      mergeConfigConflictPrompt: "Overwrite the conflicting property in the global config?\n {{ property }}: {{ values }} "
```
brandenrodgers, line 1264: You could shorten this by doing:
`Change the {{ property }} property in the global config from {{old}} to {{new}}?`

Which would show as:
`Change the httpTimeout property in the global config from 2000 to 3000? Y/n`

kemmerle: I'll ask Jono--this was placeholder text, since he was OOO on Thursday/Friday.

lang/en.lyaml
```diff
@@ -133,6 +133,16 @@ en:
     config:
       describe: "Commands for managing the CLI config file."
       subcommands:
+        migrate:
+          describe: "Migrate from the deprecated {{ deprecatedConfigPath }} file to the new global config file at {{ globalConfigPath }}."
+          migrationAlreadyCompleted: "A global configuration file already exists at {{ globalConfigPath }}. No migration necessary."
```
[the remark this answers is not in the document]
kemmerle: Made the change but also going to get Jono's feedback

commands/config/migrate.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (configPath && !fs.existsSync(configPath)) {
+    logger.log(
+      i18n(`${i18nKey}.errors.configNotFound`, {
+        configPath,
+      })
+    );
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  const deprecatedConfigExists = configFileExists(false, configPath);
```
camden11, line 41: see LDL comment but I think this will return true whenever a config flag is passed in, even if the config doesn't exist at that path

commands/config/set.ts
```diff
@@ -1,24 +1,26 @@
...
 
 const i18nKey = 'commands.config.subcommands.set';
 
-exports.command = 'set';
-exports.describe = i18n(`${i18nKey}.describe`);
+export const command = 'set';
+export const describe = i18n(`${i18nKey}.describe`);
 
-const selectOptions = async () => {
+async function selectOptions() {
```
camden11, line 20: remember return type here!

commands/account/auth.ts
```diff
@@ -7,8 +7,8 @@ import {
   createEmptyConfigFile,
   deleteEmptyConfigFile,
   getConfigDefaultAccount,
-  configFileExists,
 } from '@hubspot/local-dev-lib/config';
+import { configFileExists } from '@hubspot/local-dev-lib/config/migrate';
```
camden11, line 11: Just curious why you switched to this other `configFileExists` - seems like the one in LDL should work now with your most recent changes

kemmerle: I was under the impression that I needed to be able to pass the `path` to the file to ensure that the command would work with a `config` flag. Let me try with your suggestion.

kemmerle: As I suspected, if `configFileExists` isn't passed a `path` argument, then it assumes that there is no deprecated config file even when one does exist at the specified path in the `config` flag. See screenshot:

Screenshot 2025-03-26 at 1 45 25 PM

I'm going to resolve this--let me know if you have any questions and I can hop on a huddle.

camden11 (approved):

PR #1399, opened 2025-03-14
Adding context on how to run project commands

brandenrodgers:
## Description and Context

We've seen a couple cases where users aren't aware that they need to CD into the project directory before running the project commands. This adds a tip to the top of the "what's next" section that tells the users where to run the project commands from.

It also adds a section to explain what `hs project upload` does. IMO we should add that because it's an important command.

## Screenshots

image

### DIFF
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -540,7 +540,6 @@ en:
           logs:
             success: "Project {{#bold}}{{ projectName }}{{/bold}} was successfully created in {{ projectDest }}"
             welcomeMessage: "Welcome to HubSpot Developer Projects!"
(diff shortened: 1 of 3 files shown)


lang/en.lyaml
```diff
@@ -1174,7 +1174,7 @@ en:
       disabledUrlText: "See all HubSpot CLI commands here."
       featureHighlight:
         defaultTitle: "What's next?"
-        commandKeys:
+        featureKeys:
```
brandenrodgers, line 1177: I renamed this because these aren't all commands anymore

camden11 (approved):

joe-yeager (approved):

PR #1400, opened 2025-03-14
Update /lib to handle config revamp

camden11:
## Description and Context
This updates all files in `/lib` to use the new config methods from LDL

## Who to Notify

### DIFF
--- a/lib/buildAccount.ts
+++ b/lib/buildAccount.ts
@@ -86,13 +86,13 @@ export async function saveAccountToConfig(
 
 export async function buildDeveloperTestAccount(
   testAccountName: string,
-  parentAccountConfig: HubSpotConfigAccount,
+  parentAccount: HubSpotConfigAccount,
   env: Environment,
   portalLimit: number
 ): Promise<DeveloperTestAccount> {
   const i18nKey = 'lib.developerTestAccount.create.loading';
 
-  const parentAccountId = parentAccountConfig.accountId;
+  const parentAccountId = parentAccount.accountId;
 
   if (!parentAccountId) {
     throw new Error(i18n(`${i18nKey}.fail`));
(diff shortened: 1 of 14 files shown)


lib/buildAccount.ts
```diff
@@ -156,7 +153,7 @@ type SandboxAccount = SandboxResponse & {
 
 export async function buildSandbox(
   sandboxName: string,
-  parentAccountConfig: CLIAccount,
+  parentAccountConfig: HubSpotConfigAccount,
```
brandenrodgers, line 156: Nit, but for consistency should this be `parentAccount`?

lib/developerTestAccounts.ts
```diff
@@ -58,18 +51,18 @@ export async function validateDevTestAccountUsageLimits(
   const limit = data.maxTestPortals;
   const count = data.results.length;
   if (count >= limit) {
-    const hasDevTestAccounts = getHasDevTestAccounts(accountConfig);
+    const hasDevTestAccounts = getHasDevTestAccounts(account);
     if (hasDevTestAccounts) {
       throw new Error(
         i18n('lib.developerTestAccount.create.failure.alreadyInConfig', {
-          accountName: accountConfig.name || accountId,
+          accountName: account.name,
```
brandenrodgers, line 58: Are accounts guaranteed to have names?

lib/links.ts
```diff
@@ -130,7 +129,7 @@ export function openLink(accountId: number, shortcut: string): void {
   }
 
   const baseUrl = getHubSpotWebsiteOrigin(
-    getEnv() === 'qa' ? ENVIRONMENTS.QA : ENVIRONMENTS.PROD
```
brandenrodgers, line 133: Nice to not have to do this anymore 👌

brandenrodgers (approved):
Just a nit but lgtm!

lib/developerTestAccounts.ts
camden11: They are - in the case where there isn't a name, the account Id gets set as the name

PR #1401, opened 2025-03-17
Porting top-level commands to TS (auth and filemanager)

brandenrodgers:
## Description and Context

Ports these files to TS:
- auth
- init
- mv
- list
- remove
- open

## TODO

- [x] Add tests for the ported commands that don't already have them

### DIFF
--- a/commands/auth.ts
+++ b/commands/auth.ts
@@ -69,7 +69,9 @@ type AuthArgs = CommonArgs &
     account?: string;
   };
 
-export async function handler(args: ArgumentsCamelCase<AuthArgs>) {
+export async function handler(
+  args: ArgumentsCamelCase<AuthArgs>
+): Promise<void> {
   const {
     authType: authTypeFlagValue,
     config: configFlagValue,
--- a/commands/auth.ts
+++ b/commands/auth.ts
@@ -220,8 +220,8 @@ export async function handler(
     'accountsListCommand',
   ]);
 
-  const accountId = getAccountId(accountName);
-  await trackAuthAction('auth', authType, TRACKING_STATUS.COMPLETE, accountId!);
+  const accountId = getAccountId(accountName) || undefined;
+  await trackAuthAction('auth', authType, TRACKING_STATUS.COMPLETE, accountId);
 
   process.exit(EXIT_CODES.SUCCESS);
 }
--- a/commands/auth.ts
+++ b/commands/auth.ts
@@ -7,6 +7,7 @@ import {
 } from '@hubspot/local-dev-lib/constants/auth';
 import { ENVIRONMENTS } from '@hubspot/local-dev-lib/constants/environments';
 import { DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME } from '@hubspot/local-dev-lib/constants/config';
+import { AccessToken, CLIAccount } from '@hubspot/local-dev-lib/types/Accounts';
 import { i18n } from '../lib/lang';
 import {
   getAccessToken,
@@ -26,6 +27,7 @@ import {
   personalAccessKeyPrompt,
   OAUTH_FLOW,
   OauthPromptResponse,
+  PersonalAccessKeyPromptResponse,
 } from '../lib/prompts/personalAccessKeyPrompt';
 import { cliAccountNamePrompt } from '../lib/prompts/accountNamePrompt';
 import { setAsDefaultAccountPrompt } from '../lib/prompts/setAsDefaultAccountPrompt';
@@ -40,7 +42,12 @@ import { authenticateWithOauth } from '../lib/oauth';
 import { EXIT_CODES } from '../lib/enums/exitCodes';
 import { uiFeatureHighlight } from '../lib/ui';
 import { logError } from '../lib/errorHandlers/index';
-import { CommonArgs, ConfigArgs, TestingArgs } from '../types/Yargs';
+import {
+  AccountArgs,
+  CommonArgs,
+  ConfigArgs,
+  TestingArgs,
+} from '../types/Yargs';
 
 const i18nKey = 'commands.auth';
 
@@ -64,9 +71,9 @@ export const describe = i18n(`${i18nKey}.describe`, {
 
 type AuthArgs = CommonArgs &
   ConfigArgs &
-  TestingArgs & {
+  TestingArgs &
+  AccountArgs & {
     authType?: string;
-    account?: string;
   };
 
 export async function handler(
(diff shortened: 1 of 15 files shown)


commands/auth.ts
```diff
@@ -64,35 +57,52 @@ const ALLOWED_AUTH_METHODS = [
...
   configName: DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME,
 });
 
-exports.handler = async options => {
+type AuthArgs = CommonArgs &
+  ConfigArgs &
+  TestingArgs & {
+    authType?: string;
+    account?: string;
+  };
```
camden11, line 70: Can use `AccountArgs` here

commands/auth.ts
```diff
@@ -64,35 +57,52 @@ const ALLOWED_AUTH_METHODS = [
...
     process.exit(EXIT_CODES.ERROR);
   }
 
   trackCommandUsage('auth');
-  trackAuthAction('auth', authType, TRACKING_STATUS.STARTED);
+  trackAuthAction(
+    'auth',
+    authType,
+    TRACKING_STATUS.STARTED,
+    providedAccountId!
```
camden11, line 104: Not sure it makes sense to assert that `providedAccountId` exists since it doesn't unless the `--account` flag is used. It looks like we could update `trackAuthAction` to accept `undefined`

commands/auth.ts
```diff
@@ -154,19 +164,27 @@ exports.handler = async options => {
...
       break;
   }
 
   if (!successAuthMethod) {
-    await trackAuthAction('auth', authType, TRACKING_STATUS.ERROR);
+    await trackAuthAction(
+      'auth',
+      authType,
+      TRACKING_STATUS.ERROR,
+      providedAccountId!
```
camden11, line 178: Should this use the account that was just authed if account id wasn't provided?

commands/auth.ts
```diff
@@ -64,35 +57,52 @@ const ALLOWED_AUTH_METHODS = [
...
-  trackAuthAction('auth', authType, TRACKING_STATUS.STARTED);
+  trackAuthAction(
+    'auth',
+    authType,
+    TRACKING_STATUS.STARTED,
+    providedAccountId!
+  );
 
   let configData;
   let updatedConfig;
```
camden11, line 108: Might be good to declare types for these - they're all being interpreted as `any` currently

commands/init.ts
```diff
@@ -53,13 +53,16 @@ const TRACKING_STATUS = {
   COMPLETE: 'complete',
 };
 
-const personalAccessKeyConfigCreationFlow = async (env, account) => {
+async function personalAccessKeyConfigCreationFlow(
+  env: Environment,
+  account?: number
+) {
```
camden11, line 59: Missing return type here + in a few other functions in this file!

commands/init.ts
```diff
@@ -73,95 +76,125 @@ const personalAccessKeyConfigCreationFlow = async (env, account) => {
...
+export const describe = i18n(`${i18nKey}.describe`, {
   configName: DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME,
 });
 
-exports.handler = async options => {
+type InitArgs = CommonArgs &
+  ConfigArgs &
+  TestingArgs & {
+    authType?: string;
+    account?: string;
```
camden11, line 106: Can use `AccountArgs` here too

commands/init.ts
```diff
@@ -73,95 +76,125 @@ const personalAccessKeyConfigCreationFlow = async (env, account) => {
...
     process.exit(EXIT_CODES.ERROR);
   }
 
   if (!disableTracking) {
-    await trackAuthAction('init', authType, TRACKING_STATUS.STARTED);
+    await trackAuthAction(
+      'init',
+      authType,
+      TRACKING_STATUS.STARTED,
+      providedAccountId!
```
camden11, line 153: Same here with using the account you attempted to auth

commands/init.ts
```diff
@@ -73,95 +76,125 @@ const personalAccessKeyConfigCreationFlow = async (env, account) => {
...
+  loadConfig(configPath!, args as CLIOptions);
 
   handleExit(deleteEmptyConfigFile);
 
   try {
-    const { accountId, name } = await CONFIG_CREATION_FLOWS[authType](
-      env,
-      providedAccountId
-    );
+    let accountId: number;
```
camden11, line 179: Remember to declare types for these

commands/open.ts
```diff
@@ -1,24 +1,29 @@
...
+  EnvironmentArgs,
+  AccountArgs,
+} from '../types/Yargs';
 
 const i18nKey = 'commands.open';
 
 const separator = ' => ';
-const createListPrompt = async accountId =>
-  promptUser([
+async function createListPrompt(accountId: number) {
```
camden11, line 23: Missing return type here!

commands/init.ts
brandenrodgers: Done in [be71b49](https://github.com/HubSpot/hubspot-cli/pull/1401/commits/be71b495f9f24049001fda5d7c3328c7479fcdc8)

commands/init.ts
brandenrodgers: I don't think we can do this because we won't know the account ID until the request for the access token comes back. We'll only know it if the user passes in a `providedAccountId`. So in most cases this will be undefined, but I figured it might be nice to pass it in whenever it does exist. Not sure how much value we'll get out of it though 🤷

commands/init.ts
brandenrodgers: Done in [be71b49](https://github.com/HubSpot/hubspot-cli/pull/1401/commits/be71b495f9f24049001fda5d7c3328c7479fcdc8)

commands/init.ts
brandenrodgers: Done in [be71b49](https://github.com/HubSpot/hubspot-cli/pull/1401/commits/be71b495f9f24049001fda5d7c3328c7479fcdc8)

commands/auth.ts
brandenrodgers: Good call, done in [be71b49](https://github.com/HubSpot/hubspot-cli/pull/1401/commits/be71b495f9f24049001fda5d7c3328c7479fcdc8)

commands/auth.ts
brandenrodgers: Good call, done in [be71b49](https://github.com/HubSpot/hubspot-cli/pull/1401/commits/be71b495f9f24049001fda5d7c3328c7479fcdc8)

commands/auth.ts
brandenrodgers: Done in [be71b49](https://github.com/HubSpot/hubspot-cli/pull/1401/commits/be71b495f9f24049001fda5d7c3328c7479fcdc8)

commands/open.ts
brandenrodgers: Done in [577341a](https://github.com/HubSpot/hubspot-cli/pull/1401/commits/577341aac466df1118386022e479b9809a1104d1)

commands/auth.ts
```diff
@@ -198,12 +221,16 @@ exports.handler = async options => {
   ]);
 
   const accountId = getAccountId(accountName);
-  await trackAuthAction('auth', authType, TRACKING_STATUS.COMPLETE, accountId);
+  await trackAuthAction('auth', authType, TRACKING_STATUS.COMPLETE, accountId!);
```
camden11, line 224: nit: now that you've updated `accountId` to accept `undefined`, you don't need to assert that it exists

camden11 (dismissed):

PR #1405, opened 2025-03-21
Update remaining /lib files to use new config

camden11:
## Description and Context
This updates the lib files in `/doctor`, `/projects`, and `/prompts` to use the new non-instance based config

## Who to Notify


joe-yeager:
@camden11 Looks like the build is breaking and showing a couple files have bad imports.

camden11:
> @camden11 Looks like the build is breaking and showing a couple files have bad imports.

Ah yeah, this is just due to the WIP nature of this. I'm using an experimental version of LDL that has the new config changes, so the files I haven't updated yet will have import issues. But i'm not merging this into main, just my base branch for the config revamp

joe-yeager (approved):

PR #1406, opened 2025-03-21
feat: Add `hs app migrate` command 

joe-yeager:
## Description and Context

This PR does the following:
- Deprecates `hs project migrate-app` and `hs project clone-app`
- Moves the `hs project migrate-app` command to a new command `hs app migrate` 
- Adds a `--platform-version` flag to both commands that when added migrates the app to revamped API
- Adds the functionality to migrate a non-projects app to the revamped API

## Testing steps
- Create a non project app inside a Developer account, and leave it empty.  Don't add any components, currently only empty apps are allowed as apps is the only team that has onboarded to the system.
- Auth into that account and run `hs app migrate --platform-version unstable`
- Select the app you would like migrate and confirm what component types will be migrated
- Follow the prompts and enter a project name, a dest for the source code
- You will be prompted to enter a UID for your application, enter one and hit enter
- Wait for the migration to complete and it should download the source code in the provided location


lang/en.lyaml
```diff
@@ -566,6 +566,7 @@ en:
           header:
             text: "Migrate an app to the projects framework"
             link: "Learn more about migrating apps to the projects framework"
+          deprecationWarning: "The {{ oldCommand }} command is deprecated and will be removed. Use {{ newCommand }} going forward."
```
joe-yeager, line 569: This is what will show when `hs project migrate-app` is ran

lang/en.lyaml
```diff
@@ -582,10 +583,32 @@ en:
...
+            noApps: "No apps found in account {{ accountId }}"
+            noAppsEligible: "No apps in account {{ accountId }} are eligible for migration"
             invalidAccountTypeTitle: "{{#bold}}Developer account not targeted{{/bold}}"
             invalidAccountTypeDescription: "Only public apps created in a developer account can be converted to a project component. Select a connected developer account with {{useCommand}} or {{authCommand}} and try again."
             projectAlreadyExists: "A project with name {{ projectName }} already exists. Please choose another name."
             invalidApp: "Could not migrate appId {{ appId }}. This app cannot be migrated at this time. Please choose another public app."
+          prompt:
+            chooseApp: 'Choose the app you want to migrate: '
+            inputName: '[--name] Enter the name for the project'
+            inputDest: '[--dest] Where do you want to save the project?'
```
j0b0sapi3n, line 595: I'm thinking to either change this to `Enter the destination for the project` to make it consistent with `inputName`, or to update `inputName to `What do you want to name the project?`

Do you have a preference?

lang/en.lyaml
```diff
@@ -582,10 +583,32 @@ en:
...
+            finishingMigration: "Finishing migration"
+            migrationComplete: "Migration complete"
+            migrationFailed: "Migration failed"
+            downloadingProjectContents: "Fetching migrated project"
+            downloadingProjectContentsComplete: "Fetching migrated project complete"
+            downloadingProjectContentsFailed: "Unable to fetch migrated project"
+          migrationFailureReasons:
+            upToDate: 'App is already up to date'
+            isPrivateApp: 'Private apps are not currently migratable'
+            listedInMarketplace: 'App is listed in marketplace'
```
j0b0sapi3n, line 610: Do you know what the resolution for this one is? Do they have to unlist the app to migrate?

lang/en.lyaml
joe-yeager: Either sound fine to me, I will defer to you @j0b0sapi3n

lang/en.lyaml
joe-yeager: I just checked and the current resolution would be to wait until they are migratable, so similar to Private Apps.

lang/en.lyaml
j0b0sapi3n: Let's go with this for the prompts so they're all consistent. Let me know what you think:
```
chooseApp: 'Which app would you like to migrate?'
inputName: '[--name] What would you like to name the project?'
inputDest: '[--dest] Where would you like to save the project?'
uidForComponent: 'What UID would you like to use for {{ componentName }}?'
```

lang/en.lyaml
j0b0sapi3n: Got it. Can we update these error message to this then:

- App migration failed because this app is already up to date.
- App migration failed because the app is private. We're working on adding support for migrating private apps in the future.
- App migration failed because the app is listed in the marketplace. We're working on adding support for migrating listed apps in the future.

commands/customObject/schema/update.ts
```diff
@@ -65,7 +65,7 @@ export async function handler(
     name =
       providedName && typeof providedName === 'string'
         ? providedName
-        : await listPrompt(i18n(`commands.customObject.subcommands.schema.subcommands.update.selectSchema`), {
+        : await listPrompt<string>(i18n(`commands.customObject.subcommands.schema.subcommands.update.selectSchema`), {
```
joe-yeager, line 68: This one particular call got a little upset with the generic and was requiring an explicit `<string>` 🤷

commands/app/__tests__/migrate.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        argv: { _: ['app', 'migrate'] },
+      } as unknown as Argv;
+    });
+
+    it('should add required options', async () => {
+      await builder(mockYargs);
+
+      expect(mockYargs.options).toHaveBeenCalledWith(
+        expect.objectContaining({
+          name: expect.any(Object),
```
kemmerle, line 107: Does `expect.any(Object)` work with all of these options? I would expect `name` and `dest` to be strings, etc.

commands/app/__tests__/migrate.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        platformVersion: PLATFORM_VERSIONS.v2023_2,
+      } as ArgumentsCamelCase<MigrateAppOptions>);
+
+      expect(mockedLogger.error).toHaveBeenCalled();
+      expect(exitSpy).toHaveBeenCalledWith(EXIT_CODES.ERROR);
+      exitSpy.mockRestore();
+    });
+  });
+
+  describe('builder', () => {
```
kemmerle, line 89: Might be good to add a stub test or note to add tests for the description, as well (when we unhide the command).

commands/app/migrate.ts
```diff
@@ -0,0 +1,136 @@
...
+  logger.log(
+    uiBetaTag(
+      i18n(`commands.project.subcommands.migrateApp.header.text`),
+      false
+    )
+  );
+  logger.log(
+    uiLink(
+      i18n(`commands.project.subcommands.migrateApp.header.link`),
+      'https://developers.hubspot.com/docs/platform/migrate-a-public-app-to-projects'
```
kemmerle, line 49: Do we want to store this URL anywhere?

commands/app/migrate.ts
```diff
@@ -0,0 +1,136 @@
...
+    logger.error(
+      i18n(`commands.project.subcommands.migrateApp.errors.noAccountConfig`)
+    );
+    return process.exit(EXIT_CODES.ERROR);
+  }
+
+  logger.log('');
+  logger.log(
+    uiBetaTag(
+      i18n(`commands.project.subcommands.migrateApp.header.text`),
```
kemmerle, line 42: Can you shorten this with an `i18nKey` defined above?

commands/app/migrate.ts
```diff
@@ -0,0 +1,136 @@
...
+      'errors' in error &&
+      Array.isArray(error.errors)
+    ) {
+      error.errors.forEach(err => logError(err));
+    } else {
+      logError(error, new ApiErrorContext({ accountId: derivedAccountId }));
+    }
+    await trackCommandMetadataUsage(
+      'migrate-app',
+      { status: 'FAILURE' },
```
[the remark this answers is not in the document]
kemmerle: Same goes for below

commands/app/migrate.ts
joe-yeager: We are actively removing the string interpolation with the `i18nKeys` because they make it harder to determine which string we are using and having the full string will make the migration to the native JS object easier.

commands/project/migrateApp.ts
```diff
@@ -1,283 +1,86 @@
...
-    process.exit(EXIT_CODES.ERROR);
-  }
-  await trackCommandMetadataUsage(
-    'migrate-app',
-    { status: 'SUCCESS' },
-    derivedAccountId
-  );
-  process.exit(EXIT_CODES.SUCCESS);
-};
+export function builder(yargs: Argv): Argv<MigrateAppOptions> {
```
kemmerle, line 36: Will it confuse folks to see these options listed in the help menu when this is technically the deprecated command?

lib/app/migrate.ts
```diff
@@ -0,0 +1,584 @@
...
+import { extractZipArchive } from '@hubspot/local-dev-lib/archive';
+import { ArgumentsCamelCase } from 'yargs';
+import { MigrateAppOptions } from '../../types/Yargs';
+import chalk from 'chalk';
+import { validateUid } from '@hubspot/project-parsing-lib';
+import { MigrationApp } from '@hubspot/local-dev-lib/types/Project';
+import { UNMIGRATABLE_REASONS } from '@hubspot/local-dev-lib/constants/projects';
+import { mapToUserFacingType } from '@hubspot/project-parsing-lib/src/lib/transform';
+
+function getUnmigratableReason(reasonCode: string) {
```
kemmerle, line 47: Don't forget to add return types on these functions.

lib/app/migrate.ts
```diff
@@ -0,0 +1,584 @@
...
+  switch (reasonCode) {
+    case UNMIGRATABLE_REASONS.UP_TO_DATE:
+      return i18n(
+        'commands.project.subcommands.migrateApp.migrationNotAllowedReasons.upToDate'
```
kemmerle, line 51: `i18nKey` here?

kemmerle: Also, nit: you could change the `migrationNotAllowedReasons` key to `unmigratableReasons` for readability.

lib/app/migrate.ts
```diff
@@ -0,0 +1,584 @@
...
+    );
+  }
+
+  if (
+    appId &&
+    !allApps.some(app => {
+      return app.appId === appId;
+    })
+  ) {
+    throw new Error(
```
kemmerle, line 108: Do we want to throw an error here, before the user chooses the app to migrate? Would that be confusing?

lib/app/migrate.ts
```diff
@@ -0,0 +1,584 @@
...
+    );
+    appIdToMigrate = selectedAppId;
+  }
+
+  const selectedApp = allApps.find(app => app.appId === appIdToMigrate);
+
+  const migratableComponents: string[] = [];
+  const unmigratableComponents: string[] = [];
+
+  selectedApp?.migrationComponents.forEach(component => {
```
kemmerle, line 141: I think you mean `migratableComponents` here.

commands/app/__tests__/migrate.test.ts
joe-yeager: This is checking the options calls and not the type of the option.  I will expand it because it's a pretty generic test.

commands/app/migrate.ts
joe-yeager: It's only used in this one spot, so I think we are probably fine.

lib/app/migrate.ts
```diff
@@ -0,0 +1,584 @@
...
+      )
+    );
+  }
+
+  const projectDest =
+    dest ||
+    (await inputPrompt(
+      i18n('commands.project.subcommands.migrateApp.prompt.inputDest'),
+      {
+        defaultAnswer: path.resolve(getCwd(), projectName),
```
kemmerle, line 205: Make sure to sanitize the project name for the destination path, in case someone decides to be smart and name their project 💩 💩 💩

commands/project/migrateApp.ts
joe-yeager: The only new option added is the `--platform-version` and it is hidden, so I think we are good.

lib/app/migrate.ts
joe-yeager: This is checking if the user provided and appId that is not in the list of apps, so they have already chosen an app.

lib/app/migrate.ts
joe-yeager: Will change the key

lib/app/migrate.ts
```diff
@@ -0,0 +1,584 @@
...
+        });
+
+  try {
+    const { data: selectedApp } = await fetchPublicAppMetadata(
+      appId,
+      derivedAccountId
+    );
+    // preventProjectMigrations returns true if we have not added app to allowlist config.
+    // listingInfo will only exist for marketplace apps
+    const preventProjectMigrations = selectedApp.preventProjectMigrations;
```
kemmerle, line 458: Are we reusing the `preventProjectMigrations` property from the last migration cycle?

lib/app/migrate.ts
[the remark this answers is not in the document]
joe-yeager: Done

lib/app/migrate.ts
```diff
@@ -0,0 +1,584 @@
...
+    throw new Error(
+      i18n(`${i18nKey}.errors.projectAlreadyExists`, {
+        projectName,
+      })
+    );
+  }
+
+  await trackCommandMetadataUsage(
+    'migrate-app',
+    { status: 'STARTED' },
```
kemmerle, line 491: Same note as above--I think the closest allowed property would be `{ step: 'STARTED' }`.

lib/usageTracking.ts
```diff
@@ -22,6 +22,7 @@ type Meta = {
   type?: string | number; // "The upload type"
   file?: boolean; // "Whether or not the 'file' flag was used"
   successful?: boolean; // "Whether or not the CLI interaction was successful"
+  status?: string; // Is the action started, successful, or failed
```
kemmerle, line 25: Ooh, didn't realize this was a part of the API!

commands/app/migrate.ts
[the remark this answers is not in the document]
kemmerle: No, I was incorrect in the previous migration command, which we only discovered later. The API only accepts certain properties in the metadata.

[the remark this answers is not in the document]
joe-yeager: Gotcha, I'll update it

lib/app/migrate.ts
joe-yeager: No, we aren't reusing much which is why it was split into multiple methods.

lib/app/migrate.ts
joe-yeager: Done

package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/HubSpot/hubspot-cli",
   "dependencies": {
-    "@hubspot/local-dev-lib": "3.4.1",
+    "@hubspot/local-dev-lib": "0.3.0-experimental.0",
```
joe-yeager, line 8: Reminder to update once https://github.com/HubSpot/hubspot-local-dev-lib/pull/255 is merged and published

lib/app/migrate.ts
```diff
@@ -0,0 +1,608 @@
...
+  logger.log(
+    i18n(`${i18nKey}.errors.invalidAccountTypeDescription`, {
+      useCommand: uiCommandReference('hs accounts use'),
+      authCommand: uiCommandReference('hs auth'),
+    })
+  );
+  uiLine();
+}
+
+export async function migrateApp2023_2(
```
joe-yeager, line 453: This method is pretty much a direct copy paste

lib/app/migrate.ts
```diff
@@ -0,0 +1,638 @@
+import { CLIAccount } from '@hubspot/local-dev-lib/types/Accounts';
```
brandenrodgers, line 1: This is kind of a nit, but would you mind ordering these imports? Mixing all the third-party imports with the local imports makes it hard to figure out if there's already an existing import from any given path.

lib/app/migrate.ts
brandenrodgers: Not a blocker, but one alternate option is to move this into a dedicated legacy file just to keep the new logic clean (and make this file a little easier to parse)

commands/app/migrate.ts
```diff
@@ -0,0 +1,136 @@
...
+  logger.log(
+    uiLink(
+      i18n(`commands.project.subcommands.migrateApp.header.link`),
+      'https://developers.hubspot.com/docs/platform/migrate-a-public-app-to-projects'
+    )
+  );
+  logger.log('');
+
+  try {
+    if (platformVersion === v2025_2 || platformVersion === unstable) {
```
brandenrodgers, line 55: Should the user have to include the "v" in these? The actual value in the config doesn't include the v

lib/app/migrate.ts
joe-yeager: Yeah, I'll move it.  I had the thought of making it a second file, and it was easier to have them side by side during development, but now the file is pretty large.

commands/app/migrate.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  await trackCommandMetadataUsage(
+    'migrate-app',
+    { successful: true },
+    derivedAccountId
+  );
+  return process.exit(EXIT_CODES.SUCCESS);
+}
+
+export function builder(yargs: Argv): Argv<MigrateAppOptions> {
```
brandenrodgers, line 87: The `makeYargsBuilder` util should be available to use now! [Here's an example of how to use it](https://github.com/HubSpot/hubspot-cli/blob/main/commands/project/download.ts#L130). Even if this command isn't using a short and long description, you'll still get a benefit because it colors the command name and highlights the headers in the help output.

commands/app/migrate.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+    await trackCommandMetadataUsage(
+      'migrate-app',
+      { successful: false },
+      derivedAccountId
+    );
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  await trackCommandMetadataUsage(
```
brandenrodgers, line 79: I'm not sure what the best way to do this would be, but it might be nice to be able to differentiate between this new migration and the soon-to-be-deprecated one in our tracking

joe-yeager: Yeah, maybe I could add a tracking to the `hs project migrate-app` handler that is like `migrate-app-deprecated`?  That seems kind of gross, but we can at least subtract those from the `migrate-app` total and get an idea.

commands/app/migrate.ts
joe-yeager: @brandenrodgers It looks like this might not work for hidden commands because it is expecting that describe is defined.  Maybe I can add it when we unhide the command.  I guess I could make the `describe` argument `string | undefined` but I'm afraid that might open a can of works because I don't fully understand the `makeYargsBuilder` flow

lib/app/migrate.ts
joe-yeager: Moved

api/migrate.ts
```diff
@@ -0,0 +1,138 @@
...
+
+interface BaseMigrationApp {
+  appId: number;
+  appName: string;
+  isMigratable: boolean;
+  migrationComponents: ListAppsMigrationComponent[];
+  projectName?: string;
+}
+
+export interface MigratableApp extends BaseMigrationApp {
```
brandenrodgers, line 19: Should these go in the `types/` folder?

commands/app/migrate.ts
brandenrodgers: Yeah something like that could work. You could also call this one `app-migrate` and differentiate that way

brandenrodgers: But the more I think about it. Maybe it's not really important to differentiate. I'm not sure if we'd even need that info

brandenrodgers (approved):
Lgtm! I ran through some local tests and it all seems to be WAD

api/migrate.ts
joe-yeager: I left them here because we don't really seem to have a hard a fast rule for putting them in the `types/` folder. We seem to have types living in files next to the methods that use them as return types, so I went this that approach since they don't really make sense as standalone types.

commands/app/migrate.ts
joe-yeager: Gotcha, I'll leave it alone as is.  Actually in hindsight, we are going to deprecate and delete `hs project migrate-app` so once we release the next major we will only have `hs app migrate`

api/migrate.ts
joe-yeager: That being said I don't have a strong opinion and if we want to move this to types LMK and I will move them in a follow up, going to merge this as is.

PR #1407, opened 2025-03-25
Refactor `hs account list` and `hs account info` for override files

kemmerle:
## Description and Context
This PR refactors the `hs account list` and `hs account info` commands to account for `.hsaccount` override files with global config. 

Relies on https://github.com/HubSpot/hubspot-local-dev-lib/pull/249. I've created an experimental release for ease of testing. 

**NOTE**: It occurs to me that we need to alert the user when a `.hsaccount` override file is present when a `hubspot.config.yml` file exists, but I think we should add this change in the main PR: https://github.com/HubSpot/hubspot-cli/pull/1349. 

## Screenshots

Global config with no account override:

`hs account ls`

Screenshot 2025-03-25 at 1 08 37 PM

`hs account info`

Screenshot 2025-03-25 at 1 09 06 PM

Global config **with** account override:

`hs account ls`

Screenshot 2025-03-25 at 1 11 08 PM

`hs account info`

Screenshot 2025-03-25 at 1 17 00 PM

## TODO

- [ ] Address feedback

## Who to Notify


lib/ui/index.ts
```diff
@@ -192,3 +192,7 @@ export function uiDeprecatedMessage(
   uiDeprecatedTag(tag);
   logger.log();
 }
+
+export function indent(str: string, spacing?: string) {
```
camden11, line 196: There's actually already an `indent` function in the `Diagnosis` class used in doctor - could probably pull that one out into `lib/ui`. I think the way it accepts a number of indentions rather than a string might be a bit easier to use as well

commands/account/list.ts
```diff
@@ -114,15 +115,39 @@ export async function handler(
...
+    logger.log(i18n(`${i18nKey}.overrideFilePathTitle`));
+    logger.log(
+      indent(1) + i18n(`${i18nKey}.overrideFilePath`, { overrideFilePath })
+    );
+    logger.log(
+      indent(1) +
+        i18n(`${i18nKey}.overrideAccount`, {
+          account: getConfigDefaultAccount()!,
+        })
+    );
```
joe-yeager, line 148: Same nit for these

commands/account/list.ts
```diff
@@ -114,15 +115,39 @@ export async function handler(
...
+        i18n(`${i18nKey}.configPath`, {
+          configPath,
+        })
+    );
+    logger.log(
+      indent(1) +
+        i18n(`${i18nKey}.defaultAccount`, {
+          account: getDisplayDefaultAccount()!,
+        })
+    );
```
joe-yeager, line 132: Nit: Can we use template strings instead of string concatenation?

package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "repository": "https://github.com/HubSpot/hubspot-cli",
   "dependencies": {
-    "@hubspot/local-dev-lib": "3.3.1",
+    "@hubspot/local-dev-lib": "0.2.5-experimental.0",
```
joe-yeager, line 8: Reminder to change this before merging

joe-yeager (approved):

camden11 (approved):
Nice! LGTM

PR #1408, opened 2025-03-26
Port project deploy and add to TS

brandenrodgers:
## Description and Context

Porting a few more files to TS and adding tests:
- Project deploy
- Project add

### DIFF
--- a/commands/project/__tests__/deploy.test.ts
+++ b/commands/project/__tests__/deploy.test.ts
@@ -1,29 +1,29 @@
-// @ts-nocheck
-const { AxiosError, HttpStatusCode } = require('axios');
-const yargs = require('yargs');
-const chalk = require('chalk');
-const {
-  HubSpotHttpError,
-} = require('@hubspot/local-dev-lib/models/HubSpotHttpError');
-const { getAccountConfig } = require('@hubspot/local-dev-lib/config');
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const {
-  deployProject,
-  fetchProject,
-} = require('@hubspot/local-dev-lib/api/projects');
-const ui = require('../../../lib/ui');
-const {
+import { HttpStatusCode } from 'axios';
+import yargs, { Argv, ArgumentsCamelCase } from 'yargs';
+import chalk from 'chalk';
+import * as configUtils from '@hubspot/local-dev-lib/config';
+import { logger } from '@hubspot/local-dev-lib/logger';
+import * as projectApiUtils from '@hubspot/local-dev-lib/api/projects';
+import * as ui from '../../../lib/ui';
+import {
   addAccountOptions,
   addConfigOptions,
   addUseEnvironmentOptions,
-} = require('../../../lib/commonOpts');
-const { getProjectConfig } = require('../../../lib/projects');
-const { getProjectDetailUrl } = require('../../../lib/projects/urls');
-const { pollDeployStatus } = require('../../../lib/projects/buildAndDeploy');
-const { projectNamePrompt } = require('../../../lib/prompts/projectNamePrompt');
-const { promptUser } = require('../../../lib/prompts/promptUtils');
-const { trackCommandUsage } = require('../../../lib/usageTracking');
-const { EXIT_CODES } = require('../../../lib/enums/exitCodes');
+} from '../../../lib/commonOpts';
+import * as projectUtils from '../../../lib/projects';
+import * as projectUrlUtils from '../../../lib/projects/urls';
+import { pollDeployStatus } from '../../../lib/projects/buildAndDeploy';
+import * as projectNamePrompt from '../../../lib/prompts/projectNamePrompt';
+import * as promptUtils from '../../../lib/prompts/promptUtils';
+import { trackCommandUsage } from '../../../lib/usageTracking';
+import { EXIT_CODES } from '../../../lib/enums/exitCodes';
+import * as deployCommand from '../deploy';
+import { ProjectConfig } from '../../../types/Projects';
+import exampleProject from './fixtures/exampleProject.json';
+import {
+  mockHubSpotHttpResponse,
+  mockHubSpotHttpError,
+} from '../../../lib/testUtils';
 
 jest.mock('yargs');
 jest.mock('@hubspot/local-dev-lib/logger');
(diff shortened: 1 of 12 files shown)


commands/project/deploy.ts
```diff
@@ -135,24 +151,21 @@ exports.handler = async options => {
...
       useV3Api(projectConfig?.platformVersion)
     );
 
-    if (!deployResp || deployResp.error) {
-      logger.error(
-        i18n(`${i18nKey}.errors.deploy`, {
-          details: deployResp.error.message,
-        })
-      );
+    // TODO: Validate this change. Were we actually using deployResp.error.message?
```
brandenrodgers, line 159: I have a question about this one. `deployResp.error` doesn't exist on the response object, but we were logging out `deployResp.error.message`. If `deployResp` doesn't exist, then we'd be getting JS errors. So I assume this was never being triggered, but I wanted to double check just in case.

commands/project/__tests__/deploy.test.ts
```diff
@@ -90,127 +108,132 @@ describe('commands/project/deploy', () => {
...
-      expect(getProjectConfig).toHaveBeenCalledWith();
-    });
-
-    it('should load the project config', async () => {
-      await deployCommand.handler(options);
-      expect(getProjectConfig).toHaveBeenCalledTimes(1);
-      expect(getProjectConfig).toHaveBeenCalledWith();
+      await deployCommand.handler(args);
+      expect(getProjectConfigSpy).toHaveBeenCalledTimes(1);
+      expect(getProjectConfigSpy).toHaveBeenCalledWith();
```
kemmerle, line 178: Can this be replaced with `toHaveBeenCalled`, if it doesn't take any arguments?

commands/project/add.ts
```diff
@@ -142,5 +147,18 @@ exports.builder = yargs => {
...
+  return yargs as Argv<ProjectAddArgs>;
+}
+
+export const builder = makeYargsBuilder<ProjectAddArgs>(
+  projectAddBuilder,
+  command,
+  describe!
+);
+
+module.exports = {
```
kemmerle, line 159: Do we need this? Aren't the individual exports enough?

kemmerle (dismissed):
I had a few minor questions but otherwise LGTM :shipit:

commands/project/add.ts
brandenrodgers: Once I port `commands/projects.ts` to TS I'll be able to remove all of these `module.exports`. The issue is that the project.ts file is still commonJS and using requires, which means I'll need to keep these until I port that file. But I'm going to do that as soon as I finish up the rest of the project commands.

commands/project/__tests__/fixtures/exampleProject.json
```diff
@@ -0,0 +1,34 @@
...
+    "uploadMessage": "Upload message",
+    "autoDeployId": 123
+  },
+  "deployedBuildId": 1,
+  "createdAt": 123456,
+  "deletedAt": 123456,
+  "updatedAt": 123456,
+  "id": 123,
+  "isLocked": false,
+  "links": {},
```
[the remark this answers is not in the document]
brandenrodgers: good catch, I don't think this field even lives on the project object. Not sure why I put it in there

commands/project/__tests__/deploy.test.ts
brandenrodgers: Yep, done in [da4a9ed](https://github.com/HubSpot/hubspot-cli/pull/1408/commits/da4a9edf1299c72980f7ebfffe19ebd5c44c8f6b) 👍

commands/project/deploy.ts
```diff
@@ -1,50 +1,60 @@
...
+): boolean | string {
   if (Number(buildId) > latestBuildId) {
     return i18n(`${i18nKey}.errors.buildIdDoesNotExist`, {
       buildId: buildId,
-      projectName,
+      projectName: projectName!,
       linkToProject: uiLink(
         i18n(`${i18nKey}.errors.viewProjectsBuilds`),
-        getProjectDetailUrl(projectName, accountId)
+        getProjectDetailUrl(projectName!, accountId)!
```
camden11, line 57: I think the updates you made to the `projectNamePrompt` in your other PR should let you get rid of a lot of these non-null assertions. `validateBuildId` could _probably_ just accept a string for `projectName`

commands/project/deploy.ts
```diff
@@ -53,20 +63,22 @@ const validateBuildId = (
...
+  args: ArgumentsCamelCase<ProjectDeployArgs>
+): Promise<void> {
+  const { derivedAccountId } = args;
   const accountConfig = getAccountConfig(derivedAccountId);
-  const { project: projectOption, buildId: buildIdOption } = options;
+  const { project: projectOption, buildId: buildIdOption } = args;
   const accountType = accountConfig && accountConfig.accountType;
 
-  trackCommandUsage('project-deploy', { type: accountType }, derivedAccountId);
+  trackCommandUsage('project-deploy', { type: accountType! }, derivedAccountId);
```
camden11, line 81: `accountType` could still be undefined - not sure if this will affect tracking but would it be better to just pass in `undefined` instead of this object if `accountType` is undeinfed?

commands/project/deploy.ts
camden11: Seems safe to get rid of!

commands/project/deploy.ts
```diff
@@ -132,24 +144,21 @@ exports.handler = async options => {
...
+      logger.error(i18n(`${i18nKey}.errors.deploy`));
       return process.exit(EXIT_CODES.ERROR);
     }
 
     await pollDeployStatus(
       derivedAccountId,
-      projectName,
-      deployResp.id,
+      projectName!,
+      Number(deployResp.id),
```
camden11, line 161: I would double check what type this actually is - it might be mistyped in LDL, or we might want to update `pollDeployStatus` to accept a string

camden11 (dismissed):
Just a few minor comments but looks good!

commands/project/deploy.ts
brandenrodgers: I double checked this in the api catalog and the api does in fact return the id as a string. I could update the polling functions to take it in as a string, but it looks like it's actually more convenient to treat them as numbers in that function ([here](https://github.com/HubSpot/hubspot-cli/blob/main/lib/projects/buildAndDeploy.ts#L152), for example).

camden11 (approved):

PR #1409, opened 2025-03-27
Refactor `hs account remove` and `hs account clean` commands for default account override

kemmerle:
## Description and Context
In this PR, I've refactored the `hs account remove` and `hs account clean` commands to work with the default account override files we've introduced as part of our config revamp. 

**Note** In `hs account clean`, I've also added a prompt to replace the default account if it's removed. This may be a breaking change, so I wanted to highlight it. 

## Screenshots

`hs account clean` (with default account override present and with prompt to replace default account):

Screenshot 2025-03-27 at 10 17 19 AM

`hs account remove` (with default account override present):

Screenshot 2025-03-27 at 10 21 46 AM

## TODO

- [ ] Get OK from the team to add the prompt to update the default account.
- [ ] Get OK from Jono about the copy.

## Who to Notify

### DIFF
--- a/commands/account/clean.ts
+++ b/commands/account/clean.ts
@@ -1,12 +1,16 @@
+import fs from 'fs';
 import { Argv, ArgumentsCamelCase } from 'yargs';
 import { logger } from '@hubspot/local-dev-lib/logger';
 import { accessTokenForPersonalAccessKey } from '@hubspot/local-dev-lib/personalAccessKey';
 import {
   deleteAccount,
   getConfigAccounts,
+  getConfigDefaultAccount,
 } from '@hubspot/local-dev-lib/config';
 import { getAccountIdentifier } from '@hubspot/local-dev-lib/config/getAccountIdentifier';
 import { isSpecifiedError } from '@hubspot/local-dev-lib/errors/index';
+import { getDefaultAccountOverrideFilePath } from '@hubspot/local-dev-lib/config';
+
 import { trackCommandUsage } from '../../lib/usageTracking';
 import { i18n } from '../../lib/lang';
 import { EXIT_CODES } from '../../lib/enums/exitCodes';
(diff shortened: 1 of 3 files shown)


commands/account/clean.ts
```diff
@@ -94,22 +106,50 @@ export async function handler(
...
+    );
+
+    const accountOverride = getCWDAccountOverride();
+    const overrideFilePath = getDefaultAccountOverrideFilePath();
+    if (overrideFilePath && accountOverride) {
+      if (
+        accountsToRemove.some(
+          account =>
+            account.name === accountOverride ||
+            // @ts-ignore: Default account override files can only exist with global config
```
joe-yeager, line 124: Nit: `@ts-expect-error` has more desirable behavior than `@ts-ignore`, because when there comes a day this is no longer a ts error we will get a warning that it is unused and can be removed.  [Example](https://dev.to/maafaishal/ts-expect-error-over-ts-ignore-in-typescript-5140)

commands/account/clean.ts
```diff
@@ -94,22 +106,50 @@ export async function handler(
...
+      oneAccountFound ? `${i18nKey}.confirm.one` : `${i18nKey}.confirm.other`,
+      {
+        count: accountsToRemove.length,
+      }
+    );
+
+    const accountOverride = getCWDAccountOverride();
+    const overrideFilePath = getDefaultAccountOverrideFilePath();
+    if (overrideFilePath && accountOverride) {
+      if (
```
joe-yeager, line 120: Nit: These if statements can be merged.  Secondary nit:  The call to `.some` is a little long and might benefit from being stored in a variable

commands/account/clean.ts
```diff
@@ -94,22 +106,50 @@ export async function handler(
...
+            // @ts-ignore: Default account override files can only exist with global config
+            account.accountId === accountOverride
+        )
+      ) {
+        promptMessage =
+          promptMessage +
+          i18n(`${i18nKey}.defaultAccountOverride`, {
+            overrideFilePath,
+          });
+      }
```
joe-yeager, line 133: Nit: This can be a template string

joe-yeager (commented):
I'll go deeper after the UX discussion, but in the meantime I just have a few nits.

camden11 (approved):
Code all looks good to me. As long as we're good with the functionality changes here I think this should be good to go. Will have to figure out if this warrants a major release

PR #1410, opened 2025-03-28
Feature: Add `hs account remove-override` (Also refactor `hs account use` and `create-override`)

kemmerle:
## Description and Context
This PR finishes up the `hs account` bucket refactor as part of the global config work. After our UX sync on 3/27, we made some changes to the plans laid out in the issues. The changes are:

`hs account use`: If an override file is present, I've added a message with the file path and instructions on which commands (`hs account create-override, `hs account remove-override`) to use to change/remove it. 

`hs account create-override`: I've added a prompt if an override file already exists to confirm that the user wants to overwrite it. We decided on one command to both create/update an override file to mimic the behavior of `hs account auth`.

`hs account remove-override`: In Thursday's UX sync, we decided to add a command to remove the override file, so that users wouldn't need to manually remove it. 

## Screenshots

`hs account use` (when override file is present):

Screenshot 2025-03-28 at 12 24 09 PM

`hs account create-override` (when override file is already present):

Screenshot 2025-03-28 at 1 00 02 PM

`hs account remove-override` (when override file is present):

Screenshot 2025-03-28 at 1 03 15 PM

`hs account remove-override --force` (when override file and force flag are present):

Screenshot 2025-03-28 at 1 04 28 PM

`hs account remove-override` (when override file is **not** present):

Screenshot 2025-03-28 at 1 06 04 PM

## TODO

- [ ] Address feedback
- [ ] Get feedback on copy from Jono

## Who to Notify


j0b0sapi3n:
lgtm!

lang/en.lyaml
```diff
@@ -67,6 +80,8 @@ en:
           describe: "Set the Hubspot account to use as the default account. The default account can be overridden with the \"--account\" option."
           errors:
             accountNotFound: "The account \"{{ specifiedAccount }}\" could not be found in {{ configPath }}"
+          accountOverride: "There is an account override file at {{ overrideFilePath }} associated with account \"{{ accountOverride }}\"."
+          accountOverrideCommands: "Use {{ createOverrideCommand }} to change override accounts, or {{ removeOverrideCommand }} to remove the existing one."
```
camden11, line 84: I think this copy could maybe make it a bit more clear that the override will prevent the account you've selected from working in this project. But no strong feelings if you both are good with this @j0b0sapi3n @kemmerle

camden11: I also think in general referring to the account override file to this extent is kind of exposing implementation details to users?

camden11 (approved):
Code LGTM! Just a few thoughts about copy

lang/en.lyaml
j0b0sapi3n: @camden11 What would you suggest as an alternative? Just to compare

camden11: Hmmm maybe something along the lines of

[WARNING] This project currently has a account override set for account {{ account }}. {{account}} will continue to act as the default account for this project. Use {{ createOverrideCommand }} to change override accounts, or {{ removeOverrideCommand }} to remove the override and use your default account.

j0b0sapi3n: Yeah I don't mind hiding the implementation details. Here's one that's slightly shorter

`[WARNING] This project currently has an account override set: {{ account }}. Account {{account}} will continue to act as the default account for this project. Use {{ createOverrideCommand }} to change override accounts, or {{ removeOverrideCommand }} to remove the override and use your default account.`

PR #1412, opened 2025-03-31
Porting project download, listBuilds, and open to TS

brandenrodgers:
## Description and Context

Porting a few more files over to TS:
- project/listBuilds
- project/open
- project/download

I made some other small changes that I'll call out inline. Mostly just things to make the typing easier.

### DIFF
--- a/commands/project/download.ts
+++ b/commands/project/download.ts
@@ -1,35 +1,38 @@
-// @ts-nocheck
-const path = require('path');
-const chalk = require('chalk');
-
-const {
-  addAccountOptions,
-  addConfigOptions,
-  addUseEnvironmentOptions,
-} = require('../../lib/commonOpts');
-const { trackCommandUsage } = require('../../lib/usageTracking');
-const { getCwd, sanitizeFileName } = require('@hubspot/local-dev-lib/path');
-const { logError, ApiErrorContext } = require('../../lib/errorHandlers/index');
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const { extractZipArchive } = require('@hubspot/local-dev-lib/archive');
-const {
+import { Argv, ArgumentsCamelCase } from 'yargs';
+import path from 'path';
+import { getCwd, sanitizeFileName } from '@hubspot/local-dev-lib/path';
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { extractZipArchive } from '@hubspot/local-dev-lib/archive';
+import {
   downloadProject,
   fetchProjectBuilds,
-} = require('@hubspot/local-dev-lib/api/projects');
-const { ensureProjectExists, getProjectConfig } = require('../../lib/projects');
-const {
-  downloadProjectPrompt,
-} = require('../../lib/prompts/downloadProjectPrompt');
-const { i18n } = require('../../lib/lang');
-const { uiBetaTag } = require('../../lib/ui');
+} from '@hubspot/local-dev-lib/api/projects';
+import { logError, ApiErrorContext } from '../../lib/errorHandlers/index';
+import { getProjectConfig } from '../../lib/projects';
+import { downloadProjectPrompt } from '../../lib/prompts/downloadProjectPrompt';
+import { i18n } from '../../lib/lang';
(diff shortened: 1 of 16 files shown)


lib/ui/index.ts
```diff
@@ -107,6 +107,8 @@ export function uiFeatureHighlight(features: string[], title?: string): void {
   });
 }
 
+export function uiBetaTag(message: string, log: true): undefined;
+export function uiBetaTag(message: string, log?: false): string;
```
brandenrodgers, line 111: This makes it so that the return type is `string` whenever `log` is false

lib/prompts/projectNamePrompt.ts
```diff
@@ -1,12 +1,14 @@
...
-const { uiAccountDescription } = require('../ui');
+import { promptUser } from './promptUtils';
+import { i18n } from '../lang';
+import { ensureProjectExists } from '../projects';
+import { uiAccountDescription } from '../ui';
 
 const i18nKey = 'lib.prompts.projectNamePrompt';
 
-const projectNamePrompt = (accountId, options = {}) => {
+export function projectNamePrompt(
```
camden11, line 8: This is missing a return type - if you add one you _might_ be able to avoid some of the non null assertions for project name (in the open command for example)

commands/project/open.ts
```diff
@@ -73,5 +86,12 @@ exports.builder = yargs => {
 
   yargs.example([['$0 project open', i18n(`${i18nKey}.examples.default`)]]);
 
-  return yargs;
+  return yargs as Argv<ProjectOpenArgs>;
+}
+
+module.exports = {
```
camden11, line 92: Assuming all of these `module.exports` need to be here because the bucket hasn't been ported - could be good to add a todo to remove!

joe-yeager (approved):

commands/project/download.ts
```diff
@@ -1,74 +1,52 @@
...
-  let projectName = promptedProjectName || project;
+  const { dest, build, derivedAccountId } = args;
+  const { project: projectName } = await downloadProjectPrompt(args);
+  let buildNumberToDownload = build;
 
-  trackCommandUsage('project-download', null, derivedAccountId);
+  trackCommandUsage('project-download', undefined, derivedAccountId);
 
   try {
-    const { projectExists } = await ensureProjectExists(
```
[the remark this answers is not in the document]
camden11: Makes sense to me. Good catch!

commands/project/download.ts
```diff
@@ -83,10 +61,12 @@ exports.handler = async options => {
       }
     }
 
+    const absoluteDestPath = dest ? path.resolve(getCwd(), dest) : getCwd();
+
     const { data: zippedProject } = await downloadProject(
       derivedAccountId,
       projectName,
-      buildNumberToDownload
+      buildNumberToDownload!
```
camden11, line 69: Are you sure we can assert that the build number exists? Haven't tested but I _think_ it could be `undefined` if you had project without any builds (is that still possible?)

commands/project/open.ts
brandenrodgers: I'm actually working on a PR now that removes these so I won't forget!

commands/project/download.ts
brandenrodgers: Good call. I was pretty sure that it would always exist (b/c otherwise something above would have thrown), but I didn't test that case. I'll try that

brandenrodgers: Ok I just checked and you're right. I added an error message in [f82d7c8](https://github.com/HubSpot/hubspot-cli/pull/1412/commits/f82d7c89907fb62f41ed91af6d8953391bb79630)

lib/prompts/projectNamePrompt.ts
brandenrodgers: Added in [f82d7c8](https://github.com/HubSpot/hubspot-cli/pull/1412/commits/f82d7c89907fb62f41ed91af6d8953391bb79630) 👌

camden11 (approved):
LGTM!

PR #1416, opened 2025-04-01
Port project install-deps, upload, and logs to TS

brandenrodgers:
## Description and Context

This is branched off of: br/port-project-commands-to-ts-2

Porting and adding basic unit tests for:
- project install-deps (Already had tests)
- project upload
- project logs (already had tests)


commands/project/upload.ts
```diff
@@ -1,64 +1,70 @@
...
-      projectDir,
-      pollProjectBuildAndDeploy,
-      message,
-      useV3Api(projectConfig?.platformVersion),
-      skipValidation
-    );
+    const { result, uploadError } =
+      await handleProjectUpload<ProjectPollResult>(
+        derivedAccountId,
+        projectConfig!,
```
camden11, line 61: It's not the end of the world (and doesn't need to be fixed in this PR), but I really think we should be try to minimize our use of the `!` operator and instead write code where TS can infer these things exists. In this instance, we know `projectConfig` exists, but we could rewrite `valildateProjectConfig` in such a way that TS knows too. There are other instances where assert non-null when we _don't_ know - I think in those, manually checking is better, even if it's more verbose

brandenrodgers: @camden11 yeah makes sense. I've been mostly trying to make these changes without having any impact on the command behavior. I realized that we can get an easy win just by adding an assertion to the return type of validateProjectConfig though! I added that in [4f24372](https://github.com/HubSpot/hubspot-cli/pull/1416/commits/4f243722e75e60d1edb73c3f2a2abebda0ab5b43)

I couldn't get it to also assert projectDir unfortunately, but this is still better. From what I can tell, the only options are to either split this out into two functions that each assert one of the inputs, or to change the input args to an object that contains both inputs.

lib/projects/index.ts
```diff
@@ -82,7 +82,7 @@ export async function getProjectConfig(dir?: string): Promise<{
 export function validateProjectConfig(
   projectConfig: ProjectConfig | null,
   projectDir: string | null
-): void {
+): asserts projectConfig is ProjectConfig {
```
joe-yeager, line 85: Interesting, the `asserts` keyword is a new one for me.

joe-yeager (approved):
LGTM!

camden11 (approved):

PR #1417, opened 2025-04-01
Better task status error message

camden11:
## Description and Context
[This issue](https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/803) (or the bug report that it was based on) was a little bit misleading - the problem was not due to an HTTP timeout, but rather an unhandled error that happened inside `setTimeout` - namely an error fetching build/deploy status while polling. The error that Sejal ran into seems to have been a backend problem that I'm unable to repro, but this should make future errors a bit less confusing.

## Screenshots
Before
Screenshot 2025-03-27 at 3 40 03 PM

After (with a fake error)
Screenshot 2025-04-01 at 5 01 47 PM

## Who to Notify

### DIFF
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -1145,7 +1145,7 @@ en:
         componentCount: "Found {{ numComponents }} components in this project"
         successStatusText: "DONE"
         failedStatusText: "FAILED"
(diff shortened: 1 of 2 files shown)


brandenrodgers (approved):
lgtm!

PR #1418, opened 2025-04-02
chore: Convert lang/en.lyaml to native JS objects

joe-yeager:
## Description and Context

Does not delete the `en.lyaml` file or modify the i18n functionality at all.  This PR is purely additive so we can begin the conversion.

### DIFF
--- a/lang/en.js
+++ b/lang/en.js
@@ -1,14 +1,16 @@
+import chalk from 'chalk';
+
 export const commands = {
   generalErrors: {
     updateNotify: {
       notifyTitle: 'Update available',
-      cmsUpdateNotification:
-        '{{#bold}}The CMS CLI is now the HubSpot CLI{{/bold}}\n\nTo upgrade, uninstall {{#bold}}{{ packageName }}{{/bold}}\nand then run {{ updateCommand }}',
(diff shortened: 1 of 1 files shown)


camden11:
Is there any advantage to adding this without making the changes to actually pull data from it? Won't it just get out of date when we make copy updates?

joe-yeager:
> Is there any advantage to adding this without making the changes to actually pull data from it? Won't it just get out of date when we make copy updates?

@camden11 - good question, I was thinking when we made copy changes we could update the new file, delete the lines from the old file, and switch to use the new strings.  I'm going to work on getting all of the i18n calls ported over when I get some more time

camden11:
That makes sense to me. I guess I don't feel super strongly, but I could see it making sense to include a version of the `i18n` function that works with the JSON file with this PR so we can get started with that right away

joe-yeager:
> That makes sense to me. I guess I don't feel super strongly, but I could see it making sense to include a version of the `i18n` function that works with the JSON file with this PR so we can get started with that right away

@camden11 In the new version we wouldn't need the i18n function.  This is a js file and the functions can be directly invoked

camden11 (approved):

PR #1419, opened 2025-04-02
Update HTTP timeout error messaging

camden11:
## Description and Context
HTTP timeout errors on the CLI are usually easy to solve by increasing timeout with `hs config set`, but we don't give users any guidance towards doing that. This adds a helpful message for all timeout errors that are triggered by the default timeout.

Relies on this LDL PR: https://github.com/HubSpot/hubspot-local-dev-lib
Issue: https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/803

## Screenshots
Before (unhelpful)
Screenshot 2025-04-02 at 2 44 36 PM

After (helpful)
Screenshot 2025-04-02 at 2 44 24 PM

## Who to Notify


lib/errorHandlers/index.ts
```diff
@@ -41,6 +43,20 @@ export function logError(error: unknown, context?: ApiErrorContext): void {
     // Unknown errors
     logger.error(i18n(`${i18nKey}.unknownErrorOccurred`));
   }
+
+  if (isHubSpotHttpError(error) && error.code === 'ETIMEDOUT') {
+    const config = getConfig();
+    const defaultTimeout = config?.httpTimeout;
+
+    if (error.timeout && defaultTimeout === error.timeout) {
```
camden11, line 51: Checking this to make sure the message doesn't get displayed in cases where we explicitly set a timeout

brandenrodgers: Should we still have some sort of error override, even if we manually adjusted the timeout?

brandenrodgers (approved):
Just one question about how we handle the error in cases where we adjusted the timeout, but it's not blocking. This lgtm 👌

lib/errorHandlers/index.ts
camden11: Good question, I was thinking we'd just handle that on a case by case basis anywhere where we suspect it could be an issue (which is few or zero places as far as i know)

brandenrodgers: Ah I see what you mean. My guess is that it would be fine to open this up to apply to all timeouts, but what you have is an improvement regardless. You can make the call to keep or remove the extra check

camden11: Ohh I was just trying to avoid the case where we manually set a timeout somewhere and then a user tries to fix an error with `hs config set` but it doesn't work. I'll add a generic message for when that's not the case

brandenrodgers: I didn't think about that. Then yeah +1 to keeping this in 👌

brandenrodgers: Might be worth adding a comment so we don't forget that years from now lol

PR #1424, opened 2025-04-03
feat: Add migration for existing project

joe-yeager:
## Description and Context

Adds `hs project migrate` to migrate an existing project from `2023.2` to `unstable`

To test:
- Find an existing project using `2023.2` and cd into the project dirctory
- Run `yarn hs project migrate`
- Select the app you would like to migrate
- Confirm the migration
- If prompted for `uids` enter them
- Allow the migration to complete
- Your pre-exisiting source code should be moved to `{project-location}/archive` and your new source code will be located in `{project-location}/{project-source-dir}`.  The `hsproject.json` file should be updated to have the correct platform version


lang/en.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
joe-yeager, line 1: I updated this to `ts` file and added a nocheck so we can start using it.  I converted it to a `js` originally which was short sighted because we can't use import/exports without flipping the project type to module.

lib/app/migrate.ts
```diff
@@ -12,47 +12,48 @@ import { downloadProject } from '@hubspot/local-dev-lib/api/projects';
...
+import fs from 'fs';
+import { lib } from '../../lang/en';
 
 function getUnmigratableReason(reasonCode: string): string {
   switch (reasonCode) {
     case UNMIGRATABLE_REASONS.UP_TO_DATE:
-      return i18n(
-        'commands.project.subcommands.migrateApp.unmigratableReasons.upToDate'
-      );
+      return lib.migrate.errors.unmigratableReasons.upToDate;
```
joe-yeager, line 34: I was going a bunch of i18n key shuffling, so I figured I might as well onboard these to the new object for the adding safety of compile time checking that the key exists.

types/Yargs.ts
```diff
@@ -40,12 +40,19 @@ export type MigrateAppOptions = CommonArgs &
...
-    name: string;
-    dest: string;
-    appId: number;
+    name?: string;
+    dest?: string;
+    appId?: number;
     platformVersion: string;
   };
 
+export type ProjectMigrateOptions = CommonArgs &
```
brandenrodgers, line 49: These command specific types can live in the command file. Also this is a big nit, but for consistency we're using "args" instead of options now

joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1424/commits/bae540f2e4f2a36ff96f6a4e276d0150841e23e6

commands/project/migrate.ts
```diff
@@ -0,0 +1,83 @@
...
+import { getProjectConfig } from '../../lib/projects';
+import { PLATFORM_VERSIONS } from '@hubspot/local-dev-lib/constants/projects';
+import { logError } from '../../lib/errorHandlers';
+import { EXIT_CODES } from '../../lib/enums/exitCodes';
+
+export type ProjectMigrateArgs = CommonArgs &
+  AccountArgs &
+  EnvironmentArgs &
+  ConfigArgs & {
+    dest?: string;
```
kemmerle, line 26: I think you can delete the `dest` arg.

lang/en.ts
```diff
@@ -3453,4 +3425,60 @@ export const lib = {
   oauth: {
     missingClientId: 'Error building oauth URL: missing client ID.',
   },
+  migrate: {
+    componentsToBeMigrated: components =>
```
kemmerle, line 3429: I'm assuming that this is just temporary copy, until a designer can take a look, correct?

lang/en.ts
```diff
@@ -3453,4 +3425,60 @@ export const lib = {
...
+        isPrivateApp: 'Private apps are not currently migratable',
+        listedInMarketplace: 'Listed apps are not currently migratable',
+        generic: reasonCode => `Unable to migrate app: ${reasonCode}`,
+      },
+      noAppsEligible: (accountId, reasons) =>
+        `No apps in account ${accountId} are currently migratable${reasons.length ? `\n  - ${reasons.join('\n  - ')}` : ''}`,
+
+      invalidAccountTypeTitle: () =>
+        `${chalk.bold('Developer account not targeted')}`,
+      invalidAccountTypeDescription: (useCommand, authCommand) =>
```
kemmerle, line 3454: Ooh, could you link to the new global config here?

commands/project/migrate.ts
```diff
@@ -0,0 +1,83 @@
...
+      projectConfig
+    );
+  } catch (error) {
+    logError(error);
+    return process.exit(EXIT_CODES.ERROR);
+  }
+  return process.exit(EXIT_CODES.SUCCESS);
+}
+
+export function builder(yargs: Argv) {
```
camden11, line 62: nit: this + handler are missing return types

lang/en.ts
joe-yeager: This is the UX reviewed copy, I just relocated it to this file.

lang/en.lyaml
```diff
@@ -717,6 +714,12 @@ en:
           examples:
             default: "Create a component within your project"
             withFlags: "Use --name and --type flags to bypass the prompt."
+        migrate:
+            describe: "Migrate a project to the projects framework."
```
camden11, line 718: Does this copy need to be updated? Assuming this came from "migrate an app to the projects framework" but now it doesn't make a ton of sense

lang/en.ts
joe-yeager: I'm not following, could you unpack the suggestion?

camden11 (approved):
A few non-blocking comments, but LGM

lang/en.ts
joe-yeager: Because it has javascript functions inside of it.  So it wants those all to be typed 😞

lib/app/migrate.ts
```diff
@@ -414,25 +436,26 @@ export async function migrateApp2025_2(
...
   uiLine();
-  logger.error(i18n(`${i18nKey}.errors.invalidAccountTypeTitle`));
+  logger.error(lib.migrate.errors.invalidAccountTypeTitle);
   logger.log(
-    i18n(`${i18nKey}.errors.invalidAccountTypeDescription`, {
-      useCommand: uiCommandReference('hs accounts use'),
-      authCommand: uiCommandReference('hs auth'),
-    })
+    lib.migrate.errors.invalidAccountTypeDescription(
+      uiCommandReference('hs accounts use'),
```
kemmerle, line 456: We prefer `hs account` singular now. Also perhaps promote `hs account auth` here as well? 😁

lang/en.ts
kemmerle: Sorry, I should have said, could you link to `hs account auth` here, as well.

kemmerle (approved):
I left a few comments, mostly nits. I do think you should leave some comments in `handleMigrationSetup` for posterity before merging.

lib/app/migrate.ts
joe-yeager: I might just keep it as `hs account use` for now because I don't want a person to be mid migration and then onboard to the new config.  Might be nice to limit the moving pieces

lang/en.ts
joe-yeager: Gotcha, I think I might leave it as `hs account use` to limit the changes that are happening with a migration.  I don't want a customer to be mid-migration to the new platform and then have their config migrate as well.  I worry it might be to many pieces moving at once.

lang/en.ts
```diff
@@ -3453,4 +3425,60 @@ export const lib = {
...
+        `Only public apps created in a developer account can be converted to a project component. Select a connected developer account with ${useCommand} or ${authCommand} and try again.`,
+      appWithAppIdNotFound: appId =>
+        `Could not find an app with the id ${appId} `,
+      noAppsForProject: 'No apps associated with project {{ projectName }}',
```
brandenrodgers, line 3458: Should this use the new interpolation pattern?

joe-yeager: Whoops, great catch.  Will update this

kemmerle (dismissed):

lang/en.lyaml
```diff
@@ -676,11 +677,7 @@ en:
             copyingProjectFiles: "Copying migrated project files"
             copyingProjectFilesComplete: "Migrated project files copied"
             copyingProjectFilesFailed: "Unable to copy migrated project files"
-          unmigratableReasons:
```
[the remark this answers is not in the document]
joe-yeager: Yeah, I did some more cleanup.

brandenrodgers (approved):

brandenrodgers (approved):

PR #1428, opened 2025-04-04
Fire alarm support

brandenrodgers:
## Description and Context

**Relies on this LDL PR:** https://github.com/HubSpot/hubspot-local-dev-lib/pull/253

This is building on the discovery work that @b-ash did to support Fire Alarm warnings in the CLI. The goal here is to enable ourselves to be able to surface notifications to users on existing CLI versions. We'll be able to surface warnings without devs needing to update their CLI versions. The way I have it set up now, we can target specific CLI versions and also specific commands. This should give us the right amount of granularity. We should put some thought into how we might want to leverage this, because it will be difficult to make changes down the road. The real value in this is being able to target old CLI versions.

**Rolling this out:**
1. We release these changes, which includes some middleware to fetch fire alarms at the beginning of every command
2. From this CLI version and on, we'll be able to create Fire Alarm notifications that can target specific CLI versions and commands

**Behavior:**
- We use `querySelector` to target specific CLI commands.
  - A null value means target every command 
  - "project upload" would only surface a warning if the dev runs `hs project upload`
  - "project upload::init" would surface a warning if the dev runs `hs project upload` or `hs project init`
- The query selector doesn't support numerical values, so I put the version targeting in the message body
  - Look at the comment that I left at the top of the file for an explanation of how it works 

**Open questions:**
- How should we format these messages in the CLI output?
- Should I update the command targeting syntax to support targeting entire command buckets? (i.e. "project *")
  - Fire alarms support a severity. Should we change the styling for warnings vs outages? Maybe we could use a box around these messages (like in the update notification screenshot below) and then use the severity to determine the color of the box?
- Should devs be able to opt-out of these warnings via a config setting?
- Should we attempt to limit the frequency that we show devs these warnings? What if a dev is seeing these on every single command they run. Would that get annoying?

## Screenshots

Here's my rough first draft just to show what this can look like:
image
The messages mainly have a `title`, `severity`, and `message` that we should be surfacing. I'm not currently doing anything with the severity field.

Another thing to consider is that we have an [existing update notification system](https://github.com/HubSpot/hubspot-cli/pull/1136). We'll have to think about how this interacts with that:
image


camden11:
Before I review, did something weird happen with prettier? Looks like there's a lot of files changed that are just spacing / style changes

joe-yeager:
> Before I review, did something weird happen with prettier? Looks like there's a lot of files changed that are just spacing / style changes

@camden11 It looks like it, because none of these showed as linting errors and all my builds were clean for inlining the i18n variables.  🤔

joe-yeager:
@brandenrodgers I just fixed the linting errors in a separate PR if you pull main they should go away

lib/middleware/fireAlarmMiddleware.ts
```diff
@@ -0,0 +1,192 @@
...
+          `${acc.length > 0 ? '\n\n' : ''}${chalk.bold(alarm.title)}\n${alarm.message}`
+        );
+      }
+      return acc;
+    }, '');
+
+    let boxen;
+
+    try {
+      boxen = (await import('boxen')).default;
```
joe-yeager, line 162: OOC why are we dynamically loading the library?

brandenrodgers: That package has a type of "module" which wasn't playing nicely with our "nodeNext". Here's the error:


> The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("boxen")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to 'package.json'.ts(1479)

I'm pretty sure we don't want to change our package type to module because then we'd need to include file extensions in our imports. It creates a whole mess of other problems.

joe-yeager (approved):
The logic LGTM, just a non-blocking question

lib/middleware/fireAlarmMiddleware.ts
joe-yeager: Gotcha, thanks for explaining

lib/middleware/fireAlarmMiddleware.ts
```diff
@@ -0,0 +1,193 @@
...
+): boolean {
+  // Assume we're targeting no versions if no version string is found
+  // Target all versions using the * wildcard
+  if (!targetVersionString) {
+    return false;
+  }
+
+  // Ignore alerts on any version tags (like -beta)
+  // No need for us to send alerts for these beta CLI versions
+  if (version.includes('-') || targetVersionString.includes('-')) {
```
camden11, line 34: Hmm do you think we'll never want to target beta versions? I could see it being helpful occasionally. Maybe we could exclude them from the more generalized targeting (with `*` and `>=` etc) but let you target them by specifying the exact version?

camden11 (dismissed):

camden11:
> Should I update the command targeting syntax to support targeting entire command buckets? (i.e. "project *")

Not sure if you already did this (the logic looked like this would work) but I could definitely see a scenario where we want to target buckets. At the very least, the project bucket

brandenrodgers:
> Not sure if you already did this (the logic looked like this would work) but I could definitely see a scenario where we want to target buckets. At the very least, the project bucket

@camden11 yeah agreed. The logic does support it, so I think we should be all set. The way you'd do it is by specifying the bucket (e.g. `hs project`). I considered making it more intentional, like `hs project *`, but that felt like it would be overkill and would just complicate the logic.

lib/middleware/fireAlarmMiddleware.ts
brandenrodgers: I like that idea. This was somewhat of an arbitrary call on my part. I'll add in some support for exact version targeting for beta versions.

joe-yeager (approved):

PR #1434, opened 2025-04-07
Log upload warning for all component configs

camden11:
## Description and Context
This makes a few improvements to support new component types in `hs project dev`
* Upload warning is now logged when _any_ component config is changed. Previously, it was only logged when app and card configs were changed. 
* Similarly, the warning that is logged when you have components locally that aren't in the deployed build looks for all component types, not just apps and cards.
* Updates the copy to remove references to "public" and "production" apps

## Screenshots

Screenshot 2025-04-07 at 3 56 19 PM

## TODO
This functionality is really only a temporary fix - ideally, the dev servers should either handle this logging or tell the CLI what components they know how to handle. This will work well with the current state of unified apps, though, and is better than not logging at all.

## Who to Notify

### DIFF
--- a/lib/DevServerManagerV2.ts
+++ b/lib/DevServerManagerV2.ts
@@ -45,12 +45,10 @@ class DevServerManagerV2 {
 
   async setup({
     projectNodes,
-    onUploadRequired,
     accountId,
     setActiveApp,
   }: {
     projectNodes: { [key: string]: IntermediateRepresentationNodeLocalDev };
-    onUploadRequired: () => void;
     accountId: number;
     setActiveApp: (appUid: string | undefined) => Promise<void>;
   }): Promise<void> {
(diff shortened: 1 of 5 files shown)


lib/DevServerManagerV2.ts
```diff
@@ -64,7 +62,6 @@ class DevServerManagerV2 {
       if (serverInterface.setup) {
         await serverInterface.setup({
           components: projectNodes,
-          onUploadRequired,
```
camden11, line 67: Right now, the only way to log the upload warning for components _without_ dev servers is to do it from the CLI. To keep things simple, I decided to have the CLI handle the logging for UI extensions as well. From what I could tell, there was no additional logging logic in the dev server - it just checks if the changed file is a config file and then calls the logging function passed from the CLI

It will probably make sense to revisit this in the future

brandenrodgers (approved):

lib/DevServerManagerV2.ts
brandenrodgers: We should let the uie team know, but that sounds reasonable to me 👍

PR #1436, opened 2025-04-08
Move middleware to folder and add tests

brandenrodgers:
## Description and Context

I have some upcoming work that will add some more middlware to the CLI. The cli.js file is getting super cluttered, so I think it's time to split the middleware out into a dedicated spot. This also let me port it all to TS and add tests.

I made a few changes to the behavior of the middleware that I'll call out inline.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,31 +1,28 @@
 #!/usr/bin/env node
 
 const yargs = require('yargs');
-const updateNotifier = require('update-notifier');
-const chalk = require('chalk');
-
 const { logger } = require('@hubspot/local-dev-lib/logger');
-const { addUserAgentHeader } = require('@hubspot/local-dev-lib/http');
-const {
-  loadConfig,
-  getAccountId,
-  configFileExists,
-  getConfigPath,
-  validateConfig,
-} = require('@hubspot/local-dev-lib/config');
 const { logError } = require('../lib/errorHandlers/index');
 const { setLogLevel, getCommandName } = require('../lib/commonOpts');
-const { validateAccount } = require('../lib/validation');
 const {
   trackHelpUsage,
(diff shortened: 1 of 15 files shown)


lib/middleware/utils.ts
```diff
@@ -0,0 +1,28 @@
+type TargetCommandMap = {
+  [key: string]: {
+    target?: boolean;
+    subCommands?: TargetCommandMap;
+  };
+};
+
+export function isTargetedCommand(
```
brandenrodgers, line 8: I made a few small changes to this. I removed the nested func b/c it wasn't necessary for the recursion, and I also changed the input to an array.

This will make it easier for me to use in other places (I'll need it for the Fire Alarm work)

lib/middleware/__test__/notificationsMiddleware.test.ts
```diff
@@ -0,0 +1,9 @@
+const { notifyAboutUpdates } = require('../notificationsMiddleware');
+
+describe('lib/middleware/notificationsMiddleware', () => {
+  describe('notifyAboutUpdates()', () => {
+    it('should safely execute without throwing an error', () => {
+      expect(() => notifyAboutUpdates()).not.toThrow();
```
brandenrodgers, line 6: This was a PITA to test b/c it's commonJS and also exports a single default export that's a class that you instantiate. So I just made sure it's not throwing. I'm considering removing update-notifier as a dep soon anyways so I think this is fine for now.

bin/cli.js
```diff
@@ -1,31 +1,28 @@
 #!/usr/bin/env node
```
brandenrodgers, line 1: Much cleaner in here now!

camden11 (approved):
Nice! This is a big improvement

PR #1442, opened 2025-04-11
fix: Fix prettier commands on older node versions

joe-yeager:
## Description and Context

Fix the prettier commands on older node versions


camden11 (dismissed):

brandenrodgers (dismissed):
💅

PR #1444, opened 2025-04-11
fix: Fix project creation happening when there is a validation error

joe-yeager:
## Description and Context

Move ensure project exists calls to the upload handler, only call `ensureProjectExists` after we have validated the project.


lib/projects/upload.ts
```diff
@@ -147,11 +162,16 @@ export async function handleProjectUpload<T>(
         }
       }
 
+      await ensureProjectExists(accountId, projectConfig.name, {
+        forceCreate,
+        uploadCommand: isUploadCommand,
+      });
```
joe-yeager, line 168: Moving these here assures that we don't call it until we have validated the project.

camden11 (approved):
LGTM

PR #1446, opened 2025-04-14
fix: Disable hs project watch for v3 projects

joe-yeager:
## Description and Context

Disable `hs project watch` for projects that are using a platform version that relies on v3 projects APIs


kemmerle (approved):
LGTM!

PR #1448, opened 2025-04-15
Update hs init and hs auth for config revamp

camden11:
## Description and Context
This updates `hs init` and `hs auth` (and also app migrate) to work with the new config changes.

## Who to Notify

### DIFF
--- a/commands/auth.ts
+++ b/commands/auth.ts
@@ -7,19 +7,20 @@ import {
 } from '@hubspot/local-dev-lib/constants/auth';
 import { ENVIRONMENTS } from '@hubspot/local-dev-lib/constants/environments';
 import { DEFAULT_HUBSPOT_CONFIG_YAML_FILE_NAME } from '@hubspot/local-dev-lib/constants/config';
-import { AccessToken, CLIAccount } from '@hubspot/local-dev-lib/types/Accounts';
+import {
+  AccessToken,
+  HubSpotConfigAccount,
+} from '@hubspot/local-dev-lib/types/Accounts';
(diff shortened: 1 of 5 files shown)


commands/auth.ts
```diff
@@ -73,89 +74,63 @@ type AuthArgs = CommonArgs &
...
-  const configPath = getConfigPath();
-  if (configPath) {
-    checkAndWarnGitInclusion(configPath);
-  }
 
-  if (!getConfigPath(configFlagValue)) {
-    logger.error(i18n(`${i18nKey}.errors.noConfigFileFound`));
-    process.exit(EXIT_CODES.ERROR);
-  }
+  const configPath = getConfigFilePath();
```
camden11, line 85: No longer need to manually error - `getConfigFilePath` will throw if there's no config

commands/init.ts
```diff
@@ -163,21 +160,10 @@ export async function handler(
...
 
-  const doesOtherConfigFileExist = configFileExists(!useHiddenConfig);
-  if (doesOtherConfigFileExist) {
-    const path = getConfigPath('', !useHiddenConfig);
-    logger.error(
-      i18n(`${i18nKey}.errors.bothConfigFilesNotAllowed`, { path: path! })
-    );
-    process.exit(EXIT_CODES.ERROR);
-  }
-
```
camden11, line 174: Figured we could simplify this logic a bit. If _any_ config file exists (which we check for above), exit regardless of whether its local or global. I updated the error message to make it make sense for both cases

PR #1449, opened 2025-04-15
Finishing the project commands TS port

brandenrodgers:
## Description and Context

This wraps up the work to port all of the project command files to TS. I had to make some yargs typing changes to make things work well. I'm trying to strike the balance between getting value out of the  yargs types, and making this so clunky to use that it's too much of a hassle to follow the patterns.

I had to make a few changes:
- I realized that the `makeYargsBuilder` util was incorrectly matching on command buckets for help output. For example, running `hs project create --help` was returning the help output for `hs project --hlep`. I fixed this by matching on the last command that the user ran (e.g "create" in `hs project create --help`)
- I had to add some custom yargs types. The `CommandModule` interface is a little annoying to use because it makes the builder optional, and changes the type from a simple function to something else. I created some overrides that simplifies that type for our use cases
- All of the commands use `makeYargsBuilder` now
- All of the commands have one default export (except some of them also export their handler arg types)


brandenrodgers:
I'm going to wait for [this](https://github.com/HubSpot/hubspot-cli/pull/1424) to be merged before I put this up for review.

commands/project/download.ts
```diff
@@ -19,20 +19,21 @@ import {
...
+const command = 'download';
+const describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
 
 type ProjectDownloadArgs = CommonArgs &
   ConfigArgs &
   AccountArgs &
   EnvironmentArgs & { project?: string; dest?: string; build?: number };
 
-export async function handler(args: ArgumentsCamelCase<ProjectDownloadArgs>) {
+async function handler(args: ArgumentsCamelCase<ProjectDownloadArgs>) {
```
camden11, line 36: Missing return type here

commands/project/cloneApp.ts
```diff
@@ -53,8 +49,8 @@ export type CloneAppArgs = ConfigArgs &
     appId: number;
   };
 
-export const handler = async (options: ArgumentsCamelCase<CloneAppArgs>) => {
-  const { derivedAccountId } = options;
+async function handler(args: ArgumentsCamelCase<CloneAppArgs>) {
```
camden11, line 52: missing return type

commands/project/listBuilds.ts
```diff
@@ -114,7 +110,7 @@ async function fetchAndDisplayBuilds(
   }
 }
 
-export async function handler(args: ArgumentsCamelCase<ProjectListBuildsArgs>) {
+async function handler(args: ArgumentsCamelCase<ProjectListBuildsArgs>) {
```
camden11, line 113: missing return type

commands/project/logs.ts
```diff
@@ -90,7 +90,7 @@ export type ProjectLogsArgs = CommonArgs & {
   limit?: number;
 };
 
-export const handler = async (args: ArgumentsCamelCase<ProjectLogsArgs>) => {
+const handler = async (args: ArgumentsCamelCase<ProjectLogsArgs>) => {
```
camden11, line 93: Missing return type. This should also use `function` rather than `const`

commands/project/migrateApp.ts
```diff
@@ -1,29 +1,25 @@
...
 // TODO: Leave this as deprecated and remove in the next major release
-export const describe = uiDeprecatedTag(
+const describe = uiDeprecatedTag(
   i18n(`commands.project.subcommands.migrateApp.describe`),
   false
 );
 export const deprecated = true;
 
-export async function handler(yargs: ArgumentsCamelCase<MigrateAppArgs>) {
+async function handler(yargs: ArgumentsCamelCase<MigrateAppArgs>) {
```
camden11, line 22: missing return type

commands/project/open.ts
```diff
@@ -20,20 +14,22 @@ import {
...
+const describe = uiBetaTag(i18n(`${i18nKey}.describe`), false);
 
 type ProjectOpenArgs = CommonArgs &
   ConfigArgs &
   AccountArgs &
   EnvironmentArgs &
   TestingArgs & { project?: string };
 
-export async function handler(args: ArgumentsCamelCase<ProjectOpenArgs>) {
+async function handler(args: ArgumentsCamelCase<ProjectOpenArgs>) {
```
camden11, line 32: missing return type

lib/prompts/createProjectPrompt.ts
```diff
@@ -117,10 +140,19 @@ export async function createProjectPrompt(
...
+    if (!result.projectTemplate) {
+      throw new Error(
+        'Project template is required when projectTemplates is provided'
+      );
+    }
+    return result as CreateProjectPromptResponseWithTemplate;
+  }
+
+  return result as CreateProjectPromptResponseWithoutTemplate;
+} as CreateProjectPromptOverload;
```
camden11, line 158: Are we actually getting anything out of these new types with all this casting happening? Is it possible to do this without casting?

camden11 (commented):
Found a few minor TS issues but otherwise this is looking good!

lib/prompts/createProjectPrompt.ts
brandenrodgers: Thanks for calling this one out. I had initially tried a bunch of different variations to get this working, but I just took another look and I can do it without any casting at all. I fixed it in [85c3a79](https://github.com/HubSpot/hubspot-cli/pull/1449/commits/85c3a79987a57b1031cb52e71a1eb499a5612805). Turns out I'm just not very familiar with how overloads work 😅

types/Yargs.ts
```diff
@@ -35,3 +35,16 @@ export type ProjectDevArgs = CommonArgs & ConfigArgs & EnvironmentArgs;
 export type TestingArgs = {
   qa?: boolean;
 };
+
+// This is a workaround to make the builder method required for the CommandModule
+export type YargsCommandModule<T, U> = Omit<CommandModule<T, U>, 'builder'> & {
+  builder: (yargs: Argv) => Promise<Argv<U>>;
+};
```
joe-yeager, line 42: Nit: Since `CommandModule` is an interface we can extend it and it might simplify the TS logic here.

```typescript
export interface CommandModuleWithRequiredBuilder<T, U> extends CommandModule<T, U> {
  builder: (yargs: Argv) => Promise<Argv<U>>
}
```

lib/prompts/createProjectPrompt.ts
```diff
@@ -117,10 +138,19 @@ export async function createProjectPrompt(
...
-      projectTemplates,
+      projectTemplates!,
       promptOptions.template!
     );
   }
 
+  if (projectTemplates && projectTemplates.length > 0) {
+    if (!result.projectTemplate) {
+      throw new Error(
+        'Project template is required when projectTemplates is provided'
```
joe-yeager, line 149: Should we move this to the `en.ts`?

lib/prompts/createProjectPrompt.ts
```diff
@@ -42,14 +54,23 @@ function findTemplateByNameOrLabel(
   );
 }
 
+// Includes `projectTemplate` in the return value if `projectTemplates` is provided
+export async function createProjectPrompt(
+  promptOptions: { name?: string; dest?: string; template?: string },
```
joe-yeager, line 59: Nit: this inline type could be moved to a type and reused

joe-yeager (approved):
LGTM!  Just a few nits

camden11 (approved):

lib/prompts/createProjectPrompt.ts
brandenrodgers: Good find, moved in [df00fca](https://github.com/HubSpot/hubspot-cli/pull/1449/commits/df00fca88f867c909716a04776e8b2df7327025b)

lib/prompts/createProjectPrompt.ts
brandenrodgers: I like that idea, done in [df00fca](https://github.com/HubSpot/hubspot-cli/pull/1449/commits/df00fca88f867c909716a04776e8b2df7327025b)

PR #1450, opened 2025-04-15
Fix bug with `hs account auth`

kemmerle:
## Description and Context
Joe Kurien reported a bug that `hs account auth` returns an error when no global config file is present (ie. running the command for the first time). 

## Screenshots
BEFORE:

Screenshot 2025-04-15 at 4 23 26 PM

AFTER:

Screenshot 2025-04-15 at 4 22 42 PM

## TODO

- [ ] Address feedback
- [ ] Cut patch release to pick up the bug fix. 

## Who to Notify

### DIFF
--- a/lib/middleware/configMiddleware.ts
+++ b/lib/middleware/configMiddleware.ts
@@ -77,6 +77,15 @@ export async function loadConfigMiddleware(
   };
 
   if (
+    !configFileExists(true) &&
+    isTargetedCommand(argv._, {
+      account: { target: false, subCommands: { auth: { target: true } } },
+    })
+  ) {
+    return;
+  }
+
+  if (
     configFileExists(true) &&
     argv.config &&
     !isTargetedCommand(argv._, {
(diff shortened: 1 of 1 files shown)


lib/middleware/configMiddleware.ts
```diff
@@ -76,6 +76,15 @@ export async function loadConfigMiddleware(
     }
   };
 
+  if (
```
brandenrodgers, line 79: Could we add this to the existing `SKIP_CONFIG_VALIDATION` object? I don't remember exactly how that works, but it seems like the same use case as this.

kemmerle: `hs account auth` is a special exception, because we want to validate the config if it exists (ie. if we're just authorizing a new account). If it doesn't exist (ie. we're running the command for the first time to create the config), we don't want to validate the config, which is what this conditional stipulates.

brandenrodgers: Ah ok, thanks. Makes sense 👍

brandenrodgers (approved):

PR #1452, opened 2025-04-16
chore: Exit the app migrate or project migrateApp commands if they are ran within a project

joe-yeager:
## Description and Context

Exit the `hs app migrate` and `hs project migrateApp` commands if they are ran inside of a project directory.


brandenrodgers (approved):
Code looks good (pending design input on copy)

PR #1455, opened 2025-04-17
feat: Migrations polish

joe-yeager:
## Description and Context

Does the following:
- Dedupes components in the list of components will or will not be migrated
- When prompting for project name, validate the name in the prompt and don't fail the process if the project already exists
- Adds a message letting the user know projects migrations are one way
- Cleans up some debug logging

### DIFF
--- a/lib/app/migrate.ts
+++ b/lib/app/migrate.ts
@@ -194,10 +194,6 @@ async function selectAppToMigrate(
 
   logger.log();
 
-  if (projectConfig?.projectConfig) {
-    logger.log(lib.migrate.projectMigrationWarning);
-  }
-
   const promptMessage = projectConfig?.projectConfig
     ? `${lib.migrate.projectMigrationWarning} ${lib.migrate.prompt.proceed}`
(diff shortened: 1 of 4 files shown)


lib/app/migrate.ts
```diff
@@ -179,21 +179,30 @@ async function selectAppToMigrate(
   if (migratableComponents.length !== 0) {
     logger.log(
       lib.migrate.componentsToBeMigrated(
-        `\n - ${migratableComponents.join('\n - ')}`
+        `\n - ${[...new Set(migratableComponents)].join('\n - ')}`
```
joe-yeager, line 182: These need to be deduped before we log them

brandenrodgers: Do dupes exist when the user has multiple instances of a given component? Like multiple cards?

joe-yeager: Yeah, they can have multiple cards, webhooks, etc.

brandenrodgers (approved):
lgtm!

PR #1457, opened 2025-04-22
Fix dev test account creation

brandenrodgers:
## Description and Context

Relies on https://github.com/HubSpot/hubspot-local-dev-lib/pull/261

This fixes the double prompt that we get during project local dev if you opt to create a new dev test account. The api returns a personal access key, so there's no need for us to be re-prompting users for it.

This is a regression that I accidentally introduced a little while ago.

### DIFF
--- a/lib/localDev.ts
+++ b/lib/localDev.ts
@@ -323,14 +323,14 @@ export async function createDeveloperTestAccountForLocalDev(
       accountId
     );
 
-    const devTestAccountId = await buildDeveloperTestAccount(
+    const result = await buildDeveloperTestAccount(
       name,
       accountConfig,
       env,
       maxTestPortals
     );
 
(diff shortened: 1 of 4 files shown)


lib/localDev.ts
```diff
@@ -323,14 +323,14 @@ export async function createDeveloperTestAccountForLocalDev(
       accountId
     );
 
-    const result = await buildDeveloperTestAccount(
+    const devTestAccountId = await buildDeveloperTestAccount(
       name,
       accountConfig,
       env,
       maxTestPortals
     );
```
joe-yeager, line 332: Nit, we can just directly return the result and not use the intermediate variable.

joe-yeager (approved):
LGTM, just a minor nit.

PR #1458, opened 2025-04-22
Hide sandbox prompt for accounts with no sandbox access

camden11:
## Description and Context
When testing legacy private app local dev on an account without sandboxes, I noticed that we still showed the sandbox prompt but the only option was to proceed with your current account. This replaces that prompt with the more generic prompt we show when your default account fits the recommended requirements (i.e. is a sandbox or dev test account).

I also updated the copy of the prompt to be more in line with the rest of the language we use in `hs project dev`

## Screenshots
Before
Screenshot 2025-04-22 at 3 52 40 PM

After
Screenshot 2025-04-22 at 5 05 24 PM

## Who to Notify

### DIFF
--- a/commands/project/dev/deprecatedFlow.ts
+++ b/commands/project/dev/deprecatedFlow.ts
@@ -25,6 +25,7 @@ import {
   createInitialBuildForNewProject,
   useExistingDevTestAccount,
   checkIfParentAccountIsAuthed,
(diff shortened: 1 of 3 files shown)


joe-yeager (dismissed):
LGTM, just a nit about nesting.

PR #1459, opened 2025-04-22
feat: Incorporate UX feedback

joe-yeager:
## Description and Context

This pull request does the following:

- Changes the default platform-version for `hs app migrate` to `2025.2` 
- Adds a message when `hs project migrate` is ran outside of a project to nudge them to use `hs app migrate` if they are trying to migrate a legacy app to projects.
Screenshot 2025-04-23 at 12 20 20 PM
- Changes verbiage in log messages
- Unhides apps associated with projects when `hs app migrate` is ran, and shows a message that they are only migratable with `hs project migrate`
Screenshot 2025-04-23 at 12 22 32 PM
- Fixes a bug where the default answer was an empty string when prompting for UIDs
- Removes `unstable` as an option in the `platform-version` flag
- Adds a hidden `--unstable` flag
- Adds links to build and projects when the build is successful
- Changes default proceed with migration prompt to No

### DIFF
--- a/commands/app/migrate.ts
+++ b/commands/app/migrate.ts
@@ -20,14 +20,14 @@ import { uiBetaTag, uiCommandReference, uiLink } from '../../lib/ui';
 import { migrateApp2023_2 } from '../../lib/app/migrate_legacy';
 import { getIsInProject } from '../../lib/projects';
 
-const { v2023_2, v2025_2, unstable } = PLATFORM_VERSIONS;
-export const validMigrationTargets = [v2023_2, v2025_2, unstable];
+const { v2023_2, v2025_2 } = PLATFORM_VERSIONS;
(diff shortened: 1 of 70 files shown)


commands/app/migrate.ts
```diff
@@ -122,9 +125,14 @@ export function builder(yargs: Argv): Argv<MigrateAppArgs> {
     },
     'platform-version': {
       type: 'string',
-      choices: validMigrationTargets,
+      choices: [v2023_2, v2025_2],
+      hidden: true,
+      default: v2025_2,
+    },
+    unstable: {
```
joe-yeager, line 132: The idea is that this will only be used for internal testing and shouldn't be used by external users. We will keep this hidden and keep `unstable` out of the list of platform version choices.

kemmerle (approved):
Changes make sense to me, and I like the updated copy.

PR #1460, opened 2025-04-23
Fix: UX improvement for migration/merge prompt behavior in `hs account auth`

kemmerle:
## Description and Context
Randy S suggested this UX improvement via Slack.  

Currently, if a customer runs the `hs account auth` command and declines to migrate or merge a deprecated config, we continue to create the new global config file and add an account to it. I've changed the behavior of the command to be more intuitive for the customer and give them more info. 

## Screenshots

If customer declines to migrate the deprecated config (and no global config exists):

Screenshot 2025-04-23 at 10 41 56 AM

If the customer declines to merge a deprecated and global config:

Screenshot 2025-04-23 at 10 47 44 AM

## TODO

- [ ] Address feedback 

## Who to Notify

### DIFF
--- a/commands/account/auth.ts
+++ b/commands/account/auth.ts
@@ -33,7 +33,7 @@ import { setAsDefaultAccountPrompt } from '../../lib/prompts/setAsDefaultAccount
 import { logError } from '../../lib/errorHandlers/index';
 import { trackCommandMetadataUsage } from '../../lib/usageTracking';
 import { EXIT_CODES } from '../../lib/enums/exitCodes';
-import { uiFeatureHighlight } from '../../lib/ui';
+import { uiCommandReference, uiFeatureHighlight } from '../../lib/ui';
 import { CommonArgs, ConfigArgs } from '../../types/Yargs';
 
 const TRACKING_STATUS = {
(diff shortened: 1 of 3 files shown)


commands/account/auth.ts
```diff
@@ -105,10 +107,20 @@ export async function handler(
...
-        await handleMerge(providedAccountId);
+        mergeConfirmed = await handleMerge(providedAccountId);
+        if (!mergeConfirmed) {
+          logger.log(
+            i18n('commands.account.subcommands.auth.errors.mergeNotConfirmed', {
+              deprecatedConfigPath: getConfigPath('', false)!,
+              globalConfigPath: getConfigPath('', true)!,
+            })
+          );
+        }
```
joe-yeager, line 123: The nesting is starting to get a little deep here.  Could we split this into smaller functions?

brandenrodgers:
Should we allow users to continue with authenticating a new account if they have an existing `hubspot.config.yml` file? Having both the global config and the current config files can lead to some confusing scenarios. I'm not sure that we should allow the users to continue in the flow if they choose "no".A

I also think we might want to take a look at the way we're explaining the command's behavior. It's not super clear what selecting "yes" in this prompt will do, or why the prompt is being shown.

brandenrodgers:
Looking at the feedback:
> I kinda expected that if I say no to migration, things would stay as they are. I don’t think we should let the user keep going in that case. It should be more like, "either migrate or you can't add a new account." or we just keep using the old file

Rather than letting the user continue to migrate, we should just exit the command if they select the "no" option. That should lead to the fewest awkward edge-case scenarios. We should minimize the number of ways for people to get into a situation where they have both a non-global and a global config simultaneously. If they don't want to merge configs, then that's essentially saying that they aren't ready to opt into the new config.

If we do want to let the user continue to auth an account even though they have an existing hubspot.config.yaml file, we might want to turn this message into a warning to make it stand out more. We could also add some spacing above and below it. We should make it clear that the config will continue to exist, but it won't be used anymore because the global config takes precedence.

brandenrodgers (approved):
lgtm!

PR #1461, opened 2025-04-23
Support legacy public app development in combined accounts

camden11:
## Description and Context
Previously, converting a developer account to a combined account meant losing the ability to run `hs project dev` on legacy public apps. This is because a combined account is technically a standard account, which `hs project dev` didn't support for public apps. This updates the logic in the legacy dev flow to allow unified apps (and also updates the legacy `clone-app` and `migrate-app` commands to work with combined accounts)

See https://git.hubteam.com/HubSpot/hubspot-cli-issues/issues/806

## Who to Notify

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -84,3 +84,8 @@ export const APP_AUTH_TYPES = {
   OAUTH: 'oauth',
   STATIC: 'static',
 } as const;
+
+export const FEATURES = {
+  UNIFIED_THEME_PREVIEW: 'cms:react:unifiedThemePreview',
+  UNIFIED_APPS: 'Developers:UnifiedApps:PrivateBeta',
+} as const;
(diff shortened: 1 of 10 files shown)


lib/constants.ts
```diff
@@ -84,3 +84,8 @@ export const APP_AUTH_TYPES = {
   OAUTH: 'oauth',
   STATIC: 'static',
 } as const;
+
+export const FEATURES = {
+  UNIFIED_THEME_PREVIEW: 'cms:react:unifiedThemePreview',
+  UNIFIED_APPS: 'Developers:UnifiedApps:PrivateBeta',
+} as const;
```
joe-yeager, line 91: Thanks for moving these here 🙌

joe-yeager (approved):
LGTM!

PR #1462, opened 2025-04-24
Add type checking to lang objects

camden11:
## Description and Context
Was working with the new lang objects and found another opportunity to do some type checking and reduce copy errors. With the `satisfies` operator, we can default every function in the lang objects to expect only strings or numbers as arguments. This will help reduce accidentally passing in `undefined`, objects, or anything else unexpected. We can override lang functions on a case by case basis if they need to accept other types (see `noAppsElgible`) - I think this is rare enough that the extra checks will be worth the slightly higher effort in those cases.

## Who to Notify

### DIFF
--- a/lang/en.ts
+++ b/lang/en.ts
@@ -1,7 +1,7 @@
 // @ts-nocheck
 import chalk from 'chalk';
 
-type LangFunction = (...args: string[] | number[]) => string;
+type LangFunction = (...args: (string | number)[]) => string;
 
 type LangObject = {
   [key: string]: string | LangFunction | LangObject;
--- a/lang/en.ts
+++ b/lang/en.ts
@@ -1,6 +1,12 @@
 // @ts-nocheck
 import chalk from 'chalk';
 
+type LangFunction = (...args: string[] | number[]) => string;
+
+type LangObject = {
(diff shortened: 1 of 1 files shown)


lang/en.ts
```diff
@@ -1,6 +1,12 @@
 // @ts-nocheck
 import chalk from 'chalk';
 
+type LangFunction = (...args: string[] | number[]) => string;
+
+type LangObject = {
+  [key: string]: string | LangFunction | LangObject;
+};
+
```
joe-yeager, line 9: Nice!

lang/en.ts
```diff
@@ -1,6 +1,12 @@
 // @ts-nocheck
 import chalk from 'chalk';
 
+type LangFunction = (...args: string[] | number[]) => string;
```
joe-yeager, line 4: should this be `...args: (string | number)[]`?

joe-yeager (dismissed):
LGTM, just a question about one of the types

lang/en.ts
camden11: Yeah actually... you'd think so, but for some reason this works as expected (typing each individual arg as string or number). But I think what you suggested is clearer towards we actually want so I'll update it

PR #1465, opened 2025-04-24
Organize lib/projects and use lang objects

camden11:
## Description and Context
The upcoming work to support the Local Dev UI will require a lot of additions to the `lib/projects` folder, so I figured now would be a good time to clean things up there a bit. This PR makes a variety of housekeeping changes
* Moves all project local dev related files from`/lib` into `lib/projects/localDev``
  * `localDev` renamed to `helpers`
  * `LocalDevManager` and `LocalDevManagerV2`
  * `DevServerManager` and `DevServerManagerV2`
* Splits up `lib/projects/index` since it wasn't actually an index file
  * Most of the functions in this file are now in `lib/projects/config`
  * `ensureProjectExists` moved to `lib/projects/ensureProjectExists`
  * `logFeedbackMessage` moved to `lib/projects/ui`
* Updates all project files to use the new lang objects
  * this surfaced a lot of lang errors, so those have now been fixed

## TODO
Due to file renaming, some files lang keys no longer match their file names. I didn't want to do even more in this PR, so we can update that at a later time if it ever becomes a problem

## Who to Notify

### DIFF
--- a/lang/en.ts
+++ b/lang/en.ts
@@ -1128,8 +1128,8 @@ export const commands = {
         failedToFetchProjectDetails:
           'There was an error fetching project details',
         noFunctionsLinkText: 'Visit developer docs',
-        noFunctionsInProject: link =>
-          `There aren't any functions in this project\n\t- Run ${chalk.orange('hs project logs --help')} to learn more about logs\n\t- ${link} to learn more about serverless functions`,
+        noFunctionsInProject: (helpCommand, link) =>
+          `There aren't any functions in this project\n\t- Run ${helpCommand} to learn more about logs\n\t- ${link} to learn more about serverless functions`,
         noFunctionWithName: name => `No function with name "${name}"`,
         functionNotDeployed: name =>
           `The function with name "${name}" is not deployed`,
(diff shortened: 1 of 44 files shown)


lang/en.ts
```diff
@@ -1122,8 +1128,8 @@ export const commands = {
         failedToFetchProjectDetails:
           'There was an error fetching project details',
         noFunctionsLinkText: 'Visit developer docs',
-        noFunctionsInProject: link =>
-          `There aren't any functions in this project\n\t- Run ${chalk.orange('hs project logs --help')} to learn more about logs\n\t- ${link} to learn more about serverless functions`,
```
joe-yeager, line 1126: Haha you beat me to it, I saw that cursor made up the `chalk.orange` function 🤦

camden11: Haha cursor making things up was a common theme in this PR

joe-yeager (dismissed):
LGTM!  Thanks for splitting that up and switching these over.

joe-yeager:
> @joe-yeager Just noticed you inlining `uiCommandReference`, `uiLink` etc in the lang file - going to update this to do that as well

@camden11 yeah, I started out just converting the the function calls, and then I realized I can just move most of them to inline calls.  I think we could start small as well, just start by leaving them as variables and then slowly move them over.  Up to you if you want to do it as part of this PR.

camden11:
Yeah, I just went ahead and did it for `uiCommandReference`s that didn't involve any other variables. Let me turn a lot of lang functions into just strings

PR #1466, opened 2025-04-24
chore: Split the tracking for app migrate and project migrate-app

joe-yeager:
## Description and Context

Split the migration tracking for `hs app migrate` and `hs project migrate-app`


camden11 (dismissed):
I could see putting `handlerGenerator` in its own file, but also fine as is. LGTM

kemmerle (approved):
Changes make sense to me--thanks for DRYing up the code.

PR #1467, opened 2025-04-25
fix: Correct the formatting of the migration errors

joe-yeager:
## Description and Context

Fix an issue when when the componentErrors has length 0

image

### DIFF
--- a/lib/app/migrate.ts
+++ b/lib/app/migrate.ts
@@ -436,19 +436,21 @@ async function finalizeMigration(
     });
 
     if (isMigrationStatus(error) && error.status === MIGRATION_STATUS.FAILURE) {
-      const errorMessage = error.componentErrors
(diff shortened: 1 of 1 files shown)


lib/app/migrate.ts
```diff
@@ -436,19 +436,21 @@ async function finalizeMigration(
...
-              } = componentError;
-
-              return `${componentType}${uid ? ` (${uid})` : ''}: ${errorMessage}`;
-            })
-            .join('\n\t- ')}`
-        : error.projectErrorDetail;
+      const { componentErrors, projectErrorDetail } = error;
+      if (!componentErrors || !componentErrors.length) {
+        throw new Error(projectErrorDetail);
+      }
```
joe-yeager, line 442: I refactored a little while I was here, but the gist of it was we were missing a length check.

kemmerle (approved):

PR #1468, opened 2025-04-25
chore: Add --network-debug flag to log all http requests from LDL

joe-yeager:
## Description and Context

Add a hidden `--network-debug` flag that sets the log level to debug and sets `HUBSPOT_NETWORK_LOGGING` to true so LDL will log http traffic

### DIFF
--- a/commands/__tests__/create.test.ts
+++ b/commands/__tests__/create.test.ts
@@ -41,7 +41,6 @@ describe('commands/create', () => {
     it('should support the correct options', () => {
       createCommand.builder(yargs);
 
-      expect(yargs.option).toHaveBeenCalledTimes(3);
       expect(yargs.option).toHaveBeenCalledWith(
         'internal',
         expect.objectContaining({ type: 'boolean', hidden: true })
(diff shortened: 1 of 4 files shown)


lib/commonOpts.ts
```diff
@@ -26,12 +26,18 @@ const i18nKey = 'lib.commonOpts';
...
+    .option('debug', {
+      alias: 'd',
+      default: false,
+      describe: i18n(`${i18nKey}.options.debug.describe`),
+      type: 'boolean',
+    })
+    .option('network-debug', {
+      default: false,
+      type: 'boolean',
+      hidden: true,
```
joe-yeager, line 39: I don't see a compelling reason to unhide this.  If we think external users want to use it, I can add some copy and unhide it.  It's realistically more for our benefit

kemmerle (approved):

commands/__tests__/create.test.ts
```diff
@@ -41,7 +41,6 @@ describe('commands/create', () => {
     it('should support the correct options', () => {
       createCommand.builder(yargs);
 
-      expect(yargs.option).toHaveBeenCalledTimes(3);
```
joe-yeager, line 44: This was a fragile assertion, so I felt it was better to remove it than increment the number

kemmerle (approved):

PR #1469, opened 2025-04-25
fix: Add polling error handling after begin migration

joe-yeager:
## Description and Context

Log the error message in the event that the polling returns a failure after the begin migration

### DIFF
--- a/lib/app/migrate.ts
+++ b/lib/app/migrate.ts
@@ -390,7 +390,9 @@ async function beginMigration(
     if (isMigrationStatus(error) && error.status === MIGRATION_STATUS.FAILURE) {
       throw new Error(buildErrorMessageFromMigrationStatus(error));
     }
-    throw error;
+    throw new Error(lib.migrate.errors.migrationFailed, {
+      cause: error,
+    });
   }
 
   if (pollResponse.status !== MIGRATION_STATUS.INPUT_REQUIRED) {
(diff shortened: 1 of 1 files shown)


lib/app/migrate.ts
```diff
@@ -359,11 +378,22 @@ async function beginMigration(
...
+  let pollResponse: MigrationStatus;
+  try {
+    pollResponse = await pollMigrationStatus(derivedAccountId, migrationId, [
+      MIGRATION_STATUS.INPUT_REQUIRED,
+    ]);
+  } catch (error) {
+    SpinniesManager.fail('beginningMigration', {
+      text: lib.migrate.spinners.unableToStartMigration,
+    });
+    if (isMigrationStatus(error) && error.status === MIGRATION_STATUS.FAILURE) {
```
brandenrodgers, line 390: Not a blocker, but we do have `isSpecifiedError` in LDL that you could use here (doesn't buy you a ton tbh, outside of maybe consistency in util usage)

brandenrodgers (approved):
👌

lib/app/migrate.ts
joe-yeager: Good question.  These aren't actually `Error` objects, our polling util rejects with plain objects that look like:
```
{
   status: "SUCCESS" | "FAILURE" | "SOME_OTHER_STATUS"
   ... other fields
}
```

Wheres `isSpecifiedError` only operates on `HubSpotHttpErrors`

PR #1470, opened 2025-04-28
chore: Inline the remaining files

joe-yeager:
## Description and Context

In the remaining keys in preparation for the switch

### DIFF
--- a/commands/cms/lighthouseScore.ts
+++ b/commands/cms/lighthouseScore.ts
@@ -28,7 +28,7 @@ const DEFAULT_TABLE_HEADER = [
 ];
 
 exports.command = 'lighthouse-score [--theme]';
-exports.describe = false; // i18n(`${i18nKey}.describe`);
+exports.describe = false;
 
 const selectTheme = async accountId => {
   const { theme: selectedTheme } = await promptUser([
--- a/lib/doctor/Doctor.ts
+++ b/lib/doctor/Doctor.ts
@@ -35,8 +35,6 @@ const { i18n } = require('../lang');
 const { uiLink } = require('../ui');
 const minMajorNodeVersion = 18;
 
-const i18nKey = `lib.doctor`;
-
 export class Doctor {
   accountId: number | null;
   private diagnosis?: Diagnosis;
@@ -56,7 +54,7 @@ export class Doctor {
 
   async diagnose(): Promise<DiagnosticInfo> {
     SpinniesManager.add('runningDiagnostics', {
-      text: i18n(`${i18nKey}.runningDiagnostics`),
+      text: i18n(`lib.doctor.runningDiagnostics`),
     });
 
     this.diagnosticInfo =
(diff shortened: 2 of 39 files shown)


lib/doctor/Doctor.ts
```diff
@@ -56,7 +54,7 @@ export class Doctor {
 
   async diagnose(): Promise<DiagnosticInfo> {
     SpinniesManager.add('runningDiagnostics', {
-      text: i18n(`${i18nKey}.runningDiagnostics`),
+      text: i18n(`lib.doctor.runningDiagnostics`),
```
joe-yeager, line 57: Cursor seemed to have a hard time with this file.  It wanted to inline them like this: `${`lib.doctor`}.runningDiagnostics`, so I would pay extra special attention to this file.

lib/errorHandlers/suppressError.ts
```diff
@@ -42,17 +40,24 @@ function createPlatformVersionError(
...
-  logger.error(i18n(`${i18nKey}.platformVersionErrors.header`));
+  logger.error(
+    i18n(`lib.errorHandlers.suppressErrors.platformVersionErrors.header`)
+  );
   logger.log(
-    i18n(`${i18nKey}.platformVersionErrors.${translationKey}`, {
-      platformVersion,
-    })
+    i18n(
+      `lib.errorHandlers.suppressErrors.platformVersionErrors.${translationKey}`,
```
kemmerle, line 48: This `translationKey` needs to be deleted, correct?

joe-yeager: I left this alone because it is dynamic.  When we switch over to the new lang file, we can refactor it to be removed.  For this scope of this I was just trying to inline as much as possible

lib/prompts/selectPublicAppPrompt.ts
```diff
@@ -46,17 +46,24 @@ async function fetchPublicAppOptions(
         ? 'noAppsMigrationMessage'
         : 'noAppsCloneMessage';
       uiLine();
-      logger.error(i18n(`${i18nKey}.errors.${headerTranslationKey}`));
+      logger.error(
+        i18n(`lib.prompts.selectPublicAppPrompt.errors.${headerTranslationKey}`)
```
kemmerle, line 50: I'm assuming these translation keys are ok to leave as is, because there's some logic here, but wanted to flag just in case. There are three in this file.

commands/cms/lighthouseScore.ts
```diff
@@ -28,7 +28,7 @@ const DEFAULT_TABLE_HEADER = [
 ];
 
 exports.command = 'lighthouse-score [--theme]';
-exports.describe = false; // i18n(`${i18nKey}.describe`);
+exports.describe = false;
```
kemmerle, line 31: I think we can unhide this command now. Maybe check with @brandenrodgers

joe-yeager: The last time this came up, the answer was to let the CMS team determine the future of the commands.  I'm not sure where they landed on it

PR #1473, opened 2025-04-29
Safer copy logging POC

camden11:
## Description and Context
This creates a new typed `uiLogger` that will help prevent lang errors. The `uiLogger` is meant to be used when logging anything from the new lang objects. It only accepts strings, so it'll catch accidental `undefined` and uncalled functions.  This also updates all files that use the new lang objects to use `uiLogger`. It's already caught a few errors!

## Who to Notify


kemmerle (commented):
@camden11, looked this over, and I like the changes! I'm good if you're ready to implement this. Perhaps it would be easier to break it down into separate PRs?

lib/projects/buildAndDeploy.ts
```diff
@@ -323,15 +323,15 @@ function makePollTaskStatusFunc<T extends ProjectTask>({
...
+              uiLogger.log(
                 `${statusStrings.SUBTASK_FAIL(
                   failedSubtasks.length === 1
                     ? getSubtaskName(failedSubtasks[0])
                     : failedSubtasks.length + ' components',
                   displayId
                 )}\n`
               );
-              logger.log('See below for a summary of errors.');
+              uiLogger.log('See below for a summary of errors.');
```
kemmerle, line 334: Add this to the lang file?

lib/projects/config.ts
```diff
@@ -61,7 +61,7 @@ export async function getProjectConfig(
       projectConfig,
     };
   } catch (e) {
-    logger.error('Could not read from project config');
+    uiLogger.error('Could not read from project config');
```
kemmerle, line 64: Add to lang file?

brandenrodgers (approved):
Nice 👌 I bet we could use AI to fix the rest once Joe merges the PR to swap lang patterns

PR #1477, opened 2025-05-01
chore: Add migration warning

joe-yeager:
## Description and Context

Add a more explicit migration warning for projects:

image

### DIFF
--- a/lib/app/migrate.ts
+++ b/lib/app/migrate.ts
@@ -12,7 +12,6 @@ const inquirer = require('inquirer');
 
 import { confirmPrompt, inputPrompt, listPrompt } from '../prompts/promptUtils';
 import {
-  UI_COLORS,
   uiAccountDescription,
   uiCommandReference,
   uiLine,
@@ -48,7 +47,6 @@ import {
   getProjectDetailUrl,
 } from '../projects/urls';
 import { uiLogger } from '../ui/logger';
-import { logger } from '@hubspot/local-dev-lib/logger';
 
 export type MigrateAppArgs = CommonArgs &
   AccountArgs &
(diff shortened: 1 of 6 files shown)


lib/app/migrate.ts
```diff
@@ -264,11 +266,29 @@ async function selectAppToMigrate(
 
   uiLogger.log('');
 
-  const promptMessage = projectConfig?.projectConfig
-    ? `${lib.migrate.projectMigrationWarning} ${lib.migrate.prompt.proceed}`
-    : lib.migrate.prompt.proceed;
+  if (projectConfig?.projectConfig) {
```
joe-yeager, line 269: TODO: Update pending UX feedback

lib/ui/boxen.ts
```diff
@@ -0,0 +1,39 @@
...
+const defaultOptions: Options = {
+  titleAlignment: 'left',
+  borderColor: UI_COLORS.MARIGOLD,
+  margin: 1,
+  padding: 1,
+  textAlignment: 'left',
+  borderStyle: 'round',
+};
+
+export async function logInBox({
```
joe-yeager, line 16: I figured it made sense to make a wrapper around this so we don't need to repeat the same options and dynamic import in multiple places.

kemmerle (dismissed):
Code looks clean, and I like the warning design! :shipit:

lib/ui/boxen.ts
```diff
@@ -0,0 +1,39 @@
...
+  contents,
+  options,
+  fallBackToNoBox = false,
```
brandenrodgers, line 19: I love this variable name lol, but I'd also be fine with removing it and always falling back to no box. Or defaulting to true. We should probably be falling back to no box in the fire alarm middleware too.

joe-yeager: Haha, I had such a hard time naming it.  I'll make it fallback by default

brandenrodgers (approved):

PR #1479, opened 2025-05-05
Revamp local dev to support websocket server for UI

camden11:
## Description and Context

This revamps the local dev flow for unified apps to make it more flexible and modular. Since I've walked everyone through (and left inline comments to explain) I won't go into too much detail here, but the primary purpose of this PR is to separate the many responsibilities of the old `LocalDevManagerV2` into discrete classes.

While this makes a lot of code changes, the actual UX of running `hs project dev` should not change. Appreciate any help with testing to confirm that's true!

## TODO
* This sets things up to work with the Local Dev Websocket Server, which will be implemented in a separate PR.
* I also plan on updating `LocalDevState` to be immutable but going to do that in a separate PR as well (this is safe for now since the dev servers don't interact directly with `LocalDevState`

## Who to Notify


commands/project/dev/unifiedFlow.ts
```diff
@@ -167,7 +169,16 @@ export async function unifiedProjectDevFlow(
...
+  const watcher = new LocalDevWatcher(localDevProcess);
+  watcher.start();
+
+  handleKeypress(async key => {
+    if ((key.ctrl && key.name === 'c') || key.name === 'q') {
+      await Promise.all([localDevProcess.stop(), watcher.stop()]);
+    }
+  });
+
+  handleExit(({ isSIGHUP }) => localDevProcess.stop(!isSIGHUP));
```
camden11, line 183: `LocalDevProcess` represents the current local dev session (similar to the old `LocalDevManager`). Everything that interacts with the `LocalDevProcess` now lives _outside_ of it and uses public methods to make necessary changes to the process.

This includes
* Watcher
* Keypress listener
* In the future, the websocket server

lib/projects/localDev/AppDevModeInterface.ts
```diff
@@ -0,0 +1,193 @@
+import { fetchAppInstallationData } from '@hubspot/local-dev-lib/api/localDevAuth';
```
camden11, line 1: This is a wrapper around the `UIEDevModeInterface` and includes all app-specific logic that previously lived in `LocalDevManager`. This lets us keep `LocalDevProcess` component agnostic

lib/projects/localDev/DevServerManagerV2.ts
```diff
@@ -41,54 +52,38 @@ class DevServerManagerV2 {
     await Promise.all(this.devServers.map(devServer => callback(devServer)));
   }
 
-  async setup({
```
camden11, line 44: These lifecycle methods don't need to accept arguments anymore, since they use the `LocalDevState`

lib/projects/localDev/LocalDevLogger.ts
```diff
@@ -0,0 +1,249 @@
...
+    if (warning !== this.mostRecentUploadWarning) {
+      uiLogger.log('');
+      uiLogger.warn(warning);
+      this.logUploadInstructions();
+
+      this.mostRecentUploadWarning = warning;
+    }
+  }
+
+  addUploadWarning(warning: string): void {
```
camden11, line 74: This allows dev servers to add specific upload warnings that the logger will then log whenever the upload warning appears. Currently used when apps in a project have production installs, but in the future, other dev servers could also add messages here. For now, it lets us keep the `LocalDevLogger` component agnostic

lib/projects/localDev/LocalDevManagerV2.ts
```diff
@@ -1,557 +0,0 @@
-import path from 'path';
```
camden11, line 1: Broken up into `LocalDevProcess`, `LocalDevLogger`, `LocalDevWatcher`, and `AppDevModeInterface`

lib/projects/localDev/LocalDevProcess.ts
```diff
@@ -0,0 +1,198 @@
...
+    projectConfig,
+    projectDir,
+    projectId,
+    debug,
+    deployedBuild,
+    isGithubLinked,
+    initialProjectNodes,
+    env,
+  }: LocalDevProcessConstructorOptions) {
+    this.state = {
```
camden11, line 45: It could make sense for `LocalDevState` to be its own class in the future, especially if dev servers will be mutating it. For now, I think this object will suffice

lib/projects/localDev/LocalDevUIWebsocketServer.ts
```diff
@@ -0,0 +1,115 @@
...
+} from '@hubspot/local-dev-lib/portManager';
+import { logger } from '@hubspot/local-dev-lib/logger';
+import { LOCAL_DEV_UI_WEBSOCKET_MESSAGE_TYPES } from '../../constants';
+import { LocalDevUIWebsocketMessage } from '../../../types/LocalDev';
+import { ProjectConfig } from '../../../types/Projects';
+const SERVER_INSTANCE_ID = 'local-dev-ui-websocket-server';
+
+const LOG_PREFIX = '[LocalDevUIWebsocketServer] ';
+
+class LocalDevUIWebsocketServer {
```
camden11, line 14: This is still WIP - started working on it before I realized that the rest of the work in this PR was necessary. But figured it made sense to keep it in here and add to it in following PRs

lib/projects/localDev/LocalDevWatcher.ts
```diff
@@ -0,0 +1,70 @@
...
+import LocalDevProcess from './LocalDevProcess';
+
+const WATCH_EVENTS = {
+  add: 'add',
+  change: 'change',
+  unlink: 'unlink',
+  unlinkDir: 'unlinkDir',
+};
+
+class LocalDevWatcher {
```
camden11, line 14: This is a good example of how external sources can interact with the `LocalDevProcess`

lib/projects/localDev/LocalDevUIWebsocketServer.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (!portManagerIsRunning) {
+      throw new Error(
+        '@TODO: PortManagerServing must be running before starting LocalDevUIWebsocketServer.'
+      );
+    }
+
+    const portData = await requestPorts([{ instanceId: SERVER_INSTANCE_ID }]);
+    const port = portData[SERVER_INSTANCE_ID];
+
+    this._server = new WebSocketServer({ port });
```
joe-yeager, line 99: If we want, we can attach this `WebSocketServer` to the express server that we already use for the port manager.  This would limit the number of ports we use on the users machine and allow us to have a reliable fixed port. 

https://github.com/websockets/ws?tab=readme-ov-file#external-https-server

lib/projects/localDev/LocalDevProcess.ts
joe-yeager: I think it might be a good idea to make the object immutable so that the other dev servers can't change it.  We would want to be the source of truth for this information and probably wouldn't want it to change.

commands/project/dev/unifiedFlow.ts
kemmerle: I'd add a note with that explanation directly into the code.

lib/projects/localDev/LocalDevLogger.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    if (warning !== this.mostRecentUploadWarning) {
+      uiLogger.log('');
+      uiLogger.warn(warning);
+      this.logUploadInstructions();
+      this.mostRecentUploadWarning = warning;
+    }
+  }
+
+  fileChangeError(e: unknown): void {
```
kemmerle, line 91: You could replace the next four functions with a single function like this:

```ts
  private handleError(
    e: unknown,
    translationFn: (message: string) => string
  ): void {
    if (this.state.debug) {
      logger.error(e);
    }
    uiLogger.error(translationFn(e instanceof Error ? e.message : ''));
  }
  ```

**Review by kemmerle — COMMENTED** [review]
I only had minor comments. Overall, I think this is a major improvement for local dev--the modular code is much easier to understand now. Ping me when you're finished making changes and I can give it a final :shipit:  

One other little comment: You might want to rename the PR given that the websocket work has been removed.

File: lib/projects/localDev/LocalDevLogger.ts (hunk 10 above)
Comment (camden11) [reply lib/projects/localDev/LocalDevLogger.ts:91]: Good call!

File: lib/projects/localDev/LocalDevLogger.ts [hunk 11]
```diff
@@ -0,0 +1,230 @@
...
+    }
+    uiLogger.log(lib.LocalDevManager.uploadWarning.restartDev);
+  }
+
+  private handleError(
+    e: unknown,
+    langFunction: (message: string) => string
+  ): void {
+    if (this.state.debug) {
+      logger.error(e);
```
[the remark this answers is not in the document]
camden11: Previously, `LocalDevManagerV2` was using this pattern with using `logger.error`, I assume to make errors more visible when in `debug` mode. I decided to stick with that pattern

lang/en.ts
```diff
@@ -2689,11 +2693,20 @@ export const lib = {
       `Your project ${chalk.bold(projectName)} exists in ${accountIdentifier}, but has no deployed build. Projects must be successfully deployed to be developed locally. Address any build and deploy errors your project may have, then run ${uploadCommand} to upload and deploy your project.`,
     noComponents: 'There are no components in this project.',
     betaMessage: 'HubSpot projects local development',
-    learnMoreLocalDevServer: 'Learn more about the projects local dev server',
+    learnMoreLocalDevServer: uiLink(
```
brandenrodgers, line 2696: Are these translations in the en.lyaml file too? Would it make sense to remove them as a part of this, or should we wait and do that later once everything is ported?

brandenrodgers (approved):
I tested out a handful of different edge cases and this seems to work the same as the existing setup 🎉

lang/en.ts
camden11: I've been operating under the assumption that we'd just delete the entire file at the end, so haven't been removing things from it. I don't think there's really any need to, especially since its easy to look up what we haven't ported to the new system by searching for `i18n(`

PR #1482, opened 2025-05-07
fix: Fix platform version logs

joe-yeager:
## Description and Context

The errors for invalid platform versions logs were broken. 

### Before
Screenshot 2025-05-07 at 12 11 52 PM

### Invalid Version
Screenshot 2025-05-07 at 12 12 23 PM

### Retired version
Screenshot 2025-05-07 at 12 13 03 PM

### DIFF
--- a/lib/errorHandlers/suppressError.ts
+++ b/lib/errorHandlers/suppressError.ts
@@ -20,23 +20,17 @@ function createPlatformVersionError(
 ): void {
   let translationKey = 'unspecifiedPlatformVersion';
   let platformVersion = 'unspecified platformVersion';
-  const errorContext = err.data.context;
+  const errorContext = err?.data?.context;
 
-  switch (subCategory) {
-    case PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_RETIRED:
-      translationKey = 'platformVersionRetired';
-      if (errorContext && errorContext[subCategory]) {
-        platformVersion = errorContext[subCategory];
-      }
-      break;
-    case PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST:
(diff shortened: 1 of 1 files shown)


lib/errorHandlers/suppressError.ts
```diff
@@ -20,23 +20,17 @@ function createPlatformVersionError(
 ): void {
   let translationKey = 'unspecifiedPlatformVersion';
   let platformVersion = 'unspecified platformVersion';
-  const errorContext = err.data.context;
+  const errorContext = err?.data?.context;
```
joe-yeager, line 23: This line was causing the issue with accessing `.context` on `undefined`, because when we were invoking this function we were passing it `err.data` instead of just `err`, so it was trying to access `err.data.data.context`

lib/errorHandlers/suppressError.ts
```diff
@@ -20,23 +20,17 @@ function createPlatformVersionError(
...
-      break;
+  if (subCategory === PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_RETIRED) {
+    platformVersion = errorContext?.RETIRED_PLATFORM_VERSION ?? platformVersion;
+    translationKey = 'platformVersionRetired';
+  } else if (
+    subCategory ===
+    PLATFORM_VERSION_ERROR_TYPES.PLATFORM_VERSION_SPECIFIED_DOES_NOT_EXIST
+  ) {
+    platformVersion = errorContext?.PLATFORM_VERSION ?? platformVersion;
+    translationKey = 'nonExistentPlatformVersion';
```
joe-yeager, line 33: When I fixed the above, I also noticed that these were incorrect.  We were looking for the existence of the wrong keys so it always just said `unspecified platformVersion` instead of listing the actual invalid platform version

kemmerle (approved):
Good catches!

PR #1484, opened 2025-05-09
Port cms convert-fields to TS

brandenrodgers:
## Description and Context

Porting the `hs cms convert-fields` command to TS.

This relies on the small LDL PR here: https://github.com/HubSpot/hubspot-local-dev-lib/pull/267

### DIFF
--- a/commands/cms/convertFields.ts
+++ b/commands/cms/convertFields.ts
@@ -1,66 +1,79 @@
-// @ts-nocheck
-const path = require('path');
-const fs = require('fs');
-const { createIgnoreFilter } = require('@hubspot/local-dev-lib/ignoreRules');
-const { isAllowedExtension, getCwd } = require('@hubspot/local-dev-lib/path');
-const { logger } = require('@hubspot/local-dev-lib/logger');
-const { walk } = require('@hubspot/local-dev-lib/fs');
-const { getThemeJSONPath } = require('@hubspot/local-dev-lib/cms/themes');
-const { i18n } = require('../../lib/lang');
-const {
+import { Argv, ArgumentsCamelCase } from 'yargs';
+import path from 'path';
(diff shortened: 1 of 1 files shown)


commands/cms/convertFields.ts
```diff
@@ -1,66 +1,79 @@
...
+}
 
-exports.handler = async options => {
-  let stats;
-  let projectRoot;
-  let src;
+async function handler(
+  args: ArgumentsCamelCase<ConvertFieldsArgs>
+): Promise<void> {
+  let stats: fs.Stats | undefined;
```
camden11, line 40: nit but you can use `?:` instead of `| undefined`

camden11 (approved):

commands/cms/convertFields.ts
brandenrodgers: I'm not sure what's going on with my TS config, but for some reason my IDE is telling me that it's not a valid syntax, even though it is. Tried restarting my ts server but no luck  🤷 

image

joe-yeager (approved):

commands/cms/convertFields.ts
joe-yeager: That's not valid syntax for assignments.  Only types, function args, and types/interfaces.  

[Example](https://www.typescriptlang.org/play/?#code/PTAEDUEMBsEsBMBQAzArgOwMYBdYHt1Rk88AKSAfgC5QBnbAJ1nQHMBKUAb0UQF8eQEGAkTYAngAcApkRKgAvKC6IlSyjXpNWfAWChwkmaJFq1QAMTndVoCUwBukbDPV1GzFgG4diQQEl0RwNEaClsUFdND28gA)

joe-yeager (approved):

PR #1487, opened 2025-05-09
chore: Convert function ls command to TS

kemmerle:
## Description and Context
This converts the `hs function list` command to TypeScript. 

## Who to Notify

### DIFF
--- a/commands/function.ts
+++ b/commands/function.ts
@@ -5,6 +5,7 @@ const deploy = require('./function/deploy');
 const server = require('./function/server');
 import { i18n } from '../lib/lang';
 import { makeYargsBuilder } from '../lib/yargsUtils';
+import { YargsCommandModuleBucket } from '../types/Yargs';
 
 export const command = ['function', 'functions'];
 export const describe = i18n(`commands.function.describe`);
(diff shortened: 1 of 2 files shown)


commands/function.ts
```diff
@@ -1,16 +1,16 @@
-// @ts-nocheck
```
brandenrodgers, line 1: We're doing these command buckets differently now. We have some custom types that we're using to get some more value out of TS. Here's an example: https://github.com/HubSpot/hubspot-cli/blob/main/commands/project.ts

brandenrodgers:
Just one comment about how we're typing the command buckets, but the rest lgtm!

camden11 (approved):

commands/function.ts
```diff
@@ -4,13 +4,27 @@ import list from './function/list';
...
+function functionBuilder(yargs: Argv): Argv {
   addGlobalOptions(yargs);
   yargs.command(list).command(deploy).command(server).demandCommand(1, '');
-
   return yargs;
 }
+
+const builder = makeYargsBuilder(functionBuilder, command, describe);
+
+const functionCommand = {
```
brandenrodgers, line 20: This can use the `YargsCommandModuleBucket` type

PR #1490, opened 2025-05-09
chore: Port the module commands

joe-yeager:
## Description and Context

Port the module commands to TS

### DIFF
--- a/commands/module.ts
+++ b/commands/module.ts
@@ -1,20 +1,28 @@
-// @ts-nocheck
-const marketplaceValidate = require('./module/marketplace-validate');
-const {
-  addConfigOptions,
-  addAccountOptions,
-  addGlobalOptions,
-} = require('../lib/commonOpts');
+import { Argv } from 'yargs';
+import marketplaceValidate from './module/marketplace-validate';
+import { CommandModuleBucket } from '../types/Yargs';
+import { makeYargsBuilder } from '../lib/yargsUtils';
 
-exports.command = 'module';
-exports.describe = false;
+const command = 'module';
+const describe = undefined;
 
-exports.builder = yargs => {
-  addConfigOptions(yargs);
-  addAccountOptions(yargs);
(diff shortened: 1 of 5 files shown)


types/Yargs.ts
```diff
@@ -40,4 +41,9 @@ export interface YargsCommandModule<T, U> extends CommandModule<T, U> {
   builder: (yargs: Argv) => Promise<Argv<U>>;
 }
 
+export type CommandModuleBucket<T, U> = PartialProp<
+  YargsCommandModule<T, U>,
+  'handler'
+>;
+
```
joe-yeager, line 48: I just realized there is a type right below this for the `YargsCommandModuleBucket`.  Do we want to use that and add handlers?  Or use this and leave them as optional?

brandenrodgers: I kept `handler` on the bucket type because I ran into an issue with nested command buckets. `yargs.command()` gets mad if you pass it something that doesn't include a handler. So for nested buckets, like `hs custom-object schema <something>` the typing for `schema` needs to have a handler. I assume this will also matter once we port cli.js to TS since it calls `yargs.command` on every bucket. Maybe there's another way around this that I'm not seeing though.

commands/module.ts
```diff
@@ -1,20 +1,28 @@
...
+  useAccountOptions: true,
+});
+
+const moduleCommand: CommandModuleBucket<unknown, ModuleArgs> = {
+  command,
+  describe,
+  builder,
 };
+
+export default moduleCommand;
```
brandenrodgers, line 28: Does this also need a default `module.export`? You can test with `hs module --help` I think

types/Yargs.ts
joe-yeager: Gotcha, I'll switch to the other type and add an empty handler

kemmerle (approved):

PR #1493, opened 2025-05-09
chore: Port lang/en.ts

joe-yeager:
## Description and Context

Port the lang/en.ts

### DIFF
--- a/lang/en.ts
+++ b/lang/en.ts
@@ -149,31 +149,31 @@ export const commands = {
         },
         inactiveAccountsFound: {
           one: '1 inactive account found:',
-          other: count => `${count} inactive accounts found:`,
+          other: (count: string) => `${count} inactive accounts found:`,
         },
         confirm: {
           one: 'Remove 1 inactive account from the CLI config?',
-          other: count =>
+          other: (count: string) =>
             `Remove ${count} inactive accounts from the CLI config?`,
         },
-        removeSuccess: accountName =>
+        removeSuccess: (accountName: string) =>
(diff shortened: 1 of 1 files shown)


lang/en.ts
```diff
@@ -8,7 +7,7 @@ import {
 } from '../lib/ui';
 import { getProjectSettingsUrl } from '../lib/projects/urls';
 
-type LangFunction = (...args: (string | number)[]) => string;
+type LangFunction = (...args: never[]) => string;
```
joe-yeager, line 10: I updated this to never because it's impossible and know all of the permutations of possible arguments, and as far as the satisfies is concerned we don't really care the shape of the arguments, just just care that it is a function that returns a string

lang/en.ts
```diff
@@ -18,19 +17,23 @@ export const commands = {
   generalErrors: {
     updateNotify: {
       notifyTitle: 'Update available',
-      cmsUpdateNotification: (packageName, updateCommand) =>
+      cmsUpdateNotification: (packageName: string, updateCommand: string) =>
```
joe-yeager, line 20: I coached `Claude` to convert all of these to `string` because that is **_probably_** the most likely type.  As we switch over we can tweak the types to match reality

camden11 (approved):
I was kind of pleasantly surprised when my previous solution worked...turns out it was because of the `ts-nocheck`

This LGTM though, accomplishes all the same things and now that its all converted, it wont be that much of a pain to add new arg types when needed

PR #1494, opened 2025-05-09
chore: Convert `hs function deploy` to TS

kemmerle:
## Description and Context
This PR converts the `hs function deploy` command to TS

## Who to Notify

### DIFF
--- a/commands/function/deploy.ts
+++ b/commands/function/deploy.ts
@@ -22,6 +22,18 @@ import {
 } from '../../types/Yargs';
 import { makeYargsBuilder } from '../../lib/yargsUtils';
 
+type FunctionBuildError = {
+  status: 'ERROR';
+  errorReason: string;
+  cdnUrl: string;
+};
+
+function isFunctionBuildError(e: unknown): e is FunctionBuildError {
+  return (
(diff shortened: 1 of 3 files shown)


commands/function/deploy.ts
```diff
@@ -88,11 +105,18 @@ exports.handler = async options => {
...
-      await outputBuildLog(e.cdnUrl);
+    } else if (
+      typeof e === 'object' &&
+      e !== null &&
+      'status' in e &&
+      e.status === 'ERROR' &&
+      'cdnUrl' in e &&
+      'errorReason' in e
+    ) {
+      await outputBuildLog(e.cdnUrl as string);
```
camden11, line 116: These errors are tough to deal with - might be worth making a type + type guarding function for this so you dont have to cast stuff and do these checks inline

```
type FunctionBuildError = {
  status: 'ERROR';
  errorReason: string;
  cdnUrl: string;
};

function isFunctionBuildError(e: unknown): e is FunctionBuildError {
  return (
    typeof e === 'object' && e !== null && 'status' in e && e.status === 'ERROR'
  );
}```

camden11 (approved):
One comment, otherwise LGTM

PR #1495, opened 2025-05-09
chore: Port the fetch command to ts

joe-yeager:
## Description and Context

Port the fetch command to TS

### DIFF
--- a/commands/fetch.ts
+++ b/commands/fetch.ts
@@ -26,7 +26,6 @@ import { logError } from '../lib/errorHandlers/index';
 type FetchCommandArgs = {
   src: string;
   dest?: string;
-  derivedAccountId?: number;
   cmsPublishMode?: CmsPublishMode;
   staging?: boolean;
   assetVersion?: number;
--- a/commands/fetch.ts
+++ b/commands/fetch.ts
@@ -14,6 +14,7 @@ import { i18n } from '../lib/lang';
 import { makeYargsBuilder } from '../lib/yargsUtils';
 import {
   AccountArgs,
+  CommonArgs,
   ConfigArgs,
   EnvironmentArgs,
   YargsCommandModule,
@@ -32,7 +33,8 @@ type FetchCommandArgs = {
   overwrite?: boolean;
 } & ConfigArgs &
   AccountArgs &
-  EnvironmentArgs;
+  EnvironmentArgs &
+  CommonArgs;
 
 const command = 'fetch <src> [dest]';
 const describe = i18n('commands.fetch.describe');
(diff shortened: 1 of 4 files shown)


commands/fetch.ts
```diff
@@ -38,22 +44,27 @@ exports.handler = async options => {
...
     await downloadFileOrFolder(
-      derivedAccountId,
+      derivedAccountId!,
       src,
       resolveLocalPath(dest),
       cmsPublishMode,
-      options
+      {
+        assetVersion:
+          assetVersion !== undefined ? `${assetVersion}` : assetVersion,
```
joe-yeager, line 57: This is `type: number` and optional and it needs to be converted to `string | undefined`

commands/fetch.ts
```diff
@@ -38,22 +44,27 @@ exports.handler = async options => {
 
   trackCommandUsage('fetch', { mode: cmsPublishMode }, derivedAccountId);
 
+  const { assetVersion, staging } = options;
   try {
     // Fetch and write file/folder.
     await downloadFileOrFolder(
-      derivedAccountId,
+      derivedAccountId!,
```
joe-yeager, line 51: I'm not crazy about using the non null assertion operator, but I was trying to prevent modifying behavior because the yargs types are incorrect.  My understanding is `derivedAccountId` will always be defined

brandenrodgers: This should be fixed when you leverage the CommonArgs type

commands/fetch.ts
```diff
@@ -80,12 +91,30 @@ exports.builder = yargs => {
     },
   });
 
-  addConfigOptions(yargs);
-  addAccountOptions(yargs);
-  addOverwriteOptions(yargs);
```
brandenrodgers, line 85: Does makeYargsBuilder support `useOverwriteOptions` to replace this?

joe-yeager: No, I deleted this by mistake and have a fix pending local.  It's only used in two places, so I think calling it in the builder makes sense.

commands/fetch.ts
```diff
@@ -1,27 +1,47 @@
...
+
+import { EXIT_CODES } from '../lib/enums/exitCodes';
+import { logError } from '../lib/errorHandlers/index';
 
-const { EXIT_CODES } = require('../lib/enums/exitCodes');
-const { logError } = require('../lib/errorHandlers/index');
+type FetchCommandArgs = {
+  src: string;
+  dest?: string;
+  derivedAccountId?: number;
```
brandenrodgers, line 29: Nit, but don't need `derivedAccountId` in here anymore since it's defined in CommonArgs

brandenrodgers (dismissed):
Just a nit about a duplicate type definition, but otherwise lgtm 👌

brandenrodgers (approved):

PR #1501, opened 2025-05-09
chore: Port `hs logs`

joe-yeager:
## Description and Context

Port the hs logs command

### DIFF
--- a/lib/interpolation.ts
+++ b/lib/interpolation.ts
@@ -36,7 +36,7 @@ const delimiters: {
 };
 
 type InterpolationData = {
-  [identifier: string]: string | number;
+  [identifier: string]: unknown;
 };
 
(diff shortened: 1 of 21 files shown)


lib/interpolation.ts
```diff
@@ -35,8 +35,8 @@ const delimiters: {
   },
 };
 
-type InterpolationData = {
-  [identifier: string]: string | number;
+export type InterpolationData = {
+  [identifier: string]: unknown;
```
joe-yeager, line 39: I opened this up so we don't need to do weird stuff like:
```typescript
i18n('some.key', {
  value: `${value}`
});  
```

These values are being cast to a string already, so these types should all be ok:

https://jsbin.com/bekepus/3/edit?js,console

kemmerle (approved):

PR #1503, opened 2025-05-09
Convert hs create command bucket to TS

camden11:
## Description and Context
This converts the `hs create` command/bucket (not really a true bucket) to TS

## Who to Notify

### DIFF
--- a/types/Cms.ts
+++ b/types/Cms.ts
@@ -1,7 +1,13 @@
-// TODO: Add real type
-export type CreateArgs = {
-  branch?: string;
-};
+import { CommonArgs, ConfigArgs } from './Yargs';
+
+export type CreateArgs = CommonArgs &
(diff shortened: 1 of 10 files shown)


types/Cms.ts
```diff
@@ -1,7 +1,13 @@
-// TODO: Add real type
-export type CreateArgs = {
-  branch?: string;
-};
+import { CommonArgs, ConfigArgs } from './Yargs';
+
+export type CreateArgs = CommonArgs &
```
brandenrodgers, line 3: nit, this can live in the command file if it's only used there

brandenrodgers (approved):

types/Cms.ts
camden11: It's used in this file too!

PR #1504, opened 2025-05-09
chore: Port watch

joe-yeager:
## Description and Context

Port the `hs watch` command

### DIFF
--- a/commands/watch.ts
+++ b/commands/watch.ts
@@ -1,32 +1,52 @@
-// @ts-nocheck
-const fs = require('fs');
-const path = require('path');
+import fs from 'fs';
+import path from 'path';
+import { Argv, ArgumentsCamelCase } from 'yargs';
+import { AxiosError } from 'axios';
 
-const { watch } = require('@hubspot/local-dev-lib/cms/watch');
-const { getCwd } = require('@hubspot/local-dev-lib/path');
-const { logger } = require('@hubspot/local-dev-lib/logger');
+import { watch } from '@hubspot/local-dev-lib/cms/watch';
+import { getCwd } from '@hubspot/local-dev-lib/path';
+import { logger } from '@hubspot/local-dev-lib/logger';
 
-const {
-  addConfigOptions,
-  addAccountOptions,
-  addCmsPublishModeOptions,
-  addUseEnvironmentOptions,
-  addGlobalOptions,
-  getCmsPublishMode,
-} = require('../lib/commonOpts');
-const { uploadPrompt } = require('../lib/prompts/uploadPrompt');
-const { validateCmsPublishMode } = require('../lib/validation');
-const { trackCommandUsage } = require('../lib/usageTracking');
-const { i18n } = require('../lib/lang');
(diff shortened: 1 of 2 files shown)


commands/watch.ts
```diff
@@ -97,44 +119,47 @@ exports.handler = async options => {
...
+      disableInitial: !initialUpload,
       notify,
       commandOptions: options,
       filePaths: filesToUpload,
     },
-    postInitialUploadCallback,
+    null,
     onUploadFolderError,
-    onQueueAddError,
+    undefined,
```
joe-yeager, line 157: We were setting these variables to `null` and `undefined` and then passing them

commands/watch.ts
```diff
@@ -1,32 +1,53 @@
...
+    notify?: string;
+    convertFields?: boolean;
+    saveOutput?: boolean;
+  };
+
+const command = 'watch [src] [dest]';
+const describe = i18n(`commands.watch.describe`);
+
+const handler = async (
+  options: ArgumentsCamelCase<WatchCommandArgs>
```
brandenrodgers, line 49: nit, but for consistency we're using `args` in the handler now

commands/watch.ts
```diff
@@ -181,11 +206,31 @@ exports.builder = yargs => {
...
+const builder = makeYargsBuilder<WatchCommandArgs>(
+  watchBuilder,
+  command,
+  describe,
+  {
+    useConfigOptions: true,
+    useAccountOptions: true,
+    useGlobalOptions: true,
+    useEnvironmentOptions: true,
+    useCmsPublishModeOptions: { write: true },
```
brandenrodgers, line 221: I don't think this will work here. This will need to go in the watchBuilder

commands/watch.ts
joe-yeager: Oh dang, I've been using options everywhere

commands/watch.ts
```diff
@@ -181,11 +206,31 @@ exports.builder = yargs => {
...
+  }
+);
+
+const watchCommand: YargsCommandModuleBucket = {
```
brandenrodgers, line 225: Looks like the handler is typed weird here because this is using the YargsCommandModuleBucket type instead of the YargsCommandModule type

commands/watch.ts
joe-yeager: @kemmerle added it in one of her most recent PRs

commands/watch.ts
```diff
@@ -181,11 +206,31 @@ exports.builder = yargs => {
...
+    useGlobalOptions: true,
+    useEnvironmentOptions: true,
+    useCmsPublishModeOptions: { write: true },
+  }
+);
+
+const watchCommand: YargsCommandModule<unknown, WatchCommandArgs> = {
+  command,
+  describe,
+  handler: handler as unknown as (
```
brandenrodgers, line 228: This can just be `handler`. No need for any of the other stuff afaik

joe-yeager: Whoops, I'll remove it.

camden11 (approved):

PR #1511, opened 2025-05-12
Fix: Bugs in `account` commands

kemmerle:
## Description and Context
This PR addresses bugs that I discovered when testing out this LDL [PR](https://github.com/HubSpot/hubspot-local-dev-lib/pull/270), although the two do not rely on one another. 

1) It fixes an issue with `hs account remove`. If your default account were set to the account ID (`1234567`), but you removed the account using its `name` (ie. `hs account remove accountName`), we were not properly prompting users to choose another default account. 

2) Both `hs account create-override` and `hs account remove-override` did not properly warn customers that the commands require a global config file. 

## Screenshots
New error message for `override` commands:

Screenshot 2025-05-12 at 2 59 17 PM

## TODO

- [ ] Address feedback

## Who to Notify

### DIFF
--- a/lang/en.lyaml
+++ b/lang/en.lyaml
@@ -39,7 +39,7 @@ en:
           prompts:
             replaceOverrideFile: "Replace existing account override file?"
           errors:
-            globalConfigNotFound: "A global config file could not be found at {{ globalConfigPath }}, which is required for account overrides. Please run {{ authCommand }} to create one."
+            globalConfigNotFound: "This command is only compatible with our new global config. Run {{ authCommand }} to get started."
             accountNotFound: "The specified account could not be found in the config file {{ configPath }}"
           options:
             account:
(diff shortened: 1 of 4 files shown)


lang/en.lyaml
```diff
@@ -39,6 +39,7 @@ en:
           prompts:
             replaceOverrideFile: "Replace existing account override file?"
           errors:
+            globalConfigNotFound: "A global config file could not be found at {{ globalConfigPath }}, which is required for account overrides. Please run {{ authCommand }} to create one."
```
brandenrodgers, line 42: Copy suggestion. We could shorten this to: `This command is only compatible with our new global config. Run {{ authCommand }} to get started.`

brandenrodgers (approved):
Code lgtm. Just left one suggestion to make the error more concise

brandenrodgers (approved):

PR #1513, opened 2025-05-13
Setup local dev websocket server

camden11:
## Description and Context
This does the initial setup work for the `LocalDevWebsocketServer` and adds project upload functionality as an example of how it will work. It's still a work in progress, so leaving it commented out for now.

## Screenshots
https://hubspot.slack.com/archives/C076QV345ME/p1747160870258469

## TODO
The rest of the `LocalDevWebsocketServer` work will be added in follow up PRs. The next step is probably to update `LocalDevState` to be immutable, and to establish a pattern for `LocalDevProcess` and the dev servers to interact with the server.

### Testing
If you want, you can test this out using Postman or Insomnia or something similar.
1. Uncomment the websocket related things in `unifiedFlow.ts`
2. Run `hs project dev`
3. Grab the websocket server port, either by running `hs project dev` with `--debug` and looking at the logs, or by hitting the port manager server `localhost:8000/servers`
4. Connect to the websocket server on Insomnia
5. Send an upload message `{"type": "upload"}

## Who to Notify

### DIFF
--- a/lib/projects/localDev/LocalDevWebsocketServer.ts
+++ b/lib/projects/localDev/LocalDevWebsocketServer.ts
@@ -7,13 +7,14 @@ import { logger } from '@hubspot/local-dev-lib/logger';
 import { LOCAL_DEV_UI_WEBSOCKET_MESSAGE_TYPES } from '../../constants';
 import { LocalDevUIWebsocketMessage } from '../../../types/LocalDev';
 import LocalDevProcess from './LocalDevProcess';
+import { lib } from '../../../lang/en';
 
 const SERVER_INSTANCE_ID = 'local-dev-ui-websocket-server';
 
-const LOG_PREFIX = '[LocalDevUIWebsocketServer] ';
+const LOG_PREFIX = '[LocalDevWebsocketServer] ';
 
-class LocalDevUIWebsocketServer {
-  private _server?: WebSocketServer;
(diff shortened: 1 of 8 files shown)


brandenrodgers (approved):
Nice!

lib/projects/localDev/LocalDevWebsocketServer.ts
```diff
@@ -0,0 +1,110 @@
...
+    const portData = await requestPorts([{ instanceId: SERVER_INSTANCE_ID }]);
+    const port = portData[SERVER_INSTANCE_ID];
+
+    this.server = new WebSocketServer({ port });
+
+    this.log(lib.LocalDevWebsocketServer.logs.startup(port));
+    this.server.on('connection', ws => {
+      this._websocket = ws;
+      this.setupMessageHandlers();
+    });
```
joe-yeager, line 100: I think this might be problematic for multiple connections.  Anytime a new connection is made, I believe it will overwrite the previous connection.  So if the user were to have multiple local dev page tabs open, only the most recent tab would have a functioning websocket connection

camden11: Ohhh this is a good callout. I didn't test beyond one connection. I'll address this in another PR!

PR #1514, opened 2025-05-13
Support auto updates (opt-in)

brandenrodgers:
## Description and Context

Depends on this LDL work: https://github.com/HubSpot/hubspot-local-dev-lib/pull/272

This enables users to opt into auto-updates. The CLI will check once a day (via [update notifier](https://www.npmjs.com/package/update-notifier)) to see if the user's CLI version is outdated. If it is, it will check to see if auto updates are enabled. If they are, it will try to download the latest version automatically (currently requires the CLI to have been globally installed with npm). Otherwise, it will show the user the update notification like it does today.

I added `hs config set --allow-auto-updates` to let folks opt-into the behavior. We can keep this hidden until we're confident that it's working correctly.

The CLI will only attempt to auto update itself if the available update is a minor or patch release. We won't try to do it for major releases.


commands/config.ts
```diff
@@ -1,17 +1,29 @@
...
+const command = 'config';
+const describe = i18n('commands.config.describe');
 
+function configBuilder(yargs: Argv): Argv {
   yargs.command(set).command(migrate).demandCommand(1, '');
 
   return yargs;
 }
+
+const builder = makeYargsBuilder(configBuilder, command, describe);
```
brandenrodgers, line 17: This will make reviews slightly more annoying (sorry), but I tried to clean up the areas that I was touching while adding this feature. We need to make these changes eventually anyways.

lib/middleware/autoUpdateMiddleware.ts
```diff
@@ -0,0 +1,100 @@
+// Update-notifier is CommonJS, so we need to use require
```
brandenrodgers, line 1: This file contains the core of the auto update logic. It wasn't very difficult to do because we already had the utils to install deps. I abstracted them into the npm.ts file and leveraged them in here.

commands/config/set.ts
```diff
@@ -1,5 +1,5 @@
 import { ArgumentsCamelCase, Argv } from 'yargs';
```
brandenrodgers, line 1: Some updates to the TS patterns in here, but the work in this file adds support for `hs config set --allow-auto-updates`

lang/en.ts
```diff
@@ -2810,6 +2788,24 @@ export const lib = {
         )} to authenticate the App Developer Account ${parentAccountId} associated with ${accountIdentifier}.`,
     },
   },
+  middleware: {
```
brandenrodgers, line 2791: Just moving these from commands to lib.middleware

brandenrodgers:
I forgot to mention this in my initial comment, but you can easily test this by going into the `~/.config/configstore/update-notifier-@hubspot/cli.json` file and modifying the timestamp in there to be some time that's greater than a day ago. Then run some random CLI commands to see if it triggers an update.

lib/configOptions.ts
```diff
@@ -9,28 +9,28 @@ import { CMS_PUBLISH_MODE } from '@hubspot/local-dev-lib/constants/files';
 import { commaSeparatedValues } from '@hubspot/local-dev-lib/text';
 import { trackCommandUsage } from './usageTracking';
 import { promptUser } from './prompts/promptUtils';
-import { i18n } from '../lib/lang';
+import { lib } from '../lang/en';
+import { uiLogger } from './ui/logger';
 
-async function enableOrDisableUsageTracking(): Promise<boolean> {
+async function enableOrDisableBooleanFieldPrompt(
```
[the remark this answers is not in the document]
brandenrodgers: I thought about that, but it's so simple and it's only used in this file so I figured it was fine to keep in here. I simplified it even more just now though by using listPrompt()

brandenrodgers:
> The only potential problem i can think of - Will this work as intended if the user has npm installed on their machine but installed the CLI in some other way? From what I can tell yarn and npm global installs shouldn't conflict with each other so we're probably ok on that front

This is a good call. I should check if npm AND the cli are both globally installed before trying to auto-update

camden11 (dismissed):
Code looks good + the config set command updates working well. Is there a good way to test the actual update functionality locally?

brandenrodgers:
@camden11 It got a little buried under the other comments, but I put a quick explanation of how to test it [here](https://github.com/HubSpot/hubspot-cli/pull/1514#issuecomment-2880608310).

camden11 (approved):
Just tested and works! LGTM

PR #1519, opened 2025-05-15
Better errors for hs create

brandenrodgers:
## Description and Context

I noticed we were logging the entire axios response object whenever the clone GH repo request failed. This wraps them all in a try catch and makes sure that we're waiting for the responses.


joe-yeager (approved):
Good catch on the missing awaits

PR #1520, opened 2025-05-15
Support project profiles

brandenrodgers:
## Description and Context

An initial pass at supporting project profiles. Depends on some changes to the project parsing lib.


commands/project/dev/index.ts
```diff
@@ -50,15 +46,61 @@ export async function handler(
...
+        uiLine();
+        process.exit(EXIT_CODES.ERROR);
+      }
+
+      targetAccountId = profile.accountId;
+
+      logProfileFooter(profile);
+    } else {
+      // A profile must be specified if this project has profiles configured
+      await exitIfUsingProfiles(projectConfig, projectDir);
```
brandenrodgers, line 69: For all of the commands that leverage profiles. We will exit if you have profiles configured AND you don't specify a profile for the command. It has to be explicitly provided

lib/validation.ts
```diff
@@ -168,20 +168,17 @@ export function validateCmsPublishMode(
 }
 
 export function fileExists(_path: string): boolean {
```
brandenrodgers, line 170: I made some edits to this so I could use it in my new commands. I didn't want to include the logs. Nothing was calling this function except for `checkAndConvertToJson` so this should be a safe change. I tested it on the commands that use `checkAndConvertToJson` and everything seems to WAD

commands/project/profile/remove.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  // This should never happen
+  if (!profileName) {
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  let targetAccountId: number | undefined;
+
+  try {
+    const profileToRemove = loadHsProfileFile(projectSourceDir, profileName);
```
brandenrodgers, line 84: Intentionally not using `loadProfile` here because I don't want any of the logs if it fails to load. I'm loading it because it lets us prompt the user to remove the project from the account if we can find the `accountId` in it.

commands/project/profile/add.ts
```diff
@@ -0,0 +1,268 @@
...
+    );
+
+    if (promptResponse) {
+      profileToCopyVariablesFrom = promptResponse;
+    }
+  }
+
+  if (profileToCopyVariablesFrom) {
+    try {
+      const profileToCopyFileContent = loadHsProfileFile(
```
brandenrodgers, line 197: Not using `loadProfile` here because we don't want to log anything on errors. We'll make the variables section in the new config empty if this fails.

commands/project/__tests__/deploy.test.ts
kemmerle, line 1: I would suggest also adding a test for the `conflicts` option.

commands/project/dev/index.ts
```diff
@@ -70,13 +112,22 @@ export async function handler(
...
+  });
+
   yargs.example([
     [
       '$0 project dev',
       i18n(`commands.project.subcommands.dev.examples.default`),
     ],
   ]);
 
+  yargs.conflicts('profile', 'account');
```
kemmerle, line 129: Have we added a test for `conflicts` here?

commands/project/profile/add.ts
```diff
@@ -0,0 +1,268 @@
...
+  commands.project.profile.add.verboseDescribe,
+  false
+);
+
+type ProjectProfileAddArgs = CommonArgs & {
+  name?: string;
+  targetAccount?: number;
+};
+
+async function handler(
```
kemmerle, line 39: Could we split some of the `handler` code into smaller functions for readability?

lang/en.ts
```diff
@@ -907,6 +907,90 @@ export const commands = {
...
+        },
+        positionals: {
+          name: 'The name of the project profile',
+        },
+        options: {
+          targetAccount: 'The target account ID for this profile',
+        },
+      },
+      remove: {
+        describe: 'Remove an exisiting project profile',
```
kemmerle, line 956: Typo

brandenrodgers: Thanks for catching this one 👌 fixed in [9494987](https://github.com/HubSpot/hubspot-cli/pull/1520/commits/94949878726ed9bab4a22eb036076b79a2817582)

commands/project/dev/index.ts
brandenrodgers: Good call, done in [9494987](https://github.com/HubSpot/hubspot-cli/pull/1520/commits/94949878726ed9bab4a22eb036076b79a2817582)

commands/project/profile/add.ts
brandenrodgers: I moved all the logic to select a profile to copy the variables from into a dedicated util in the add.ts file. I think it should make the handler a littler easier to follow [7184fbf](https://github.com/HubSpot/hubspot-cli/pull/1520/commits/7184fbf223790b4f4f1246338071a5af248e9523)

kemmerle (approved):

PR #1521, opened 2025-05-16
Support private oauth apps in hs project dev

camden11:
## Description and Context
This updates the `AppDevModeInterface` to remove some of the marketplace app specific logic and support private oauth apps (and theoretically, static auth apps as well, though these aren't officially supported yet)

Most of the changes here are just updating variable names to not be marketplace specific

## Who to Notify

### DIFF
--- a/lib/projects/localDev/AppDevModeInterface.ts
+++ b/lib/projects/localDev/AppDevModeInterface.ts
@@ -28,8 +28,8 @@ type AppDevModeInterfaceConstructorOptions = {
 class AppDevModeInterface {
   localDevState: LocalDevState;
   localDevLogger: LocalDevLogger;
-  _app?: AppIRNode | null;
-  marketplaceAppData?: PublicApp;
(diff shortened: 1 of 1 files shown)


lib/projects/localDev/AppDevModeInterface.ts
```diff
@@ -144,24 +145,29 @@ class AppDevModeInterface {
...
-    if (!this.app) {
+    if (!this.appNode) {
       return;
     }
 
-    if (this.app?.config.distribution === APP_DISTRIBUTION_TYPES.MARKETPLACE) {
-      try {
-        await this.fetchMarketplaceAppData();
+    try {
+      await this.fetchAppData();
```
camden11, line 153: The only change to logic is in the `setup` function. We now fetch app data (which is required to check installation and get the install link) and check if the app is installed for _all_ apps rather than just marketplace apps

PR #1524, opened 2025-05-19
TS updates for the account commands

brandenrodgers:
## Description and Context

Updating the TS patterns in the account commands

### DIFF
--- a/commands/account/rename.ts
+++ b/commands/account/rename.ts
@@ -1,15 +1,15 @@
 import { Argv, ArgumentsCamelCase } from 'yargs';
 import { logger } from '@hubspot/local-dev-lib/logger';
 import { renameAccount } from '@hubspot/local-dev-lib/config';
-import { addConfigOptions, addAccountOptions } from '../../lib/commonOpts';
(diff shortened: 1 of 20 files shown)


commands/account/rename.ts
```diff
@@ -63,3 +60,23 @@ export function builder(yargs: Argv): Argv<AccountRenameArgs> {
 
   return yargs as Argv<AccountRenameArgs>;
 }
+
+const builder = makeYargsBuilder<AccountRenameArgs>(
+  accountRenameBuilder,
+  command,
+  describe,
+  {
+    useGlobalOptions: true,
```
camden11, line 69: Non-blocking (and doesn't make sense to be part of this PR) but we may just want to update `useGlobalOptions` to default to true. Not sure if there are any commands where it's false

camden11 (approved):

commands/account/rename.ts
brandenrodgers: Yeah that's a good idea. Agreed 👍

PR #1525, opened 2025-05-19
Update hs feedback destination

brandenrodgers:
## Description and Context

`hs feedback` will now send the user here: https://developers.hubspot.com/feedback

This removes the options, but this isn't a command that would ever be used as a part of an automated workflow. So I don't think this warrants a breaking change.

### DIFF
--- a/commands/feedback.ts
+++ b/commands/feedback.ts
@@ -20,7 +20,7 @@ async function handler() {
   );
 
   if (!shouldOpen) {
-    logger.error(
+    logger.log(
       i18n(`commands.project.subcommands.feedback.error`, {
         url: uiLink('the developer feedback form', FEEDBACK_URL),
       })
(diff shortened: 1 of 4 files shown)


commands/feedback.ts
```diff
@@ -1,77 +1,44 @@
...
   );
 
-  if (shouldOpen || usedTypeFlag) {
-    // NOTE: for now, all feedback should go to the hubspot-cli repository
-    const url = FEEDBACK_URLS.BUG;
-    open(url, { url: true });
-    logger.success(
-      i18n(`commands.project.subcommands.feedback.success`, { url })
+  if (!shouldOpen) {
+    logger.error(
```
camden11, line 23: Feel like this should just be a regular log and not an error

camden11 (approved):
One minor comment, otherwise LGTM

camden11 (approved):

PR #1529, opened 2025-05-21
Upload projects to test accounts in hs project dev

camden11:
## Description and Context
This uploads the `hs project dev` flow for platform versions `2025.2` and above to default to uploading projects to the developer test account chosen for testing. It also allows users to go through the dev flow with a test account as their default account,  and will skip the account selection prompt in this case.

## Testing
There are unfortunately still a few pain points with this flow, especially when using a new developer test account. Here's what you current need to do to get through the new `hs project dev` flows that this PR enables

#### New developer test account
Also applicable for existing developer test accounts that your project hasn't been uploaded to yet
1. Use a standard account with unified apps beta access as your default account
2. When prompted, create a new developer test account. After the account is created, you'll be prompted to upload your project.
3. Before uploading, ungate your new test account for `Developers:UnifiedApps:PrivateBeta`
4. Proceed with uploading your project. Next, you'll be prompted to install your app
5. Before attempting to install your app, update your backend with the client id and secret from your newly created app
6. Install your app in your test acconut
7. Proceed with local development

#### Separate project account and testing account
You may want to run `hs project dev` and test on an account other than the one where the project lives (similar to the current default flow). To do this, run `hs project dev --targetProjectAccount=[project account id or name] --targetTestingAccount=[testing account id or name]`

## TODO
The next step of this work is supporting sandboxes in the account selection prompt. This will be added in another PR

## Who to Notify
@brandenrodgers @joe-yeager

### DIFF
--- a/commands/project/dev/index.ts
+++ b/commands/project/dev/index.ts
@@ -31,8 +31,8 @@ async function handler(
   const {
     derivedAccountId,
     providedAccountId,
-    targetTestingAccount,
-    targetProjectAccount,
+    testingAccount,
+    projectAccount,
   } = args;
 
   const { projectConfig, projectDir } = await getProjectConfig();
@@ -47,21 +47,21 @@ async function handler(
 
   // If either targetTestingAccount or targetProjectAccount is provided, the other must be provided
   if (
-    (targetTestingAccount && !targetProjectAccount) ||
-    (targetProjectAccount && !targetTestingAccount)
+    (testingAccount && !projectAccount) ||
+    (projectAccount && !testingAccount)
   ) {
-    uiLogger.error('TODO');
+    uiLogger.error(commands.project.dev.errors.invalidAccountFlags);
     process.exit(EXIT_CODES.ERROR);
   }
 
   // Legacy projects do not support targetTestingAccount and targetProjectAccount
-  if (targetTestingAccount && targetProjectAccount && !useV3) {
-    uiLogger.error('TODO');
+  if (testingAccount && projectAccount && !useV3) {
+    uiLogger.error(commands.project.dev.errors.unsupportedAccountFlags);
     process.exit(EXIT_CODES.ERROR);
   }
 
   let targetProjectAccountId =
-    providedAccountId || getAccountId(targetProjectAccount);
+    providedAccountId || getAccountId(projectAccount);
 
   let profile: HsProfileFile | undefined;
 
@@ -91,7 +91,7 @@ async function handler(
   }
 
   const targetTestingAccountId =
-    getAccountId(targetTestingAccount) || targetProjectAccountId;
+    getAccountId(testingAccount) || targetProjectAccountId;
 
   trackCommandUsage('project-dev', {}, targetProjectAccountId);
 
@@ -100,21 +100,21 @@ async function handler(
   uiLogger.log(commands.project.dev.logs.learnMoreLocalDevServer);
 
   if (useV3) {
-    await unifiedProjectDevFlow(
+    await unifiedProjectDevFlow({
       args,
-      targetProjectAccountId,
-      targetTestingAccountId,
+      initialTargetProjectAccountId: targetProjectAccountId,
+      initialTargetTestingAccountId: targetTestingAccountId,
       projectConfig,
       projectDir,
-      profile
-    );
+      profileConfig: profile,
+    });
   } else {
-    await deprecatedProjectDevFlow(
+    await deprecatedProjectDevFlow({
       args,
-      targetProjectAccountId,
+      accountId: targetProjectAccountId,
       projectConfig,
-      projectDir
-    );
+      projectDir,
+    });
   }
 }
 
@@ -128,14 +128,12 @@ function projectDevBuilder(yargs: Argv): Argv<ProjectDevArgs> {
 
   yargs.options('targetTestingAccount', {
     type: 'string',
-    // TODO: We may want to add messaging that this only works on certain platform versions
     description: commands.project.dev.options.targetTestingAccount,
     hidden: true,
   });
 
   yargs.options('targetProjectAccount', {
     type: 'string',
-    // TODO: We may want to add messaging that this only works on certain platform versions
     description: commands.project.dev.options.targetProjectAccount,
     hidden: true,
   });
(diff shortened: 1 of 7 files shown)


commands/project/dev/unifiedFlow.ts
```diff
@@ -25,27 +26,20 @@ import SpinniesManager from '../../../lib/ui/SpinniesManager';
...
+import { uiLogger } from '../../../lib/ui/logger';
+import { commands } from '../../../lang/en';
+import { isDeveloperTestAccount, isSandbox } from '../../../lib/accountTypes';
 // import LocalDevWebsocketServer from '../../../lib/projects/localDev/LocalDevWebsocketServer';
 
 export async function unifiedProjectDevFlow(
   args: ArgumentsCamelCase<ProjectDevArgs>,
-  accountConfig: CLIAccount,
+  initialTargetProjectAccountId: number,
+  initialTargetTestingAccountId: number,
```
joe-yeager, line 37: Nit: This is a pretty high number of args, might be better to have it take an object as an arg

commands/project/dev/index.ts
```diff
@@ -21,36 +19,53 @@ import {
...
+    uiLogger.error(commands.project.dev.errors.noProjectConfig);
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  // If either targetTestingAccount or targetProjectAccount is provided, the other must be provided
+  if (
+    (targetTestingAccount && !targetProjectAccount) ||
+    (targetProjectAccount && !targetTestingAccount)
+  ) {
+    uiLogger.error('TODO');
```
joe-yeager, line 53: Placeholder log message

commands/project/dev/index.ts
```diff
@@ -21,36 +19,53 @@ import {
...
     process.exit(EXIT_CODES.ERROR);
   }
 
-  let targetAccountId = providedAccountId;
+  // Legacy projects do not support targetTestingAccount and targetProjectAccount
+  if (targetTestingAccount && targetProjectAccount && !useV3) {
+    uiLogger.error('TODO');
```
joe-yeager, line 59: Placeholder log message

commands/project/dev/index.ts
```diff
@@ -115,18 +122,31 @@ function projectDevBuilder(yargs: Argv): Argv<ProjectDevArgs> {
...
+  });
+
+  yargs.options('targetTestingAccount', {
+    type: 'string',
+    // TODO: We may want to add messaging that this only works on certain platform versions
+    description: commands.project.dev.options.targetTestingAccount,
+    hidden: true,
+  });
+
+  yargs.options('targetProjectAccount', {
```
joe-yeager, line 136: Nit: The naming of these is very similar.  I wonder if we should differentiate them more.  Maybe just removing the target prefix.  Reviewing this PR I found myself doing double takes a lot.

commands/project/dev/unifiedFlow.ts
```diff
@@ -55,84 +49,93 @@ export async function unifiedProjectDevFlow(
...
 
-  if (!derivedAccountIsRecommendedType && !profileConfig) {
-    logger.error(
-      i18n(`commands.project.subcommands.dev.errors.invalidUnifiedAppsAccount`),
-      {
-        authCommand: uiCommandReference('hs auth'),
-      }
+  if (!targetProjectAccountConfig) {
+    uiLogger.error(
+      commands.project.dev.errors.noAccount(initialTargetProjectAccountId)
```
joe-yeager, line 80: Thanks for updating all of this to the uiLogger and the new object syntax 🙌

joe-yeager (commented):
LGTM from a code perspective, just a few nits and a couple placeholder logs.  I haven't had a chance to run it yet

commands/project/dev/index.ts
```diff
@@ -21,36 +19,53 @@ import {
...
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  let targetProjectAccountId =
+    providedAccountId || getAccountId(targetProjectAccount);
 
   let profile: HsProfileFile | undefined;
 
-  if (!targetAccountId && useV3Api(projectConfig.platformVersion)) {
+  if (!targetProjectAccountId && useV3) {
```
brandenrodgers, line 68: This if condition can never be undefined because `getAccountId(targetProjectAccount)` will fall back to using the default account. I think we want to specifically check if `providedAccountId  || args.targetProjectAccountId` is defined

brandenrodgers:
Left one comment about something that was causing profiles to not WAD, but otherwise everything lgtm during testing 👍

commands/project/dev/index.ts
camden11: Ohhh good catch. So many of these config related functions are so confusing...can't wait to get rid of them

commands/project/dev/index.ts
```diff
@@ -21,36 +19,54 @@ import {
...
 
-  let targetAccountId = providedAccountId;
+  // Legacy projects do not support targetTestingAccount and targetProjectAccount
+  if (testingAccount && projectAccount && !useV3) {
+    uiLogger.error(commands.project.dev.errors.unsupportedAccountFlags);
+    process.exit(EXIT_CODES.ERROR);
+  }
+
+  let targetProjectAccountId =
+    (providedAccountId && derivedAccountId) ||
```
brandenrodgers, line 64: iirc `derivedAccountId` will always exist if `providedAccountId` exists

brandenrodgers (approved):
One nit about some potentially unnecessary logic, but otherwise lgtm

commands/project/dev/index.ts
camden11: That's what I thought initially, but `providedAccountId` is kind of a misnomer - its whatever value the user passed in to the account flag, so if it's a name, we don't convert it to and id. If it exists, then `derivedAccountId` will be the provided account id converted to an id if necessary.

There are a LOT of unnecessary conversions between account name / id / full config object so trying to cut back on those where possible

camden11:
Going to hold off on merging this until we get confirmation from Laurie that we want to include this in the bug bash

commands/project/dev/index.ts
brandenrodgers: > If it exists, then derivedAccountId will be the provided account id converted to an id if necessary.

If that's true, then isn't checking `providedAccountId` the same as checking `providedAccountId && derivedAccountId`?

camden11:
Closing since we are not moving forward with this flow until a later date

PR #1532, opened 2025-05-27
chore: Add prompt

joe-yeager:

### DIFF
--- a/lib/projects/localDev/AppDevModeInterface.ts
+++ b/lib/projects/localDev/AppDevModeInterface.ts
@@ -139,22 +139,28 @@ class AppDevModeInterface {
       return;
     }
 
-    const {
-      data: { isInstalledWithScopeGroups, previouslyAuthorizedScopeGroups },
-    } = await fetchAppInstallationData(
-      this.localDevState.targetTestingAccountId,
-      this.localDevState.projectId,
-      this.appNode.uid,
-      this.appNode.config.auth.requiredScopes,
-      this.appNode.config.auth.optionalScopes
-    );
+    if (this.appNode.config.auth.type === APP_AUTH_TYPES.OAUTH) {
+      const {
+        data: { isInstalledWithScopeGroups, previouslyAuthorizedScopeGroups },
+      } = await fetchAppInstallationData(
+        this.localDevState.targetTestingAccountId,
+        this.localDevState.projectId,
+        this.appNode.uid,
+        this.appNode.config.auth.requiredScopes,
+        this.appNode.config.auth.optionalScopes
+      );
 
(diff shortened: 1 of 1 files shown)


brandenrodgers (approved):

lib/projects/localDev/AppDevModeInterface.ts
```diff
@@ -70,10 +79,27 @@ class AppDevModeInterface {
...
+          this.appNode?.uid
+        )
+      );
+    }
+
     return getStaticAuthAppInstallUrl({
       targetAccountId: this.localDevState.targetTestingAccountId,
       env: this.localDevState.env,
-      appId: this.appNode!.uid,
+      appId: `${app.id}`,
```
joe-yeager, line 102: This needs to be the internal app id and not the uid

lib/projects/localDev/AppDevModeInterface.ts
```diff
@@ -70,10 +79,27 @@ class AppDevModeInterface {
         redirectUrls: this.appNode.config.auth.redirectUrls,
       });
     }
+
+    const {
+      data: { results },
+    } = await fetchPublicAppsForPortal(
+      this.localDevState.targetProjectAccountId
+    );
+    const app = results.find(app => app.sourceId === this.appNode?.uid);
```
joe-yeager, line 88: Fetch the apps for the portal and find the correct app by `uid`

lib/projects/localDev/AppDevModeInterface.ts
```diff
@@ -139,6 +165,10 @@ class AppDevModeInterface {
       return;
     }
 
+    if (this.appNode.config.auth.type === APP_AUTH_TYPES.STATIC) {
+      return installAppPrompt(await this.getAppInstallUrl(), false);
+    }
```
joe-yeager, line 170: This always prompts the user to install the app into portal when it is static auth

lib/projects/localDev/AppDevModeInterface.ts
```diff
@@ -87,7 +88,12 @@ class AppDevModeInterface {
     const app = results.find(app => app.sourceId === this.appNode?.uid);
 
     if (!app) {
-      throw new Error('Unable to find app in the project account');
+      throw new Error(
```
brandenrodgers, line 91: We could also just log an error here and exit the process

joe-yeager: https://github.com/HubSpot/hubspot-cli/pull/1532/commits/7b53a192bf602653575cdc26dde0fb3eadecbf2c

brandenrodgers (approved):

PR #1534, opened 2025-05-28
Show warning in profiles if there are duplicate ids in use

brandenrodgers:
## Description and Context

This adds a warning to `profile add` that lets the user know about the potential issues of targeting the same account with multiple profiles.

### DIFF
--- a/commands/project/profile/add.ts
+++ b/commands/project/profile/add.ts
@@ -23,6 +23,7 @@ import {
   confirmPrompt,
 } from '../../../lib/prompts/promptUtils';
 import { fileExists } from '../../../lib/validation';
+import { debugError } from '../../../lib/errorHandlers';
 
 const command = 'add [name]';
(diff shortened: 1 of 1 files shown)


commands/project/profile/add.ts
```diff
@@ -189,17 +188,42 @@ async function handler(
...
+  for (const profile of existingProfiles) {
+    try {
+      const loadedProfile = loadHsProfileFile(projectSourceDir, profile);
+      if (loadedProfile?.accountId === targetAccountId) {
+        existingTargetAccountIdSelected = true;
+        break;
+      }
+    } catch (err) {
+      // Skip profiles that can't be loaded
+      continue;
```
joe-yeager, line 207: Nit: The `continue;` is redundant if it's at the end of the block.  Maybe this could be a debug log instead.

joe-yeager (approved):
LGTM, just a nit.

PR #1535, opened 2025-05-28
[WIP] Support dev on dev sandbox accounts if the portal has access

brandenrodgers:
## Description and Context

Add support for selecting dev sandbox accounts during the `hs project dev` init flow if the user's account has access to them.

### DIFF
--- a/commands/project/dev/unifiedFlow.ts
+++ b/commands/project/dev/unifiedFlow.ts
@@ -3,6 +3,7 @@ import util from 'util';
 import { ArgumentsCamelCase } from 'yargs';
 import { logger } from '@hubspot/local-dev-lib/logger';
 import { getAccountIdentifier } from '@hubspot/local-dev-lib/config/getAccountIdentifier';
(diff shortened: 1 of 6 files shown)


commands/project/dev/unifiedFlow.ts
```diff
@@ -86,47 +88,74 @@ export async function unifiedProjectDevFlow(
...
+          env
+        );
+      }
+    } else if (accountType === HUBSPOT_ACCOUNT_TYPES.DEVELOPMENT_SANDBOX) {
+      const sandboxAccountPromptResponse =
+        await selectSandboxTargetAccountPrompt(accounts!, accountConfig);
+
+      targetTestingAccountId = sandboxAccountPromptResponse.targetAccountId;
+    } else {
+      targetTestingAccountId = targetProjectAccountId;
```
joe-yeager, line 158: Non-blocking nit:  This section might benefit from a refactor to split this up into smaller pieces.

brandenrodgers: I may hold off and let Camden break this up however he wants to when he gets back. He was another WIP pr that touches this code too.

PR #1537, opened 2025-05-30
Support app secrets

brandenrodgers:
## Description and Context

Supporting app secrets. Adds simple CRUD functionality for them, and tests.


lib/prompts/selectAppPrompt.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (appsResponse.data.results) {
+      availableApps = appsResponse.data.results;
+    }
+  } catch (err) {
+    debugError(err);
+  }
+
+  if (availableApps.length === 0) {
+    logger.error(lib.prompts.selectAppPrompt.errors.noApps);
+    process.exit(EXIT_CODES.ERROR);
```
joe-yeager, line 26: Just food for thought.  I've been trying to avoid calling `process.exit` outside of the handler methods and throwing errors instead.  The handler can then catch/log/exit.  I feels like that should be the responsibility of the handler and not the prompt method.  It also makes these methods easier to test.

joe-yeager (commented):
LGTM, just nit/suggestion.  It looks like there might be some type errors affecting the build

lib/prompts/selectAppPrompt.ts
brandenrodgers: Yeah that's a good call. I updated this to just return null instead, and then exit from the handler if there's no app returned in [ad1136d](https://github.com/HubSpot/hubspot-cli/pull/1537/commits/ad1136dbf2636380611e6db491b3ed6653c7e7f9)

Issue #702, opened 2022-06-02
add option to decrease verbosity of hs fetch, for CI/CD usage


techieshark:
## Description and Context

`hs fetch` emits one line of output for every file/folder it writes. This seems excessive for CI/CD use cases; it'd be great to have an option to decrease verbosity.

This with hs version 4.0.0.

### Steps to reproduce

1. Run `hs fetch` command to fetch files
2. Observe one line of output per file/folder

## Expected behavior

There is some way to decrease the output volume; perhaps a `--quiet` option.

chiragchadha1:
Going to close this one out. It's been open for 4 years without a taker and we don't have plans to prioritize this. Feel free to re-open if you'd like to contribute a PR, and we'd be happy to help review. Thanks!

PR #1576, opened 2026-01-10
Add --quiet option to hs cms fetch to reduce output verbosity

citruscai:
addresses issue #702


chiragchadha1:
Thanks for the PR @citruscai! I'm going to close as the HubSpot CLI is managed in an internal source repo that syncs out to this public one. We don't have a workflow set up for accepting external code contributions. I should have flagged that when I closed the original issue. 

We really appreciate your work. If `--quiet` ends up prioritized in the future, your patch here is a useful reference implementation. Thanks again!

PR #1600, opened 2026-04-15
fix(migrate): download project files when post-migration build fails

### LINKED ISSUE #1599: hs project migrate fails if initial build fails
### CLI Version

8.4.0

### Node.js Version

24.2.0

### Operating System

Debian Linux

### Description

When `hs project migrate` runs, the HubSpot server performs the source
transformation and then immediately attempts an automatic build. If that
build fails (e.g. due to a transient server error), the migration status
API returns FAILURE with the message:

  "Your project was migrated, but the initial build failed. Check the
   CLI logs or your project build logs to fix any issues, then try
   building again."

This status unambiguously indicates that the migration itself succeeded —
only the automatic post-migration build attempt failed. However, the CLI
treats this FAILURE identically to a true migration failure: it throws and
exits before ever calling `downloadProjectFiles`. As a result, the
developer is left with an empty project directory containing only
`hsproject.json`, and there is no way to recover without re-running the
migration or manually reconstructing the migrated sources.

### Steps to Reproduce

Run `hs project migrate` from Platfrom 2025.1 with serverless functions built with esbuild

### Expected Behavior

hs project migrate downloads the changed source code for manual fixing

### Screenshots

_No response_

### Additional Context

_No response_

klemens-ninja:
When the migration API returns FAILURE with 'initial build failed', the migration itself succeeded but the automatic post-migration build hit a transient server error. The CLI was treating this FAILURE status identically to a real migration failure: it threw in finalizeAppMigration before downloadProjectFiles was ever called, leaving the developer with only an empty hsproject.json and no way to recover the migrated sources.

Changes:
- Add fetchProject to imports from @hubspot/local-dev-lib/api/projects
- Introduce FinalizeResult discriminated union type for finalizeAppMigration
- In finalizeAppMigration, detect the 'initial build failed' FAILURE case, emit a warning instead of throwing, and return the sentinel { buildId: undefined, migrationSucceededBuildFailed: true }
- In migrateApp, when the sentinel is present, call fetchProject to resolve the latest build ID and proceed to downloadProjectFiles as normal

Fixes #1599

### DIFF
--- a/lib/app/migrate.ts
+++ b/lib/app/migrate.ts
@@ -10,6 +10,7 @@ import { AUTO_GENERATED_COMPONENT_TYPES } from '@hubspot/project-parsing-lib/con
 import { MIGRATION_STATUS } from '@hubspot/local-dev-lib/types/Migration';
 import {
   downloadProject,
+  fetchProject,
   fetchProjectComponentsMetadata,
 } from '@hubspot/local-dev-lib/api/projects';
 import { Separator } from '@inquirer/prompts';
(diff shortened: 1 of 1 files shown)


chiragchadha1:
Thanks for digging into this and putting together a fix — we appreciate the effort. After investigating the backend, we've found that the `FAILURE` status is intentional here: when the build fails, the migration hasn't actually been finalized (the platform version isn't updated), so the CLI is correct to treat it as a failure.

The underlying issue is the error message from the server is misleading. We're working with the backend team to improve that. We're also discussing whether the transformed source files should be made available for download on build failure.

We've asked for `--debug` output on the issue to confirm exactly what's happening in your case. Going to close this PR for now, but thanks again for flagging the problem.

